PR #57, opened 2018-05-25
First go at a PolyFill

pipobscure:
This is a first go at a polyfill. 

Let's discuss and I'll continuously adapt the implementation as we come to conclusions.

### DIFF
--- a/polyfill/README.md
+++ b/polyfill/README.md
@@ -1,0 +1,37 @@
+# Temporal Polyfill
+
+This is an implementation of a polyfill for the [TC39 Termporal Proposal](https://github.com/tc39/proposal-temporal).
+
+## Implementation Issues:
+
+## Proposed Changes / Additions to the Proposal
+
+### `Instant` constructor to take single `BigInt` argument
+
+### `Instant.toDate()` & `Instant.fromDate()` for interoperability with `Date()`
+
+### `Instant.prototype.valueOf()` & `ZonedInstant.prototype.valueOf()`
+
+To facilitate comparisons between `Intant`s and/or `ZonedInstant`s it would be extremely useful to implement `valueOf()`.
+Considering the fact that these have nanosecond precision, the proposed `[BigInt](https://github.com/tc39/proposal-bigint)`
+seems to be the right choice of primitive.
+
+### `Instant.prototype.toDate()` & `ZonedInstant.prototype.toDate()`
+
+To facilitate better interoperation with existing `Date` APIs it would be useful to add a casting method to
+`Instant` & `ZonedInstant`.
+
+An example of such APIs would be the `Intl.DateTimeFormat` apis which work well with Dates, but not (yet?) with `Instant` &
+`ZonedInstant`. Though for this specific case it might be better to add `format()` (see below).
+
+### `Instant.prototype.format(locale = navigator.language, options = {})` & `ZonedInstant.prototype.format(locale = navigator.language, options = {})`
+
+Considering the frequency of converting date/time values to customized strings, the `format` convenience method seems
+to make sense. It should follow the same signature as `Intl.DateTimeFormat.prototype.format` with the notable exception
+that `locale` ise defaulted to `navigator.language` and the `timeZone` option is overriddent to the `ZonedInstant`'s time
+zone or UTC for `Instant`s.
+
+### static `Instant.now()`
+
+This would facilitate easier creation of Instant objects eliminating the need to work with `Date` objects. In addition,
+when implemented natively, this would enable a higher precision timer.
--- a/polyfill/package.json
+++ b/polyfill/package.json
@@ -1,0 +1,33 @@
+{
+  "name": "temporal",
+  "version": "1.0.0",
+  "description": "TC39 Temporal Proposal Polyfill",
+  "main": "index.js",
+  "module": "lib/index.mjs",
+  "directories": {
+    "lib": "lib",
+    "test": "test"
+  },
+  "scripts": {
+    "test": "for FILE in $(find test -type f -name '*.js' | grep -v 'test/node.js'); do node --harmony-bigint -r ./test/node.js $FILE; done",
+    "prepare": "npm run build",
+    "build": "rollup -c rollup.config.js"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://bbgithub.dev.bloomberg.com/javascript-guild/temporal-polyfill.git"
+  },
+  "keywords": [
+    "TC39",
+    "Polyfill",
+    "Temporal"
+  ],
+  "author": "Philipp Dunkel <pdunkel1@bloomberg.net>",
+  "license": "UNLICENSED",
+  "dependencies": {},
+  "devDependencies": {
+    "rollup": "^0.59.1",
+    "rollup-plugin-license": "^0.6.0",
+    "rollup-plugin-node-resolve": "^3.3.0"
+  }
+}
--- a/polyfill/package.json
+++ b/polyfill/package.json
@@ -9,7 +9,9 @@
     "test": "test"
   },
   "scripts": {
-    "test": "for FILE in $(find test -type f -name '*.js' | grep -v 'test/node.js'); do node --harmony-bigint -r ./test/node.js $FILE; done",
+    "test:bigint": "for FILE in $(find test/temporal -type f -name '*.js' | grep -v 'test/node.js'); do node -r ./test/node.js $FILE; if [ $? = 0 ]; then echo 'OK ' $FILE; else echo 'FAIL ' $FILE; fi; done",
+    "test:compat": "tape test/compat/**/*.js",
+    "test": "npm run test:compat && npm run test:bigint",
     "prepare": "npm run build",
     "build": "rollup -c rollup.config.js"
   },
@@ -23,11 +25,12 @@
     "Temporal"
   ],
   "author": "Philipp Dunkel <pdunkel1@bloomberg.net>",
-  "license": "UNLICENSED",
+  "license": "MIT",
   "dependencies": {},
   "devDependencies": {
     "rollup": "^0.59.1",
     "rollup-plugin-license": "^0.6.0",
-    "rollup-plugin-node-resolve": "^3.3.0"
+    "rollup-plugin-node-resolve": "^3.3.0",
+    "tape": "^4.9.1"
   }
 }
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+node_modules/
+polyfill/index.js
+polyfill/index.js.map
--- a/polyfill/LICENSE
+++ b/polyfill/LICENSE
@@ -1,0 +1,6 @@
+Copyright (C) 2018 Bloomberg LP. All rights reserved.
+
+This software will be open-source, however a specific license has not yet been decided.
+
+Until such a time that this License file contains an valid open-source license, please
+consider this software UNLICENSED and proprietary. All rights are reserved.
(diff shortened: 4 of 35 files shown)


polyfill/lib/civildatetime.js
```diff
@@ -0,0 +1,51 @@
+import { plus, pad, getMsSinceEpoch  } from './util.js';
+import { CivilDate } from './civildate.js';
+import { CivilTime } from './civiltime.js';
+import { Instant } from './instant.js';
+import { ZonedInstant } from './zonedinstant.js';
+
+const DATA = Symbol('data');
```
ljharb: Instead of only using a Symbol, what about using a `WeakMap` (with a Symbol → string fallback)? That way the state would be truly private when possible.

polyfill/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "version": "1.0.0",
+  "description": "TC39 Temporal Proposal Polyfill",
+  "main": "index.js",
+  "directories": {
+    "lib": "lib",
+    "test": "test"
+  },
+  "scripts": {
+    "test": "for FILE in $(find test -type f -name '*.js' | grep -v 'test/node.js'); do node -r ./test/node.js $FILE; done",
+    "build": "rollup -c rollup.config.js"
```
ljharb: this should also have `"prepublish": "npm run build"`

polyfill/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "keywords": [
+    "TC39",
+    "Polyfill",
+    "Temporal"
+  ],
+  "author": "Philipp Dunkel <pdunkel1@bloomberg.net>",
+  "license": "UNLICENSED",
+  "dependencies": {},
+  "devDependencies": {
+    "rollup": "^0.59.1",
```
ljharb: do we need to use rollup here? that prevents people from deep-importing just the parts they need.

maggiepint:
I am excited and will review tomorrow 😀

maggiepint:
Also, what is Bloomberg's license policy? And ECMAs? Maybe this should get pulled out to a personal repo or even a JS Foundation one?

ljharb:
Probably should go in a separate repo, then it can be MIT-licensed without having to ask anyone.

polyfill/README.md
```diff
@@ -0,0 +1,41 @@
...
+This is an implementation of a polyfill for the [TC39 Termporal Proposal](https://github.com/tc39/proposal-temporal).
+
+## Implementation Issues:
+
+### Edge-Case: Getting *epoch-milliseconds*
+
+Getting the correctly time-zoned *milliseconds since epoch* for a `CivilDateTime` when calling `withZone` is currently
+based on `Intl.DateTimeFormat` fumbling. However this fails for the edgecase where the instant to be extracted falls
+within one timezone offset of a DST-Change.
+
```
maggiepint: @icambron do you have a work around for this?

polyfill/README.md
```diff
@@ -0,0 +1,41 @@
...
+The cost of implementing this correctly would be to bring in the TZ-Database (like momentjs has done); or alternatively using momentjs directly. However neither are worth the weight for the purpose of exploring the API.
+
+## Proposed Changes / Additions to the Proposal
+
+### `Instant.prototype.valueOf()` & `ZonedInstant.prototype.valueOf()`
+
+To facilitate comparisons between `Intant`s and/or `ZonedInstant`s it would be extremely useful to implement `valueOf()`.
+Considering the fact that these have nanosecond precision, the proposed `[BigInt](https://github.com/tc39/proposal-bigint)`
+seems to be the right choice of primitive.
```
maggiepint: Yes! I was reluctant to depend on bigint when it was early stage, but its done at this point.

pipobscure:
Hi all,

1. Symbol vs. WeakMap : I don’t think actual privacy is a feature (realising I’m alone with this). In reality we should be using class-private fields for that. I’ll recode for that and transpile for now. For now this makes debugging so much easier than WeakMap that I’m inclined to leave that until the last minute.

2. Edge-Case: I just wanted to get something to play with quickly. I’m in the process of generating a proper version of it as we speak. (I have all the bits just need a bunch of conversion/simplification)

3. License: from a conversation with the powers that be I got the impression that it’s nit as much of an issue and we’ll just conform to whatever the proposal is under. I based this on a proposal from Dominic and he was using BSD.

4. Build/Rollup: Yes this is not in a distributable state yet. I just want to be able to have an easy way to make this work in all environments. prepublish / potentially minifies prebuilts will be a thing.

5. BigInt: Great. Should the API change because of this? (Instant Constructor)? Do we want static .fromDate() .fromNanoseconds() .fromMilliseconds(millis, nanos=0)? rather than the generic. (Please discuss)

6. Merging: DO NOT! This is for now just something we can play with to work out the details of the proposal. Once the proposal is finalised (i.e. in a shape all the proponents are happy with) I’ll make sure this is in a distributable and strictly compliant state.

Cheers,
Philipp

polyfill/README.md
icambron: Kind of. I don't quite understand why "within one timezone offset of a DST change" is the bounds of the edge case. Maybe the way Luxon uses Intl to do this pretty different?

For Luxon, there are two edge cases:
 
1. The CivilDateTime is inside spring forward and thus doesn't exist. This is easy to detect: guess the offset (Luxon uses "what's the offset for `now()` in that zone?"), compute the epoch millis, then test whether the resulting date has the same offset as your guess. If yes, it's good to go, if not, adjust the epoch millis by the difference. The way Luxon does this, you end up pushing the date forward in these cases. The code for that is (mostly) here:  https://github.com/moment/luxon/blob/baf63e639b1b635f10ef54744c2def23ecfb9140/src/datetime.js#L72.

2. The CivilDateTime is an ambiguous fall back time. This one is tougher. Here's what Luxon does: https://moment.github.io/luxon/docs/manual/zones.html#ambiguous-times. Codewise, that just uses the code above; if the offset of the generated time matches the guess, we're done. That means we could end up with either side of the ambiguity (just by having different guesses) and never even know it, but c'est la vie.

pipobscure:
TL;DR
the edgecase is just a result of this coming out of a hackathon. I just mentioned it to avoid justified criticism. Will be fixed shortly, just haven’t gotten to it yet.

The difference to luxon is that I don’t know the offset. To find it I basically take the UTC and format it in the TZ. That gives me the offset unless it’s within 1 offset of a DST change. Since I didn’t want to bother including a full TZ database, this was a quick and dirty fix that works for the 99% case and doesn’t require including the full DB.

maggiepint:
Phillip,

This is also what Isaac does with luxon.

icambron:
Yeah, you really should check out the Luxon details. It also uses the Intl API to compute offsets with no knowledge of TZ database info, but I think I've hammered down all the edge cases (at least, all the ones I know about!)

polyfill/package.json
keithamus: [`prepublish` is deprecated in favour of `prepare`](https://docs.npmjs.com/misc/scripts#prepublish-and-prepare)

polyfill/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "temporal",
+  "version": "1.0.0",
+  "description": "TC39 Temporal Proposal Polyfill",
+  "main": "index.js",
```
keithamus: As this is using ES modules; and rollup to generate a umd bundle, it'd be great to have a `module` keyword for build tools that can understand ES modules.

pipobscure:
This is a fascinating experience. The whole point of pointing out the EdgeCase thing was to prevent needless discussion of it, because it was a known hack in there only because it's Hackathon code 😄 

Well that didn't work, it ended up being the most discussed item.

---

Any case, I have eliminated the edgecase by basically doing the same thing as luxon (thanks for pointing me at it).

I've also changed the code to work with BigInt and thereby made some assumptions on how the API would change given BigInt.

---

Can we discuss whether the API should change in the proposed ways? Or do we want it to remain as speced? Or...

polyfill/package.json
ljharb: prepare won’t work in older npm versions; prepublish still works in everything.

littledan:
For license, would you be OK with using [Ecma's software license](https://www.ecma-international.org/memento/Policies/Ecma_Policy_on_Submission_Inclusion_and_Licensing_of_Software.htm) (BSD)?

ljharb:
Is there a reason not to use MIT? It’s far more common in the ecosystem.

pganssle:
@littledan @ljharb If you're discussing licenses, can I put in a plug for Apache 2.0 instead of one of the other two popular permissive licenses? Apache 2.0 has essentially the same obligations as the other licenses, but makes explicit a lot of what is implicit in MIT and BSD.

ljharb:
@pganssle I’d prefer MIT > Apache > BSD, in that order, but I’m fine with multiple licenses. In practice, none of the “implicitness” you’re concerned about in MIT is a legal problem, in my experience.

pganssle:
@ljharb Well, very few of these licenses has really been tested in court. And there's a certain "mutually assured destruction" if they are interpreted another way, but there's still a lot of grey area on the margin. Interestingly, the whole "BSD + Patents" fiasco was essentially because Facebook provided you *more* rights than what BSD or MIT provides you out of the box (BSD gives you *no* patent grant, BSD + Patents gives you a patent grant revocable under certain conditions.), but the way it was structured people realized how exposed they were to patent claims from Facebook. 

Why prefer MIT to Apache? Both are very common OSI approved licenses. Almost everything that's in Apache is basically a *requirement* to feasibly use MIT in your project (patent grant, contributors are assumed to release their contributions under the same license, etc), so Apache provides no additional burden and heads off a lot of possible issues at the pass.

ljharb:
Because in the JS ecosystem, MIT is (by far) the most common choice.

pipobscure:
Thanks for the input on licensing. Ultimately that will be up to the powers that be (that are footing the bill) to decide (in progress). I'll put the arguments made to them. I personally (it's not my call to make) think that considering it should be a polyfill/implementation (non-normative) in conjunction with a TC39 proposal as well as tests (normative) that ECMA's Software License may be most appropriate. Though I think discussing licenses is both premature as well as out of expertise. Let's get this into a shape first that resembles what we want to push at TC39; in the meantime I'll bug the experts on licensing to solve that issue.

ljharb:
@pipobscure for reference, all my proposals’ polyfills (and all the polyfills i maintain for other features) are MIT.

pganssle:
@ljharb I don't really understand how "other projects use MIT" is an argument in favor of MIT. Apache and MIT are both permissive licenses and are fully compatible. You don't get anything out of having two MIT licensed projects used together because you need to retain the individual copyright notices for each project *anyway*.

ljharb:
@pganssle it's not an argument against Apache, certainly - but "do what everyone else does" is definitely a strong argument in favor of that thing, especially from the legal perspective (something my company's legal team encourages me to consider).

pipobscure:
Licensing has been dealt with. I checked with our legal eagles, and they told me to go with MIT. So MIT it is.

Truth be told, I don’t think it matters much one way or another; so long as it’s permissive enough to be universally useful. And MIT definitely meets that criterion

maggiepint:
Do we want to continue to keep this in this repository? Not sure how we want to move forward with this polyfill - though it would be great to get people using it and hear feedback.

pipobscure:
I think it’s the right place. Let’s discuss synchronously.

littledan:
@ljharb All source code included in the ECMAScript specification is (not necessarily exclusively) licensed under the Ecma software license, which looks MIT-style to me. If you have concerns with this policy, I recommend discussing them within Ecma--for example, you could come to a GA meeting and explain your concerns.

PR #72, opened 2018-07-24
initial stab at zonedinstant  @maggiepint

RedSquirrelious:
Largely adapted from @pipobscure 's Instant.  Still must add algorithms for toCivil*

### DIFF
--- a/spec-zonedinstant.html
+++ b/spec-zonedinstant.html
@@ -1,0 +1,165 @@
++<pre class=metadata> 
+  title: Temporal proposal: ZonedInstant
+  toc: false 
+  stage: 1 
+  contributors: Maggie Pint, Brian Terlson, Matt Johnson, Sasha Pierson, Philipp Dunkel
+</pre> 
+<emu-intro id="intro"> 
+  <h1>ZonedInstant Objects</h1>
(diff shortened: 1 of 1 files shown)


spec-zonedinstant.html
```diff
@@ -0,0 +1,165 @@
...
+</emu-clause>
+<emu-clause id="sec-temporal-zonedinstant-alg">
+  <h1>ZonedInstant(*instant*, *timeZone*)</h1>
+  <p>This description only applies if the ZonedInstant constructor is called with at least 2 arguments.</p>
+  <p>When the ZonedInstant constructor function is called, the following steps are taken:</p>
+  <emu-alg>
+    1. Let numberOfArgs be the number of arguments passed to this function call. 
+    1. Assert: numberOfArgs == 2. 
+    1. If NewTarget is not undefined, then 
+      a. Let ms be instant.milliseconds.
```
pipobscure: We are not storing MilliSeconds & NanoSeconds separately, but rather a single BigInt value. So should this be:


a. Let ns be instant.valueOf().
a. Let tz be timeZone.
a. Let O be ? OrdinaryCreateFromConstructor(NewTarget, "%ZonedInstantPrototype%",<<[[Value,TimeZoneValue]]>>)
a. Set O.[[Value]] to ns
a. Return O.

PR #88, opened 2018-09-21
move polyfill to separate repo

pipobscure:
@ljharb I've moved the polyfill to another repo. This should ease releasing to npm.


pipobscure:
The move and refactor is complete. Thought the polyfill is a bit ahead of the proposal in that it implements a few things I am advocating, so as to be able to play with them.

 * `CivilXXX` objects renamed to `ISOXXX`
 * `toCivilXXX` methods removed from `ZonedInstant` and replaced by `fromZonedInstant` statics on these.
 * `millisecond`, `microsecond`, `nanosecond` are the same value at different granularities rather than smaller units of the same.
 * Probably more. **PLEASE HAVE A LOOK!!!**

pipobscure:
@maggiepint @mj1856 can one of you merge this?

PR #94, opened 2018-09-28
Correct optional constructor arguments for CivilTime and CivilDateTime.

IgnoredAmbience:
The previous forms of the constructor optional arguments gave the
nanosecond field the highest priority, with the second field the lowest.
This patch reverses the optional brace nesting to match intended usage.

Fixes #75.


IgnoredAmbience:
Note: I have not regenerated spec-rendered.html in this patch.

PR #97, opened 2018-10-08
Update Instant & Renamed ZonedInstant to ZonedDateTime

pipobscure:


rxaviers:
Should README.md be updated accordingly? Thanks

ljqx:
what about the name `ZonedCivilDateTime`?

spec/zoneddatetime.html
```diff
@@ -0,0 +1,164 @@
...
+      Hour: `00` through `24`
+      Minute: `00` through `59`
+      Second: `00` through `59`
+      Nanoseconds: `000000000` through `999999999`
+      OffsetSign: `+` or `-`
+      TimeOffset: OffsetSign Hour Second
+      IANAZoneName: a geographic IANA timezone name
+      IANAZone: `[` IANAZoneName `]`
+      TimeZone: OffsetSign [ IANAZone ] (IANAZone is only included if a geographic IANA timezone was specified)
+      IsoDateTime: Year Month Day `T` Hour Minute Second `.` Nanoseconds `Z`
```
ljqx: if `IANA` is all uppercase, then `Iso` should be too

pipobscure: Done

README.md
```diff
@@ -237,14 +238,24 @@ Represents an absolute point in time, with an associated time zone.
 
 ### Constructor
 ```js
-new ZonedInstant(instant, timeZone)
+new ZonedDateTime(instant, timeZone)
```
gibson042: The heading still says ZonedInstant.

spec/zoneddatetime.html
```diff
@@ -0,0 +1,164 @@
...
+      <p>This description only applies if the ZonedDateTime constructor is called with at least 2 arguments.</p>
+      <p>When the ZonedDateTime constructor function is called, the following steps are taken:</p>
+      <emu-alg>
+        1. Let numberOfArgs be the number of arguments passed to this function call.
+        1. Assert: numberOfArgs == 2.
+        1. If NewTarget is not undefined, then
+          1. Let ms be instant.milliseconds.
+          1. Let ns be instant.nanoseconds.
+          1. Let tz be timeZone.
+          1. Let O be ? OrdinaryCreateFromConstructor(NewTarget, "%ZonedInstantPrototype%", <<[[MillisecondValue,
```
gibson042: There are still several remaining references to ZonedInstant.

README.md
```diff
@@ -237,14 +238,24 @@ Represents an absolute point in time, with an associated time zone.
...
-let timeZone = zonedInstant.timeZone;
+let year = zonedInstant.year;
+let month = zonedInstant.month;
+let day = zonedInstant.day;
+let hour = zonedInstant.hour;
+let minute = zonedInstant.minute;
+let second = zonedInstant.second;
+let millisecond = zonedInstant.millisecond; // 0-999
+let microsecond = zonedInstant.microsecond; // 0-999
+let nanosecond = zonedInstant.nanosecond; // 0-999
```
gibson042: I think it would be very confusing for civil values and instant values to have such similarly-named properties with such wildly different semantics.

Unit | CivilTime and CivilDateTime | ZonedDateTime | Instant
---- | ---- | ---- | ----
millisecond | whole milliseconds since last second transition | whole milliseconds since last second transition | whole POSIX milliseconds since epoch
microsecond | _absent_ | whole microseconds since last millisecond transition | whole POSIX microseconds since epoch
nanosecond | whole nanoseconds since last millisecond transition | whole nanoseconds since last _microsecond_ transition | whole POSIX nanoseconds since epoch

README.md
```diff
@@ -212,22 +212,23 @@ Counted as number of nanoseconds from `1970-01-01T00:00:00.000000000Z`.
...
 ### Constructor
 ```js
-new Instant(milliseconds[, nanoseconds])
+new Instant(nanoseconds)
 ```
 
 #### Parameters
- - `milliseconds` : Integer value representing the number of milliseconds elapsed from 1970-01-01 00:00:00.000 UTC, without regarding leap seconds.
- - `nanoseconds` : Optional. Integer value representing the nanosecond within the millisecond.
+ - `nanoseconds` : BigInt value representing the number of nanoseconds elapsed from 1970-01-01 00:00:00.000 UTC, without regarding leap seconds.
```
gibson042: This PR is about renaming ZonedInstant; unrelated changes to Instant should be removed from it.

spec/instant.html
```diff
@@ -8,48 +8,47 @@ <h1>Instant Objects</h1>
   <emu-clause id=sec-temporal-instant-constructor>
     <h1>new Instant(*nanoseconds*)</h1>
     <p>When called as a constructor creates a new Instant object. *nanoseconds* is a BigInt of the nanoseconds since
-      epoch.</p>
+      epoch. This value is stored in the [[value]] slot.</p>
```
gibson042: I don't think Instant changes belong in this PR, but as an aside: if we want to use internal slots at all (which is debatable), [[InstantData]] or maybe [[InstantValue]] would be more consistent with the rest of ECMAScript.

pipobscure:
Subsumed by #111

PR #128, opened 2019-04-03
Define the Temporal module.

Ms2ger:
Fixes #98.

I don't consider this the optimal solution in terms of layering, but let's at least get the discussion started.

### DIFF
--- a/biblio.json
+++ b/biblio.json
@@ -1,0 +1,14 @@
+{
+  "https://proposal-javascript-standard-library-domenic.now.sh/": [
+    {
+      "type": "op",
+      "aoid": "CreateSyntheticModule",
+      "id": "sec-createsyntheticmodule"
+    },
+    {
+      "type": "op",
+      "aoid": "SetSyntheticModuleExport",
+      "id": "sec-setsyntheticmoduleexport"
+    }
+  ]
+}
--- a/spec.html
+++ b/spec.html
@@ -7,11 +7,14 @@ stage: 1
 contributors: Maggie Pint, Matt Johnson, Brian Terlson, Daniel Ehrenberg, Philipp Dunkel, Sasha Pierson
 </pre>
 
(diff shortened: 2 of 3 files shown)


ljharb:
The proposal reached consensus for stage 2 with "globals or built-in modules".

Built-in modules do not yet exist, and the spec has no convention for declaring specifiers - and in fact that is completely left up to implementations.

I find it very inappropriate for a proposal in stage 2 to, without achieving consensus to do so, to change a major semantic (the sort of thing that is required to be decided prior to stage 2), especially to depend on a stage 1 feature that has an uncertain future.

I appreciate that this PR is only trying to write spec text for the module approach, but since built-in modules aren't stage 2, there's no spec to build off of.

littledan:
@ljharb You've explained your thoughts on built-in modules in other threads in this repisotry. It seems clear that there's a disagreement here. Would it be OK for this repository to continue development from the champion group, and then we can discuss your feedback at the next TC39 meeting?

ljharb:
I won’t be able to attend the next one, but my feedback is that it should be returned to stage 1 if it’s not going to include spec text for how one would access the feature in the language that exists. Obviously you can make whatever changes you like in the repo.

littledan:
OK, thanks for explaining your point. Let's see if we can discuss this further in the July or September meetings. If we agree that this proposal repository can continue development, I don't think there's any need for further clarification.

ryzokuken:
@ljharb did you already discuss this and reach consensus?

ljharb:
The July and September meetings haven’t happened yet.

ryzokuken:
Sure! I was just checking in to see if you had perhaps reached a conclusion over online discussions.

littledan:
My understanding of the current plan is that we will proceed with classes under a `Temporal` namespace object, rather than a module, since built-in modules are still at Stage 1.

Issue #12, opened 2017-04-19
What should this be named? Is it a built in module or a global namespace?


maggiepint:
Currently the working name is "temporal", but concerns were voiced in #10 about this.

Agree it may not be a perfect name. What does make sense?

Turbo87:
I had a few unstructured thoughts about this:

- the internationalization code is grouped in the `Intl` namespace now, which is capitalized. Maybe it should be `Temporal` instead of `temporal` then?

- should we use `DateTime.Local` and `DateTime.Zoned` instead? but then again for the current proposal that might make sense but if we want to add functionality in the future it might not fit as well.

apaprocki:
If the built-in route is chosen, the name would also have to be a reserved keyword. Something like `Temporal` is probably much less common than other words, but I'm not sure about how research was done before adding prior new reserved keywords. The built-in module route is certainly much safer.

domenic:
Why would it have to be a reserved keyword? It can just be any name, like all the other globals we've added recently (e.g. Reflect).

apaprocki:
@domenic Sorry, was under the impression that the built-ins were reserved in some manner. Strike that!

hollowdoor:
I happen to like the name. As far as module vs. global I think all apis that aren't primitives should be modules, but that's impossible. :)

mj1856:
Sticking with `temporal`.  Thanks for the feedback!

PR #136, opened 2019-06-02
Document Mental-Model and Naming choices

pipobscure:
Describe the underlying mental model as well as the reasioning and discussion behin the naming of Civil objects.


civil.md
```diff
@@ -0,0 +1,57 @@
...
+
+Famous quote:
+
+    There are only two hard things in Computer Science: cache invalidation and naming things.
+
+    -- Phil Karlton
+
+Infamous quote:
+
+    Cache invalidation is easy if you don't use a caches.
```
ljharb:
```suggestion
    Cache invalidation is easy if you don't use a cache.
```

civil.md
```diff
@@ -0,0 +1,57 @@
...
+
+    -- Phil Dunkel
+
+Which leaves us with a *hard problem in computer science*: the naming of the `Civil` family of objects.
+
+**There are a set of objects that represents dates, times as well as aggregations and parts thereof which do not contain enough data to tie them to a sepcific timezone or locality: what do we call them?**
```
ljharb:
```suggestion
**There are a set of objects that represents dates, times, as well as aggregations and parts thereof which do not contain enough data to tie them to a sepcific timezone or locality: what do we call them?**
```

civil.md
```diff
@@ -0,0 +1,57 @@
...
+
+**Options:**
+
+ * `Gregorian*` since they are specific to the Gregorian Calendar.
+ * `Local*` since they look like what we use to talk about our local time.
+ * `Civil*` since they look like what we use in civil life.
+
+## Gregorian
+
+**Pro:** The reason `Gregorian` was suggested was in view of the fact that we are limiting ourselves to the gregorian calendar. Future implementations of these object could them be named appropriately according to the calendar they implement and be on equal footing to the current set of objects that we intend to support.
```
ljharb:
```suggestion
**Pro:** The reason `Gregorian` was suggested was in view of the fact that we are limiting ourselves to the gregorian calendar. Future implementations of these objects could then be named appropriately according to the calendar they implement and be on equal footing to the current set of objects that we intend to support.
```

mentalmodel.md
```diff
@@ -0,0 +1,94 @@
...
+  datetime --- time
+  date --- yearmonth
+  date --- monthday
+```
+![Mental Model Diagram](./objectrelation.png)
+
+Before continuing I'd also like to call out the following API principles to keep in mind when evaluating the proposal:
+
+1. All objects are immutable
+2. Methods beginning with `with` will result in objects that contain equal or more data and are therefore safe in terms of lossing information.
```
ljharb:
```suggestion
2. Methods beginning with `with` will result in objects that contain equal or more data and are therefore safe in terms of losing information.
```

mentalmodel.md
```diff
@@ -0,0 +1,94 @@
...
+3. Methods beginning with `get` will result in retrieving an aspect of the object. Thus the value retrieved will have less information that the original object.
+4. The calculation methods `plus`, `minus` and `difference` act with or result in objects of the same type. `difference` is only permissible with objects of the same information content, while `plus` and `minus` are permissible only with `Durations` (or objects containing the same level of information) generated using the `difference` method.
```
ljharb:
```suggestion
4. The calculation methods `plus`, `minus`, and `difference` act with or result in objects of the same type. `difference` is only permissible with objects of the same information content, while `plus` and `minus` are permissible only with `Durations` (or objects containing the same level of information) generated using the `difference` method.
```

mentalmodel.md
```diff
@@ -0,0 +1,94 @@
...
+
+**Please also see the detailed [API description of these temporal objects](objects.md)!!!**
+
+## Instant
+
+`Instant` objects are meant to represent a specific point in absolute time. For convenience we have chosen that point to be *midnight of January the 1st 1970 UTC* as our *epoch*. The choice fell upon that specific date and time due to the precendent of the [POSIX Epoch](http://pubs.opengroup.org/onlinepubs/9699919799/mindex.html) and Unix Epoch based on it.
+
+We decided to ignore  leap seconds since they are arbitrarily decided upon and unpredictable. An `Instant` can therefore not be said to be on the TAE timeline, but rather on the UTC timeline. In terms of leap-seconds we accept them a input (`fromString`) but immediately convert them to the relevant numeric representation and therefore ignoring the fact that this came from a leap-second.
```
ljharb:
```suggestion
We decided to ignore leap seconds since they are arbitrarily decided upon and unpredictable. An `Instant` can therefore not be said to be on the TAE timeline, but rather on the UTC timeline. In terms of leap-seconds we accept them a input (`fromString`) but immediately convert them to the relevant numeric representation and therefore ignoring the fact that this came from a leap-second.
```

mentalmodel.md
```diff
@@ -0,0 +1,94 @@
...
+
+## OffsetDateTime
+
+`OffsetDateTime` objects represent a point in time together with a specific offset from UTC. Given that these offets are only valid on Earth, they are tied to a place. However they do not have a timezone or dayligh-davings adjustments or anything of that nature.
+
+As such they can be considered at best to be tied to the planet Earth rather than a specific locality. However because they are tied to the planet, they do have a concept of time units uch as hours and days. And because we restrict ourselves to the proleptic gregorian calendar, they do have months and years that correspond with our western understanding of these ideas.
+
+## ZonedDateTime
+
+`ZonedDateTime` objects represent a point in time in a specific IANA timezone. Since these timezones are tied to a specific locality (city/country), these are more precise and observe daylight savings changes.
```
ljharb:
```suggestion
`ZonedDateTime` objects represent a point in time in a specific IANA timezone. Since these timezones are tied to a specific locality (city/state/country), these are more precise and observe daylight savings changes.
```

mentalmodel.md
```diff
@@ -0,0 +1,94 @@
...
+
+Contrary to `ZonedDateTime` and `OffsetDateTime` the `CivilDateTime` objects are not tied to an absolute point in time. An example for this would be the statement *the next new year is at 2020-01-01T00:00:00.000000000*. This is true independent of where on the planet one is and independent of any locality. It also represents a different point in time when combined with a timezone:
+
+ * Berlin: `2019-12-31T23:00:00.000000000Z`
+ * London: `2020-01-01T00:00:00.000000000Z`
+ * New York: `2020-01-01T05:00:00.000000000Z`
+
+## CivilDate
+
+This contain just the date part. It can be mentally considered the *leaf of a desk-calendar*. It can be placed anywhere on the planet, can mean any time within the 24-hour day, but the extent of the time period cannot be acurately specified as it has a different meaning in Australia and Europe for example.
```
ljharb:
```suggestion
This contains just the date part. It can be mentally considered the *leaf of a desk-calendar*. It can be placed anywhere on the planet, can mean any time within the 24-hour day, but the extent of the time period cannot be acurately specified as it has a different meaning in Australia and Europe for example.
```

mentalmodel.md
```diff
@@ -0,0 +1,94 @@
...
+
+This has usually been represented by absuing a datetime object of some sort. However this leads to all kinds of problems when doing date/time maths. This is a more accurate and specific representation of the concept.
```
ljharb:
```suggestion
This has usually been represented by abusing a datetime object of some sort. However this leads to all kinds of problems when doing date/time maths. This is a more accurate and specific representation of the concept.
```

mentalmodel.md
```diff
@@ -0,0 +1,94 @@
...
+
+## CivilTime
+
+This represents a time without a date or relation to a specific location. 12:00:00 is noon every day and in every place. It can be considered like the display of an unwound watch. It is on the wrist of a wearer, and unless someone adjusts it, winds it or otherwise influences it, it will be stable, but devoid of meaning relative to any actual timeline.
+
+## CivilYearMonth
+
+This represents an entire month of a year. It's an incompete date, but nevertheless useful in many contexts. An example use would be the statement: *By January 2020 we want to have completed the temporal specification*
```
ljharb:
```suggestion
This represents an entire month of a year. It's an incomplete date, but is nevertheless useful in many contexts. An example use would be the statement: *By January 2020 we want to have completed the temporal specification*
```

mentalmodel.md
```diff
@@ -0,0 +1,94 @@
...
+
+In terms of interoperability, the TAG review yielded the information that HTML form elements can yield dates like this. And this concept is included in other specifications. To provide a full and compatible set of objects we determined that it was within scope of this proposal to include this type.
+
+## CivilMonthDay
+
+Similarly this is a day and month without a year. An example would be *new years day* / *the 1st of January* where the year is considered irelevant as it is a valid statement for any year.
```
ljharb:
```suggestion
Similarly this is a day and month without a year. An example would be *new years day* / *the 1st of January* where the year is considered irrelevant as it is a valid statement for any year.
```

mentalmodel.md
```diff
@@ -0,0 +1,94 @@
...
+
+In terms of interoperability, the TAG review yielded the information that HTML form elements can yield dates like this. And this concept is included in other specifications. To provide a full and compatible set of objects we determined that it was within scope of this proposal to include this type.
+
+## Durations
+
+`Duration` objects are only yielded by taking the difference between two like temporal objects. Durations are always absolute values and can be used for date/time maths using the `plus` and `minus` operation on temporal objects.
+
+The danger with these objects is that they can incorreclty be applied to unlike objects. Consider applying the difference between `2019-01-02T12:00:00.000000000` and `2019-02-02T13:00:00.000000000` which is `1month 1hour` to a date of `2019-05-01` which drops the hours entirely or a time of `15:00:00.000000000` which ignores the month.
```
gibson042:
```suggestion
The danger with these objects is that they can incorrectly be applied to unlike objects. Consider applying `P1MT1H` (1 month and 1 hour) to a date of `2019-05-01` (which ignores the hour) or a time of `15:00:00.000000000` (which ignores the month).
```

mentalmodel.md
```diff
@@ -0,0 +1,94 @@
...
+```mermaid
+graph TB
+  instant((Instant))
+    offset((OffsetDateTime))
+    zoned((ZonedDateTime))
+      datetime((CivilDateTime))
+      time((CivilTime))
+      date((CivilDate))
+      yearmonth((CivilYearMonth))
+      monthday((CivilMonthDay))
```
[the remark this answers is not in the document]
pipobscure: No, it's just a much simpler version. will bring #12 in as well

mentalmodel.md
```diff
@@ -0,0 +1,94 @@
...
+  date --- yearmonth
+  date --- monthday
+```
+![Mental Model Diagram](./objectrelation.png)
+
+Before continuing I'd also like to call out the following API principles to keep in mind when evaluating the proposal:
+
+1. All objects are immutable
+2. Methods beginning with `with` will result in objects that contain equal or more data and are therefore safe in terms of lossing information.
+3. Methods beginning with `get` will result in retrieving an aspect of the object. Thus the value retrieved will have less information that the original object.
```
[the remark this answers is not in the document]
pipobscure: Tried writing explainers / examples and found it extremely bad experience. This version of `get` is just a suggestion. Will be opening issue soonish

ljharb:
It kind of defeats the purpose of a pull request if reviewers aren’t given time to rereview changes :-/

PR #137, opened 2019-06-04
Cleaning README by externalising tech-notes

pipobscure:


techrecord.md
```diff
@@ -0,0 +1,79 @@
...
+ * Temporal objects should not have an inheritance relationship
+ * offsetSeconds as a measurement is a relic of `Date` and should be discontinued.
+
+### Augmentation / Information lossiness
+
+All temporal objects should be information stable. All methods that increase or keep stable the amount of data should have the `with` prefix. All methods that decrease the amount of data (lossy operations) should have the `get` prefix.
+
+### Durations are absolute values with integer units
+
+Negative durations make no sense. Therefore there is a need to have both a `plus` and a `minus` method for date/time arithmetics.
```
gibson042: I don't dispute the value of having both `plus` and `minus`, but negative durations make _tons_ of sense!

techrecord.md
```diff
@@ -0,0 +1,79 @@
...
+Only integer units are supported in durations and arithmetic. Expressing fractions requires the use of smaller units.
```
gibson042: We still need to resolve Number vs. BigInt. `new Duration({ nanoseconds: Number.MAX_SAFE_INTEGER })` is only 104 days.

techrecord.md
```diff
@@ -0,0 +1,79 @@
...
+OffsetDateTime.fromString rejects input with a bracketed time zone, since it only cares about offset and cannot necessarily verify consistency (e.g., because the implementation doesn't have time zone data). For example, OffsetDateTime.fromString("2019-04-02T22:58:36.123456789-04:00[America/New_York]") does not result in a successful parse.
+
+ZonedDateTime.fromString requires a bracketed time zone and consistency between that value and the offset. For example, ZonedDateTime.fromString("2019-04-02T22:58:36.123456789-05:00[America/New_York]") does not result in a successful parse because the UTC offset and time zone are inconsistent at the given instant.
+
+For eliminating the need to use Date, fromString functions accept reduced precision values, e.g. CivilTime.fromString("10:23") and OffsetDateTime.fromString("2019-04-03T02:30Z") and maybe CivilDate.fromString("2019-04").
+
+### Other decisions / discussion from 2019-03
+
+ * Time-Unit properties should be named as below (plural in Duration and singular in every other type)?
+ * Temporal objects should not have an inheritance relationship
```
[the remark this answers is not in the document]
pipobscure: What I mean here is that:

`new CivilDateTime(...) instanceof CivilDate === false`
`new ZonedDateTime(...) instanceof OffsetDateTime === false`

and so on.

techrecord.md
pipobscure: `Duration` is not limited to nanoseconds. Also `Duration` cannot be created with `new`.

So `{ years: 5, months:3, days:28, minutes: 5, nanoseconds: 100 }` is fine and therefore numeric members are entirely fine here.

gibson042: I think it is a mistake to have no Duration constructor, but that doesn't actually matter for this point because the plan #120 includes a `largestElementInResult` parameter (which means dealing with something like `dateTime.plus({years: 1}).difference(dateTime, "nanoseconds")`).

PR #140, opened 2019-07-05
Specify the integration with the Intl.DateTimeFormat API.

Ms2ger:
This is a work in progress. Please let me know if this is roughly the direction to go in.

CC #129.

### DIFF
--- a/spec/intl.html
+++ b/spec/intl.html
@@ -87,9 +87,27 @@
         1. Let _r_ be ResolveLocale( %DateTimeFormat%.[[AvailableLocales]], _requestedLocales_, _opt_, %DateTimeFormat%.[[RelevantExtensionKeys]], _localeData_).
         1. Set _dateTimeFormat_.[[Locale]] to _r_.[[locale]].
         1. Set _dateTimeFormat_.[[Calendar]] to _r_.[[ca]].
-        1. Set _dateTimeFormat_.[[HourCycle]] to _r_.[[hc]].
+        1. <del>Set _dateTimeFormat_.[[HourCycle]] to _r_.[[hc]].</del>
+        1. <ins>Let _dataLocale_ be _r_.[[dataLocale]].</ins>
+        1. <ins>Let _dataLocaleData_ be _localeData_.[[&lt;_dataLocale_&gt;]].</ins>
+        1. <ins>Let _hcDefault_ be _dataLocaleData_.[[hourCycle]].</ins>
+        1. <ins>Let _hc_ be _r_.[[hc]].</ins>
+        1. <ins>If _hc_ is *null*, then</ins>
+          1. <ins>Set _hc_ to _hcDefault_.</ins>
+        1. <ins>If _hour12_ is not *undefined*, then</ins>
+          1. <ins>If _hour12_ is *true*, then</ins>
+            1. <ins>If _hcDefault_ is `"h11"` or `"h23"`, then</ins>
+              1. <ins>Set _hc_ to `"h11"`.</ins>
+            1. <ins>Else,</ins>
+              1. <ins>Set _hc_ to `"h12"`.</ins>
+          1. <ins>Else,</ins>
+            1. <ins>Assert: _hour12_ is *false*.</ins>
+            1. <ins>If _hcDefault_ is `"h11"` or `"h23"`, then</ins>
+              1. <ins>Set _hc_ to `"h23"`.</ins>
+            1. <ins>Else,</ins>
+              1. <ins>Set _hc_ to `"h24"`.</ins>
         1. Set _dateTimeFormat_.[[NumberingSystem]] to _r_.[[nu]].
-        1. Let _dataLocale_ be _r_.[[dataLocale]].
+        1. <del>Let _dataLocale_ be _r_.[[dataLocale]].</del>
         1. Let _timeZone_ be ? Get(_options_, `"timeZone"`).
         1. If _timeZone_ is not *undefined*, then
           1. Let _timeZone_ be ? ToString(_timeZone_).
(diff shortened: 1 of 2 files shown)


spec/intl.html
```diff
@@ -0,0 +1,372 @@
...
+          1. Set _dateTimeFormat_.[[HourCycle]] to _hc_.
+          1. If _dateTimeformat_.[[HourCycle]] is `"h11"` or `"h12"`, then
+            1. Let _pattern_ be _bestFormat_.[[pattern12]].
+          1. Else,
+            1. Let _pattern_ be _bestFormat_.[[pattern]].
+        1. Else,
+          1. Set _dateTimeFormat_.[[HourCycle]] to *undefined*.
+          1. Let _pattern_ be _bestFormat_.[[pattern]].
+        1. Set _dateTimeFormat_.[[Pattern]] to _pattern_.
+        1. <ins>For each row in <emu-xref href="#table-temporal-patterns"></emu-xref>, except the header row, in table order, do</ins>
```
sffc: How about making a new abstract operation `GetDateTimeFormatPattern(limitedOptions, formats)` that runs this logic but also the logic for the regular non-Temporal case (in which limitedOptions == options)?  I did something similar in Intl.NumberFormat with a new abstract operation `GetNumberFormatPattern`.

spec/intl.html
```diff
@@ -0,0 +1,372 @@
...
+          1. <ins>For each field _field_ of _opts_, do</ins>
+            1. <ins>If _field_ is in _fields_, then</ins>
+              1. <ins>Set _limitedOptions_.[[&lt;field>]] to _opts_.[[&lt;field>]].</ins>
+          1. <ins>If _matcher_ is `"basic"`, then</ins>
+            1. <ins>Let _bestFormat_ be BasicFormatMatcher(_limitedOptions_, _formats_).</ins>
+          1. <ins>Else,</ins>
+            1. <ins>Let _bestFormat_ be BestFitFormatMatcher(_limitedOptions_, _formats_).</ins>
+          1. <ins>If _bestFormat_ has any fields that are in _fields_, then</ins>
+            1. <ins>If _bestFormat_ has an [[hour]] field, then</ins>
+              1. <ins>Do the hourCycle dance.</ins>
```
sffc: If you make a new abstract operation, then you don't have to duplicate complex parts of the spec like the "hourCycle dance".

spec/intl.html
```diff
@@ -0,0 +1,372 @@
...
+            1. <ins>Else,</ins>
+              1. <ins>Let _pattern_ be _bestFormat_.[[pattern]].</ins>
+          1. <ins>Else,</ins>
+            1. <ins>Let _pattern_ be *null*.</ins>
+          1. <ins>Set _dateTimeFormat_'s internal slot whose name is the Pattern column of the row to _pattern_.</ins>
+        1. Return _dateTimeFormat_.
+      </emu-alg>
+
+      <emu-table id="table-temporal-patterns">
```
sffc: You can wrap the whole table in `<ins class="block">` (also for IsTemporalObject)

spec/intl.html
```diff
@@ -0,0 +1,372 @@
...
+          <tr>
+            <th>[[CivilDatePattern]]</td>
+            <td>[[weekday]], [[era]], [[year]], [[month]], [[day]]</td>
+          </tr>
+        </table>
+      </emu-table>
+
+    </emu-clause>
+
+    <emu-clause id="sec-formatnumberstring" aoid="IsTemporalObject">
```
sffc: If you add this here, it changes the section numbers for all following sections.  Probably safest to just add new operations at the end.

spec/intl.html
```diff
@@ -0,0 +1,372 @@
...
+        1. Let _nfOptions_ be ObjectCreate(*null*).
+        1. Perform ! CreateDataPropertyOrThrow(_nfOptions_, `"useGrouping"`, *false*).
+        1. Let _nf_ be ? Construct(%NumberFormat%, &laquo; _locale_, _nfOptions_ &raquo;).
+        1. Let _nf2Options_ be ObjectCreate(*null*).
+        1. Perform ! CreateDataPropertyOrThrow(_nf2Options_, `"minimumIntegerDigits"`, 2).
+        1. Perform ! CreateDataPropertyOrThrow(_nf2Options_, `"useGrouping"`, *false*).
+        1. Let _nf2_ be ? Construct(%NumberFormat%, &laquo; _locale_, _nf2Options_ &raquo;).
+        1. <del>Let _tm_ be ToLocalTime(_x_, _dateTimeFormat_.[[Calendar]], _dateTimeFormat_.[[TimeZone]]).</del>
+        1. Let _result_ be a new empty List.
+        1. Let _patternParts_ be PartitionPattern(<del>_dateTimeFormat_.[[Pattern]]</del><ins>_pattern_</ins>).
```
sffc: I don't see the section from here down in the current published spec or in the draft on GitHub.  Where did it come from?  For example, here you have,

```html
For each element _patternPart_ of _patternParts_, in List order, do
```

but the current spec parses the pattern in the same function and has,

```html
Repeat, while _beginIndex_ is an integer index into _pattern_
```

spec/intl.html
```diff
@@ -0,0 +1,372 @@
...
+          1. <ins>Let _tm_ be ToLocalTime(_x_, _dateTimeFormat_.[[Calendar]], _dateTimeFormat_.[[TimeZone]]).</ins>
+        1. Let _locale_ be _dateTimeFormat_.[[Locale]].
+        1. Let _nfOptions_ be ObjectCreate(*null*).
+        1. Perform ! CreateDataPropertyOrThrow(_nfOptions_, `"useGrouping"`, *false*).
+        1. Let _nf_ be ? Construct(%NumberFormat%, &laquo; _locale_, _nfOptions_ &raquo;).
+        1. Let _nf2Options_ be ObjectCreate(*null*).
+        1. Perform ! CreateDataPropertyOrThrow(_nf2Options_, `"minimumIntegerDigits"`, 2).
+        1. Perform ! CreateDataPropertyOrThrow(_nf2Options_, `"useGrouping"`, *false*).
+        1. Let _nf2_ be ? Construct(%NumberFormat%, &laquo; _locale_, _nf2Options_ &raquo;).
+        1. <del>Let _tm_ be ToLocalTime(_x_, _dateTimeFormat_.[[Calendar]], _dateTimeFormat_.[[TimeZone]]).</del>
```
sffc: It might be worth specifying in more detail how ToLocalTime works.  Right now it's very hand-wavy.

spec/intl.html
```diff
@@ -0,0 +1,372 @@
...
+            1. <ins>Let _pattern_ be _dtf_.[[ZonedDateTimePattern]].</ins>
+            1. Let _offsetDateTime_ be _date_.getOffsetDateTime().
+            1. <ins>Let _tm_ be { [[weekday]]: _offsetDateTime_.dayOfWeek, [[year]]: _offsetDateTime_.year, [[month]]: _offsetDateTime_.month, [[day]]: _offsetDateTime_.day, [[hour]]: _offsetDateTime_.hour, [[minute]]: _offsetDateTime_.minute, [[second]]: _offsetDateTime_.second }.</ins>
+          1. <ins>If _pattern_ is *null*, throw a *TypeError* exception.</ins>
+        1. <ins>Else,</ins>
+          1. <ins>Let _pattern_ be _dtf_.[[Pattern]].</ins>
+          1. If _date_ is *undefined*, then
+            1. Let _x_ be Call(%Date_now%, *undefined*).
+          1. Else,
+            1. Let _x_ be ? ToNumber(_date_).
```
sffc: You may need to resolve that _x_ is a different type here than in the Temporal case.  It looks like _x_ is only used in one other place in this operation, down at the bottom in the fallback block.

spec/intl.html
```diff
@@ -0,0 +1,372 @@
...
+            1. <ins>Let _pattern_ be _dtf_.[[CivilTimePattern]].</ins>
+            1. <ins>Let _tm_ be { [[hour]]: _date_.hour, [[minute]]: _date_.minute, [[second]]: _date_.second }.</ins>
+          1. <ins>Else if _date_ is a CivilDateTime object, then</ins>
+            1. <ins>Let _pattern_ be _dtf_.[[CivilDateTimePattern]].</ins>
+            1. <ins>Let _tm_ be { [[weekday]]: _date_.dayOfWeek, [[year]]: _date_.year, [[month]]: _date_.month, [[day]]: _date_.day, [[hour]]: _date_.hour, [[minute]]: _date_.minute, [[second]]: _date_.second }.</ins>
+          1. <ins>Else if _date_ is a ZonedDateTime object, then</ins>
+            1. <ins>Let _pattern_ be _dtf_.[[ZonedDateTimePattern]].</ins>
+            1. Let _offsetDateTime_ be _date_.getOffsetDateTime().
+            1. <ins>Let _tm_ be { [[weekday]]: _offsetDateTime_.dayOfWeek, [[year]]: _offsetDateTime_.year, [[month]]: _offsetDateTime_.month, [[day]]: _offsetDateTime_.day, [[hour]]: _offsetDateTime_.hour, [[minute]]: _offsetDateTime_.minute, [[second]]: _offsetDateTime_.second }.</ins>
+          1. <ins>If _pattern_ is *null*, throw a *TypeError* exception.</ins>
```
sffc: I support throwing an exception here, but Dan mentioned that maybe the other approach would be to use a fallback for _pattern_ that includes all of the fields if none of the fields for a particular Temporal type were specified.  That would probably be logic for InitializeDateTimeFormat.  Keep it as-is for now, and we should follow up later on exactly which behavior we want here.

spec/intl.html
```diff
@@ -0,0 +1,372 @@
...
+      <p>
+        The PartitionDateTimePattern abstract operation is called with arguments _dateTimeFormat_ (which must be an object initialized as a DateTimeFormat) and <del>_x_</del><ins>_date_</del> (which must be <del>a Number</del><ins>an ECMAScript</ins> value), <del>interprets _x_ as a time value as specified in ES2015, <emu-xref href="#sec-time-values-and-time-range"></emu-xref>,</del> and creates the corresponding parts according to the effective locale and the formatting options of _dateTimeFormat_. The following steps are taken:
+      </p>
+
+      <emu-alg>
+        1. <ins>If IsTemporalObject(_date_) is *true*,</ins>
+          1. <ins>Let _x_ be _date_.</ins>
+          1. <ins>If _date_ is a CivilDate object, then</ins>
+            1. <ins>Let _pattern_ be _dtf_.[[CivilDatePattern]].</ins>
+            1. <ins>Let _tm_ be { [[weekday]]: _date_.dayOfWeek, [[year]]: _date_.year, [[month]]: _date_.month, [[day]]: _date_.day }.</ins>
```
sffc: Are you allowed to access the internal slots of the Temporal type like this?  Make sure you are accessing the internal slots in a way that is consistent with the rest of the spec.

domenic:
FWIW this is not the direction I want to go in: see https://github.com/tc39/proposal-temporal/issues/129#issuecomment-480277289

spec/intl.html
sffc: FYI, this is being taken care of in https://github.com/tc39/ecma402/pull/348

spec/intl.html
Ms2ger: This is https://github.com/tc39/ecma402/pull/366.

spec/intl.html
Ms2ger: The rest of the spec doesn't really exist yet; once the shape of these objects is defined formally, we can update this.

sffc:
Why did you close this?  This PR is still relevant.

pipobscure:
I was asked to rename master to main (git branches) that seems to have done that automatically.

ljharb:
You may be able to edit the PR to change the base branch, and reopen it?

sffc:
Ping @Ms2ger or @pipobscure, can one of you reopen this PR? (I would but it seems I don't have write access to this repo)

PR #160, opened 2019-10-07
Add Polyfill based on es-abstract

pipobscure:


README.md
```diff
@@ -2,6 +2,8 @@
 
 Provides standard objects and functions for working with dates and times.
 
+**NOTE: The current working version is the [Polyfill](./polyfill). Specification text is being written based on that!**
```
ljharb: maybe worth explicitly clarifying this as not for production use?

polyfill/lib/absolute.mjs
```diff
@@ -0,0 +1,138 @@
...
+import { SLOT_EPOCHNANOSECONDS } from "./slots.mjs";
+
+export function Absolute(epochNanoSeconds) {
+  if (!(this instanceof Absolute)) return new Absolute(epochNanoseconds);
+  this[SLOT_EPOCHNANOSECONDS] = epochNanoSeconds;
+}
+
+Absolute.prototype.getEpochSeconds = function getEpochSeconds() {
+  let epochNanoSeconds = this[SLOT_EPOCHNANOSECONDS];
+  let epochSecondsBigInt = epochNanoSeconds / BigInt(1_000_000_000);
```
ljharb: why not use bigint syntax?

polyfill/lib/absolute.mjs
```diff
@@ -0,0 +1,138 @@
...
+  epochNanosecondsParam
+) {
+  let epochNanoseconds = BigInt(epochNanosecondsParam);
+  let resultObject = new Absolute(epochNanoseconds);
+  return resultObject;
+};
+Absolute.fromString = function fromString(isoString) {};
+
+Object.defineProperty(Absolute.prototype, Symbol.toStringTag, {
+  get: () => "Temporal.Absolute"
```
ljharb: why are these getters? if they are getters, can they brand check? if not, why not data properties?

polyfill/lib/absolute.mjs
```diff
@@ -0,0 +1,138 @@
+import { ES } from "./ecmascript.mjs";
+import { SLOT_EPOCHNANOSECONDS } from "./slots.mjs";
+
+export function Absolute(epochNanoSeconds) {
+  if (!(this instanceof Absolute)) return new Absolute(epochNanoseconds);
```
ljharb: why not use class syntax for these?

polyfill/lib/date.mjs
```diff
@@ -0,0 +1,286 @@
...
+  let month = ES.ISODateTimePartString(this[SLOT_MONTH]);
+  let day = ES.ISODateTimePartString(this[SLOT_DAY]);
+  let resultString = `${year}-${month}-${day}`;
+  return resultString;
+};
+Date.prototype.toLocaleString = function toLocaleString(...args) {
+  return new Intl.DateTimeFormat(...args).format(this);
+};
+Date.prototype.toJSON = function toJSON() {
+  return this.toString();
```
ljharb: should this observably call toString on the receiver, or should it just === Date.prototype.toString?

polyfill/lib/absolute.mjs
pipobscure: Because most other intrinsic objects can be created/cast while omitting the `new`. For consistency sake, I wanted the same behaviour. In additions `Time(15, 30, 45)` without the new actually feels more natural/convenient, especially given that they are immutable objects.

polyfill/lib/absolute.mjs
pipobscure: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/toStringTag

pipobscure: Will change to value property

polyfill/lib/absolute.mjs
pipobscure: Because the transformation I use can understand BigInt syntax (123n) and the Number Separator proposal (1_000) but not the combination thereof. So this is purely for convenience.

README.md
```diff
@@ -2,6 +2,9 @@
...
+***This is NOT suitable for production use as it has several performance issues that are in place for the sake of clarity!!!***
```
littledan: Can I ask why? I think it'll be good to have a production-ready polyfill at some point.

jasonwilliams: I agree, it would be useful to have an official polyfill developers can use in their applications, which would make the transition from Date to Temporal even easier.

polyfill/lib/absolute.mjs
littledan: I'd really discourage this pattern as well, though it's a minor point. Did you file an issue against whoever upstream fails to support the combination of BigInt and numeric separators?

polyfill/lib/absolute.mjs
littledan: ES6 and later tends to require `new` for classes, rather than making them callable. I think we should follow this convention in Temporal.

pipobscure:
Just to clarify: the point of this polyfill is to almost read like spectext, so as to enable easier writing which will happen next.  Because some of the abstracts (especially around timezones) are simple and cheap in C++ bit expensive in JS right now, this chooses clarity over performance.

Once this is done, I will create a version that can be used in production where I can use all the tricks in the book at the cost of clarity.

polyfill/lib/absolute.mjs
pipobscure: I disagree because of patterns like:

dateTime.with({ day: 5 })
dateTime.with(Date(2019,8,10))
Time(dateTime) // casting to Time
Date(dateTime) // casting to Date

I think this makes things so much more ergonomic. So if you feel strongly, can you open an issue so we can have that discussion properly.

polyfill/lib/absolute.mjs
pipobscure: Changed and fixed tooling

README.md
ljharb: It would be useful, but a) proposals generally aren’t recommended for use in production prior to stage 3 anyways, and b) a production-ready polyfill is maximally backwards compatible, which would include not using bigints and symbols etc directly.

polyfill/lib/absolute.mjs
littledan: Agreed with @ljharb. Let's go over these sorts of decisions in the Temporal meeting today; I put it on the agenda.

PR #171, opened 2019-10-10
Update spec text for Temporal.Duration

ryzokuken:
Started tinkering on this, please feel free to review and give suggestions.

/cc @pipobscure @littledan @Ms2ger

### DIFF
--- a/spec/duration.html
+++ b/spec/duration.html
@@ -2,55 +2,442 @@
 <meta charset="utf8" />
 
 <emu-clause id="sec-temporal-duration">
-    <h1>Duration Objects</h1>
-    <p>
-    Duration objects describe the difference between two like Date, Time, or DateTime objects.
-    They are only created via the <i>.minus()</i> methods of these objects.
-    </p>
+  <h1>Temporal.Duration Objects</h1>
+
+  <p>
+    A Temporal.Duration object describes the difference between two like Date,
+    Time, or DateTime objects. Objects of this type are only created via the
+    <i>.difference()</i> methods of these objects.
+  </p>
+
+  <!-- <ul>
+    <li><mark>TODO: what's the consensus on this?</mark></li>
+    <li>
+      The <i>.difference()</i> operation on Temporal.Date objects creates a
+      Temporal.Duration that only has years, months, and days properties.
+    </li>
+    <li>
+      The <i>.difference()</i> operation on Temporal.Time objects creates a
+      Temporal.Duration that only has hours, minutes, seconds, milliseconds,
+      microseconds, and nanoseconds properties.
+    </li>
+    <li>
+      The <i>.difference()</i> operation on Temporal.Instant objects creates a
+      Temporal.Duration like that of Time objects.
+    </li>
+    <li>
+      The <i>.difference()</i> operation on Temporal.DateTime objects creates a
+      Temporal.Duration that has all member properties, except if passed an
+      option indicating it should be created as a Time object or Date object.
+    </li>
+  </ul>
+  <p>
+    When created, the smaller units are combined into full larger units (i.e.
+    1000 milliseconds are recorded as 1 seconds) up to the largest available
+    unit. The result of this is that the largest available unit may have a
+    larger value in some circumstances. Durations created from Time objects may
+    have an hours property larger than 24, which would be invalid for a Duration
+    created from a DateTime object.
+  </p>
+  <p>
+    When a Duration is applied to an object via the <i>.plus()</i> method, only
+    the relevant fields are taken into consideration. Such that a Date objects
+    receiving a DateTime duration will entirely ignore the values of the hours
+    and below fields. As a consequence A Date object receiving a Duration from
+    the <i>.difference()</i> of Time objects will be a no-op, since it will not have
+    any relevant fields even in the case where the hours are larger than 24
+    which would equate.
+  </p>
+  <p>
+    All properties of Durations are positive integers. When coerced from an
+    object that looks like a Duration the following rules are applied:
+  </p>
+  <ol>
+    <li>All fields whose names match those of a Duration are extracted.</li>
+    <li>Each value is coerced:</li>
     <ul>
-        <li>The <i>.minus()</i> operation on Date objects creates a Duration that only has years, months, and days properties.</li>
-        <li>The <i>.minus()</i> operation on Time objects creates a Duration that only has hours, minutes, seconds, milliseconds, microseconds, and nanoseconds properties.</li>
-        <li>The <i>.minus()</i> operation on Instant objects creates a Duration like that of Time objects.</li>
-        <li>The <i>.minus()</i> operation on DateTime objects creates a Duration that has all member properties, except if passed an option indicating it should be created as a Time object or Date object.</li>
+      <li>If the field is numeric it is rounded to the nearest integer.</li>
+      <li>If the field is non-numeric it is set to undefined.</li>
     </ul>
+    <li>A Duration is created with the available values.</li>
+    <li>
+      If one of years, months, or days is numeric, the others are set to 0.
+    </li>
+    <li>
+      If one of hours, minutes, seconds, milliseconds, microseconds, or
+      nanoseconds is numeric, the others are set to 0.
+    </li>
+    <li>
+      The Duration is normalized through adding/subtracting values of adjacent
+      units.
+    </li>
+    <li>
+      If the resulting Duration has a negative value that value is converted to
+      its absolute value.
+    </li>
+  </ol> -->
+
+  <emu-clause id="sec-temporal-duration-constructor">
+    <h1>The Temporal.Duration Constructor</h1>
     <p>
-    When created, the smaller units are combined into full larger units (i.e. 1000 milliseconds are recorded as 1 seconds) up to the largest available unit.
-    The result of this is that the largest available unit may have a larger value in some circumstances. Durations created from Time objects may have an hours property larger than 24, which would be invalid for a Duration created from a DateTime object.
+      The Temporal.Duration constructor is the %Temporal.Duration% intrinsic
+      object. When called as a constructor, it creates and initializes a new
+      Temporal.Duration object.
     </p>
     <p>
-    When a Duration is applied to an object via the <i>.plus()</i> method, only the relevant fields are taken into consideration.
-    Such that a Date objects receiving a DateTime duration will entirely ignore the values of the hours and below fields. As a consequence
-    A Date object receiving a Duration from the <i>.minus()</i> of Time objects will be a no-op, since it will not have any relevant fields even in the case where the hours are larger than 24 which would equate.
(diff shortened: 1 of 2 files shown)


pipobscure:
note: .minus was changed to .difference

.plus & .minus are now two arithmetic operations

also duration is creatable and also castable (string, number, bigint)

ryzokuken:
@pipobscure I'm getting rid of stuff and following the polyfill, thanks for telling me about these.

spec/duration.html
```diff
@@ -2,44 +2,78 @@
...
+    created from a DateTime object.
+  </p>
+  <p>
+    When a Duration is applied to an object via the <i>.plus()</i> method, only
+    the relevant fields are taken into consideration. Such that a Date objects
+    receiving a DateTime duration will entirely ignore the values of the hours
+    and below fields. As a consequence A Date object receiving a Duration from
+    the <i>.minus()</i> of Time objects will be a no-op, since it will not have
+    any relevant fields even in the case where the hours are larger than 24
+    which would equate.
```
Yay295: Wouldn't it be easier and more consistent for all Duration objects to have the same fields? What's the rationale for it behaving differently?

spec/duration.html
```diff
@@ -2,44 +2,78 @@
...
+      option indicating it should be created as a Time object or Date object.
+    </li>
+  </ul>
+  <p>
+    When created, the smaller units are combined into full larger units (i.e.
+    1000 milliseconds are recorded as 1 seconds) up to the largest available
+    unit. The result of this is that the largest available unit may have a
+    larger value in some circumstances. Durations created from Time objects may
+    have an hours property larger than 24, which would be invalid for a Duration
+    created from a DateTime object.
```
Yay295: Wouldn't it be easier and more consistent for all Duration objects to have the same fields? What's the rationale for it behaving differently?

Yay295: Also, specifically on "Durations created from Time objects may have an hours property larger than 24", how? The difference between the earliest valid time and the latest valid time would be 24 hours, so this isn't even an issue.

Yay295: Also also, how does this work with days and months? How can you combine a number of days together to get a month?

pipobscure: If you take

duration = datetime.difference(datetime)

and then
time.plus(duration)

it should throw if the duration has more days,months,years

and for

date.plus(duration)

it should throw if it has hours,minutes,...

you can get a duration with hours > 24 by

absolute.difference(absolute)

in that case time.plus(duration) should wrap.

Yay295: Is Absolute an Instant?

pipobscure: Yes. But it only balances up to hours, Since days/months/years are variable length units. can be problematic

spec/duration.html
```diff
@@ -52,5 +86,81 @@ <h1>Duration Objects</h1>
...
+    <p>The Temporal.Duration constructor has the following properties:</p>
+    <emu-clause id="sec-temporal-date-prototype">
+      <h1>Temporal.Duration.prototype</h1>
+      <p>
+        The initial value of Temporal.Duration.prototype is
+        %Temporal.Duration.prototype%.
+      </p>
+      <p>
+        This property has the attributes { [[Writable]]: false, [[Enumerable]]:
+        false, [[Configurable]]: false }.
```
Ms2ger: Nit: `false` should be `*false*`.

spec/duration.html
```diff
@@ -52,5 +86,81 @@ <h1>Duration Objects</h1>
...
+      </p>
+    </emu-clause>
+    <emu-clause id="sec-temporal-date-prototype-fromstring">
+      <h1>Temporal.Duration.fromString(_isoString_)</h1>
```
Ms2ger:
```suggestion
      <h1>Temporal.Duration.fromString ( _isoString_ )</h1>
```

spec/duration.html
```diff
@@ -52,5 +86,81 @@ <h1>Duration Objects</h1>
...
+      <p>
+        The `fromString` method takes one argument _isoString_.
+        The following steps are taken:
+      </p>
+      <emu-alg>
+        1. Let _isoString_ be ? ToString(_isoString_)
```
Ms2ger:
```suggestion
        1. Let _isoString_ be ? ToString(_isoString_).
```

spec/duration.html
```diff
@@ -52,5 +86,81 @@ <h1>Duration Objects</h1>
...
+        1. Let _valid_ be *true* if _isoString_ conforms to the "duration" format specified in RFC 3339 (https://www.ietf.org/rfc/rfc3339.txt) Appendix A.
```
Ms2ger: …and false otherwise

spec/duration.html
```diff
@@ -52,5 +86,81 @@ <h1>Duration Objects</h1>
...
+        1. If _valid_ is *false*, then
+          1. Throw a *RangeError* exception.
+        1. Let _years_, _months_, _days_, _hours_, _minutes_, _seconds_, _milliseconds_, _microseconds_ and _nanoseconds_ be the respective parts of _isoString_ interpreted as decimal numbers.
+        1. Return ? CreateDuration (_years_, _months_, _days_, _hours_, _minutes_, _seconds_, _milliseconds_, _microseconds_, _nanoseconds_).
```
Ms2ger:
```suggestion
        1. Return ? CreateDuration(_years_, _months_, _days_, _hours_, _minutes_, _seconds_, _milliseconds_, _microseconds_, _nanoseconds_).
```

spec/duration.html
```diff
@@ -2,44 +2,78 @@
...
+    <li>
+      The <i>.minus()</i> operation on Temporal.Instant objects creates a
+      Temporal.Duration like that of Time objects.
+    </li>
+    <li>
+      The <i>.minus()</i> operation on Temporal.DateTime objects creates a
+      Temporal.Duration that has all member properties, except if passed an
+      option indicating it should be created as a Time object or Date object.
+    </li>
+  </ul>
```
Ms2ger: I'm not sure there is consensus around this. Maybe add a note about that?

spec/duration.html
```diff
@@ -2,44 +2,78 @@
...
-    </p>
+  <h1>Temporal.Duration Objects</h1>
+  <p>
+    A Temporal.Duration object describes the difference between two like Date,
+    Time, or DateTime objects. Objects of this type are only created via the
+    <i>.minus()</i> methods of these objects.
+  </p>
+  <ul>
+    <li>
+      The <i>.minus()</i> operation on Temporal.Date objects creates a
```
Ms2ger: `difference()`?

spec/duration.html
ryzokuken: These existed from before I started working on `Duration`, I kept it because I wasn't sure, but I'll add a note just to make sure.

spec/duration.html
```diff
@@ -52,5 +87,182 @@ <h1>Duration Objects</h1>
...
+        1. Let _h_ be ? ToInteger(_hours_).
+        1. Let _m_ be ? ToInteger(_minutes_).
+        1. Let _s_ be ? ToInteger(_seconds_).
+        1. Let _ms_ be ? ToInteger(_milliseconds_).
+        1. Let _mis_ be ? ToInteger(_microseconds_).
+        1. Let _ns_ be ? ToInteger(_nanoseconds_).
+        1. Let _bt_ be ? BalanceTime(_h_, _m_, _s_, _ms_, _mis_, _ns_).
+        1. Let _y_ be ? ToInteger(_years_).
+        1. Let _mo_ be ? ToInteger(_months_).
+        1. Let _d_ be ? ToInteger(_days_).
```
Ms2ger: Actually, another thought. I think we should do all the ToInteger calls before the BalanceTime call, and do them in argument order

spec/duration.html
```diff
@@ -52,5 +87,182 @@ <h1>Duration Objects</h1>
...
+          The `from` method takes one argument _arg_.
+          The following steps are taken:
+        </p>
+        <emu-alg>
+          1. If Type(_arg_) is String, then
+            1. Return ? DurationFromString(_arg_).
+          1. If Type(_arg_) is BigInt, then
+            1. Return ? CreateDuration(0, 0, 0, 0, 0, 0, 0, 0, _arg_).
+          1. If Type(_arg_) is Number, then
+            1. Return ? CreateDuration(0, 0, 0, 0, 0, 0, +_arg_, 0, 0).
```
Ms2ger: I think we should hold off on those. Overloading bigint and number is an antipattern anyway, and making them do something different entirely seems like a bad idea too.

spec/duration.html
```diff
@@ -52,5 +87,182 @@ <h1>Duration Objects</h1>
...
+            }.
+            1. For each row of <emu-xref href="#table-temporal-durationlike-properties"></emu-xref>, except the header row, in table order, do
+              1. Let _prop_ be the Property value of the current row.
+              1. Let _val_ be ? Get(_arg_, _prop_).
+              1. If _val_ is *undefined*, then
+                1. Let _val_ be *0*.
+              1. Else
+                1. Let _val_ be ? ToInteger(_val_).
+              1. Set _result_'s internal slot whose name is the Internal Slot value of the current row to _val_.
+            1. Return ? CreateDuration(_result_.[[Years]], _result_.[[Months]], _result_.[[Days]], _result_.[[Hours]], _result_.[[Minutes]], _result_.[[Seconds]], _result_.[[Milliseconds]], _result_.[[Microseconds]], _result_.[[Nanoseconds]])
```
Ms2ger:
```suggestion
            1. Return ? CreateDuration(_result_.[[Years]], _result_.[[Months]], _result_.[[Days]], _result_.[[Hours]], _result_.[[Minutes]], _result_.[[Seconds]], _result_.[[Milliseconds]], _result_.[[Microseconds]], _result_.[[Nanoseconds]]).
```

spec/duration.html
```diff
@@ -52,5 +87,182 @@ <h1>Duration Objects</h1>
...
+      <table class="real-table">
+        <thead>
+          <tr>
+            <th>Internal Slot</th>
+            <th>Property</th>
+          </tr>
+        </thead>
+        <tbody>
+          <tr>
+            <td>[[Years]]</td>
```
Ms2ger: Sort these alphabetically, please

spec/duration.html
ryzokuken: Just following the polyfill here. Maybe @pipobscure can explain his reasoning?

spec/duration.html
ryzokuken: Sure!

spec/duration.html
ryzokuken: Oops, sorry.

spec/duration.html
ryzokuken: Okay.

ryzokuken:
Apart from #195, this feels good to go from my side. @pipobscure @littledan @Ms2ger care to take a quick look?

ryzokuken:
Merging on @Ms2ger's approval.

PR #185, opened 2019-10-13
Fix minor bugs around Date.

Ms2ger:


spec/abstractops.html
```diff
@@ -151,7 +151,7 @@ <h1>CreateMonthDay ( _month_, _day_ )</h1>
   </emu-clause>
 
   <emu-clause id=sec-temporal-CreateDateTime aoid=CreateDateTime>
-    <h1>CreateDateTime ( …args )</h1>
+    <h1>CreateDateTime ( …_args_ )</h1>
```
ljharb: this should use three dots and not an ellipsis; but also eventually it should presumably be explicit

PR #187, opened 2019-10-13
Specify Temporal.Time.

Ms2ger:


spec/abstractops.html
```diff
@@ -129,6 +129,68 @@ <h1>ToDurationLike ( _duration_ )</h1>
     </emu-table>
   </emu-clause>
 
+  <emu-clause id=sec-temporal-ToTimeDurationLike aoid=ToTimeDurationLike>
+    <h1>ToTimeDurationLike ( _duration_ )</h1>
+    <emu-alg>
+      1. Let _result_ be {
```
ljharb: is this meant to be a Record, or an object?

spec/abstractops.html
```diff
@@ -157,10 +219,25 @@ <h1>CreateDateTime ( …args )</h1>
...
-    <h1>ToTimeLike ( _time_ )</h1>
-    <p>
-      <mark>TODO: Define.</mark>
-    </p>
+  <emu-clause id=sec-temporal-FormatSecondsStringPart aoid=FormatSecondsStringPart>
+    <h1>FormatSecondsStringPart ( _second_, _millisecond_, _microsecond_, _nanosecond_ )</h1>
+    <emu-alg>
+      1. If _second_, _millisecond_, _microsecond_, and _nanosecond_ are all 0, then
+        1. Return `""`.
+      1. Let _nanos_, _micros_, and _millis_ be `""`.
```
ljharb: Do we typically set more than one variable at a time like this?

spec/abstractops.html
```diff
@@ -157,10 +219,25 @@ <h1>CreateDateTime ( …args )</h1>
...
+      1. If _nanosecond_ is not 0, then
+        1. Set _nanos_ to _nanosecond_ formatted as a three-digit decimal number, padded to the left with zeroes if necessary.
```
ljharb: i believe these can use https://tc39.es/ecma262/#sec-stringpad instead of prose

spec/civiltime.html
```diff
@@ -1,230 +1,558 @@
...
+          1. Let _nanosecond_ be _timelike_.[[Nanosecond]].
+        1. Else
+          1. Let _nanosecond_ be _time_.[[Nanosecond]].
+        1. If ! ValidateTime(_hour_, _minute_, _second_, _millisecond_, _microsecond_, _nanosecond_) is *false*, then
+          1. Throw a *TypeError* exception.
+        1. If _disambiguation_ is not *undefined*, then
+          1. Set _disambiguation_ to ? ToString(_disambiguation_).
+        1. Else
+          1. Set _disambiguation_ to `"constrain"`.
+        1. Let _result_ be ? RegulateTime(_hour_, _minute_, _second_, _millisecond_, _microsecond_, _nanosecond_, _disambiguation_).
```
ljharb: this seems to be repeating a lot of the steps from the constructor. Can they be consolidated?

spec/civiltime.html
```diff
@@ -1,230 +1,558 @@
...
+
+    <emu-clause>
+      <h1>Temporal.Time.prototype.difference ( _otherTime_ )</h1>
+      <p>
+        The `difference` method takes one argument _otherTime_.
+        The following steps are taken:
+      </p>
+      <emu-alg>
+        1. Let _time_ be the *this* value.
+        1. If Type(_time_) is not Object or _time_ does not have an [[InitializedTemporalTime]] internal slot, then
```
ljharb: this “if” and throw is repeated many times. Perhaps like RequireObjectCoercible, it’d  be helpful to have a RequireTime abstract op?

spec/civiltime.html
```diff
@@ -1,230 +1,558 @@
...
+          1. Throw a *TypeError* exception.
+        1. Let _otherTime_ be ? ToTime(_otherTime_).
```
ljharb: why is the receiver brand checked, but the argument coerced?

spec/civiltime.html
```diff
@@ -1,230 +1,558 @@
...
+    <emu-clause>
+      <h1>Temporal.Time.prototype.toLocaleString ( …_args_ )</h1>
+      <p>
+        The following steps are taken:
+      </p>
+      <emu-alg>
+        1. Let _time_ be the *this* value.
+        1. If Type(_time_) is not Object or _time_ does not have an [[InitializedTemporalTime]] internal slot, then
+          1. Throw a *TypeError* exception.
+        1. Let _dateFormat_ be ? Construct(%DateTimeFormat%, _args_).
```
ljharb: is this intrinsic from intl? Intl is not necessarily available to 262 engines.

spec/abstractops.html
Ms2ger: I'm not sure. I can make it more verbose if you prefer.

spec/abstractops.html
Ms2ger: StringPad does a bunch of things that can throw exceptions, so I'd kinda prefer not to. Maybe we can introduce a separate abstract operation and also use it in the dozen or so places that ecma262 uses this wording.

spec/civiltime.html
Ms2ger: Not all that much, actually. I think in this case it's better to leave the argument coercion to the method itself rather than trying to extract something here.

spec/civiltime.html
Ms2ger: I can do that if people agree it makes sense.

spec/civiltime.html
Ms2ger: The receiver is checked for consistency with the rest of the spec and other new APIs. It also doesn't add much value to support it; it just means you'd be able to do
```js
Temporal.Time.prototype.compare.call({...}, [otherDate]);
```
rather than
```js
Temporal.Time.from({...}).compare(otherDate);
```
It's rather more helpful to allow passing `{...}` rather than `Temporal.Time.from({...})` as an argument.

However, I'd be fine with requiring actual branded objects as well.

spec/civiltime.html
Ms2ger: Yes. We can copy some wording from 262 to make this sorta optional.

spec/abstractops.html
Ms2ger: Record.

spec/abstractops.html
ljharb: None of those exceptions are observable with primitive string inputs, with a length of 3 - I'm not sure why we'd want to make a different one.

Ms2ger: Will wait for a response on https://github.com/tc39/ecma262/issues/1737

spec/civiltime.html
Ms2ger: https://github.com/tc39/proposal-temporal/issues/207

pipobscure:
I'm merging this for progress

ljharb:
Note that merging things makes review harder, because there’s no interface for suggestions etc short of a PR.

Ms2ger:
You could also file issues.

ljharb:
Indeed, and that’s all that’s left to me after a flurry of merges, but that is much less ergonomic than posting comments on PRs and staying up to date on a large, complex, and fast-changing spec by understanding master, and also understanding each significant PR. The current approach means i have to reset my entire understanding every couple weeks and start over from scratch by rereading all of master.

Ms2ger:
My recommendation would be to wait a week or two; at that point there should a mostly stable spec to review.

ljharb:
Sure, but I’d hoped to not have to review much of anything then because I’d kept up on all the changes; as it is, I’ll have to do all that review from scratch.

PR #194, opened 2019-10-15
Define Date.from().

Ms2ger:


spec/civildate.html
```diff
@@ -466,28 +461,48 @@ <h1>CreateDate ( _y_, _m_, _d_ [ , _newTarget_ ] )</h1>
...
+    <emu-clause id=sec-temporal-DateFromString aoid=DateFromString>
+      <h1>DateFromString ( _date_ )</h1>
+      <emu-alg>
+        1. Assert: Type(_date_) is String.
+        1. Let _validSyntax_ be *true* if _isostring_ conforms to the format `YYYY-MM-DD`, and *false* otherwise, where:
+          * `YYYY` is either four decimal digits or an expanded year of `"+"` or `"-"` followed by six decimal digits;
+          * `-` is the code unit 0x002D (HYPHEN-MINUS);
+          * `MM` is two decimal digits;
+          * `-` is the code unit 0x002D (HYPHEN-MINUS);
+          * `DD` is two decimal digits.
```
ryzokuken: While it's more doable to define this format right here for `Date` as opposed to `Duration`, should we point this to ISO 8601/RFC 3339 for the sake of consistency?

Ms2ger: That would be fine with me as well.

pipobscure: Dangerous, since this only supports a subset of ISO8601. We were thinking as describing the string parsing as “the date part of a string of the format produced by Absolute, Date, DateTime”

ryzokuken: @pipobscure fair enough, but how do we plan to specify the format for the string produced by `Absolute`? Should we link to ISO or specify the format ourselves? What are the benefits here of specifying ourselves since the ISO standard is pretty good for our purpose and doesn't seem to be going anywhere anytime soon?

pipobscure: @littledan : agreed.

The subset for date-parts is: we only support the regular version `YYYY-MM-DD` and not the ordinal (`YYYY-DDD`) or weekdate (`YYYY-Www-D`) versions and we always require the hyphen (`-`) which is sometimes options in ISO8601

The subset for time-parts is: we always require the colon (`:`) separator. In adddition the precision for fractional seconds is higher (nanoseconds). Furthermore the timezone can be denoted as `+HH:MM` or with an IANA name `+01:00[Europe/Berlin]` as well as the default UTC (`Z`)

The change for durations is: we only support the `P(n)Y(n)M(n)DT(n)H(n)M(n)S` version. The hypenated (` P[YYYY]-[MM]-[DD]T[hh]:[mm]:[ss]`) version is not supported. The numeric value for seconds in the string is not restricted to an integer, but can contain a fractional part which is translated to milliseconds, microseconds and nanoseconds.

These go both for parsing, as well as for `toString()` methods of any kind. (While I think we should leave it open for parsers to accept more iso formats, the stringifiers should always use these formats)

Which leaves me with the question whether we want to make supporting parsing of these additional ISO8601 formats mandatory/optional/invalid.

ljharb: Please either mandatory or invalid, not optional :-)

Ms2ger: I've filed https://github.com/tc39/proposal-temporal/issues/198 for this. Let's not block this PR on that, though.

ryzokuken: @Ms2ger I never suggested blocking it anyway, just wanted to make it consistent and wanted to know what you thought was the best way to do it. Marking as resolved :)

PR #204, opened 2019-10-26
polyfill.intl: Correct typo 'ammend'

apaprocki:


polyfill/lib/intl.mjs
```diff
@@ -121,19 +121,19 @@ function timeAmmend(options) {
...
       month: 'numeric',
       day: 'numeric'
     });
   }
-  options = ammend(options, ammendments);
+  options = amend(options, amendments);
   return options;
 }
-function datetimeAmmend(options) {
+function datetimeAmend(options) {
```
ryzokuken: Unrelated to this change, but should this be `dateTimeAmend`?

PR #208, opened 2019-10-28
SubClassing should adhere to TC39 design practice

### LINKED ISSUE #196: Review subclassing support
In #195, @littledan suggested that `Duration.fromNanoseconds` and friends should be able to return subclasses of `Duration`. This seems somewhat reasonable, but we should consider what happens in other subclassing cases. For example, what happens to `Time.withDate`? Is there any way that can return a subclass of `DateTime`? How? If not, do we want to support that by replacing it with `DateTime.from`?

pipobscure:
Resolves #196


polyfill/lib/ecmascript.mjs
```diff
@@ -31,7 +31,20 @@ export const ES = ObjectAssign(ObjectAssign(ObjectAssign({}, Cast), ES2019), {
...
+    let any = !anyof.length;
+    for (let prop of anyof) {
+      const value = ES.ToNumber(bag[prop]);
+      any = any || Number.isFinite(value);
+    }
+    let none = false;
+    for (let prop of noneof) {
+      const value = ES.ToNumber(bag[prop]);
+      none = none || (Number.isFinite(value) && !!value);
+    }
```
Ms2ger: I think the entire concept of "this object is not allowed to have these keys" is an anti-pattern. Does it exist anywhere else in the platform?

It's especially weird because you're allowing `{foo: NaN}` and `{foo: 0}` even when `foo` is forbidden.

polyfill/lib/date.mjs
```diff
@@ -53,24 +53,29 @@ export class Date {
...
+    }
     const { year = GetSlot(this, YEAR), month = GetSlot(this, MONTH), day = GetSlot(this, DAY) } = dateLike;
     return new Date(year, month, day, disambiguation);
   }
   plus(durationLike = {}, disambiguation = 'constrain') {
     const duration = ES.CastDuration(durationLike);
-    let { year, month, day } = this;
-    let { years, months, days, hours, minutes, seconds, milliseconds, microseconds, nanoseconds } = duration;
-    if (hours || minutes || seconds || milliseconds || microseconds || nanoseconds)
+    if (!ES.ValidPropertyBag(duration, [], [ 'hours', 'minutes', 'seconds', 'milliseconds', 'microseconds', 'nanoseconds' ])) {
```
Ms2ger: I don't understand the value of this check. In any case, it should use the internal slots, not the (possibly tampered-with) getters.

polyfill/lib/monthday.mjs
```diff
@@ -33,45 +33,46 @@ export class MonthDay {
...
-    )
+    if (!ES.ValidPropertyBag(duration, [], [ 'years', 'hours', 'minutes', 'seconds', 'milliseconds', 'microseconds', 'nanoseconds' ])) {
       throw new RangeError('invalid duration');
-    ({ year, month, day } = ES.AddDate(year, month, day, years, months, days, disambiguation));
-    ({ month, day } = ES.BalanceDate(1970, month, day));
+    }
+    let { month, day } = this;
+    const { months, days } = duration;
+    const year = 1970; // non-leap year
+    ({ month, day } = ES.AddDate(year, month, day, years, months, days, disambiguation));
```
Ms2ger: Where does `years` come from?

polyfill/lib/monthday.mjs
```diff
@@ -33,45 +33,46 @@ export class MonthDay {
...
+    ({ month, day } = ES.SubtractDate(year, month, day, years, months, days, disambiguation));
+    ({ month, day } = ES.BalanceDate(year, month, day));
     return new MonthDay(month, day);
   }
-  difference(other, disambiguation = 'constrain') {
+  difference(other) {
     other = ES.CastMonthDay(other);
     const [one, two] = [this, other].sort(MonthDay.compare);
     let months = two.month - one.month;
     let days = (two.days = one.days);
```
Ms2ger: Pretty sure there should be more than one `=` here.

polyfill/lib/date.mjs
pipobscure: The reason is that `date.plus({ seconds: 60 })` does not make sense. So the duration you pass in is not allowed to have a positive `seconds` value.

My mistake in this PR (**will change**): I combined the duration checking and the validity checking into 1 operation.

polyfill/lib/monthday.mjs
pipobscure: Where?

polyfill/lib/date.mjs
```diff
@@ -53,24 +53,29 @@ export class Date {
     return ES.LeapYear(GetSlot(this, YEAR));
   }
   with(dateLike = {}, disambiguation = 'constrain') {
+    if (!ES.ValidPropertyBag(dateLike, [ 'year', 'month', 'day' ])) {
```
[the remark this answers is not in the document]
pipobscure: Then `!ES.ValidPropertyBag(`

polyfill/lib/monthday.mjs
pipobscure: Nowhere. Thanks for the catch.

polyfill/lib/monthday.mjs
pipobscure: Ah see it. It should be `-` rather than `=`

pipobscure:
@ms2ger : I believe to have addressed your remarks by changing the code. Thanks for the catches!

polyfill/lib/ecmascript.mjs
```diff
@@ -31,7 +31,31 @@ export const ES = ObjectAssign(ObjectAssign(ObjectAssign({}, Cast), ES2019), {
...
   },
-
+  ValidPropertyBag: (bag, anyof = []) => {
+    if (!bag || 'object' !== typeof bag) return false;
+    let any;
+    for (let prop of anyof) {
+      if (prop in bag) {
+        const value = ES.ToNumber(bag[prop]);
+        if (Number.isFinite(value)) {
+          any = any || {};
```
Ms2ger:
```suggestion
          any = any || Object.create(null);
```

PR #210, opened 2019-10-28
built-in module -> global object

hax:
GitHub also auto change the encoding from UTF-16LE to UTF-8.


hax:
@ljharb  See my original comment. 

And u could use rich diff to see the changes.

ljharb:
utf-8 isn’t a binary file, and the diffs would show.

ljharb:
I do see that on the desktop site, they show up, but not the mobile site.

ljharb:
Can that be reverted? It shouldn’t be necessary to make a markdown file “binary” on github to make prose changes.

Ms2ger:
Please rebase on main; I fixed the encoding there.

hax:
it seems no way to rebase in github... i just close this and will create new pr.

PR #215, opened 2019-10-28
Add duration.toLocaleString

### LINKED ISSUE #170: Add toLocaleString for duration
ICU has the ability to format durations, but this feature is not fully supported in ECMA-402.  This does not need to be a Stage 3 blocker, but I would like to see the Intl proposal move quickly.  Upstream issue: https://github.com/tc39/ecma402/issues/47

pipobscure:
Resolves #170


polyfill/lib/duration.mjs
```diff
@@ -130,6 +130,9 @@ export class Duration {
     if (!dateParts.length && !timeParts.length) return 'PT0S';
     return `P${dateParts.join('')}${timeParts.join('')}`;
   }
+  toLocaleString(...args) {
+    return new Intl.DateTimeFormat(...args).format(this);
```
ljharb: shouldn’t this brand check the receiver first?

ljharb: @pipobscure ?

Ms2ger: #224

Issue #207, opened 2019-10-28
Consider using abstract operations for brand checks


Ms2ger:
See https://github.com/tc39/proposal-temporal/pull/187#discussion_r334302663

ljharb:
Turns out `RequireInternalSlot` already exists, so that's what should be used here: https://tc39.es/ecma262/#sec-requireinternalslot

PR #228, opened 2019-10-29
Add YearMonth spec.

Ms2ger:


polyfill/lib/casts.mjs
```diff
@@ -143,14 +143,11 @@ export function CastTime(arg, aux) {
 
 export function CastYearMonth(arg, aux) {
   const YearMonth = ES.GetIntrinsic('%Temporal.YearMonth%');
-  if (HasSlot(arg, YEAR, MONTH)) {
-    if (!HasSlot(arg, DAY)) return arg;
-    return new YearMonth(GetSlot(arg, year), GetSlot(arg, MONTH));
+  if (HasSlot(arg, YEAR, MONTH) && !HasSlot(arg, DAY)) {
+    return arg;
   }
   if ('string' === typeof arg) {
```
ljharb: nbd but why the yoda condition?

Ms2ger: @pipobscure

pipobscure: I don't understand the question: because it allows `TemporalObject.from(isostring)` to be valid. This is very useful for making `.from` be the universal "try hard to cast this" operation.

littledan: I think this was a very minor stylistic question, and we don't need to follow up further.

spec/yearmonth.html
```diff
@@ -3,5 +3,511 @@
...
+      <h1>get Temporal.YearMonth.prototype.daysInMonth</h1>
+      <p>
+        Temporal.YearMonth.prototype.daysInMonth is an accessor property whose set accessor function is undefined.
+        Its get accessor function performs the following steps:
+      </p>
+      <emu-alg>
+        1. Let _yearMonth_ be the *this* value.
+        1. If Type(_yearMonth_) is not Object or _yearMonth_ does not have an [[InitializedTemporalYearMonth]] internal slot, then
+          1. Throw a *TypeError* exception.
+        1. Return DaysInMonth(_yearMonth_.[[Year]], _yearMonth_.[[Month]]).
```
ryzokuken:
```suggestion
        1. Return ? DaysInMonth(_yearMonth_.[[Year]], _yearMonth_.[[Month]]).
```

spec/yearmonth.html
```diff
@@ -3,5 +3,511 @@
...
+        1. Else
+          1. Set _disambiguation_ to `"constrain"`.
+        1. If _otherYearMonth_.[[Year]] is not *undefined*, then
+          1. Let _y_ be _otherYearMonth_.[[Year]].
+        1. Else
+          1. Let _y_ be _yearMonth_.[[Year]].
+        1. If _otherYearMonth_.[[Month]] is not *undefined*, then
+          1. Let _m_ be _otherYearMonth_.[[Month]].
+        1. Else
+          1. Let _m_ be _yearMonth_.[[Month]].
```
ryzokuken: Why can't this be a single condition?

spec/yearmonth.html
```diff
@@ -3,5 +3,511 @@
...
+        1. Let _m_ be _yearMonth_.[[Month]] - _duration_.[[Months]].
+        1. Let _result_ be ? RegulateYearMonth(_y_, _m_, _disambiguation_).
+        1. Return ? CreateYearMonth(_result_.[[Year]], _result_.[[Month]]).
+      </emu-alg>
+    </emu-clause>
+
+    <emu-clause>
+      <h1>Temporal.YearMonth.prototype.difference ( _otherYearMonth_ )</h1>
+      <p>
+        The `difference` method takes two arguments, _otherYearMonth_ and _disambiguation_.
```
ryzokuken:
```suggestion
        The `difference` method takes one argument, _otherYearMonth_.
```

spec/yearmonth.html
```diff
@@ -3,5 +3,511 @@
...
+        1. Else,
+          1. Let _greater_ be _yearMonth_.
+          1. Let _smaller_ be _otherYearMonth_.
+        1. Let _years_ be _greater_.[[Year]] - _smaller_.[[Year]].
+        1. Let _months_ be _greater_.[[Month]] - _smaller_.[[Month]].
+        1. If _months_ &lt; 0, then
+          1. Set _years_ to _years_ - 1.
+          1. Set _months_ to _months_ + 12.
+        1. <mark>Assert: _months_ &ge; 0.</mark>
+        1. Return ? CreateDuration(_years_, _months_, 0).
```
ryzokuken:
```suggestion
        1. Return ? CreateDuration(_years_, _months_).
```

polyfill/lib/casts.mjs
ljharb: It was stylistic and minor; to clarify, I’m asking why “string equals arg” instead of the more common “arg equals string”. Colloquially, putting the variable last is called a “yoda condition”.

pipobscure:
Should we merge this?

spec/yearmonth.html
```diff
@@ -3,5 +3,511 @@
...
+              <td>`"year"`</td>
+            </tr>
+          </tbody>
+        </table>
+      </emu-table>
+    </emu-clause>
+    <emu-clause id=sec-temporal-RegulateYearMonth aoid=RegulateYearMonth>
+      <h1>RegulateYearMonth ( _year_, _month_, _disambiguation_ )</h1>
+      <emu-alg>
+        1. Assert: _disambiguation_ is one of `"constrain"`, `"balance"`, or `"reject"`.
```
ljharb: this assertion is invalid - it's called with an arbitrary string value supplied by the user.

It seems like what you want is an assertion that it's a string, and then to throw a TypeError if it's not one of the allowed values.

spec/yearmonth.html
```diff
@@ -3,5 +3,511 @@
...
+            <tr>
+              <td>[[Year]]</td>
+              <td>`"year"`</td>
+            </tr>
+          </tbody>
+        </table>
+      </emu-table>
+    </emu-clause>
+    <emu-clause id=sec-temporal-RegulateYearMonth aoid=RegulateYearMonth>
+      <h1>RegulateYearMonth ( _year_, _month_, _disambiguation_ )</h1>
```
ljharb: this is missing assertions that year and month are integers.

spec/yearmonth.html
```diff
@@ -3,5 +3,511 @@
...
+      <emu-alg>
+        1. Assert: Type(_yearMonth_) is String.
+        1. Let _valid_ be *true* if _yearMonth_ matches <a href="https://github.com/tc39/proposal-temporal/blob/main/polyfill/lib/regex.mjs">the "yearmonth" regular expression in the polyfill</a> and *false* otherwise.
+        1. <mark>TODO: Actually specify this somehow.</mark>
+        1. If _valid_ is *false*, then
+          1. Throw a *RangeError* exception.
+        1. Let _year_ and _month_ be the respective parts of _yearMonth_ interpreted as decimal numbers.
+        1. Return the new Record {
+          [[Year]]: _year_,
+          [[Month]]: _month_,
```
ljharb: all these multiline records seem indented incorrectly; the slots should get an additional level

spec/yearmonth.html
```diff
@@ -3,5 +3,511 @@
...
+          }.
+      </emu-alg>
+    </emu-clause>
+
+    <emu-clause id="sec-temporal-YearMonthFromString" aoid="YearMonthFromString">
+      <h1>YearMonthFromString ( _yearMonth_ )</h1>
+      <emu-alg>
+        1. Assert: Type(_yearMonth_) is String.
+        1. Let _result_ be ParseYearMonthString(_yearMonth_).
```
ljharb:
```suggestion
        1. Let _result_ be ? ParseYearMonthString(_yearMonth_).
```

spec/yearmonth.html
```diff
@@ -3,5 +3,511 @@
...
+        1. Let _month_ be _yearMonth_.[[Month]] formatted as a two-digit decimal number, padded to the left with a zero if necessary.
+        1. Return the string-concatenation of _year_, the code unit 0x002D (HYPHEN-MINUS), and _month_.
+      </emu-alg>
+    </emu-clause>
+
+    <emu-clause id=sec-temporal-YearMonthCompare aoid=YearMonthCompare>
+      <h1>YearMonthCompare ( _one_, _two_ )</h1>
+      <emu-alg>
+        1. Assert: Type(_one_) is Object.
+        1. Assert: _one_ has an [[InitializedTemporalYearMonth]] internal slot.
```
ljharb: it seems like we should have an `IsYearMonth` abstract op, rather than constantly needing to assert that it's both an object and has the right internal slots.

spec/yearmonth.html
```diff
@@ -3,5 +3,511 @@
...
+
+    <emu-clause>
+      <h1>Temporal.YearMonth.prototype.difference ( _otherYearMonth_ )</h1>
+      <p>
+        The `difference` method takes one argument, _otherYearMonth_.
+        The following steps are taken:
+      </p>
+      <emu-alg>
+        1. Let _yearMonth_ be the *this* value.
+        1. If Type(_yearMonth_) is not Object or _yearMonth_ does not have an [[InitializedTemporalYearMonth]] internal slot, then
```
ljharb: IsYearMonth could be used here

spec/yearmonth.html
```diff
@@ -3,5 +3,511 @@
...
+          1. Throw a *TypeError* exception.
+        1. Let _otherYearMonth_ be ? ToYearMonth(_otherYearMonth_).
+        1. If YearMonthCompare(_yearMonth_, _otherYearMonth_) &lt; 0, then
```
ljharb:
```suggestion
        1. If ! YearMonthCompare(_yearMonth_, _otherYearMonth_) &lt; 0, then
```

spec/yearmonth.html
Ms2ger: I'll handle this separately across the entire spec.

spec/yearmonth.html
Ms2ger: See https://github.com/tc39/proposal-temporal/issues/207.

PR #247, opened 2019-10-31
Add DateTime spec.

Ms2ger:


spec/date.html
```diff
@@ -597,5 +592,17 @@ <h1>PadYear ( _y_ )</h1>
         1. Return the string-concatenation of _yearSign_ and _year_.
       </emu-alg>
     </emu-clause>
+
+    <emu-clause id=sec-temporal-AddDate aoid=AddDate>
+      <h1>AddDate ( _year_, _month_, _day_, _years_, _months_, _days_, _disambiguation_ )</h1>
```
ljharb: This abstract op should add assertions for all its argument types.

spec/time.html
```diff
@@ -649,5 +644,18 @@ <h1>TimeCompare ( _one_, _two_ )</h1>
         1. Return +0.
       </emu-alg>
     </emu-clause>
+
+    <emu-clause id=sec-temporal-AddTime aoid=AddTime>
+      <h1>AddTime ( _hour_, _minute_, _second_, _millisecond_, _microsecond_, _nanosecond_, _hours_, _minutes_, _seconds_, _milliseconds_, _microseconds_, _nanoseconds_ )</h1>
```
ljharb: This also needs argument type assertions.

spec/datetime.html
```diff
@@ -1,352 +1,824 @@
...
+    <emu-clause>
+      <h1>Temporal.DateTime.compare ( _one_, _two_ )</h1>
+      <p>
+        The `compare` method takes two arguments, _one_ and _two_.
+        The following steps are taken:
+      </p>
+      <emu-alg>
+        1. Let _one_ be ? ToDateTime(_one_).
+        1. Let _two_ be ? ToDateTime(_two_).
+        1. Return DateTimeCompare(_one_, _two_).
```
ljharb: ? or ! prefix?

spec/datetime.html
```diff
@@ -1,352 +1,824 @@
...
+      <h1>get Temporal.DateTime.prototype.daysInMonth</h1>
+      <p>
+        Temporal.DateTime.prototype.daysInMonth is an accessor property whose set accessor function is undefined.
+        Its get accessor function performs the following steps:
+      </p>
+      <emu-alg>
+        1. Let _dateTime_ be the *this* value.
+        1. If Type(_dateTime_) is not Object or _dateTime_ does not have an [[InitializedTemporalDateTime]] internal slot, then
+          1. Throw a *TypeError* exception.
+        1. Return DaysInMonth(_dateTime_.[[Year]], _dateTime_.[[Month]]).
```
ljharb: ? or ! prefix?

ryzokuken:
@Ms2ger in light of https://github.com/tc39/proposal-temporal/pull/248, could you please rename `DateTime.prototype.inZone` to `DateTime.prototype.inTimeZone`? Thanks.

spec/datetime.html
```diff
@@ -1,352 +1,824 @@
...
+    <emu-clause>
+      <h1>Temporal.DateTime.prototype.toLocaleString ( [ _locales_ [ , _options_ ] ] )</h1>
+      <p>
+        The following steps are taken:
+      </p>
+      <emu-alg>
+        1. Let _dateTime_ be the *this* value.
+        1. If Type(_dateTime_) is not Object or _dateTime_ does not have an [[InitializedTemporalDateTime]] internal slot, then
+          1. Throw a *TypeError* exception.
+        1. If the implementation does not include the ECMA-402 Internationalization API, then
```
ljharb: I don’t think this is a realistic mechanism; instead, the fallback logic should be defined as an abstract operation, and 402 should override/replace that abstract operation.

spec/datetime.html
```diff
@@ -1,352 +1,824 @@
...
+    <h1>Abstract operations</h1>
+
+    <emu-clause id=sec-temporal-ToPartialDateTime aoid=ToPartialDateTime>
+      <h1>ToPartialDateTime ( _dateTime_ )</h1>
+      <emu-alg>
+        1. If Type(_dateTime_) is String, then
+          1. Return ? ParseDateTimeString(_dateTime_).
+        1. If Type(_dateTime_) is Object, then
+          1. Let _result_ be the new Record {
+            [[Year]]: *undefined*,
```
ljharb:
```suggestion
              [[Year]]: *undefined*,
```

spec/datetime.html
```diff
@@ -1,352 +1,824 @@
...
+            [[Nanosecond]]: *undefined*,
+            }.
+          1. Let _any_ be *false*.
+          1. For each row of <emu-xref href="#table-temporal-datetimelike-properties"></emu-xref>, except the header row, in table order, do
+            1. Let _property_ be the Property value of the current row.
+            1. Let _value_ be ? Get(_dateTime_, _property_).
+            1. If _value_ is not *undefined*, then
+              1. Set _any_ to *true*.
+              1. Set _value_ to ? ToInteger(_value_).
+              1. Set _result_'s internal slot whose name is the Internal Slot value of the current row to _value_.
```
ljharb: If there’s going to be a lot of occurrences of this magic meta prose looking up things from the table, let’s please find a way to consolidate it to a single abstract operation (fine to file a followup for this)

spec/datetime.html
```diff
@@ -1,352 +1,824 @@
...
+          1. If _any_ is *false*, then
+            1. Throw a *TypeError* exception.
+          1. Return _result_.
+        1. Throw a *TypeError* exception.
+      </emu-alg>
+
+      <emu-table id="table-temporal-datetimelike-properties">
+        <emu-caption>Properties of a DateTimeLike</emu-caption>
```
ljharb: This caption doesn’t seem quite accurate

spec/datetime.html
```diff
@@ -1,352 +1,824 @@
...
+          1. Return *false*.
+        1. Return *true*.
+      </emu-alg>
+    </emu-clause>
+
+    <emu-clause id=sec-temporal-createdatetime aoid=CreateDateTime>
+      <h1>CreateDateTime ( _year_, _month_, _date_, _hour_, _minute_, _second_, _millisecond_, _microsecond_, _nanosecond_ [, _newTarget_] )</h1>
+      <emu-alg>
+        1. Assert: ! ValidateDateTime(_year_, _month_, _date_, _hour_, _minute_, _second_, _millisecond_, _microsecond_, _nanosecond_) is *true*.
+        1. If _newTarget_ is not given, set it to %Temporal.DateTime%
```
ljharb:
```suggestion
        1. If _newTarget_ is not given, set it to %Temporal.DateTime%.
```

spec/datetime.html
```diff
@@ -1,352 +1,824 @@
...
+        1. Let _object_ be ? OrdinaryCreateFromConstructor(_newTarget_, `"%Temporal.DateTime.prototype%"`, « [[InitializedTemporalDateTime]], [[Year]], [[Month]], [[Day]], [[Hour]], [[Minute]], [[Second]], [[Millisecond]], [[Microsecond]], [[Nanosecond]] »).
```
ljharb:
```suggestion
        1. Let _object_ be ? OrdinaryCreateFromConstructor(_newTarget_, `%Temporal.DateTime.prototype%`, « [[InitializedTemporalDateTime]], [[Year]], [[Month]], [[Day]], [[Hour]], [[Minute]], [[Second]], [[Millisecond]], [[Microsecond]], [[Nanosecond]] »).
```

spec/datetime.html
```diff
@@ -1,352 +1,824 @@
...
+      <h1>ParseDateTimeString ( _dateTime_ )</h1>
+      <emu-alg>
+        1. Assert: Type(_dateTime_) is String.
+        1. Let _valid_ be *true* if _dateTime_ matches <a href="https://github.com/tc39/proposal-temporal/blob/main/polyfill/lib/regex.mjs">the "datetime" regular expression in the polyfill</a> and *false* otherwise.
+        1. <mark>TODO: Actually specify this somehow.</mark>
+        1. If _valid_ is *false*, then
+          1. Throw a *RangeError* exception.
+        1. Let _year_, _month_, _day_, _hour_, _minute_, _second_, _millisecond_, _microsecond_, _nanosecond_ be the respective parts of _dateTime_ interpreted as decimal numbers.
+        1. Return the new Record {
+          [[Year]]: _year_,
```
ljharb:
```suggestion
            [[Year]]: _year_,
```

spec/datetime.html
```diff
@@ -1,352 +1,824 @@
...
+    <emu-clause id=sec-temporal-ToDateTime aoid=ToDateTime>
+      <h1>ToDateTime ( _dateTime_ )</h1>
+      <emu-alg>
+        1. If Type(_dateTime_) is String, then
+          1. Return ? DateTimeFromString(_dateTime_).
+        1. If Type(_dateTime_) is Object, then
+          1. If _dateTime_ has an [[InitializedTemporalDateTime]] internal slot, then
+            1. Return _dateTime_.
+          1. Let _result_ be the new Record {
+            [[Year]]: *undefined*,
```
ljharb:
```suggestion
              [[Year]]: *undefined*,
```

polyfill/lib/casts.mjs
```diff
@@ -44,37 +44,13 @@ export function CastAbsolute(arg, aux) {
 }
 
 export function CastDateTime(arg, aux) {
+  if ('string' === typeof arg) {
+    return DateTime.fromString(arg);
+  }
   const DateTime = ES.GetIntrinsic('%Temporal.DateTime%');
-  if (HasSlot(arg, YEAR, MONTH, DAY, HOUR, MINUTE, SECOND, MILLISECOND, MICROSECOND, NANOSECOND)) {
+  if (ES.IsDateTime(arg)) {
```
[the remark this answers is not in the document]
Ms2ger: #232

spec/datetime.html
```diff
@@ -1,352 +1,824 @@
...
+          1. Let _microsecond_ be 0.
+        1. If _nanosecond_ is not *undefined*, then
+          1. Let _nanosecond_ be ? ToInteger(_nanosecond_).
+        1. Else
+          1. Let _nanosecond_ be 0.
+        1. If _disambiguation_ is not *undefined*, then
+          1. Set _disambiguation_ to ? ToString(_disambiguation_).
+        1. Else
+          1. Set _disambiguation_ to `"constrain"`.
+        1. Let _result_ be ? RegulateDateTime(_year_, _month_, _day_, _hour_, _minute_, _second_, _millisecond_, _microsecond_, _nanosecond_, _disambiguation_).
```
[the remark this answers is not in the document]
Ms2ger: Same reply as elsewhere.

spec/datetime.html
```diff
@@ -1,352 +1,824 @@
...
-  </emu-clause>
+    <emu-clause>
+      <h1>Temporal.DateTime.fromString ( _isostring_ )</h1>
+      <p>
+        The `fromString` method takes one argument _isostring_.
+        The following steps are taken:
+      </p>
+      <emu-alg>
+        1. Let _isostring_ be ? ToString(_isostring_).
+        1. Return ? DateTimeFromString(_isostring_).
```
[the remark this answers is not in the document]
Ms2ger: Also #232.

spec/datetime.html
```diff
@@ -1,352 +1,824 @@
...
-      and may not be NaN.</p>
-  </emu-clause>
+    <emu-clause>
+      <h1>Temporal.DateTime.from ( _arg_ )</h1>
+      <p>
+        The `from` method takes one argument _arg_.
+        The following steps are taken:
+      </p>
+      <emu-alg>
+        1. Return ? ToDateTime(_arg_).
```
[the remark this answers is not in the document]
Ms2ger: Also #232.

spec/datetime.html
```diff
@@ -1,352 +1,824 @@
...
+      <h1>get Temporal.DateTime.prototype.daysInYear</h1>
+      <p>
+        Temporal.DateTime.prototype.daysInYear is an accessor property whose set accessor function is undefined.
+        Its get accessor function performs the following steps:
+      </p>
+      <emu-alg>
+        1. Let _dateTime_ be the *this* value.
+        1. If Type(_dateTime_) is not Object or _dateTime_ does not have an [[InitializedTemporalDateTime]] internal slot, then
+          1. Throw a *TypeError* exception.
+        1. If ! IsLeapYear(_dateTime_.[[Year]]) is *true*, then
```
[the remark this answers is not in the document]
Ms2ger: Is there a reason you didn't look at the spec to check? There isn't one now, but I'll add one after this PR merges.

spec/datetime.html
Ms2ger: I'm not sure what you mean by "realistic" in this context. (I also notice that ECMA262 doesn't use an abstract operation.)

spec/datetime.html
Ms2ger: Just the one line?

spec/datetime.html
Ms2ger: I'm not sure what you mean by "magic meta prose". This follows precedent from (at least) ECMA402.

spec/datetime.html
Ms2ger: Why not?

spec/datetime.html
Ms2ger: ECMA262 uses the quotes throughout.

spec/datetime.html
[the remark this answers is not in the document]
ljharb: Is there a tracking issue?

spec/datetime.html
ljharb: Precedent from 402 doesn't necessarily dictate 262. I mean referring to the table instead of hardcoding the slot names.

spec/datetime.html
ljharb: All of them, but GitHub doesn’t yet have multiline suggestions.

spec/datetime.html
ljharb: Is there an existing place in 262 where 402 is mentioned in this way?

spec/datetime.html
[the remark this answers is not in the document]
ljharb: I’ll assume you’re not being salty here, so “I’m reviewing on mobile”.

https://tc39.es/ecma262/#eqn-DaysInYear

spec/datetime.html
ljharb: Because the left hand column is internal slots, and because it’s a lookup table of DateTimeLike properties to internal slot names, not just properties of a DateTimeLike.

spec/datetime.html
ljharb: ah, in that case please disregard.

spec/datetime.html
Ms2ger: No, neither have precedent. Let's keep this wording for now; we can do the more complicated thing when merging into the main specs.

spec/datetime.html
Ms2ger: I really don't want to inline the lists of slots and property names in the algorithm(s). Let's keep this for now.

spec/datetime.html
ljharb: The more complicated thing will be required to pass editor review, required for stage 3, but that certainly doesn’t have to be in this PR.

spec/datetime.html
littledan: This text would all go in 402.

spec/datetime.html
littledan: The current pattern is to have a dummy definition in 262 which is overridden in 402. I think we can work this out when we are editing the PRs to merge into the upstream specs.

polyfill/lib/casts.mjs
```diff
@@ -44,37 +44,13 @@ export function CastAbsolute(arg, aux) {
 }
 
 export function CastDateTime(arg, aux) {
+  if ('string' === typeof arg) {
+    return DateTime.fromString(arg);
```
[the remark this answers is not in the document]
littledan: I don't think this is a rule... There are lots of cases of abstract ops calling into properties.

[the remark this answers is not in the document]
ljharb: When necessary for a protocol, you’re right - but in this case, it’s not looking up a property on arg. Replacing DateTime.fromString shouldn’t change what this operation does.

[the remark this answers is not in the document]
ljharb: This shouldn't be resolved, unless the spec is intending to allow replacing `Date.fromString` to alter this algorithm.

polyfill/lib/casts.mjs
```diff
@@ -129,10 +99,28 @@ export function CastDate(arg, aux) {
   throw new RangeError(`invalid date ${arg}`);
 }
 
+export function TimeFromString(arg, Construct) {
```
ljharb: super minor, but "Construct" seems like it's a verb, I'd expect "Constructor" or "C" here

spec/datetime.html
```diff
@@ -1,352 +1,824 @@
...
-  </emu-clause>
+    <emu-clause>
+      <h1>get Temporal.DateTime.prototype.year</h1>
+      <p>
+        Temporal.DateTime.prototype.year is an accessor property whose set accessor function is undefined.
+        Its get accessor function performs the following steps:
+      </p>
+      <emu-alg>
+        1. Let _dateTime_ be the *this* value.
+        1. If Type(_dateTime_) is not Object or _dateTime_ does not have an [[InitializedTemporalDateTime]] internal slot, then
```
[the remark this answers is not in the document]
ljharb: Turns out `RequireInternalSlot` already exists, so this should be:
```suggestion
        1. Perform ? RequireInternalSlot(_dateTime_, [[InitializedTemporalDateTime]]).
```
(same throughout)

polyfill/lib/casts.mjs
```diff
@@ -44,37 +45,13 @@ export function CastAbsolute(arg, aux) {
...
-      GetSlot(aux, MONTH),
-      GetSlot(aux, DAY),
-      GetSlot(arg, HOUR),
-      GetSlot(arg, MINUTE),
-      GetSlot(arg, SECOND),
-      GetSlot(arg, MILLISECOND),
-      GetSlot(arg, MICROSECOND),
-      GetSlot(arg, NANOSECOND)
-    );
-  }
```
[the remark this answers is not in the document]
Ms2ger: There's only one argument in [the docs](https://github.com/tc39/proposal-temporal/blob/cefd79ca974e097c993dc580c1da45e11dc59d9c/docs/datetime.md#temporaldatetimefromthing-string--object--temporaldatetime), and there isn't a way to pass a second argument from most call sites.

It's also very easy to use `date.withTime(time)` rather than `DateTime.from(date, time)`, so I didn't think it was essential to add.

polyfill/lib/casts.mjs
```diff
@@ -44,37 +45,13 @@ export function CastAbsolute(arg, aux) {
...
 export function CastDateTime(arg, aux) {
+  if ('string' === typeof arg) {
+    return DateTime.fromString(arg);
+  }
   const DateTime = ES.GetIntrinsic('%Temporal.DateTime%');
-  if (HasSlot(arg, YEAR, MONTH, DAY, HOUR, MINUTE, SECOND, MILLISECOND, MICROSECOND, NANOSECOND)) {
+  if (ES.IsDateTime(arg)) {
     return arg;
   }
-  if (HasSlot(arg, EPOCHNANOSECONDS)) return arg.inTimeZone(aux);
```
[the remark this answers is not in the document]
Ms2ger: Same issue about needing two arguments as below, and also not all that high-value. I also think it makes more sense that all `.from` methods are consistent in the arguments they take (just strings and option bags, and at this point existing objects of the type we're constructing).

polyfill/lib/casts.mjs
```diff
@@ -44,37 +45,13 @@ export function CastAbsolute(arg, aux) {
 }
 
 export function CastDateTime(arg, aux) {
+  if ('string' === typeof arg) {
+    return DateTime.fromString(arg);
```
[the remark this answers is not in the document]
Ms2ger: Note, this code is moved from below the object-handling code in this function; that avoids the intrinsic lookup and also matches the spec. I'm hoping this can land before #252 in any case.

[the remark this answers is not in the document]
ryzokuken: Yeah, let's probably make a TODO comment directed to me or sth? To make sure I remember to fix this while rebasing/later on (although a failed test should be good enough, but just in case).

polyfill/lib/casts.mjs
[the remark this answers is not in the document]
ryzokuken: So the projections are going to be one-way only in that case? Sounds fair to me.

polyfill/lib/casts.mjs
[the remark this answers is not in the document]
ryzokuken: Fair enough, thanks.

polyfill/test/time.mjs
```diff
@@ -268,6 +268,17 @@ describe('Time', () => {
...
+      it('Time.from("15:23")', () => {
+        const original = Time.fromString;
+        Time.fromString = () => { throw TypeError("Should not call fromString") };
+        try {
+          equal(`${Time.from('15:23')}`, '15:23');
+        } finally {
+          Time.fromString = original;
+        }
+      });
+    });
```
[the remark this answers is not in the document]
ryzokuken: Going with the above resolution, please add a comment here so I remember to remove it later.

Issue #235, opened 2019-10-30
Bikeshedding the name of the inZone method


littledan:
Now that we have `Temporal.TimeZone`, should we call this method `inTimeZone`?

ryzokuken:
In a quick albeit shallow search of the codebase (unfortunately the search functionality in VSCode doesn't allow negative lookahead), I failed to come up with the usage of "zone" anywhere apart from timezones, so +1 to this.

maggiepint:
No opposition, especially if it creates API parity.

ryzokuken:
@pipobscure do you have any contradicting opinions on this by any chance? Assigning this to myself, will make a PR for it today.

PR #248, opened 2019-11-01
[api] rename inZone to inTimeZone

ryzokuken:
Rename `Absolute.prototype.inZone` and `DateTime.prototype.inZone` to
`Absolute.prototype.inTimeZone` and `DateTime.prototype.inTimeZone`
respectively, since Temporal.TimeZone now exists.

Fixes: https://github.com/tc39/proposal-temporal/issues/235


ryzokuken:
Thanks! Merging.

PR #249, opened 2019-11-01
[spec] fix all occurances of [[Day]] in BalanceTime

ryzokuken:
In the spec, BalanceTime was returning an object with a [[Day]] internal
slot, which was neither inline with the polyfill nor the intended
behavior, this PR renames that slot to [[Days]], which better represents
is actual significance (to denote the delta days).

Fixes: https://github.com/tc39/proposal-temporal/issues/234

### DIFF
--- a/spec/time.html
+++ b/spec/time.html
@@ -228,7 +228,7 @@
         1. Let _millisecond_ be _time_.[[Millisecond]] + _duration_.[[Milliseconds]].
         1. Let _microsecond_ be _time_.[[Microsecond]] + _duration_.[[Microseconds]].
         1. Let _nanosecond_ be _time_.[[Nanosecond]] + _duration_.[[Nanoseconds]].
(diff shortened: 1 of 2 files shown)


spec/time.html
```diff
@@ -228,7 +228,7 @@ <h1>Temporal.Time.prototype.plus ( _duration_ )</h1>
         1. Let _millisecond_ be _time_.[[Millisecond]] + _duration_.[[Milliseconds]].
         1. Let _microsecond_ be _time_.[[Microsecond]] + _duration_.[[Microseconds]].
         1. Let _nanosecond_ be _time_.[[Nanosecond]] + _duration_.[[Nanoseconds]].
-        1. Let _result_ be BalanceTime(_hour_, _minute_, _second_, _millisecond_, _microsecond_, _nanosecond_).
+        1. Let _result_ be ? BalanceTime(_hour_, _minute_, _second_, _millisecond_, _microsecond_, _nanosecond_).
```
Ms2ger: While you're fixing this, I think `!` would be better (throughout).

Issue #179, opened 2019-10-11
Should Date.withTime have a disambiguation argument?


Ms2ger:
@pipobscure said no in https://github.com/tc39/proposal-temporal/pull/161/files#r333621271, but the polyfill has it:

https://github.com/tc39/proposal-temporal/blob/3a51ea6e5cb697ed55c44531106d47b0b03fc468/polyfill/lib/date.mjs#L108-L116

pipobscure:
I am inconsolably inconsistent.

The reason for this is that originally we allowed only property Time instances as arguments to this method. In which case disambiguation is not necessary, since the argument is perforce consistent.

However we have since changed that to allow pretty much anything that can be cast to a timeLike value. At which point disambiguation does become relevant.

However since Casting can actually be done, I wonder whether we should just reject inconsistent values and require casting if needed. But then again it;s a hassle.

You see the conundrum?
Your thoughts are as good as mine.
Opinions?

PR #250, opened 2019-11-01
[polyfill] add disambiguation param to withYear

ryzokuken:
Add a disambiguation parameteter to `MonthDay.prototype.withYear`,
finally enabling disambiguation across all `with*` methods in the API.

Fixes: https://github.com/tc39/proposal-temporal/issues/179

(Since we decided it was best to keep `disambiguation` consistently across the API wherever necessary, this puts all our disambiguation worries to an end. All constructors, `plus` methods, `minus` methods, and `with*` methods now have a disambiguation parameter).

### DIFF
--- a/docs/monthday.md
+++ b/docs/monthday.md
@@ -20,7 +20,7 @@
 
 ## monthDay.toLocaleString(locale?: string, options?: object) : string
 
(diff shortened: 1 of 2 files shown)


ryzokuken:
Thanks! Merging.

PR #259, opened 2019-11-11
Set up codecov.io integration.

Ms2ger:


Ms2ger:
There's an issue here with it reporting "Unable to find commit in GitHub". Maybe merging into main will fix that.

PR #266, opened 2019-11-18
Adding documentation to Absolute & Date

jasonwilliams:
Making a start here.
Documentation for Temporal.Absolute

* `from` is missing, i think i need to understand this one a bit more
* * I can't see to get from(obj) working properly, so would be interested in some help on this
* `toString` has an example, im not sure if we want these but toying with the idea
* same with Absolute the polyfill for Date.from() doesn't seem to work properly


ryzokuken:
@jasonwilliams can you update as per #252?

ryzokuken:
@jasonwilliams thank you so much!

PR #274, opened 2019-11-21
Add Absolute spec.

Ms2ger:


Ms2ger:
I'm pretty confused by which parts use BigInt and which parts use Number and why, so please review that in particular.

spec/absolute.html
```diff
@@ -2,103 +2,374 @@
...
+    </emu-clause>
+
+    <emu-clause id="sec-temporal.absolute.fromepochnanoseconds">
+      <h1>Temporal.Absolute.fromEpochNanoseconds ( _arg_ )</h1>
+      <p>
+        The `fromEpochNanoseconds` method takes one argument _arg_.
+        The following steps are taken:
+      </p>
+      <emu-alg>
+        1. Let _epochNanoseconds_ be ? ToBigInt(_arg_).
```
pipobscure: Again this should NOT cast ToBigInt

spec/absolute.html
```diff
@@ -2,103 +2,374 @@
...
+    <emu-clause id="sec-temporal.absolute.prototype.getepochseconds">
+      <h1>Temporal.Absolute.prototype.getEpochSeconds ( )</h1>
+      <p>
+        The following steps are taken:
+      </p>
+      <emu-alg>
+        1. Let _absolute_ be the *this* value.
+        1. Perform ? RequireInternalSlot(_absolute_, [[InitializedTemporalAbsolute]]).
+        1. Let _ns_ be _absolute_.[[Nanoseconds]].
+        1. Let _s_ be floor(_ns_ / 1,000,000,000).
```
pipobscure: Integer maths has no `floor`. And for negative numbers `floor` would be wrong. So that step is superfluous. 

`_absolute_.[[Nanoseconds]]` is a 64bit integer.

spec/absolute.html
```diff
@@ -2,103 +2,374 @@
...
+    <emu-clause id="sec-temporal.absolute.prototype.getepochmilliseconds">
+      <h1>Temporal.Absolute.prototype.getEpochMilliseconds ( )</h1>
+      <p>
+        The following steps are taken:
+      </p>
+      <emu-alg>
+        1. Let _absolute_ be the *this* value.
+        1. Perform ? RequireInternalSlot(_absolute_, [[InitializedTemporalAbsolute]]).
+        1. Let _ns_ be _absolute_.[[Nanoseconds]].
+        1. Let _ms_ be floor(_ns_ / 1,000,000).
```
pipobscure: See above

spec/absolute.html
```diff
@@ -2,103 +2,374 @@
...
+    <emu-clause id="sec-temporal.absolute.prototype.getepochmicroseconds">
+      <h1>Temporal.Absolute.prototype.getEpochMicroseconds ( )</h1>
+      <p>
+        The following steps are taken:
+      </p>
+      <emu-alg>
+        1. Let _absolute_ be the *this* value.
+        1. Perform ? RequireInternalSlot(_absolute_, [[InitializedTemporalAbsolute]]).
+        1. Let _ns_ be _absolute_.[[Nanoseconds]].
+        1. Let _mcs_ be floor(_ns_ / 1,000).
```
pipobscure: See above

spec/absolute.html
```diff
@@ -2,103 +2,374 @@
...
+    </emu-clause>
+
+    <emu-clause id="sec-temporal.absolute.fromepochmicroseconds">
+      <h1>Temporal.Absolute.fromEpochMicroseconds ( _arg_ )</h1>
+      <p>
+        The `fromEpochMicroseconds` method takes one argument _arg_.
+        The following steps are taken:
+      </p>
+      <emu-alg>
+        1. Let _arg_ be ? ToBigInt(_arg_).
```
pipobscure: Again this should NOT cast ToBigInt. It should throw if not given a BigInt.

spec/absolute.html
```diff
@@ -2,103 +2,374 @@
...
+
+    <emu-clause id="sec-temporal.absolute">
+      <h1>Temporal.Absolute ( _epochNanoseconds_ )</h1>
+      <p>
+        When the `Temporal.Absolute` function is called, the following steps are taken:
+      </p>
+      <emu-alg>
+        1. If NewTarget is *undefined*, then
+          1. Throw a *TypeError* exception.
+        1. Let _epochNanoseconds_ be ? ToBigInt(_epochNanoseconds_).
```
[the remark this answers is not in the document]
Ms2ger: Why?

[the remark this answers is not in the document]
littledan: That's what ToBigInt does. https://tc39.es/ecma262/#sec-tobigint

spec/absolute.html
littledan: Math in the JS spec was "mathematical values" by default. Now it's Numbers by default, but you can opt into BigInts if you want. I think this calculation is correct, since _ns_ will be positive anyway.

pipobscure: @littledan OK, in that case we need to specify integer maths.

And **NO this will not be always positive**. The ns value here would be negative for anything before the epoch. (ie: 1969-12-31T23:59:59.999999999Z has ns value of -1n)

Ms2ger: So, what's actually the rounding we want? Towards zero? Round to nearest (and what with .5? Bankers' rounding?) Or do we want to return fractions in some of these?

pipobscure: Round toward 0

littledan: See the JS spec's mathematical notation in https://tc39.es/ecma262/#sec-mathematical-operations . We could add more (harder) or we could use some awkward expression in words to describe exactly what we want here (easier).

PR #278, opened 2019-11-23
Move system-info to Temporal.now (Fixes #236)

pipobscure:


polyfill/lib/now.mjs
```diff
@@ -0,0 +1,31 @@
...
+
+export const now = {
+    absolute: getAbsolute,
+    dateTime: getDateTime,
+    date: getDate,
+    time: getTime,
+    timeZone: getTimeZone
+};
+
+function getAbsolute() {
```
littledan: Nit: These functions will have the wrong name property.

polyfill/test/exports.mjs
```diff
@@ -18,8 +18,9 @@ import * as Temporal from 'tc39-temporal';
 
 describe('Exports', () => {
   const named = Object.keys(Temporal);
-  it(`should be 13 things`, () => {
-    equal(named.length, 13);
+  console.error(named);
```
Ms2ger: Probably want to drop this again.

polyfill/lib/now.mjs
```diff
@@ -0,0 +1,31 @@
...
+};
+
+function absolute() {
+    const ns = ES.SystemUTCEpochNanoSeconds();
+    return new Absolute(ns);
+}
+function dateTime(zone = timeZone()) {
+    zone = ES.ToTimeZone(zone);
+    const abs = absolute();
+    let dateTime = zone.getDateTimeFor(abs);
```
Ms2ger: Your use of let vs const confuses me :)

polyfill/lib/now.mjs
```diff
@@ -0,0 +1,31 @@
...
+    return dateTime;
+}
+function date(zone) {
+    return dateTime(zone).getDate();
```
Ms2ger: This doesn't seem to be robust against author code messing with prototypes; I didn't quite follow if that was a goal at this point.

polyfill/test/exports.mjs
```diff
@@ -18,8 +18,9 @@ import * as Temporal from 'tc39-temporal';
...
+  it(`should be 9 things`, () => {
```
Ms2ger: Wanna add `now` to the tests below?

polyfill/lib/now.mjs
pipobscure: Me too 😄

polyfill/lib/now.mjs
pipobscure: Should probably discuss. Though I think it's fine for now.

Issue #239, opened 2019-10-30
The specification text should use SpeciesConstructor for methods like plus


littledan:
This is part of implementing subclassing support #196. The polyfill seems to implement this well, e.g., https://github.com/tc39/proposal-temporal/blob/main/polyfill/lib/date.mjs#L86 . However, the spec text still calls things like CreateDate directly, e.g., in Temporal.Date.prototype.plus

ptomato:
This was done by Ms2ger some time ago.

PR #281, opened 2019-11-26
Add initial spec for TimeZone.

Ms2ger:


spec/timezone.html
```diff
@@ -3,5 +3,240 @@
...
+    <emu-clause id="sec-temporal.timezone">
+      <h1>Temporal.TimeZone ( _identifier_ )</h1>
+      <p>
+        When the `Temporal.TimeZone` function is called, the following steps are taken:
+      </p>
+      <emu-alg>
+        1. If NewTarget is *undefined*, then
+          1. Throw a *TypeError* exception.
+        1. Let _identifier_ be ? ToString(_identifier_).
+        1. If _identifier_ matches `/([+-][01]?[0-9]):?([0-5][0-9])/`, then
```
ljharb: We don’t have any precedent for this kind of comparison (regex notation); I’d expect either making an actual regex and calling abstract ops to test it, or, a combination of string ops and prose.

spec/timezone.html
```diff
@@ -3,5 +3,240 @@
...
+    </emu-clause>
+
+    <emu-clause id="sec-temporal.timezone.from">
+      <h1>Temporal.TimeZone.from ( _arg_ )</h1>
+      <p>
+        The `from` method takes one argument _arg_.
+        The following steps are taken:
+      </p>
+      <emu-alg>
+        1. Return ? ToTimeZone(_arg_).
```
ljharb: reminder for #239 / #196

spec/timezone.html
```diff
@@ -3,5 +3,240 @@
...
+  </emu-clause>
+
+  <emu-clause id="sec-temporal-timezone-abstract-ops">
+    <h1>Abstract operations</h1>
+
+    <emu-clause id="sec-temporal-timezonefromstring" aoid="TimeZoneFromString">
+      <h1>TimeZoneFromString ( _timeZone_ )</h1>
+      <emu-alg>
+        1. Assert: Type(_timeZone_) is String.
+        1. Let _valid_ be *true* if _timeZone_ matches <a href="https://github.com/tc39/proposal-temporal/blob/main/polyfill/lib/regex.mjs">the "timezone" regular expression in the polyfill</a> and *false* otherwise.
```
ljharb: same comment here re regex notation and alternatives (Not sure if the todo after this line, applies to it)

spec/timezone.html
```diff
@@ -3,5 +3,240 @@
...
+        1. <mark>TODO: Actually specify this somehow.</mark>
+        1. If _valid_ is *false*, then
+          1. Throw a *RangeError* exception.
+        1. Let _identifier_ be the relevant part of _timeZone_.
```
ljharb: “the relevant part”?

spec/timezone.html
ryzokuken: +1

ryzokuken: +1

spec/timezone.html
ryzokuken: Is this inline with the polyfill though? I got confused here since I'm not sure if someone would subclass `TimeZone` but honestly that shouldn't matter given that we mention

>       The Temporal.TimeZone constructor is designed to be subclassable.

above.

littledan:
I think @ljharb 's suggestions about avoiding using RegExp notation here are good, and we can address them in a follow-on PR. Can we land this?

Ms2ger:
I filed https://github.com/tc39/proposal-temporal/issues/312 for the regular expressions.

PR #305, opened 2020-01-03
Addition of cookbook fixes #240

jasonwilliams:
Fixes #240.

### DIFF
--- a/cookbook/README.md
+++ b/cookbook/README.md
@@ -1,0 +1,7 @@
+# Temporal Cookbook
+
+## Construction
+
+[Absolute from legacy Date](./absoluteFromLegacyDate.mjs)
+
+Map a legacy ECMAScript Date instance into a Temporal.Absolute instance corresponding to the same instant in absolute time.
--- a/cookbook/README.md
+++ b/cookbook/README.md
@@ -1,5 +1,11 @@
 # Temporal Cookbook
 
+## Running the cookbook files
+
+Running cookbook files: `node --experimental-modules --no-warnings --icu-data-dir ./polyfill/node_modules/full-icu/ -r ./polyfill/index.js ${cookbookFile}`
+
+_The above code allows `Temporal` to exist as a global object before the cookbook file runs._
+
 ## Construction
 
 ### [Absolute from legacy Date](./absoluteFromLegacyDate.mjs)
--- a/cookbook/README.md
+++ b/cookbook/README.md
@@ -2,6 +2,10 @@
 
 ## Construction
 
-[Absolute from legacy Date](./absoluteFromLegacyDate.mjs)
+### [Absolute from legacy Date](./absoluteFromLegacyDate.mjs)
 
 Map a legacy ECMAScript Date instance into a Temporal.Absolute instance corresponding to the same instant in absolute time.
+
+### [Zoned instant from instant and time zone](./getParseableZonedStringAtInstant.mjs)
+
+Map a Temporal.Absolute instance and a time zone name into a string serialization of the local time in that zone corresponding to the instant in absolute time.
--- a/cookbook/absoluteFromLegacyDate.mjs
+++ b/cookbook/absoluteFromLegacyDate.mjs
@@ -1,5 +1,3 @@
-import { Temporal } from "../polyfill/lib/index.mjs";
-
 /**
  * Converts a Date instance into a Temporal.Absolute instance
  *
@@ -11,5 +9,5 @@ function getAbsoluteFromDate(esDate) {
     return date;
 }
 
-const date = new Date();
-getAbsoluteFromDate(date); // Absolute [Temporal.Absolute] {}
+const date = new Date("1970-01-01T00:00:01Z");
+const absolute = getAbsoluteFromDate(date); // Absolute [Temporal.Absolute] {}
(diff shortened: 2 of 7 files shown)


cookbook/absoluteFromLegacyDate.mjs
```diff
@@ -0,0 +1,15 @@
+import { Temporal } from "../polyfill/lib/index.mjs";
```
littledan: I'd prefer if code samples are written in terms of the new standard (with Temporal already there as a global), rather than importing the polyfill. That is, just deleting this line and leaving everything else the same.

cookbook/README.md
```diff
@@ -0,0 +1,11 @@
+# Temporal Cookbook
+
+## Construction
+
+### [Absolute from legacy Date](./absoluteFromLegacyDate.mjs)
```
littledan: It'd be nice if we could have a nice documentation page that included code samples inline, but we can build this later as a second step.

cookbook/absoluteFromLegacyDate.mjs
jasonwilliams: After a conversation with Philip and others the decision was to have runnable examples that developers could checkout and run themselves, I think until this is in the spec it needs to import the polyfill for that to happen.

Not sure if @pipobscure has any thoughts

littledan: Feel free to go ahead with landing this if you prefer.

ljharb: Why not make them runnable by installing a global as part of the runner setup?

jasonwilliams:
> Would it be interesting to run these examples in CI to make sure they stay up to date?

to answer your question yes, but i think that should be a separate issue

cookbook/absoluteFromLegacyDate.mjs
littledan: In a few conversations about this topic, @pipobscure and I agreed that this should be deleted, replaced by a comment explaining how to run this in Node at the command line.

cookbook/getParseableZonedStringAtInstant.mjs
```diff
@@ -0,0 +1,18 @@
...
+function getParseableZonedStringAtInstant(absolute, ianaTimeZoneName) {
+    // inTimeZone accepts either a timezone string or a Temporal.TimeZone object
+    // It returns a DateTime object
+    const localDateTime = absolute.inTimeZone(ianaTimeZoneName);
+    return localDateTime.toString();
+}
+
+const absoluteTime = Temporal.Absolute.from("2020-01-03T10:41:51Z");
+const timeZone = "Europe/Paris";
+getParseableZonedStringAtInstant(absoluteTime, timeZone); // '2020-01-03T11:41:51'
```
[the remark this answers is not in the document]
ptomato: In that case you probably want `absolute.toString(ianaTimeZoneName)`

cookbook/absoluteFromLegacyDate.mjs
jasonwilliams: > Why not make them runnable by installing a global as part of the runner setup?

Is that possible with something like this https://github.com/tc39/proposal-temporal/blob/main/polyfill/lib/initialise.js which uses a promise, seems like nodejs' preload doesn't wait for any promise to resolve

jasonwilliams: imports have gone and this will be mentioned in the README.md instead

cookbook/README.md
jasonwilliams: From what i understand we can generate a documentation page from these individual files at a later point, so im going to resolve this for now

cookbook/absoluteFromLegacyDate.mjs
```diff
@@ -0,0 +1,15 @@
...
+ * @param {Date} esDate This is a Date object
+ * @returns {Temporal.Absolute} Absolute Object
+ */
+function getAbsoluteFromDate(esDate) {
+    const date = Temporal.Absolute.fromEpochSeconds(esDate.getTime());
+    return date;
+}
+
+const date = new Date();
+getAbsoluteFromDate(date); // Absolute [Temporal.Absolute] {}
```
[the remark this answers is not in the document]
jasonwilliams: interestingly `absolute.getEpochMilliseconds()` is returning `{ value: 1000n }` is this correct? thats what the latest polyfill is doing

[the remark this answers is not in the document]
jasonwilliams: so i can't copy what you've wrote there for now

pipobscure:
I'm merging this as a step forward.

cookbook/absoluteFromLegacyDate.mjs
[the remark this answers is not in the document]
gibson042: No, that looks like a double bug in the polyfill.

cookbook/getParseableZonedStringAtInstant.mjs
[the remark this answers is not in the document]
gibson042: @pdunkel you merged this incorrect definition of getParseableZonedStringAtInstant: #328

cookbook/absoluteFromLegacyDate.mjs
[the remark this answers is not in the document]
ptomato: https://github.com/tc39/proposal-temporal/pull/326 has a fix for the `{ value: foo }` bug.

PR #316, opened 2020-01-16
Temporal.TimeZone docs

ptomato:
MDN-style documentation of Temporal.TimeZone, plus minor bug fixes so that the examples work.

### DIFF
--- a/docs/timezone.md
+++ b/docs/timezone.md
@@ -1,19 +1,213 @@
 # `Temporal.TimeZone`
 
-A representation of a time zone giving access to information about offsets & dst-changes as well as allowing for conversions.
+A `Temporal.TimeZone` is a representation of a time zone: either an [IANA time zone](https://www.iana.org/time-zones), including information about the time zone such as the offset between the local time and UTC at a particular time, and daylight saving time (DST) changes; or simply a particular UTC offset with no DST.
 
-## new Temporal.TimeZone(timeZone: string) : Temporal.TimeZone
+Since `Temporal.Absolute` and `Temporal.DateTime` do not contain any time zone information, a `Temporal.TimeZone` object is required to convert between the two.
+
+Finally, the `Temporal.TimeZone` object itself provides access to a list of the time zones in the IANA time zone database.
+
+## new Temporal.TimeZone(timeZoneIdentifier: string) : Temporal.TimeZone
+
+**Parameters:**
+- `timeZoneIdentifier` (string): A description of the time zone; either its IANA name, or a UTC offset.
+
+**Returns:** a new `Temporal.TimeZone` object.
+
+For a list of IANA time zone names, see the current version of the [IANA time zone database](https://www.iana.org/time-zones).
+A convenient list is also available [on Wikipedia](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones), although it might not reflect the latest official status.
+
+The string `timeZoneIdentifier` is canonicalized before being used to determine the time zone.
+For example, values like `+100` will be understood to mean `+01:00`, and capitalization will be corrected.
+If no time zone can be determined from `timeZoneIdentifier`, then a `RangeError` is thrown.
+
+Usage examples:
+```javascript
+tz = new Temporal.TimeZone('UTC');
+tz = new Temporal.TimeZone('Africa/Cairo');
+tz = new Temporal.TimeZone('america/VANCOUVER');
+tz = new Temporal.TimeZone('Asia/Katmandu');  // alias of Asia/Kathmandu
+tz = new Temporal.TimeZone('-04:00');
+tz = new Temporal.TimeZone('+645');
+/* WRONG */ tz = new Temporal.TimeZone('local');  // not a time zone, throws
+```
+
+### Difference between IANA time zones and UTC offsets
+
+The returned time zone object behaves slightly differently depending on whether an IANA time zone name (e.g. `Europe/Berlin`) is given, or a UTC offset (e.g. `+01:00`).
+IANA time zones may have DST transitions, and UTC offsets do not.
+For example:
+
+```javascript
+tz1 = new Temporal.TimeZone('-08:00');
+tz2 = new Temporal.TimeZone('America/Vancouver');
+now = Temporal.now.absolute();
+tz1.getTransitions(now).next().done;  // => true
+tz2.getTransitions(now).next().done;  // => false
+```
 
 ## timeZone.name : string
 
+The `name` property gives an unambiguous identifier for the time zone.
+Effectively, this is the canonicalized version of whatever `timeZoneIdentifier` was passed as a parameter to the constructor.
+
 ## timeZone.getOffsetFor(absolute: Temporal.Absolute) : string
+
+**Parameters:**
+- `absolute` (`Temporal.Absolute`): The time for which to compute the time zone's UTC offset.
+
+**Returns**: a string indicating the UTC offset at the given time.
+
+Since the UTC offset can change throughout the year in time zones that employ DST, this method queries the UTC offset at a particular time.
+
+Note that only `Temporal.TimeZone` objects constructed from an IANA time zone name may have DST transitions; those constructed from a UTC offset do not.
+If `timeZone` is a UTC offset time zone, the return value of this method is effectively the same as `timeZone.name`.
+
+Example usage:
+```javascript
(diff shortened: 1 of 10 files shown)


docs/timezone.md
```diff
@@ -1,19 +1,213 @@
...
+
+```javascript
+tz = new Temporal.TimeZone('Europe/Berlin');
+dt = new Temporal.DateTime(2019, 3, 31, 2, 45);
+tz.getAbsoluteFor(dt, 'earlier');  // => 2019-03-31T00:45Z
+tz.getAbsoluteFor(dt, 'later');    // => 2019-03-31T01:45Z
+tz.getAbsoluteFor(dt, 'reject');   // throws
+```
+
+In this example, the wall-clock time 2:45 doesn't exist, so it is treated as either 1:45 +01:00 or 3:45 +02:00.
```
gibson042: This seems wrong to me... it means that `tz.getAbsoluteFor(skipped_02_45.minus("PT50M"), "earlier")` is _later_ than `tz.getAbsoluteFor(skipped_02_45, "earlier")`! Shouldn't the "two possible absolute times" for a time of day skipped by a positive transition be "immediately before the transition" vs. "the matching time in the pre-transition offset" (e.g., converting a skipped **02:45** to **01:59:59.999999999** vs. **03:45**)?

docs/timezone.md
```diff
@@ -1,19 +1,213 @@
...
+tz = new Temporal.TimeZone('America/Sao_Paulo');
+dt = new Temporal.DateTime(2019, 2, 16, 23, 45);
+tz.getAbsoluteFor(dt, 'earlier');  // => 2019-02-17T01:45Z
+tz.getAbsoluteFor(dt, 'later');    // => 2019-02-17T02:45Z
+tz.getAbsoluteFor(dt, 'reject');   // throws
+```
+
+In this example, the wall-clock time 23:45 exists twice.
+
+> *Compatibility Note*: The built-in behaviour of the Moment Timezone and Luxon libraries is to give the same result as `earlier` when turning the clock back, and `later` when setting the clock forward.
```
gibson042: This seems to indicate a gap in the Temporal API that should be corrected, and closely relates to [Preserving local time](https://github.com/tc39/proposal-temporal/issues/240#user-content-preserving-local-time) (cc @pdunkel).

docs/timezone.md
ptomato: I don't have an opinion yet on which way is correct, but FWIW the behaviour described here is what the polyfill currently does. (The spec is marked TODO on this point.) The moment-timezone package also works like this. So maybe we should open an issue to resolve that question separately.

ryzokuken: +1 to following up on this.

polyfill/lib/duration.mjs
```diff
@@ -53,17 +53,17 @@ export class Duration {
         [years, months, days, hours, minutes, seconds, milliseconds, microseconds, nanoseconds] = arr;
         break;
       case 'balance':
-        let tdays;
+        let deltaDays;
         ({
-          days: tdays,
+          deltaDays,
```
ryzokuken: Would this work as expected? Do tests pass?

polyfill/lib/ecmascript.mjs
```diff
@@ -39,7 +39,7 @@ import {
   NANOSECONDS
 } from './slots.mjs';
 
-const DAYMILLIS = 3600000;
+const DAYMILLIS = 86400000;
```
ryzokuken: How'd we get this wrong?

polyfill/lib/timezone.mjs
```diff
@@ -4,6 +4,8 @@ import { IDENTIFIER, EPOCHNANOSECONDS, CreateSlots, GetSlot, SetSlot } from './s
 import { ZONES } from './zones.mjs';
 import { timezone as STRING } from './regex.mjs';
 
+import bigInt from 'big-integer';
```
ryzokuken: why wasn't this failing? Let me guess, no tz tests?

polyfill/lib/timezone.mjs
```diff
@@ -80,7 +82,7 @@ export class TimeZone {
 
     if (!~['earlier', 'later'].indexOf(disambiguation)) throw new RangeError(`no such absolute found`);
 
-    const { ms: utcms } = ES.GetEpochFromParts(
+    const utcns = ES.GetEpochFromParts(
```
ryzokuken: Would this work?

docs/timezone.md
```diff
@@ -1,19 +1,213 @@
 # `Temporal.TimeZone`
 
-A representation of a time-zone giving access to information about offsets & dst-changes as well as allowing for conversions.
+A `Temporal.TimeZone` is a representation of a time zone: either an [IANA time zone](https://www.iana.org/time-zones), including information about the time zone such as the offset between the local time and UTC at a particular time, and daylight saving time (DST) changes; or simply a particular UTC offset with no DST.
 
-## new Temporal.TimeZone(timeZone: string) : Temporal.TimeZone
+Since `Temporal.Absolute` and `Temporal.DateTime` do not contain any time zone information, a `Temporal.TimeZone` object is required to convert between the two.
+
+Finally, the `Temporal.TimeZone` object itself provides access to a list of the time zones in the IANA time zone database.
```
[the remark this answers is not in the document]
ptomato: It's a stub in the spec (https://tc39.es/proposal-temporal/#sec-temporal.timezone.prototype-@@iterator) but it does look like it's intentionally there. It was also mentioned elsewhere in the stub documentation before I expanded it.

docs/timezone.md
```diff
@@ -1,19 +1,213 @@
...
+
+Usage examples:
+```javascript
+tz = new Temporal.TimeZone('UTC');
+tz = new Temporal.TimeZone('Africa/Cairo');
+tz = new Temporal.TimeZone('america/VANCOUVER');
+tz = new Temporal.TimeZone('Asia/Katmandu');  // alias of Asia/Kathmandu
+tz = new Temporal.TimeZone('-04:00');
+tz = new Temporal.TimeZone('+645');
+// tz = new Temporal.TimeZone('local');  // not a time zone, throws
```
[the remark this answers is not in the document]
ptomato: Not necessarily, I thought it would be useful to show that it's incorrect so that people don't copy-paste it, but if that's confusing then probably the "throws" comment at the end is enough. Or we could do something like this:
```js
/* WRONG */ tz = new Temporal.TimeZone('local');  // not a time zone, throws
```

polyfill/lib/duration.mjs
ptomato: Yes, this wasn't covered by any tests before, so it wasn't obvious that it was broken. The new tests that I added do cover it.

polyfill/lib/timezone.mjs
ptomato: Yes, the existing code was broken but it wasn't covered by any tests. My guess is that the existing code was correct at one point but then the return value of `GetEpochFromParts` must have changed.

polyfill/lib/timezone.mjs
ptomato: This was actually not a problem in the existing code, because bigInt wasn't used in this file before I made these changes. But since the return value of `GetEpochFromParts` changed to return a bigInt (see comment below) we now have to import this to do math on it.

polyfill/lib/ecmascript.mjs
ptomato: :man_shrugging: it has test coverage now, so it should be okay from now on

docs/timezone.md
ptomato: Opened https://github.com/tc39/proposal-temporal/issues/317 to follow up.

docs/timezone.md
ptomato: Opened https://github.com/tc39/proposal-temporal/issues/318 for this.

docs/timezone.md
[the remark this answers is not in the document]
ptomato: I went with the latter.

docs/timezone.md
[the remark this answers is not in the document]
littledan: Does anyone know the motivation for this feature? @pipobscure

docs/timezone.md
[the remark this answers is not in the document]
ryzokuken: Thanks.

Issue #276, opened 2019-11-22
docs/duration.md claims Duration constructor supports option bag/string arguments


Ms2ger:
It doesn't – that's `Duration.from`.

pipobscure:
Sorry, that is a carry-over from earlier concepts. The constructor doe **NOT** support options bags, that's what we made the `.from` for.

pipobscure:
Can we close this? (I will for now. reopen if someone thinks it should be)

ptomato:
I think it should be reopened. It's still a defect in the docs, which I am working on a patch for.

pipobscure:
Reopening per request by @ptomato

ptomato:
Ah, I already reopened it myself and then closed it when my #320 got merged, I think this is all good now :smile:

PR #320, opened 2020-01-21
Fix Duration constructor docs

ptomato:
Corrects the signature of `new Temporal.Duration()` in the reference docs and spec to match the polyfill.

Closes: #276

### DIFF
--- a/spec/duration.html
+++ b/spec/duration.html
@@ -92,7 +92,7 @@
       initialize subclass instances with the necessary internal slots.
     </p>
     <emu-clause id="sec-temporal.duration">
-      <h1>Temporal.Duration ( _years_, _months_, _days_, _hours_, _minutes_, _seconds_, _milliseconds_, _microseconds_, _nanoseconds_, _disambiguation_ )</h1>
+      <h1>Temporal.Duration ( [ _years_ [ , _months_ [ , _days_ [ , _hours_ [ , _minutes_ [ , _seconds_ [ , _milliseconds_ [ , _microseconds_ [ , _nanoseconds_ [ , _disambiguation_ ] ] ] ] ] ] ] ] ] ] )</h1>
       <p>
         When the `Temporal.Duration` function is called, the following steps are taken:
       </p>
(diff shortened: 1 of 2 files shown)


spec/duration.html
```diff
@@ -92,7 +92,7 @@ <h1>The Temporal.Duration Constructor</h1>
       initialize subclass instances with the necessary internal slots.
     </p>
     <emu-clause id="sec-temporal.duration">
-      <h1>Temporal.Duration ( _years_, _months_, _days_, _hours_, _minutes_, _seconds_, _milliseconds_, _microseconds_, _nanoseconds_, _disambiguation_ )</h1>
+      <h1>Temporal.Duration ( [ _years_ [ , _months_ [ , _days_ [ , _hours_ [ , _minutes_ [ , _seconds_ [ , _milliseconds_ [ , _microseconds_ [ , _nanoseconds_ [ , _disambiguation_ ] ] ] ] ] ] ] ] ] ] )</h1>
```
pipobscure: This seems wrong as it implies that to specify a _month_ the _year_ is mandatory which it's not.

Maybe:

```
<h1>Temporal.Duration ( [_years_] , [_months_ ] , [_days_] , [_hours_] , [_minutes_] , [_seconds_] , [_milliseconds_] , [_microseconds_] , [_nanoseconds_] , [_disambiguation_])</h1>
```

littledan: I'm very confused by this suggestion. The syntax seems like a new innovation, and it looks to me like you could omit them and the system would figure out which parameter is which (which isn't the case--in this spec, it's position-based).

Isn't it mandatory to provide something in that argument position, to fill in the year? You may provide the value `undefined`, though.

pipobscure: Correct. But we ask for the values to be numeric, so undefined is not an option per se. Also this syntax is not new. the `[` thing has always just meant it's an optional argument. By going in several levels and collecting all the `]` at the end what you are expressing is that the options are right to left optional. Or in short that for the second argument may only be specified if the first is there. Which is not quite this.

pipobscure: *(But I will certainly not make an issue of it. If this is clear to everyone but me then by all means)*

I just wanted to express that `new Temporal.Duration(undefined, 5, undefined 7)` is perfectly allowed. The syntax to me suggested that it wouldn't be but rather only 
```
new Temporal.Duration()
new Temporal.Duration(1)
new Temporal.Duration(1, 2)
new Temporal.Duration(1, 2, 3)
new Temporal.Duration(1, 2, 3, 4)
```

and so on were ok.

littledan: Let's make it clear that `undefined` is a valid parameter. I think Philip's syntax fits better with existing conventions.

ljharb: The syntax in the PR as is loos right to me; undefined is always a valid potential parameter, and the years param is indeed required to provide months. Brackets don’t denote “can be undefined”, they denote “can be omitted entirely”.

ptomato: According to the spec ToInteger(`undefined`) is +0, so I don't think any change to the spec is necessary. Maybe it would be better to make sure that the reference docs (when I write them) mention that `undefined` is a valid value and give an example like `new Temporal.Duration(undefined, 3)`

ptomato:
Any remaining objections to merging this?

PR #321, opened 2020-01-21
Fix and clarify TimeZone.from()

ptomato:
Closes: #313


docs/timezone.md
```diff
@@ -45,6 +47,51 @@ tz1.getTransitions(now).next().done;  // => true
...
+
+**Parameters:**
+- `thing` (string or object): A `Temporal.TimeZone` object or a string from which to create a `Temporal.TimeZone`.
+
+**Returns:** a new `Temporal.TimeZone` object.
+
+This static method creates a new time zone from another value.
+If the value is a string, it can be:
+- a string that is accepted by `new Temporal.TimeZone()`;
+- a string in the ISO 8601 format including a time zone part, optionally extended with an IANA time zone name in square brackets appended to it;
```
gibson042:
```suggestion
- a string in the ISO 8601 format including a UTC offset, optionally extended with an IANA time zone name in square brackets appended to it;
```

docs/timezone.md
```diff
@@ -45,6 +47,51 @@ tz1.getTransitions(now).next().done;  // => true
...
+- or, a string with _only_ the time zone part of the ISO 8601 format, again extended with optional IANA name in brackets.
```
gibson042: I remain opposed to accepting "-08:00[America/Vancouver]" or "Z". What is the use case? Where would they come from, if not an ISO 8601 string that should be used in entirety?

docs/timezone.md
```diff
@@ -45,6 +47,51 @@ tz1.getTransitions(now).next().done;  // => true
...
+Or, if the value is an object, it can be another `Temporal.TimeZone` object, which is returned directly.
```
gibson042: What about a non-Temporal.TimeZone object with the right properties, in analogy to `Temporal.Date.from({year: 2020, month: 1, day: 23})` and `Temporal.Duration.from({hours: 24})`?

docs/timezone.md
```diff
@@ -45,6 +47,51 @@ tz1.getTransitions(now).next().done;  // => true
...
+
+Usage examples:
+```javascript
+// IANA time zone names and UTC offsets
+tz = Temporal.TimeZone.from('UTC');
+tz = Temporal.TimeZone.from('Africa/Cairo');
+tz = Temporal.TimeZone.from('america/VANCOUVER');
+tz = Temporal.TimeZone.from('Asia/Katmandu');  // alias of Asia/Kathmandu
+tz = Temporal.TimeZone.from('-04:00');
+tz = Temporal.TimeZone.from('+645');
```
gibson042: This also is not valid ISO 8601 and therefore suspect. Why should Temporal.TimeZone.from accept it?

docs/timezone.md
```diff
@@ -45,6 +47,51 @@ tz1.getTransitions(now).next().done;  // => true
...
+// ISO 8601 string with time zone part
+tz = Temporal.TimeZone.from('2020-01-14T00:31:00.065858086Z');
+tz = Temporal.TimeZone.from('2020-01-13T16:31:00.065858086-08:00');
+tz = Temporal.TimeZone.from('2020-01-13T16:31:00.065858086-08:00[America/Vancouver]');
+
+// Only time zone part of ISO 8601 string
+tz = Temporal.TimeZone.from('Z');
+tz = Temporal.TimeZone.from('-08:00[America/Vancouver]');
+// If the offset and IANA time zone name disagree, the name takes precedence
+Temporal.TimeZone.from('-05:00[UTC]').name;  // => UTC
```
gibson042: What?!?

docs/timezone.md
ptomato: Not sure I agree with this change. A time zone part can be a UTC offset optionally extended with an IANA name, but it can also be `Z`.

docs/timezone.md
ptomato: I also feel that these shouldn't be accepted, however, this is the status quo and it seems this is still being debated in #313 (and the precedent is https://github.com/tc39/proposal-temporal/issues/198#issuecomment-547552203); in any case, I've put this on the agenda for the next call.

docs/timezone.md
ptomato: The only property is `name`, so you'd have `Temporal.TimeZone.from({name: 'America/Vancouver'})` which seems a bit redundant with `Temporal.TimeZone.from('America/Vancouver')`. However, it seems reasonable to add this capability for symmetry, if people feel strongly about it.

docs/timezone.md
ptomato: Because we have decided that `Temporal.TimeZone.from` should accept a UTC offset string (regardless of whether we decide that a bare ISO 8601 time zone part is okay) and the polyfill currently considers that a UTC offset string. (The spec says nothing yet, since that's still marked TODO.) I think I agree that a legal UTC offset string should be the same as a legal UTC offset in an ISO 8601 time zone part, so I'll add another commit to this branch to bring that in line.

docs/timezone.md
ptomato: Well, we have to do _something_ if they disagree, and this is what the polyfill currently does :stuck_out_tongue_closed_eyes: 

I agree that it seems off, though. Other options would be:
- consider the string invalid if they disagree (although that opens a can of worms because an IANA zone can have multiple offsets throughout the year and throughout history, so what does disagreement mean? We should _definitely_ not allow a bare time zone part if that's the case, since agreement would only be well-defined for a specific date and time)
- change our extension to ISO 8601 so that it's not possible to give both an offset and a name

docs/timezone.md
gibson042: Surely "2020-01-25T15:15Z[Etc/UTC]" is invalid, right? "-05:00" is not a time zone, it is a difference between local time and UTC (which can be followed by a bracketed time zone name). And "Z" in that position is also not a time zone, it is a UTC designator (which cannot be).

docs/timezone.md
gibson042: If the only property is `name`, then we have a separate problem of time zone objects not being constructible. Isn't a Temporal.TimeZone object intended to represent a [tzdata](https://data.iana.org/time-zones/tz-how-to.html) entity?

docs/timezone.md
gibson042: > * We should _definitely_ not allow a bare time zone part if that's the case, since agreement would only be well-defined for a specific date and time)

What do you mean? `Temporal.TimeZone.from('-05:00')` should return the time zone that is always 5 hours behind UTC, `Temporal.TimeZone.from('UTC')` should return the time zone that is always UTC, and `Temporal.TimeZone.from('America/New_York')` should return the US Eastern Time Zone (whose relationship to UTC is variable).

docs/timezone.md
ptomato: How about "a string in the ISO 8601 format including a time zone part" and then clarify in the text below that a time zone part can be "either the UTC designator `Z`, or a UTC offset optionally extended with an IANA name in brackets"?

docs/timezone.md
ptomato: Not sure if I understand what you mean with time zone objects not being constructible? If you mean that an object can have a UTC offset and no name, then maybe that means that the `name` property is misnamed; since `Temporal.TimeZone.from('-08:00').name === '-08:00'` so presumably if we allowed construction from an options bag we'd allow `{ name: '-08:00' }`.

docs/timezone.md
ptomato: > > * We should _definitely_ not allow a bare time zone part if that's the case, since agreement would only be well-defined for a specific date and time)
> 
> What do you mean? `Temporal.TimeZone.from('-05:00')` should return the time zone that is always 5 hours behind UTC, `Temporal.TimeZone.from('UTC')` should return the time zone that is always UTC, and `Temporal.TimeZone.from('America/New_York')` should return the US Eastern Time Zone (whose relationship to UTC is variable).

Sorry, I should have clarified that better; my thinking is that we have two somewhat overlapping possibilities for designating a time zone, which I'm calling
- "time zone identifier" (offset or IANA name), and
- "bare time zone part of ISO 8601 string" (whatever we'd allow as an ISO 8601 string, but with the date and time parts chopped off).
Strings like `UTC` and `America/Vancouver` are only valid "time zone identifiers". Strings like `Z` and `-08:00[America/Vancouver]` are only valid "bare time zone parts". Strings like `-08:00` are valid as both.

My opinion is that because a "bare time zone part" can actually contradict itself (`-05:00[UTC]`) that's another argument for not allowing them when constructing a time zone object. I suspect we agree about this.

> > change our extension to ISO 8601 so that it's not possible to give both an offset and a name
> 
> It's only a problem when constructing a time zone object—in all other cases, time zone names are only valid after UTC offsets, which theirselves are only valid after datetimes (and therefore provide access to sufficient information for detecting disagreement and rejecting the input when found).

Agreed.

docs/timezone.md
gibson042: That structure definitely works, but I think I would still call it an "offset part" rather than a "time zone part", because it's fundamentally coming from ISO 8601 as a UTC offset and only being _extended_ to include time zone information.

docs/timezone.md
gibson042: No, I mean that time zone objects consist fundamentally of a collection of rules describing when and how local time transitions between (UTC) offsets. Although a time zone can be _looked up_ from a common database by name, it's not a name—it's a data structure, and constructing a time zone from scratch is an important use case (cf. [`getTimeZoneObjectFromRules`](https://github.com/tc39/proposal-temporal/issues/240#user-content-timezone-instance), among others).

ptomato: Constructing "synthetic" time zone objects is an API that we describe neither in the spec nor in the polyfill right now. I'd consider that out of scope for fixing the current problems with `TimeZone.from()`, but I opened another issue for it: #332

ptomato:
I've updated this pull request to reflect the changes that we discussed in the meeting this morning.

docs/timezone.md
gibson042: Given the undeveloped state of those objects, I'm fine with keeping them out of scope for this PR.

docs/timezone.md
```diff
@@ -45,6 +48,47 @@ tz1.getTransitions(now).next().done;  // => true
...
+**Parameters:**
+- `thing` (string or object): A `Temporal.TimeZone` object or a string from which to create a `Temporal.TimeZone`.
+
+**Returns:** a new `Temporal.TimeZone` object.
+
+This static method creates a new time zone from another value.
+If the value is a string, it can be:
+- a string that is accepted by `new Temporal.TimeZone()`;
+- a string in the ISO 8601 format including a time zone offset part.
+Or, if the value is an object, it can be another `Temporal.TimeZone` object, which is returned directly.
```
gibson042: #232 calls for `from` to always return a new object. But if the other `from`s currently do direct returns, then :+1: to this for now and they can all be fixed together.

spec/timezone.html
```diff
@@ -191,6 +191,11 @@ <h1>Abstract operations</h1>
       <h1>TimeZoneFromString ( _timeZone_ )</h1>
       <emu-alg>
         1. Assert: Type(_timeZone_) is String.
+        1. If _timeZone_ matches `/([+-][01]?[0-9]):?([0-5][0-9])/`, then
```
gibson042: What is the rationale for these bounds? -19:59 through +19:59 is a strange cutoff.

pipobscure: I believe this came from current maximum being +14:00

Such that I didn’t add a third option for the first numeric digit in parsing.

The absolute maximum would be +-24:00 is the theoretical possibility of that more important?

ptomato: This line I've just copied from another place in the spec (the Temporal.TimeZone constructor) assuming it'll get replaced as part of #312.

docs/timezone.md
ptomato: They do, as far as I can tell.

spec/timezone.html
gibson042: ISO 8601 does not establish bounds as far as I can tell, so I'd align with [RFC 3339](https://tools.ietf.org/html/rfc3339#section-5.6) on -23:59 through +23:59.

ptomato: OK, will address this as part of #312 then.

ptomato:
I think all the questions and comments are resolved here, can anyone approve and/or merge?

ptomato:
Added one more fix to reject date/time strings where the UTC offset doesn't agree with the bracketed IANA name.

pipobscure:
are there any tests for the reges changes? I'm thinking there should be something in `polyfill/test/regex.mjs`

ptomato:
`polyfill/test/regex.mjs` is currently commented out in `polyfill/test/all.mjs`, so those tests are not run... however, I'm happy to turn it back on. I'm not a fan of testing the regex directly though, since it's an implementation detail and I think it's more important to test the public API. I'd rather change the regex tests to use `X.from()` and consolidate any other `X.from("some-string")` tests into that file.

ptomato:
Since the regex tests PR got approved in the meantime, I'll add some tests to that file before I merge this.

PR #323, opened 2020-01-24
Use internal slots

ptomato:
Closes: #231.

Open question from that issue: what about `dateTime.with(date)`, should that use the internal slots of `date`?

### DIFF
--- a/polyfill/lib/date.mjs
+++ b/polyfill/lib/date.mjs
@@ -67,10 +67,7 @@ export class Date {
   }
   with(dateLike = {}, disambiguation = 'constrain') {
     if (!ES.IsDate(this)) throw new TypeError('invalid receiver');
-    const props = ES.ValidPropertyBag(dateLike, ['year', 'month', 'day']);
-    if (!props) {
-      throw new RangeError('invalid date-like');
-    }
+    const props = ES.ToPartialDate(dateLike);
     const { year = GetSlot(this, YEAR), month = GetSlot(this, MONTH), day = GetSlot(this, DAY) } = props;
     const Construct = ES.SpeciesConstructor(this, Date);
     return new Construct(year, month, day, disambiguation);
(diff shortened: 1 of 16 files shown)


spec/abstractops.html
```diff
@@ -82,8 +82,11 @@ <h1>ToDurationLike ( _duration_ )</h1>
       1. If Type(_duration_) is not Object, then
         1. Throw a *TypeError* exception.
       1. For each row of <emu-xref href="#table-temporal-durationlike-properties"></emu-xref>, except the header row, in table order, do
-        1. Let _property_ be the Property value of the current row.
-        1. Let _value_ be ? ToInteger(? Get(_duration_, _property_)).
+        1. If _duration_ has an [[InitializedTemporalDuration]] internal slot, then
+          1. Let _value_ be the value of _duration_'s internal slot whose name is the Internal Slot value of the current row.
```
gibson042: Are we **sure** we want conditional Get bypass on an argument? [`Array.prototype.concat`](https://tc39.es/ecma262/#sec-array.prototype.concat) could have done this but did not, and cases that do seem limited to signatures that _require_ typed input (e.g., [`new DataView`](https://tc39.es/ecma262/#sec-dataview-buffer-byteoffset-bytelength)). I'm not convinced that the benefits of such branches are sufficient to justify the complexity of including them.

ptomato: Indeed, I'm still wondering about code like what I wrote in [this comment](https://github.com/tc39/proposal-temporal/issues/231#issuecomment-577495048)...

ptomato:
Since #362 landed, this is going to need a bit of revision.

ptomato:
Revised. I ended up removing the changes to Duration since #362 already made the spec so that internal slots are consulted when passing in a Duration object.

The next step is still to determine whether we actually want this or not...

ptomato:
Meeting, Feb. 27: We don't actually want this, at least not in this form.

ljharb:
@ptomato can you elaborate? i very much want all important data for a temporal type to be handled with internal slots.

PR #326, opened 2020-01-25
Absolute docs

ptomato:
Reference docs for Temporal.Absolute as well as a quick hack to add Prism.js syntax highlighting to the reference docs.


ptomato:
Merge conflict fixed.

PR #333, opened 2020-01-29
Add cutoff parameter to Date.difference() and DateTime.difference()

ptomato:
Closes: #307

### DIFF
--- a/docs/absolute.md
+++ b/docs/absolute.md
@@ -278,23 +278,49 @@ oneDay = new Temporal.Duration(0, 0, 1);
 Temporal.now.absolute().minus(oneDay);
 ```
 
-### absolute.**difference**(_other_: Temporal.Absolute) : Temporal.Duration
+### absolute.**difference**(_other_: Temporal.Absolute, _largestUnit_: string = 'seconds') : Temporal.Duration
 
 **Parameters:**
 - `other` (`Temporal.Absolute`): Another time with which to compute the difference.
+- `largestUnit` (optional string): The largest unit of time to allow in the resulting `Temporal.Duration` object.
+  Valid values are `'days'`, `'hours'`, `'minutes'`, and `'seconds'`.
+  The default is `"seconds"`.
 
 **Returns:** a `Temporal.Duration` representing the difference between `absolute` and `other`.
 
 This method computes the difference between the two times represented by `absolute` and `other`, and returns it as a `Temporal.Duration` object.
 The difference is always positive, no matter the order of `absolute` and `other`, because `Temporal.Duration` objects cannot represent negative durations.
 
+The `largestUnit` parameter controls how the resulting duration is expressed.
+The returned `Temporal.Duration` object will not have any nonzero fields that are larger than the unit in `largestUnit`.
+A difference of two hours will become 7200 seconds when `largestUnit` is `"seconds"`, for example.
+However, a difference of 30 seconds will still be 30 seconds even if `largestUnit` is `"hours"`.
+
+By default, the largest unit in the result is seconds.
+Unlike other Temporal types, months and years are not allowed.
+This is because in the ISO calendar, months and years can be different lengths depending on which month is meant and whether the year is a leap year.
+`Temporal.Absolute` is intended to be calendar-independent and therefore free of these ambiguities.
+
 Example usage:
 ```js
 startOfMoonMission = Temporal.Absolute.from('1969-07-16T13:32:00Z');
 endOfMoonMission = Temporal.Absolute.from('1969-07-24T16:50:35Z');
-missionLength = startOfMoonMission.difference(endOfMoonMission);  // => P8DT3H18M35S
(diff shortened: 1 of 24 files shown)


ptomato:
Putting this up for initial review; I assume the cutoff parameter should be added to all the other difference methods for consistency as well?

ptomato:
Updated to reflect the latest discussion in #307, the last commit can be squashed into the second-last pending discussion.

docs/time.md
```diff
@@ -34,7 +34,22 @@ A representation of wall-clock time.
 
 ### time.**minus**(_duration_: string | object, _disambiguation_: 'constrain' | 'balance' | 'reject' = 'constrain') : Temporal.Time
 
-### time.**difference**(_other_: Temporal.Time) : Temporal.Duration
+### time.**difference**(_other_: Temporal.Time, _largestUnit_: string = 'days') : Temporal.Duration
+
+**Parameters:**
+- `other` (`Temporal.DateTime`): Another date/time with which to compute the difference.
```
gibson042:
```suggestion
- `other` (`Temporal.Time`): Another time with which to compute the difference.
```

docs/date.md
```diff
@@ -71,10 +71,35 @@ Returns a new `Date` object which is the sum of the current object plus the addi
...
+  The default is `days`.
 
 Returns a new `Duration` object which is the difference between the current `Date` object and the argument `Date` value.
 
+**TODO:** Rest of the documentation in another pull request, this to be rebased
+
+The `largestUnit` parameter controls how the resulting duration is expressed.
+The returned `Temporal.Duration` object will not have any nonzero fields that are larger than the unit in `largestUnit`.
+A difference of two hours will become 7200 seconds when `largestUnit` is `"seconds"`, for example.
+However, a difference of 30 seconds will still be 30 seconds even if `largestUnit` is `"hours"`.
```
gibson042: This text needs to be changed for Temporal.Date, which does not support a largest unit of "hours".

docs/time.md
```diff
@@ -34,7 +34,22 @@ A representation of wall-clock time.
...
+  The default is `days`.
+
+**TODO:** Rest of the documentation in another pull request, this to be rebased
+
+The `largestUnit` parameter controls how the resulting duration is expressed.
+The returned `Temporal.Duration` object will not have any nonzero fields that are larger than the unit in `largestUnit`.
+A difference of two hours will become 7200 seconds when `largestUnit` is `"seconds"`, for example.
+However, a difference of 30 seconds will still be 30 seconds even if `largestUnit` is `"hours"`.
+
+For consistency with the other types' `difference` methods, the default largest unit in the result is days, although this method never returns any duration longer than 12 hours.
```
gibson042: I don't see the benefit of this consistency, especially given that it is already broken by e.g. Temporal.Date difference not supporting "hours" or lower.

docs/absolute.md
```diff
@@ -271,23 +271,48 @@ oneDay = new Temporal.Duration(0, 0, 1);
...
 endOfMoonMission = Temporal.Absolute.from('1969-07-24T16:50:35Z');
 missionLength = startOfMoonMission.difference(endOfMoonMission);  // => P8DT3H18M35S
 endOfMoonMission.difference(startOfMoonMission);  // => P8DT3H18M35S
 missionLength.toLocaleString();  // example output: '8 days 3 hours 18 minutes 35 seconds'
+
+// A billion (10^9) seconds since the epoch in different units
+epoch = new Temporal.Absolute(0n);
+billion = Temporal.Absolute.fromEpochSeconds(1e9);
+epoch.difference(billion);  // => P11574DT1H46M40S
+epoch.difference(billion, 'seconds')  // => PT1000000000S
```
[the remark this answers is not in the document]
ptomato: I switched them instead, to show the default and then show the result for `hours`.

docs/time.md
ptomato: I've changed the default value to `hours` for Time.

spec/abstractops.html
```diff
@@ -40,6 +40,20 @@ <h1>ToTimeZoneDisambiguation ( _disambiguation_ )</h1>
     </emu-alg>
   </emu-clause>
 
+  <emu-clause id="sec-temporal-tolargestunit" aoid="ToLargestUnit">
+    <h1>ToLargestUnit ( _largestUnit_, _disallowedUnits_ )</h1>
+    <emu-alg>
+      1. If _largestUnit_ is *undefined*, then
+        1. Return `"days"`.
```
gibson042: This assumption works, but it doesn't sit right with me. I'd rather see either every use provide a non-undefined value (having already checked for undefined and replacing it with a default as is done by Absolute) or for the operation to be more like [ECMA-402 GetOption](http://tc39.es/ecma402/#sec-getoption) in accepting default value as input (e.g., ```ToLargestUnit(_largestUnit_, « `"hours"`, `"minutes"`, `"seconds"` », `"seconds"`)```).

And in anticipation of adding this text to the full ECMA-262, a more specific operation name such as ToLargestTemporalUnit would probably make sense.

PR #334, opened 2020-01-31
adding getTimeZoneObjectFromIanaName

jasonwilliams:


littledan:
I have to say, I find it a bit strange to abstract thigns like this into methods. I thought a cookbook would consist of snippets like

> Here's how to get a TimeZone from a name:
```js
new Temporal.TimeZone("Europe/London")
```

Cookbooks are designed to be copy-pasted and changed. It wouldn't make sense to copy-paste this kind of function.

Maybe we should differentiate API usage samples like this and larger example code, with only the latter being in separate files, extracted into functions, etc.

jasonwilliams:
So this was taken from  https://github.com/tc39/proposal-temporal/issues/240, im not sure if this one warrants a method, as its literally just that one line.

btw, I've separated these PRs out so conversations like this can happen with blocking future additions

cookbook/getTimeZoneObjectFromIanaName.js
```diff
@@ -0,0 +1,11 @@
...
+ *
+ * @param ianaTimeZoneName {string} TimeZone String
+ * @returns {Temporal.Timezone}
+ * @example getTimeZoneObjectFromIanaName("Europe/London")
+ */
+function getTimeZoneObjectFromIanaName(ianaTimeZoneName) {
+    return new Temporal.TimeZone(ianaTimeZoneName);
+}
+
+getTimeZoneObjectFromIanaName("Europe/London").toString(); // Europe/London
```
Ms2ger: Maybe it would be useful to show how to use some of the methods of the `TimeZone` objects here.

In any case, please rebase and add the new test to `docs/cookbook/all.mjs`

jasonwilliams:
@Ms2ger @ptomato done

jasonwilliams:
> Good with me, though @littledan's concern is still pending. I'm fine to merge this and reorganize things later, though, to keep it moving.

I think we should do that

PR #336, opened 2020-01-31
sort Absolute Instants

jasonwilliams:
Addition of Sort Absolute Instants to the cookbook

### DIFF
--- a/docs/cookbook/sortAbsoluteInstants.mjs
+++ b/docs/cookbook/sortAbsoluteInstants.mjs
@@ -1,0 +1,28 @@
+/**
+ * getSortedInstants will sort an array of strings (each of which is parseable as a Temporal.Absolute and may or may not include an IANA time zone name)
+ * by the corresponding absolute time (e.g., for presenting global log events sequentially).
+ *
+ *
+ * @param {string[]} parseableAbsoluteStrings This is a Date instance
+ * @param {boolean} reverse accending or descending order
+ * @returns {Temporal.Absolute[]} Temporal.Absolute instance
+ */
+function getSortedInstants(parseableAbsoluteStrings, reverse = false) {
+    const absoluteTimes = parseableAbsoluteStrings.map(v => {
+        return Temporal.Absolute.from(v);
+    });
+
+    const sortedAbsoluteTimes = absoluteTimes.sort((a, b) => {
+        return a.compare(b);
+    });
+
+    return reverse ? sortedAbsoluteTimes.reverse() : sortedAbsoluteTimes;
+}
+
+let a = "2020-01-23T17:04:36.491865121-08:00";
+let b = "2020-02-10T17:04:36.491865121-08:00";
+let c = "2019-03-30T01:45:00+01:00[Europe/Berlin]";
+let d = "2019-03-30T01:45:00+00:00[Europe/London]";
+let e = "2019-03-30T01:45:00-06:00[America/New_York]";
+
+getSortedInstants([a, b, c, d, e]);
--- a/docs/cookbook/sortAbsoluteInstants.mjs
+++ b/docs/cookbook/sortAbsoluteInstants.mjs
@@ -7,7 +7,7 @@ import assert from "assert";
  *
  * @param {string[]} parseableAbsoluteStrings - a group of ISO Strings
  * @param {boolean} reverse ascending or descending order
- * @returns {string[]} Temporal.Absolute instance
+ * @returns {string[]} the array from parseableAbsoluteStrings, sorted
  */
 function getSortedInstants(parseableAbsoluteStrings, reverse = false) {
     const sortedAbsoluteTimes = parseableAbsoluteStrings
--- a/docs/cookbook/sortAbsoluteInstants.mjs
+++ b/docs/cookbook/sortAbsoluteInstants.mjs
@@ -1,28 +1,33 @@
+import assert from "assert";
+
 /**
  * getSortedInstants will sort an array of strings (each of which is parseable as a Temporal.Absolute and may or may not include an IANA time zone name)
  * by the corresponding absolute time (e.g., for presenting global log events sequentially).
  *
  *
- * @param {string[]} parseableAbsoluteStrings This is a Date instance
- * @param {boolean} reverse accending or descending order
+ * @param {string[]} parseableAbsoluteStrings - a group of ISO Strings
+ * @param {boolean} reverse ascending or descending order
  * @returns {Temporal.Absolute[]} Temporal.Absolute instance
  */
 function getSortedInstants(parseableAbsoluteStrings, reverse = false) {
-    const absoluteTimes = parseableAbsoluteStrings.map(v => {
-        return Temporal.Absolute.from(v);
-    });
-
-    const sortedAbsoluteTimes = absoluteTimes.sort((a, b) => {
-        return a.compare(b);
-    });
+    const sortedAbsoluteTimes = parseableAbsoluteStrings.sort(
+        Temporal.Absolute.compare
+    );
 
     return reverse ? sortedAbsoluteTimes.reverse() : sortedAbsoluteTimes;
 }
 
-let a = "2020-01-23T17:04:36.491865121-08:00";
-let b = "2020-02-10T17:04:36.491865121-08:00";
-let c = "2019-03-30T01:45:00+01:00[Europe/Berlin]";
-let d = "2019-03-30T01:45:00+00:00[Europe/London]";
-let e = "2019-03-30T01:45:00-06:00[America/New_York]";
+const a = "2020-01-23T17:04:36.491865121-08:00";
+const b = "2020-02-10T17:04:36.491865121-08:00";
+const c = "2019-03-30T01:45:00+01:00[Europe/Berlin]";
+const d = "2019-03-16T01:45:00+00:00[Europe/London]";
+const e = "2019-03-25T01:45:00-06:00[America/New_York]";
 
-getSortedInstants([a, b, c, d, e]);
+const results = getSortedInstants([a, b, c, d, e]);
+
+// results will have correct order
+assert.equal(results[0], d);
+assert.equal(results[1], e);
+assert.equal(results[2], c);
+assert.equal(results[3], a);
+assert.equal(results[4], b);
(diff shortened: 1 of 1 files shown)


cookbook/sortAbsoluteInstants.mjs
```diff
@@ -0,0 +1,28 @@
...
+ * @param {boolean} reverse accending or descending order
+ * @returns {Temporal.Absolute[]} Temporal.Absolute instance
+ */
+function getSortedInstants(parseableAbsoluteStrings, reverse = false) {
+    const absoluteTimes = parseableAbsoluteStrings.map(v => {
+        return Temporal.Absolute.from(v);
+    });
+
+    const sortedAbsoluteTimes = absoluteTimes.sort((a, b) => {
+        return a.compare(b);
```
ptomato: Same thing as in the other pull request, this would need to be `absoluteTimes.sort(Temporal.Absolute.compare)` to fix the "compare is not a function" problem

cookbook/sortAbsoluteInstants.mjs
```diff
@@ -0,0 +1,28 @@
+/**
+ * getSortedInstants will sort an array of strings (each of which is parseable as a Temporal.Absolute and may or may not include an IANA time zone name)
+ * by the corresponding absolute time (e.g., for presenting global log events sequentially).
+ *
+ *
+ * @param {string[]} parseableAbsoluteStrings This is a Date instance
```
ptomato: Docstring seems wrong for this parameter

cookbook/sortAbsoluteInstants.mjs
```diff
@@ -0,0 +1,28 @@
...
+ * @param {boolean} reverse accending or descending order
```
ptomato: typo: ascending

cookbook/sortAbsoluteInstants.mjs
```diff
@@ -0,0 +1,28 @@
...
+function getSortedInstants(parseableAbsoluteStrings, reverse = false) {
+    const absoluteTimes = parseableAbsoluteStrings.map(v => {
+        return Temporal.Absolute.from(v);
+    });
+
+    const sortedAbsoluteTimes = absoluteTimes.sort((a, b) => {
+        return a.compare(b);
+    });
+
+    return reverse ? sortedAbsoluteTimes.reverse() : sortedAbsoluteTimes;
```
gibson042: #240 asks for the _strings_ to be sorted by corresponding absolute time, not for them to be converted into Temporal.Absolute instances first.

cookbook/sortAbsoluteInstants.mjs
```diff
@@ -0,0 +1,28 @@
...
+}
+
+let a = "2020-01-23T17:04:36.491865121-08:00";
+let b = "2020-02-10T17:04:36.491865121-08:00";
+let c = "2019-03-30T01:45:00+01:00[Europe/Berlin]";
+let d = "2019-03-30T01:45:00+00:00[Europe/London]";
+let e = "2019-03-30T01:45:00-06:00[America/New_York]";
+
+getSortedInstants([a, b, c, d, e]);
```
gibson042: Where is the assertion or demonstration of the result?

jasonwilliams: Are people happy for me to import [assert](https://nodejs.org/api/assert.html) for each of these?

jasonwilliams:
@ptomato @gibson042 updated..

How is this even working?
Looks like `compare` only accepts Temporal.Absolute accord to the documentation, is it being created from a `string` somewhere? 
Does this need to be documented?

cookbook/sortAbsoluteInstants.mjs
ptomato: The way I've been doing it in the snippets in the documentation is just putting the result in a comment after the code.
I'd use assert if we plan to run these cookbook functions as part of the tests or something like that.

ptomato:
Looking good now, I guess I'd defer to Richard though on what the cookbook recipe is actually supposed to do :smile:
Not sure I understand your question. If the idea is to sort the strings (so that the expected return value is an array of strings sorted by time) then I'd use something like decorate-sort-undecorate.

Incidentally it occurred to me that to implement `reverse` without looping through the array a second time you could use something like
```js
let compareFunc = Temporal.Absolute.compare;
if (reverse)
    compareFunc = (a, b) => Temporal.Absolute.compare(b, a)
```

jasonwilliams:
Yeah good idea on the compare.
@ptomato what editor are you using? I have a VSCode task which runs these for me, I can commit the .vscode folder if that’s any use.

It’s working for me, do you get an error? Or the wrong order

jasonwilliams:
For me it’s working which is surprising as it doesn’t look like compare is supposed to accept strings as it’s argument

ptomato:
I get an error, "TypeError: Cannot read property 'slot-epochNanoSeconds' of undefined".

I feel a bit silly for not having thought of this before, but actually `compare` did previously accept strings, although it wasn't supposed to; see #237. And in fact I actually fixed that issue myself :flushed: So maybe your task is using an old version of the polyfill?

I'm using Sublime Text but I run it from the command line, using `node --experimental-modules --no-warnings --icu-data-dir ./polyfill/node_modules/full-icu/ -r ./polyfill/index.js cookbook/sortAbsoluteInstants.mjs`.

jasonwilliams:
Ok built and ran again i now have 
`TypeError: invalid Absolute object`

cookbook/sortAbsoluteInstants.mjs
```diff
@@ -0,0 +1,33 @@
...
+ * getSortedInstants will sort an array of strings (each of which is parseable as a Temporal.Absolute and may or may not include an IANA time zone name)
+ * by the corresponding absolute time (e.g., for presenting global log events sequentially).
+ *
+ *
+ * @param {string[]} parseableAbsoluteStrings - a group of ISO Strings
+ * @param {boolean} reverse ascending or descending order
+ * @returns {Temporal.Absolute[]} Temporal.Absolute instance
+ */
+function getSortedInstants(parseableAbsoluteStrings, reverse = false) {
+    const sortedAbsoluteTimes = parseableAbsoluteStrings.sort(
```
Ms2ger: How about this?

```suggestion
    const sortedAbsoluteTimes = parseableAbsoluteStrings.map(Temporal.Absolute.from).sort(
```

jasonwilliams:
I seem to be having an issue right now
```
Uncaught RangeError: "+01:00" is not a valid value for the disambiguation argument
```

Its not happy with the `2019-03-30T01:45:00+01:00[Europe/Berlin]` string

cookbook/sortAbsoluteInstants.mjs
jasonwilliams: that won't work because the second argument to from() will be the index, and from will incorrectly assume this to be something else. (timezone?)
Leading to this error:
```
    return this === Absolute ? result : new this(GetSlot(result, EPOCHNANOSECONDS));
                                        ^

TypeError: this is not a constructor
```

Ms2ger:
> I seem to be having an issue right now
> 
> ```
> Uncaught RangeError: "+01:00" is not a valid value for the disambiguation argument
> ```
> 
> Its not happy with the `2019-03-30T01:45:00+01:00[Europe/Berlin]` string

This is a bug in the polyfill `ToAbsolute` implementation:

```js
    return datetime.inTimeZone(zone, match[11] ? match[10] : 'earlier');
```

jasonwilliams:
The bug happens here:
https://github.com/tc39/proposal-temporal/blob/main/polyfill/lib/timezone.mjs#L41

disambiguation is set to '+01:00' but it expects either `'earlier', 'later', 'reject'`

jasonwilliams:
@Ms2ger zone is already set for that `inTimeZone` call, what should the second argument be in this case? "earlier" ?

cookbook/sortAbsoluteInstants.mjs
ptomato: The second argument can be safely deleted, it does nothing and isn't supposed to be there. However this error is the result of something more subtle. The `this === Absolute` check is, I guess, supposed to check whether you are calling this method as `Temporal.Absolute.from()` or `SomeSubclass.from()`... but `parseableAbsoluteStrings.map(Temporal.Absolute.from)` overrides the `this` value that `Temporal.Absolute.from()` is called with. That's why the arrow function works, as well as any of these others:
```js
parseableAbsoluteStrings.map(v => Temporal.Absolute.from(v))
parseableAbsoluteStrings.map(Temporal.Absolute.from.bind(Temporal.Absolute))
parseableAbsoluteStrings.map(Temporal.Absolute.from, Temporal.Absolute)
```
I think this is actually a serious gotcha with the API, since `map(Temporal.Absolute.from)` seems quite natural to do.

ptomato:
I think the second argument should probably be `"reject"` there. The disambiguation argument is for `datetime.inTimeZone` in case you have a wall-clock time that doesn't exist, or exists more than once, on that date in that time zone, which happens during DST changes or other discontinuities. But in ToAbsolute where we're parsing an ISO string, the string should not be ambiguous and if it is, I think it's preferable to throw.

cookbook/sortAbsoluteInstants.mjs
gibson042: This is still incorrect; it returns an array of Temporal.Absolute instances rather than an array of the input strings in an order determined by a comparison of Absolute instances derived from them. Like @ptomato mentioned, the expected pattern is decorate-sort-undecorate.

jasonwilliams:
Updated

docs/cookbook/sortAbsoluteInstants.mjs
```diff
@@ -0,0 +1,34 @@
+import assert from "assert";
+
+/**
+ * getSortedInstants will sort an array of strings (each of which is parseable as a Temporal.Absolute and may or may not include an IANA time zone name)
+ * by the corresponding absolute time (e.g., for presenting global log events sequentially).
+ *
+ *
+ * @param {string[]} parseableAbsoluteStrings - a group of ISO Strings
+ * @param {boolean} reverse ascending or descending order
+ * @returns {string[]} Temporal.Absolute instance
```
ptomato: One more thing, this isn't a Temporal.Absolute instance anymore, I'd say something like "the array from parseableAbsoluteStrings, sorted"

PR #346, opened 2020-02-07
Make parameters optional in constructors and from()

ptomato:
Closes: #340


polyfill/lib/date.mjs
```diff
@@ -139,6 +139,7 @@ export class Date {
     return new MonthDay(GetSlot(this, MONTH), GetSlot(this, DAY));
   }
   static from(arg) {
+    if (typeof arg === 'string') return ES.DateFromString(arg);
     let result = ES.ToDate(arg);
```
Ms2ger: I think we agreed to call `ToString` for non-object values, but didn't get around to updating the spec to say that. We should probably preserve that behavior. It looks like this changes also causes us to call `ToObject` on non-string values; I don't think that's desirable. Consider:

```js
Number.prototype.year = 2002;
let { year } = 7;
assert(year === 2002);
// And also
assert(Temporal.Date.from(7).year === 2002);
```

Also, the `*FromString` abstract operations in the spec date back to the time before #252, when we also had `Date.fromString`. I'm not sure they still serve a purpose.

polyfill/lib/ecmascript.mjs
```diff
@@ -107,32 +107,20 @@ export const ES = ObjectAssign(ObjectAssign({}, ES2019), {
...
+      year = 0,
+      month = 1,
+      day = 1,
+      hour = 0,
+      minute = 0,
+      second = 0,
+      millisecond = 0,
+      microsecond = 0,
+      nanosecond = 0,
+    } = item;
```
Ms2ger: Somewhat obscure, but the spec intentionally requires that the `Get` calls are done in lexicographical order. Then again, the current code doesn't do that either.

spec/date.html
```diff
@@ -16,16 +16,22 @@ <h1>The Temporal.Date Constructor</h1>
...
     <emu-clause id="sec-temporal.date">
-      <h1>Temporal.Date ( _year_, _month_, _day_, _disambiguation_ )</h1>
+      <h1>Temporal.Date ( [ _year_ [ , _month_ [ , _day_ [ , _disambiguation_ ] ] ] ] )</h1>
       <p>
         When the `Temporal.Date` function is called, the following steps are taken:
       </p>
       <emu-alg>
         1. If NewTarget is *undefined*, then
           1. Throw a *TypeError* exception.
         1. Let _y_ be ? ToInteger(_year_).
```
Ms2ger: I guess we're now consistently relying on the property `ToInteger(undefined) === 0`. Please consider whether it's worth documenting that.

spec/date.html
```diff
@@ -449,23 +455,20 @@ <h1>DateFromString ( _date_ )</h1>
...
-          1. Return ? DateFromString(_date_).
         1. If Type(_date_) is Object, then
           1. Let _result_ be the Record {
-            [[Year]]: *undefined*,
-            [[Month]]: *undefined*,
-            [[Day]]: *undefined*,
+            [[Year]]: 0,
+            [[Month]]: 1
+            [[Day]]: 1
             }.
```
Ms2ger: Consider whether it would be clearer to put the default values in the table, and if we could use something like "a new Record with all the internal slots given in the Internal Slot column in #table".

polyfill/lib/date.mjs
Ms2ger: ... and we should have tests.

spec/date.html
ptomato: You mean as a side note for clarity? I figured one could easily click through to the actual spec and it would be easy to miss the special "1" values if they looked visually the same as checking for "undefined" and assigning "0" values on the rest of the elements.

ptomato:
> Also, we should probably check if anyone has objections to the change in the title.

I figured creating a pull request and tagging reviewers would be the best way to do that, rather than waiting on comments on #340.

polyfill/lib/date.mjs
ptomato: OK, having that context changes things quite a bit :smile: I'll change this back and change the spec instead.

I think this part is pretty well covered by tests (except for checking that non-object arguments are converted to strings, which I'll add.)

I think having `*FromString` as separate abstract operations makes the spec easier to read, but I'm not sure what's usual regarding when to make something an abstract operation. I've just been winging it :smile:

polyfill/lib/ecmascript.mjs
ptomato: I'll add another commit fixing this.

polyfill/lib/date.mjs
ptomato: Testing that non-object arguments are converted to strings can't currently work because as far as I know there are no non-object, non-string values that would produce a valid ISO string ... until we resolve #198 / #312 when presumably a number like `20200208` would be accepted. I'll add skipped tests for now.

ptomato:
Rebased.

@pipobscure @gibson042 Any objections to making this change?

ptomato:
Updated to implement what we discussed in the 2020-02-13 Temporal meeting. Now year, month, and day are required, and the rest of the parameters are optional.

polyfill/lib/ecmascript.mjs
```diff
@@ -310,6 +311,22 @@ export const ES = ObjectAssign(ObjectAssign({}, ES2019), {
     }
     return any ? any : false;
   },
+  ValidDateTimeFrom: (bag, required, optional = []) => {
+    if (!bag || 'object' !== typeof bag) return false;
+    let result = {};
+    for (let prop of required) {
+      if (!(prop in bag)) throw new TypeError(`required property '${prop}' missing`);
+      const value = ES.ToNumber(bag[prop]);
+      if (Number.isFinite(value)) result[prop] = value;
```
gibson042: This is a bit weird... doesn't it mean that e.g. `Temporal.DateTime.from({ month: 1, day: 1 })` throws an error while `Temporal.DateTime.from({ year: undefined, month: 1, day: 1 })` returns a Temporal.DateTime instance for 0000-01-01T00:00Z?

<details><summary><code>Temporal.DateTime.from({ year: undefined, month: 1, day: 1 })</code></summary>

1. [Temporal.DateTime.from(arg)](https://github.com/tc39/proposal-temporal/blob/8bbaf4c1569ceede161f17bd79d1670e3d20aca3/polyfill/lib/datetime.mjs#L298): `let result = ES.ToDateTime(arg)`
   1. [ES.ToDateTime(item)](https://github.com/tc39/proposal-temporal/blob/8bbaf4c1569ceede161f17bd79d1670e3d20aca3/polyfill/lib/ecmascript.mjs#L110): `const props = ES.ValidDateTimeFrom(item, ['year', 'month', 'day'], …)`
      1. [ES.ValidDateTimeFrom(bag, required, optional = [])](https://github.com/tc39/proposal-temporal/blob/8bbaf4c1569ceede161f17bd79d1670e3d20aca3/polyfill/lib/ecmascript.mjs#L319): `const value = ES.ToNumber(bag[prop])`
         1. [ToNumber(argument)](https://tc39.es/ecma262/#sec-tonumber): argument type undefined ⇒ return **NaN**
      1. [ES.ValidDateTimeFrom(bag, required, optional = [])](https://github.com/tc39/proposal-temporal/blob/8bbaf4c1569ceede161f17bd79d1670e3d20aca3/polyfill/lib/ecmascript.mjs#L320): `if (Number.isFinite(value)) result[prop] = value;` (_condition not satisfied for **NaN**, so property not assigned_)
   1. [ES.ToDateTime(item)](https://github.com/tc39/proposal-temporal/blob/8bbaf4c1569ceede161f17bd79d1670e3d20aca3/polyfill/lib/ecmascript.mjs#L123): `const { hour = 0, day, microsecond = 0, millisecond = 0, minute = 0, month, nanosecond = 0, second = 0, year, } = props` (_`year` is **undefined**_)
   1. [ES.ToDateTime(item)](https://github.com/tc39/proposal-temporal/blob/8bbaf4c1569ceede161f17bd79d1670e3d20aca3/polyfill/lib/ecmascript.mjs#L134): `return new TemporalDateTime(year, month, day, hour, minute, second, millisecond, microsecond, nanosecond, 'reject');`
      1. [Temporal.DateTime.prototype.constructor( year, month, day, … )](https://github.com/tc39/proposal-temporal/blob/8bbaf4c1569ceede161f17bd79d1670e3d20aca3/polyfill/lib/datetime.mjs#L33): `year = ES.ToInteger(year);`
         1. [ToInteger(argument)](https://tc39.es/ecma262/#sec-tointeger): Let <var>number</var> be ?&nbsp;<emu-xref aoid="ToNumber" id="_ref_1810"><a href="https://tc39.es/ecma262/#sec-tonumber">ToNumber</a></emu-xref>(<var>argument</var>).
            1. [ToNumber(argument)](https://tc39.es/ecma262/#sec-tonumber): argument type undefined ⇒ return **NaN**
         1. [ToInteger(argument)](https://tc39.es/ecma262/#sec-tointeger): If <var>number</var> is <emu-val>NaN</emu-val>, <emu-val>+0</emu-val>, or <emu-val>-0</emu-val>, return <emu-val>+0</emu-val>. (_condition satisfied for `year`_)
      1. [Temporal.DateTime.prototype.constructor( year, month, day, … )](https://github.com/tc39/proposal-temporal/blob/8bbaf4c1569ceede161f17bd79d1670e3d20aca3/polyfill/lib/datetime.mjs#L76): `SetSlot(this, YEAR, year);`
</details>

Why should there be a difference between implicit vs. explicit `undefined`?

ljharb: There should not; the spec generally uses `Get`, not `HasProperty`.

ptomato: Fixed this, and added tests where one item of the required properties for Date, DateTime, YearMonth, and MonthDay is undefined, ensuring that from() throws in that case.

PR #353, opened 2020-02-12
Re-enable regex tests

ptomato:
- Re-enable regex tests which were commented out
- Fix bug in generating test expressions which was giving durations with multiple seconds components

### DIFF
--- a/polyfill/test/regex.mjs
+++ b/polyfill/test/regex.mjs
@@ -356,12 +356,14 @@ describe('fromString regex', ()=>{
       day = day.concat(day.map((p) => `${p}${n}`));
     }
 
-    const tp = ['4H', '5M', '6S', '7.123S', '8.123456S', '9.123456789S', '0.123S', '0.123456S', '0.123456789S'];
+    const tp = ['4H', '5M'];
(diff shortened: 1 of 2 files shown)


ptomato:
@pipobscure Re. the discussion in https://github.com/tc39/proposal-temporal/pull/321, let's re-enable those tests for a start.

polyfill/test/regex.mjs
```diff
@@ -356,12 +356,14 @@ describe('fromString regex', ()=>{
...
     }
 
-    const tp = ['4H', '5M', '6S', '7.123S', '8.123456S', '9.123456789S', '0.123S', '0.123456S', '0.123456789S'];
+    const tp = ['4H', '5M'];
+    const sec = ['', '6S', '7.123S', '8.123456S', '9.123456789S', '0.123S', '0.123456S', '0.123456789S'];
     let tim = [''];
     while (tp.length) {
       const n = tp.shift();
       tim = tim.concat(tim.map((p) => `${p}${n}`));
     }
```
Ms2ger: I found this pretty hard to read. Consider writing it out as
```js
let tim = ["", "4H", "5M", "4H5M"];
```
instead.

PR #356, opened 2020-02-13
Changing symbols to strings for Calendar

sffc:
See #310

### DIFF
--- a/docs/calendar-draft.md
+++ b/docs/calendar-draft.md
@@ -42,7 +42,7 @@ class MyCalendar {
 	) : Temporal.Date;
 
 	/** A string identifier for this calendar */
-	id] : string;
+	id : string;
 
 	//////////////////
 	//  Arithmetic  //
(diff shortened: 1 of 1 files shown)


docs/calendar-draft.md
```diff
@@ -32,38 +32,38 @@ class MyCalendar {
...
 
 	/** Returns the projection of isoDate in the custom calendar */
-	[Temporal.Calendar.fromISO](
+	fromISO(
 		isoDate: Temporal.Date
 	) : Temporal.Date;
 
 	/** A string identifier for this calendar */
-	[Temporal.Calendar.id] : string;
+	id] : string;
```
ljharb:
```suggestion
	id : string;
```

altho this should ideally be an accessor on the prototype, rather than an own property?

sffc:
The `id` field is super provisional right now, and may be removed once we reach a conclusion on #293.

PR #363, opened 2020-02-14
Editorial: Define DateToString abstract operation.

Ms2ger:


spec/date.html
```diff
@@ -558,6 +551,18 @@ <h1>PadYear ( _y_ )</h1>
       </emu-alg>
     </emu-clause>
 
+    <emu-clause id="sec-temporal-datetostring" aoid="DateToString">
+      <h1>DateToString ( _date_ )</h1>
+      <emu-alg>
+        1. Assert: Type(_date_) is Object.
+        1. Assert: _date_ has an [[InitializedTemporalDate]] internal slot.
+        1. Let _year_ be ! PadYear(_date_.[[Year]]).
```
ptomato: As a side note, I don't think the definition for PadYear is currently correct, but that doesn't have any bearing on whether or not to merge this.

PR #364, opened 2020-02-14
Start filling in the specification around RegulateDate.

Ms2ger:


spec/abstractops.html
```diff
@@ -264,6 +264,13 @@ <h1>NonNegativeModulo ( _x_, _y_ )</h1>
     </emu-alg>
   </emu-clause>
 
+  <emu-clause id="sec-temporal-constraintorange" aoid="ConstrainToRange">
+    <h1>ConstrainToRange ( _x_, _minimum_, _maximum_ )</h1>
```
ljharb: This kind of “ClampToRange” operation seems like it’d be generally useful in the spec; there's 9 instances of `min(max(x, minimum), maximum)` in there (and 1 of `max(min(x, maximum), minimum)`). any interest in making an editorial PR, in advance of Temporal?

spec/date.html
```diff
@@ -546,6 +552,29 @@ <h1>ValidateDate ( _y_, _m_, _d_ )</h1>
       </emu-alg>
     </emu-clause>
 
+    <emu-clause id="sec-temporal-constraindate" aoid="ConstrainDate">
+      <h1>ConstrainDate ( _year_, _month_, _day_ )</h1>
+      <emu-alg>
+        1. Assert: _year_, _month_, and _day_ are integer Number values.
+        1. Set _year_ to ConstrainToRange(_year_, -999999, 999999).
```
ptomato: See #198, I'm not sure it's settled that this is the range of years. Then again, I think this is what the polyfill currently does, so no need to block merging on that.

spec/date.html
```diff
@@ -546,6 +552,29 @@ <h1>ValidateDate ( _y_, _m_, _d_ )</h1>
...
+        1. Set _day_ to ConstrainToRange(_day_, 1, DaysInMonth(_year_, _month_));
+        1. Return the Record {
+          [[Year]]: _year_,
+          [[Month]]: _month_,
+          [[Day]]: _day_,
+          }.
+      </emu-alg>
+    </emu-clause>
+
+    <emu-clause id="sec-temporal-balancedate" aoid="BalanceDate">
```
ptomato: I have a BalanceDate draft in https://github.com/tc39/proposal-temporal/pull/333/commits/e79b53a9efc8e597bb1728f24b2f86640c704718 which is in my pull request #333, consider cherry picking that commit into this pull request if it looks reasonable?

spec/monthday.html
```diff
@@ -325,17 +325,12 @@ <h1>ToPartialMonthDay ( _monthDay_ )</h1>
...
-        1. If _disambiguation_ is `"balance"`, then
-          1. Return ! BalanceMonthDay(_month_, _day_).
-        1. If _disambiguation_ is `"reject"`, then
-          1. If ! ValidateMonthDay(_month_, _day_) is *false*, then
-            1. Throw a *RangeError* exception.
-          1. Return the new Record {
-              [[Month]]: _month_,
-              [[Day]]: _day_
-            }.
+        1. Let _leapYear_ be 1972.
```
ptomato: Any reason for 1972 in particular?

Ms2ger: Any leap year will do; this one was close to 1970 which was used in the polyfill before (presumably based on the epoch).

spec/abstractops.html
Ms2ger: No interest on my side, but feel free to take the idea and run with it.

PR #365, opened 2020-02-14
Regex tests

ptomato:
Refactor the regex tests to test the behaviour of from() instead of the implementation (regex match).


ptomato:
@pipobscure What do you think about this, since you wrote these tests? It's what I mentioned in https://github.com/tc39/proposal-temporal/pull/321#issuecomment-585349356. My feeling is that only checking that the regex matches, and not what was done with the match, was hiding some bugs like the other patch fixed here, where the capturing groups in the regex had gone out of sync with the indices accessed on the match object in the code.

ptomato:
OK, thanks for the reviews, with two approving I think I'll merge this, happy to revisit if there's some intention of the original code that I'm missing.

Issue #24, opened 2017-07-08
TODO: Define min/max values


mj1856:
All of the objects (except perhaps `ZonedInstant`) should have a `MIN_VALUE` and `MAX_VALUE` constant.

mj1856:
For `Instant`, we will use the same range as the old `Date` type, which is ±100,000,000 days surrounding the epoch.  Since we have nanosecond precision, that works out to be `-271821-04-20T00:00:00.000000000Z` through `275760-09-13T00:00:00.000000000Z`.  We should probably restrict `LocalDate` and `LocalDateTime` to values at or near here, but we should consider the edge cases also.

ryzokuken:
@maggiepint is this behavior still undefined? I couldn’t see any maximum or minimum values on the polyfill.

littledan:
Do we actually want this API? I'm not convinced we do. I don't think "Number has it" is a good reason, and I don't understand the motivation. It'd be good to hear from @pdunkel  on this question.

gibson042:
We definitely want to _have_ minimum and maximum values and it makes sense to align them with Date, but that doesn't mean they must be exposed as properties.

littledan:
Why will we have minimum and maximum values?

ryzokuken:
I guess this isn't relevant anymore since we are already adhering to a subset of the ISO format, `Duration` for example already implements the restrictions mentioned above. Closing, but feel free to reopen if necessary.

ptomato:
Based on discussion in #198 I think we should reopen this. It was decided so far in the thread that Temporal.Absolute has the same range as old-style Date (epoch ± 1e8 days) but this behaviour needs to be reflected in the spec.

Also we need to decide on a range for Temporal.DateTime, Temporal.Date, and Temporal.YearMonth; should they be the same as Temporal.Absolute (in which case you have to throw a RangeError when doing `Temporal.Absolute.from(maxAbsolute).inTimeZone('+01:00')`, for example) or should they encompass an extra day around either end (in which case you have to throw a RangeError when doing `Temporal.DateTime.from(maxDateTime).inTimeZone('UTC')`)?

Temporal.Duration doesn't currently have any limits at all: `new Temporal.Duration(Infinity).years` is accepted.

gibson042:
[...]

PR #366, opened 2020-02-16
Documentation for Temporal.Date

ptomato:
Brings the documentation in line with the current state of the polyfill.

### DIFF
--- a/docs/date.md
+++ b/docs/date.md
@@ -1,96 +1,392 @@
 # Temporal.Date
 
-A representation of a calendar date.
+A `Temporal.Date` represents a calendar date.
+"Calendar date" refers to the concept of a date as expressed in everyday usage, independent of any time zone.
+For example, it could be used to represent an event on a calendar which happens during the whole day no matter which time zone it's happening in.
+
+`Temporal.Date` refers to the whole of a specific day; if you need to refer to a specific time on that day, use `Temporal.DateTime`.
+A `Temporal.Date` can be converted into a `Temporal.DateTime` by combining it with a `Temporal.Time` using the `withTime()` method.
+
+`Temporal.YearMonth` and `Temporal.MonthDay` carry less information than `Temporal.Date` and should be used when complete information is not required.
 
 ## Constructor
 
-### **new Temporal.Date**(_year_: number, _month_: number, _day_: number, _disambiguation_: 'constrain' | 'balance' | 'reject' = 'constrain') : Temporal.Date
+### **new Temporal.Date**(_isoYear_: number, _isoMonth_: number, _isoDay_: number, _disambiguation_: 'constrain' | 'balance' | 'reject' = 'constrain') : Temporal.Date
 
-Creates a new `Date` object that represents a calendar date
+**Parameters:**
+- `isoYear` (number): A year.
+- `isoMonth` (number): A month, ranging between 1 and 12 inclusive.
+- `isoDay` (number): A day of the month, ranging between 1 and 31 inclusive.
+- `disambiguation` (optional string): How to deal with out-of-range values of the other parameters.
+  Allowed values are `constrain`, `balance`, and `reject`.
+  The default is `constrain`.
+
+**Returns:** a new `Temporal.Date` object.
+
+Use this constructor if you have the correct parameters for the date already as individual number values, or you need the disambiguation behaviour.
+Otherwise, `Temporal.Date.from()`, which accepts more kinds of input, is probably more convenient.
+
+All values are given as reckoned in the [ISO 8601 calendar](https://en.wikipedia.org/wiki/ISO_8601#Dates).
+
+The `disambiguation` parameter works as follows:
+- In `constrain` mode (the default), any out-of-range values are clamped to the nearest in-range value.
+- In `balance` mode, any out-of-range values are resolved by balancing them with the next highest unit.
+- In `reject` mode, the presence of out-of-range values will cause the constructor to throw a `RangeError`.
+
+Usage examples:
+```javascript
+// Pi day in 2020
+date = new Temporal.Date(2020, 3, 14)  // => 2020-03-14
+
+// Different disambiguation modes
+date = new Temporal.Date(2001, 13, 1, 'constrain')  // => 2001-12-01
+date = new Temporal.Date(2001, -1, 1, 'constrain')  // => 2001-01-01
+date = new Temporal.Date(2001, 13, 1, 'balance')  // => 2002-01-01
+date = new Temporal.Date(2001, -1, 1, 'balance')  // => 2000-11-01
+date = new Temporal.Date(2001, 13, 1, 'reject')  // throws
+date = new Temporal.Date(2001, -1, 1, 'reject')  // throws
+```
 
 ## Static methods
 
 ### Temporal.Date.**from**(_thing_: string | object) : Temporal.Date
 
+**Parameters:**
+- `thing` (string or object): The value representing the desired date.
+
+**Returns:** a new `Temporal.Date` object (or the same object if `thing` was a `Temporal.Date` object.)
+
+This static method creates a new `Temporal.Date` object from another value.
+If the value is a string, it must be in ISO 8601 format.
+If the string designates a time zone, the time zone will be ignored.
+If the value is another `Temporal.Date` object, the same object is returned.
+If the value is any other object, it must have `year`, `month`, and `day` properties, and a `Temporal.Date` will be constructed from them.
+
+Note that any time or time zone part of an ISO 8601 string passed to this function is optional, and will be ignored.
+
+Example usage:
+```javascript
+date = Temporal.Date.from('2006-08-24');  // => 2006-08-24
+date = Temporal.Date.from('2006-08-24T15:43:27');  // => 2006-08-24
+date = Temporal.Date.from('2006-08-24T15:43:27Z');  // => 2006-08-24
+date = Temporal.Date.from('2006-08-24T15:43:27+01:00[Europe/Brussels]');
+  // => 2006-08-24
+date === Temporal.Date.from(date)  // => true
+
+date = Temporal.Date.from({year: 2006, month: 8, day: 24});  // => 2006-08-24
+date = Temporal.Date.from(Temporal.DateTime.from('2006-08-24T15:43:27'));
+  // => same as above; Temporal.DateTime has year, month, and day properties
+```
+
 ### Temporal.Date.**compare**(_one_: Temporal.Date, _two_: Temporal.Date) : number
 
-Allows for easier comparison of `Date` objects, returns:
+**Parameters:**
+- `one` (`Temporal.Date`): First date to compare.
+- `two` (`Temporal.Date`): Second date to compare.
 
-- `-1` if the first object represents a lower value than the second
-- `0` if the 2 objects represent the same value
-- `1` if the first object represents a higher value than the second
+**Returns:** &minus;1, 0, or 1.
+
+Compares two `Temporal.Date` objects.
+Returns an integer indicating whether `one` comes before or after or is equal to `two`.
+- &minus;1 if `one` comes before `two`;
+- 0 if `one` and `two` are the same;
+- 1 if `one` comes after `two`.
+
+This function can be used to sort arrays of `Temporal.Date` objects.
+For example:
+```javascript
+one = Temporal.Date.from('2006-08-24');
+two = Temporal.Date.from('2015-07-14');
+three = Temporal.Date.from('1930-02-18');
+sorted = [one, two, three].sort(Temporal.Date.compare);
+sorted.join(' ');  // => 1930-02-18 2006-08-24 2015-07-14
+```
 
 ## Properties
 
 ### date.**year** : number
 
-Returns the year this `Date` represents
-
 ### date.**month** : number
-
-Returns the month this `Date` represents
 
 ### date.**day** : number
 
-Returns the day this `Date` represents
+The above read-only properties allow accessing each component of the date individually.
+
+Usage examples:
+```javascript
+date = Temporal.Date.from('2006-08-24');
+date.year   // => 2006
+date.month  // => 8
+date.day    // => 24
+```
 
 ### date.**dayOfWeek** : number
 
-Returns the day of the week this `Date` represents
+The `dayOfWeek` read-only property gives the weekday number that the date falls on.
+The weekday number is defined as in the ISO 8601 standard: a value between 1 and 7, inclusive, with Monday being 1, and Sunday 7.
+For an overview, see [ISO 8601 on Wikipedia](https://en.wikipedia.org/wiki/ISO_8601#Week_dates).
 
-```js
-const date = new Temporal.Date(2019, 11, 18); // Monday
-date.dayOfWeek; // 1
+Usage example:
+```javascript
+date = Temporal.Date.from('2006-08-24');
+['MON', 'TUE', 'WED', 'THU', 'FRI', 'SAT', 'SUN'][date.dayOfWeek - 1]  // => THU
+```
+
+### date.**dayOfYear** : number
+
+The `dayOfYear` read-only property gives the ordinal day of the year that the date falls on.
+This is a value between 1 and 365, or 366 in a leap year.
+
+Usage example:
+```javascript
+date = Temporal.Date.from('2006-08-24');
+// ISO ordinal date
+console.log(date.year, date.dayOfYear);  // 2006 236
 ```
 
 ### date.**weekOfYear** : number
 
-Returns the week in a calendar year
+The `weekOfYear` read-only property gives the ISO week number of the date.
+This is normally a value between 1 and 52, but in a few cases it can be 53 as well.
+ISO week 1 is the week containing the first Thursday of the year.
+For more information on ISO week numbers, see for example the Wikipedia article on [ISO week date](https://en.wikipedia.org/wiki/ISO_week_date).
+
+Usage example:
+```javascript
+date = Temporal.Date.from('2006-08-24');
+// ISO week date
+console.log(date.year, date.weekOfYear, date.dayOfWeek);  // 2006 34 4
+```
 
 ### date.**daysInMonth** : number
 
-Returns the number of days in the month the `Date` Object represents
+The `daysInMonth` read-only property gives the number of days in the month that the date falls in.
+This is 28, 29, 30, or 31, depending on the month and whether the year is a leap year.
+
+Usage example:
+```javascript
+// Attempt to write some mnemonic poetry
+const monthsByDays = {};
+for (let month = 1; month <= 12; month++) {
+    const date = Temporal.now.date().with({month});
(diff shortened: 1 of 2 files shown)


docs/date.md
```diff
@@ -1,96 +1,398 @@
...
+
+`Temporal.YearMonth` and `Temporal.MonthDay` carry less information than `Temporal.Date` and should be used when complete information is not required.
 
 ## Constructor
 
 ### **new Temporal.Date**(_year_: number, _month_: number, _day_: number, _disambiguation_: 'constrain' | 'balance' | 'reject' = 'constrain') : Temporal.Date
 
-Creates a new `Date` object that represents a calendar date
+**Parameters:**
+- `year` (number): A year, ranging between &minus;271821 and 275760 inclusive.
```
sffc: Where do these numbers come from?

docs/date.md
```diff
@@ -1,96 +1,398 @@
...
+- `month` (number): A month, ranging between 1 and 12 inclusive.
```
sffc: I would rename the parameters to `isoYear`, `isoMonth`, and `isoDay` according to what we agreed regarding the constructor behavior.

docs/date.md
```diff
@@ -1,96 +1,398 @@
...
+- `day` (number): A day of the month, ranging between 1 and 31 inclusive.
+- `disambiguation` (optional string): How to deal with out-of-range values of the other parameters.
```
sffc: No disambiguation parameter on the constructors.

docs/date.md
```diff
@@ -1,96 +1,398 @@
...
 
 ## Constructor
 
 ### **new Temporal.Date**(_year_: number, _month_: number, _day_: number, _disambiguation_: 'constrain' | 'balance' | 'reject' = 'constrain') : Temporal.Date
 
-Creates a new `Date` object that represents a calendar date
+**Parameters:**
+- `year` (number): A year, ranging between &minus;271821 and 275760 inclusive.
+- `month` (number): A month, ranging between 1 and 12 inclusive.
+- `day` (number): A day of the month, ranging between 1 and 31 inclusive.
```
sffc: Please add calendar as an argument.

docs/date.md
```diff
@@ -1,96 +1,398 @@
...
+- `disambiguation` (optional string): How to deal with out-of-range values of the other parameters.
+  Allowed values are `constrain`, `balance`, and `reject`.
+  The default is `constrain`.
+
+**Returns:** a new `Temporal.Date` object.
+
+Use this constructor if you have the correct parameters for the date already as individual number values, or you need the disambiguation behaviour.
```
sffc: No disambiguation on the constructor.

docs/date.md
```diff
@@ -1,96 +1,398 @@
...
+**Parameters:**
+- `thing` (string or object): The value representing the desired date.
+
+**Returns:** a new `Temporal.Date` object (or the same object if `thing` was a `Temporal.Date` object.)
+
+This static method creates a new `Temporal.Date` object from another value.
+If the value is a string, it must be in ISO 8601 format.
+If the string designates a time zone, the time zone will be ignored.
+If the value is another `Temporal.Date` object, the same object is returned.
+If the value is any other object, it must have `year`, `month`, and `day` properties, and a `Temporal.Date` will be constructed from them.
```
sffc: This is not precisely correct.  The fields in .from will be passed verbatim to the default calendar, or to a custom calendar if it is provided in the option bag.  For example, `Temporal.Date.from({ era: "reiwa", year: 1, month: 1, day: 1, calendar: "japanese" })` is a valid thing to do; in this case, the options bag has an extra field.

docs/date.md
```diff
@@ -1,96 +1,398 @@
...
-Returns the year this `Date` represents
-
 ### date.**month** : number
 
-Returns the month this `Date` represents
-
 ### date.**day** : number
 
-Returns the day this `Date` represents
+The above read-only properties allow accessing each component of the date individually.
```
sffc: They are getters that return the calendar-specific year/month/day, per #291.  They don't return raw data.

docs/date.md
```diff
@@ -1,96 +1,398 @@
...
+date = Temporal.Date.from('2006-08-24');
+date.year   // => 2006
+date.month  // => 8
+date.day    // => 24
+```
 
 ### date.**dayOfWeek** : number
 
-Returns the day of the week this `Date` represents
+The `dayOfWeek` read-only property gives the weekday number that the date falls on.
```
sffc: **Here and elsewhere:** In general, you should say that these properties are calendar-specific.  Something like:

> The `dayOfWeek` getter returns the weekday number that the date falls on, calling the corresponding function on the `calendar` field.

docs/date.md
```diff
@@ -1,96 +1,398 @@
...
+The weekday number is a value between 1 and 7, inclusive: Monday is 1, and Sunday is 7.
```
sffc: According to the ISO 8601 convention.

docs/date.md
```diff
@@ -1,96 +1,398 @@
...
+The `dayOfYear` read-only property gives the ordinal day of the year that the date falls on.
+This is a value between 1 and 365, or 366 in a leap year.
+
+Usage example:
+```javascript
+date = Temporal.Date.from('2006-08-24');
+// Print date in ISO ordinal date format
+year = `${date.year}`.padStart(4, '0');
+day = `${date.dayOfYear}`.padStart(3, '0');
+`${year}-${day}`  // => 2006-236
```
sffc: **Here and elsewhere:**  I see what you're trying to do, but please, let's not put string formatting in the docs.  We don't want to encourage that behavior.  People should use to[ISO]String or toLocaleString.  Just do something like `console.log(date.year, date.dayOfYear)` without the string formatting.

docs/date.md
```diff
@@ -1,96 +1,398 @@
...
+```javascript
+date = Temporal.now.date();
+`The year is ${Math.round(date.dayOfYear / date.daysInYear * 100)}% over!`
+// example output: "The year is 10% over!"
+```
 
 ### date.**leapYear** : boolean
 
-Returns a boolean indicating whether the `Date` is in a leap year or not
+The `leapYear` read-only property tells whether the year that the date falls in is a leap year or not.
```
sffc: Should this be named `isLeapYear` instead, to emphasize that it is a yes or no answer?

docs/date.md
```diff
@@ -1,96 +1,398 @@
...
+date.with({year: 2100}).leapYear  // => false
+```
 
 ## Methods
 
 ### date.**with**(_dateLike_: object, _disambiguation_: 'constrain' | 'balance' | 'reject' = 'constrain') : Temporal.Date
 
-### date.**plus**(_duration_: Temporal.Duration | object | string, _disambiguation_: 'constrain' | 'balance' | 'reject' = 'constrain') : Temporal.Date
+**Parameters:**
+- `dateLike` (object): an object with some or all of the properties of a `Temporal.Date`.
```
sffc: Again, calendar-specific fields.

docs/date.md
```diff
@@ -1,96 +1,398 @@
...
+- `thing` (string or object): The value representing the desired date.
+
+**Returns:** a new `Temporal.Date` object (or the same object if `thing` was a `Temporal.Date` object.)
+
+This static method creates a new `Temporal.Date` object from another value.
+If the value is a string, it must be in ISO 8601 format.
+If the string designates a time zone, the time zone will be ignored.
+If the value is another `Temporal.Date` object, the same object is returned.
+If the value is any other object, it must have `year`, `month`, and `day` properties, and a `Temporal.Date` will be constructed from them.
+
```
sffc: Where's the disambiguation option in .from?

docs/date.md
```diff
@@ -1,96 +1,398 @@
...
 
-Returns a new `Date` object which is the sum of the current object plus the additional argument.
+The `duration` argument can be any value that could be passed to `Temporal.Duration.from()`:
+- a `Temporal.Duration` object;
+- any object with properties denoting a duration, such as `{ days: 5 }`;
+- a string in ISO 8601 duration format, such as `P5D`.
 
-### date.**minus**(_duration_: Temporal.Duration | object | string, _disambiguation_: 'constrain' | 'balance' | 'reject' = 'constrain') : Temporal.Date
+However, the `duration` argument must be in whole days.
+For example, `{ days: 5, hours: 0 }` is allowed, but `{ days: 5, hours: 12 }` will throw a `RangeError`.
```
sffc: Let's not document this until we've reached a conclusion on #324.

docs/date.md
```diff
@@ -1,96 +1,398 @@
...
+date = Temporal.Date.from('2019-03-31')
+date.minus({months: 1}, 'constrain')  // => 2019-02-28
+date.minus({months: 1}, 'balance')  // => 2019-03-03
+date.minus({months: 1}, 'reject')  // => throws
+```
 
 ### date.**difference**(_other_: Temporal.Date) : Temporal.Duration
 
-Returns a new `Duration` object which is the difference between the current `Date` object and the argument `Date` value.
+**Parameters:**
```
sffc: How about the new cutoff parameter?  I assume it still defaults to "days".

docs/date.md
```diff
@@ -1,96 +1,398 @@
...
+Usage example:
+```javascript
+date = Temporal.Date.from('2006-08-24');
+date.difference(Temporal.Date.from('2019-01-31'))  // => P12Y5M7D
+```
 
 ### date.**toString**() : string
 
-Returns an ISO 8601 string representing the current `Date` object
+**Returns:** a string in the ISO 8601 date format representing `date`.
```
sffc: We're discussing the behavior of this method in #262

docs/date.md
```diff
@@ -1,96 +1,398 @@
...
+```js
+date = Temporal.Date.from('2006-08-24');
+date.toString();  // => 2006-08-24
+```
 
 ### date.**toLocaleString**(_locale_?: string, _options_?: object) : string
 
-Returns a string with a locally sensitive representation of the specified `Date` object. Overrides the `Object.prototype.toLocaleString()` method.
+**Parameters:**
+- `locales` (optional string or array of strings): A string with a BCP 47 language tag with an optional Unicode extension key, or an array of such strings.
```
sffc: If you haven't already done so, you can copy the boilerplate that we have on, e.g., `Date.prototype.toLocaleString`.

docs/date.md
jasonwilliams: I'd be interested to know this too

docs/date.md
```diff
@@ -1,96 +1,398 @@
...
+date = new Temporal.Date(2001, -1, 1, 'reject')  // throws
+```
 
 ## Static methods
 
 ### Temporal.Date.**from**(_thing_: string | object) : Temporal.Date
 
-### Temporal.Date.**compare**(_one_: Temporal.Date, _two_: Temporal.Date) : number
+**Parameters:**
+- `thing` (string or object): The value representing the desired date.
```
jasonwilliams: Do we want to give more detail here on what the string should/should not be?
Is it just an ISO 8601 string? the output from Date() also?

docs/date.md
jasonwilliams: weren't `isoYear`, `isoMonth`, and `isoDay` the internal slots of Date, not neccesarily the arguments passed in?

docs/date.md
gibson042: It comes from legacy Date, cf. https://tc39.es/ecma262/#sec-time-values-and-time-range
> A Number can exactly represent all integers from -9,007,199,254,740,992 to 9,007,199,254,740,992 (<a href="https://tc39.es/ecma262/#sec-number.min_safe_integer">20.1.2.8</a> and <a href="https://tc39.es/ecma262/#sec-number.max_safe_integer">20.1.2.6</a>). A time value supports a slightly smaller range of -8,640,000,000,000,000 to 8,640,000,000,000,000 milliseconds. This yields a supported time value range of exactly -100,000,000 days to 100,000,000 days relative to midnight at the beginning of 01 January, 1970 UTC.

An explanatory reference would be useful.

ptomato:
@sffc I'm primarily trying to document what is, not what ought to be, for two reasons — 1, because I think documenting things that don't exist yet would be off-putting for curious JS programmers trying out the polyfill; 2, because I believe it's more practical for me to get these pull requests landed ASAP in order to provide a base for others to add onto, and have new features be documented as part of landing those features.

If it's OK with you, for that reason I'd prefer that the changes for things like calendar support and moving the disambiguation out of the constructors be documented as part of solving those respective tickets.

docs/date.md
ptomato: I checked in the notes from the last call and what @jasonwilliams said is correct.

docs/date.md
ptomato: This came from #24. However according to my own principle below I should've left this out — currently the limit is -999999 to +999999 in the polyfill and spec. I'll remove this for this revision and it can be added to the docs as part of #24.

docs/date.md
ptomato: Point taken, although that does make these examples less exciting; and people can't actually get these date formats from toString or toLocaleString.

docs/date.md
ptomato: From reading that issue it seems to only apply to toLocaleString?

docs/date.md
ptomato: Opened https://github.com/tc39/proposal-temporal/issues/369 for this.

docs/date.md
ptomato: For now I've included an abridged version of https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DateTimeFormat with a link to the whole thing. I do think we should include more complete information on each doc page eventually, but I'd like to see how the implementation of calendars ends up affecting this method before going into more detail.

docs/date.md
ptomato: Currently the text says "string must be in ISO 8601 format" and that's what the polyfill accepts. The spec is unclear until we solve #312 and #198. I'm not overly excited about accepting the output of Date.toString() since according to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toString it was implementation-dependent before ES2018, and I can also affect what is printed by changing my `LANG` or `LC_ALL` environment variables.

I opened https://github.com/tc39/proposal-temporal/issues/370 for this.

sffc:
Although I agree in principle with taking baby steps, this PR amounts to a rewrite of the Temporal.Date docs, and we do already have agreement on most of the components of Calendar.  You can refer to calendar-draft.md for more detail.  This PR seems like the right time to update the docs to reflect the reality in the Calendar world.

docs/date.md
sffc: Yes, `isoYear`, `isoMonth`, and `isoDay` are internal slots, but the constructor is where you directly set values to them.  It is misleading to name them `year`, `month`, and `day` since those names are for the calendar-specific year, month, and day.

docs/date.md
```diff
@@ -1,96 +1,389 @@
...
+for (let month = 1; month <= 12; month++) {
+    const date = Temporal.now.date().with({month});
+    monthsByDays[date.daysInMonth] = (monthsByDays[date.daysInMonth] || []).concat(date);
+}
+
+const strings = monthsByDays[30].map(date => date.toLocaleString('en', {month: 'long'}));
+// Shuffle to improve poem as determined empirically
+strings.unshift(strings.pop());
+const last = strings.pop();
+const poem = `Thirty days hath ${strings.join(', ')}, and ${last}`;
```
sffc: You use `toLocaleString()`, which is nice, but the way to format a list is to use Intl.ListFormat.

docs/date.md
```diff
@@ -1,96 +1,389 @@
...
 ### date.**daysInYear** : number
 
-Returns the number of days in the year the `Date` Object represents
+The `daysInYear` read-only property gives the number of days in the year that the date falls in.
+This is 365 or 366, depending on whether the year is a leap year.
+
+Usage example:
+```javascript
+date = Temporal.now.date();
+`The year is ${Math.round(date.dayOfYear / date.daysInYear * 100)}% over!`
```
sffc: Better (not perfect, but better):

```javascript
const date = Temporal.now.date();
const percent = date.dayOfYear / date.daysInYear;
console.log(`The year is ${percent.toLocaleString("en", { style: "percent" })} over!`);
```

docs/date.md
sffc: Ok, thanks

docs/date.md
sffc: No, that issue is specifically about toString.  It has nothing to do with toLocaleString.

You can leave this as-is for the time being though since we don't yet have a concrete conclusion on #262.

littledan:
This PR makes a lot of improvements in terms of documenting things we agree on--thanks for the good work. We've agreed on a lot of details of calendars, and I don't think we necessarily need to hold back the docs to the current state of the polyfill. But I think we'd benefit letting @ptomato decide how his work is chunked up into PRs--this makes things easier to develop and review. The plan would still be to hold back on a big release announcement until everything is done, so until then, I think it's fine to proceed incrementally.

docs/date.md
sffc: (later)

docs/date.md
sffc: (later)

docs/date.md
sffc: (later)

docs/date.md
sffc: (later)

docs/date.md
sffc: (later)

docs/date.md
sffc: (later)

docs/date.md
sffc: (later)

docs/date.md
sffc: (later)

sffc:
Ok, this PR is updating the docs to reflect the current state of the *polyfill*, which is behind the decisions from recently resolved issues.  Can you update the PR description to say that?  I resolved most of my issues with "(later)", assuming they will be fixed when the corresponding issues are resolved.  I still have a few small ones that we can fix now.

ptomato:
Sure! Sorry, it wasn't intentional to imply that the PR reflected the current consensus. Only that I feel it should be kept in sync with the current state of the polyfill.

I've also opened #382 to add a warning label to the docs, to make this clearer both to us and to casual readers.

docs/date.md
ptomato: OK, I understand now.

docs/date.md
ptomato: I assume it's OK to do this later along with the rest?

docs/date.md
ptomato: I've read over #262 several times now and can't see how it relates to toString... I must be missing something, but I'm happy to leave this until that issue is resolved.

docs/date.md
ptomato: I did not know about that! It seems to be supported in the latest version of Node, despite MDN saying it isn't (out of date?)

docs/date.md
sffc: ok

docs/date.md
sffc: Gah, that's the wrong issue.  Take a look at #293. :)

PR #371, opened 2020-02-18
Remove arithmetic from MonthDay

ptomato:
As discussed in the ticket, we'll consider MonthDay arithmetic undefined
without a year attached, because of not knowing whether the year is a
leap year or not.

Also remove MonthDay.daysInMonth which is meaningless without a year and
was not in the spec or polyfill anyway.

Closes: #261

### DIFF
--- a/docs/monthday.md
+++ b/docs/monthday.md
@@ -16,17 +16,9 @@
 
 ### monthDay.**day** : number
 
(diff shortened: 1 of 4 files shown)


ptomato:
Ran into this while trying to document MonthDay and found the open issue.

PR #374, opened 2020-02-18
Bring calendar-draft.md up to date with #290 and #291

sffc:


docs/calendar-draft.md
```diff
@@ -219,23 +236,34 @@ Temporal.Date.prototype.toString = function() {
...
 ```
 
-For objects with time components (such as Temporal.DateTime), the calendar would be appended to the end of the string.  It would be distinguishable from the time zone because it does not contain a slash.  For example: `2019-12-06T16:23+00:50[America/NewYork][hebrew]`.  @gibson042 points out that this could be probematic: "There are many aliases without / ... [including] [#156](https://github.com/tc39/proposal-temporal/issues/156). And it gets worse with author-defined time zone and calendar names."
+In this scenario, `Temporal.Date.from` would take a new optional `options` argument, with a single field `idToCalendar` specifying a function to map from identifiers to Calendar objects.  If not present, the `Intl.Calendar` namespace will be searched.  Example call site with custom calendars:
+
+```javascript
+const fooCalendar = new FooCalendar();
 
-Alternatively, we may consider changing the syntax to add `c=` for calendars and `z=` for zones.  `2019-12-06T16:23+00:50[z=America/NewYork][c=hebrew]`
+Temporal.Date.from("2019-12-03[foo]", {
```
ptomato: I'd guess this should be `[c=foo]` (and we specify that Temporal strips the `c=` before passing the identifier to `idToCalendar`)?

docs/calendar-draft.md
```diff
@@ -43,6 +35,11 @@ class MyCalendar {
 		isoDate: Temporal.Date
 	) : Temporal.Date;
 
+	/** Constructs a Temporal.Date from a free-form option bag */
+	fromFields(
```
ptomato: Minor bikeshedding, but IMO invoking this as `someCalendar.fromFields(someFields)` doesn't sound like what this method is described to do. Maybe `dateFromFields()` would be clearer?

sffc: Done.  Also added an example for how this method would be called from `Temporal.Date.from`.

docs/calendar-draft.md
sffc: Fixed

PR #375, opened 2020-02-18
Specify ConstrainYearMonth.

Ms2ger:


spec/yearmonth.html
```diff
@@ -385,6 +385,19 @@ <h1>BalanceYearMonth ( _year_, _month_ )</h1>
       </emu-alg>
     </emu-clause>
 
+    <emu-clause id="sec-temporal-constraindate" aoid="ConstrainDate">
+      <h1>ConstrainYearMonth ( _year_, _month_ )</h1>
+      <emu-alg>
+        1. Assert: _year_ and _month_ are integer Number values.
+        1. Set _year_ to ConstrainToRange(_year_, -999999, 999999).
```
ptomato: I think these need a `!` for "assert no abrupt completion"? Or is that optional when it's just a mathematical function?

spec/date.html
```diff
@@ -540,12 +540,11 @@ <h1>ValidateDate ( _y_, _m_, _d_ )</h1>
       <h1>ConstrainDate ( _year_, _month_, _day_ )</h1>
       <emu-alg>
         1. Assert: _year_, _month_, and _day_ are integer Number values.
-        1. Set _year_ to ConstrainToRange(_year_, -999999, 999999).
-        1. Set _month_ to ConstrainToRange(_month_, 1, 12).
-        1. Set _day_ to ConstrainToRange(_day_, 1, DaysInMonth(_year_, _month_));
+        1. Let _yearMonth_ be ! ConstrainYearMonth(_year_, _month_).
+        1. Set _day_ to ConstrainToRange(_day_, 1, DaysInMonth(_yearMonth_.[[Year]], _yearMonth_.[[Month]]));
```
ptomato: Same here, `!`

spec/yearmonth.html
ljharb: At the moment, we prefer the prefix on every abstract operation call.

PR #380, opened 2020-02-18
Drop support for passing strings to with().

Ms2ger:
This is not a particularly useful thing to do, as the string parsing will
fill in all the fields, making the call little more than a more expensive
way to call from().

The polyfill does not support passing strings.

### DIFF
--- a/spec/date.html
+++ b/spec/date.html
@@ -460,24 +460,24 @@
     <emu-clause id="sec-temporal-topartialdate" aoid="ToPartialDate">
       <h1>ToPartialDate ( _date_ )</h1>
       <emu-alg>
-        1. If Type(_date_) is Object, then
-          1. Let _result_ be {
-            [[Year]]: *undefined*,
-            [[Month]]: *undefined*,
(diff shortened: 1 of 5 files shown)


spec/date.html
```diff
@@ -460,21 +460,24 @@ <h1>ToDate ( _date_ )</h1>
     <emu-clause id="sec-temporal-topartialdate" aoid="ToPartialDate">
       <h1>ToPartialDate ( _date_ )</h1>
       <emu-alg>
-        1. Let _result_ be {
-          [[Year]]: *undefined*,
-          [[Month]]: *undefined*,
-          [[Day]]: *undefined*,
-          }.
-        1. If Type(_date_) is Undefined or Null, then
+        1. If Type(_date_) is Object, then
```
ptomato: Minor editorial suggestion, I find early returns easier to read in code, and unless the spec writing style discourages that, I'd probably find it easier here:
```
1. If Type(_date_) is not Object, then
1. Let _result_ be...
```

PR #381, opened 2020-02-18
Specify DateTime#minus().

Ms2ger:


spec/datetime.html
```diff
@@ -383,7 +383,10 @@ <h1>Temporal.DateTime.prototype.minus ( _duration_ [ , _disambiguation_ ] )</h1>
         1. Perform ? RequireInternalSlot(_dateTime_, [[InitializedTemporalDateTime]]).
         1. Let _duration_ be ? ToLimitedDuration(_duration_, « »).
         1. Set _disambiguation_ to ? ToDisambiguation(_disambiguation_).
-        1. <mark>TODO</mark>
+        1. Let _timePart_ be ? SubtractTime(_dateTime_.[[Hour]], _dateTime_.[[Minute]], _dateTime_.[[Second]], _dateTime_.[[Millisecond]], _dateTime_.[[Microsecond]], _dateTime_.[[Nanosecond]], _duration_.[[Hours]], _duration_.[[Minutes]], _duration_.[[Seconds]], _duration_.[[Milliseconds]], _duration_.[[Microseconds]], _duration_.[[Nanoseconds]]).
+        1. Let _datePart_ be ? SubtractDate(_dateTime_.[[Year]], _dateTime_.[[Month]], _dateTime_.[[Day]], _duration_.[[Years]], _duration_.[[Months]], _duration_.[[Days]] - _timePart_.[[Day]], _disambiguation_).
+        1. Let _result_ be ! ConstrainDateTime(_datePart_.[[Year]], _datePart_.[[Month]], _datePart_.[[Day]], _timePart_.[[Hour]], _timePart_.[[Minute]], _timePart_.[[Second]], _timePart_.[[Millisecond]], _timePart_.[[Microsecond]], _timePart_.[[Nanosecond]]).
```
ptomato: The polyfill doesn't have this step. Is it necessary, since SubtractDate already has a constrain/balance/reject step?

Ms2ger: The polyfill does do this, because it calls the constructor with the default "constrain" argument, while CreateDateTime asserts. However, I agree it doesn't seem to be needed.

PR #382, opened 2020-02-18
Add "incomplete" banner to the top of polyfill docs

ptomato:
This makes it clearer that the docs are not an authoritative reference
for the proposal, they just reflect the state of the polyfill.

### DIFF
--- a/docs/head.html.part
+++ b/docs/head.html.part
@@ -23,7 +23,11 @@
         --variable: #39a33c;
       }
       .mermaid svg { height: 13em; }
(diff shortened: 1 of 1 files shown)


ptomato:
Thanks for the review, I'll go ahead and merge this.

littledan:
FWIW I feel like the wording of this banner is a bit funny. I think of the documentation as representing the proposal, not the polyfill (even if we're developing it in concert with the polyfill). Anyway this banner is good.

PR #384, opened 2020-02-19
Refactor regex

ptomato:
This is something that I'll use for #312, and I've been rebasing it for some time while gradually working on the actual spec and having the discussion in #198. I think it would be useful to get this in, even though it doesn't change any functionality.


ptomato:
I changed this a bit, to abstract the parsing rules for each type into its own `ES.ParseXString` method.

PR #385, opened 2020-02-19
Use ToLimitedDuration in polyfill

ptomato:
ToLimitedDuration was recently added to the spec. This brings the polyfill implementation more in line with the spec and adds some test coverage.

### DIFF
--- a/polyfill/lib/ecmascript.mjs
+++ b/polyfill/lib/ecmascript.mjs
@@ -16,6 +16,7 @@ import { Duration as TemporalDuration } from './duration.mjs';
 import bigInt from 'big-integer';
 
 import {
+  GetSlot,
   HasSlot,
   EPOCHNANOSECONDS,
   IDENTIFIER,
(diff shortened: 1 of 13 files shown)


polyfill/lib/ecmascript.mjs
```diff
@@ -311,6 +312,14 @@ export const ES = ObjectAssign(ObjectAssign({}, ES2019), {
       'reject'
     );
   },
+  ToLimitedDuration: (item, disallowedSlots = []) => {
+    const duration = ES.ToDuration(item);
+    for (let slot of disallowedSlots) {
+      if (GetSlot(duration, slot) !== 0)
+        throw new RangeError(`invalid duration field ${slot}`);
```
Ms2ger: Nit: I think we consistently brace single-line ifs. (We should perhaps lint this code.)

ptomato: I think the current state is pretty inconsistent, how about I open a separate PR adding an eslintrc and the results of running it over the code?

ptomato: See #394 for the linter PR; I'll make this change and merge this PR now.

Issue #330, opened 2020-01-25
Temporal.Time.prototype.difference should not exceed PT12H


gibson042:
`difference` methods return the absolute difference between the receiver and the argument, regardless of how they compare with each other. And since time of day is modular, such comparison isn't even possible without an attached day (e.g., 10:00 is earlier than 23:00 on the same day, but later than 23:00 on the preceding day). Inferring same-day semantics makes sense for `compare`, but seems wrong for `difference`—why should 10:00 vs. 23:00 get reported as PT13H while the equal span 11:00 vs. 00:00 gets reported as PT11H? In general, why should output values greater than PT12H be possible output only when one input value happens to be in the first half of a day?

Consistent comparison-ignoring results from `difference` should include proper modular arithmetic for Temporal.Time values, in which no two values can be separated by more than 12 hours.

PR #386, opened 2020-02-19
Fix Time.difference to always return the smaller possible duration

ptomato:
Closes: #330

### DIFF
--- a/spec/time.html
+++ b/spec/time.html
@@ -284,7 +284,10 @@
         1. Let _microsecond_ be _greater_.[[Microsecond]] - _smaller_.[[Microsecond]].
         1. Let _nanosecond_ be _greater_.[[Nanosecond]] - _smaller_.[[Nanosecond]].
         1. Let _result_ be ! BalanceTime(_hour_, _minute_, _second_, _millisecond_, _microsecond_, _nanosecond_).
-        1. Return ? CreateDuration(0, 0, _result_.[[Days]], _result_.[[Hour]], _result_.[[Minute]], _result_.[[Second]], _result_.[[Millisecond]], _result_.[[Microsecond]], _result_.[[Nanosecond]]).
+        1. If _result_.[[Hour]] &geq; 12, then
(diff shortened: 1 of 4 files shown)


ptomato:
Note that this does break what was until now an invariant:
```js
- smaller.plus(larger.difference(smaller)) === larger
- larger.minus(larger.difference(smaller)) === smaller
```
e.g.
```js
const larger = Temporal.Time.from('23:00');
const smaller = Temporal.Time.from('10:00');
smaller.plus(larger.difference(smaller))  // => 21:00
larger.minus(larger.difference(smaller))  // => 12:00
```
This is because 23:00 is actually the "smaller" time here.

spec/time.html
```diff
@@ -284,7 +284,10 @@ <h1>Temporal.Time.prototype.difference ( _otherTime_ )</h1>
         1. Let _microsecond_ be _greater_.[[Microsecond]] - _smaller_.[[Microsecond]].
         1. Let _nanosecond_ be _greater_.[[Nanosecond]] - _smaller_.[[Nanosecond]].
         1. Let _result_ be ! BalanceTime(_hour_, _minute_, _second_, _millisecond_, _microsecond_, _nanosecond_).
-        1. Return ? CreateDuration(0, 0, _result_.[[Days]], _result_.[[Hour]], _result_.[[Minute]], _result_.[[Second]], _result_.[[Millisecond]], _result_.[[Microsecond]], _result_.[[Nanosecond]]).
+        1. If _result_.[[Hour]] &geq; 12, then
+          1. Set _result_ to ! BalanceDuration(0, 24 &minus; _result_.[[Hour]], &minus;_result_.[[Minute]], &minus;_result_.[[Second]], &minus;_result_.[[Millisecond]], &minus;_result_.[[Microsecond]], &minus;_result_.[[Nanosecond]]);
```
Ms2ger: In prose, we don't tend to use semicolons :)
```suggestion
          1. Set _result_ to ! BalanceDuration(0, 24 &minus; _result_.[[Hour]], &minus;_result_.[[Minute]], &minus;_result_.[[Second]], &minus;_result_.[[Millisecond]], &minus;_result_.[[Microsecond]], &minus;_result_.[[Nanosecond]]).
```

PR #394, opened 2020-02-20
Add eslint configuration and CI job

ptomato:
- Fix existing linter complaints
- Add check to brace all single-line ifs and fix existing instances
- Also fix an actual bug that was uncovered by linting

### DIFF
--- a/polyfill/lib/ecmascript.mjs
+++ b/polyfill/lib/ecmascript.mjs
@@ -87,8 +87,9 @@ export const ES = ObjectAssign(ObjectAssign({}, ES2019), {
     const result = new TemporalTimeZone(zone);
     if (offset && ianaName) {
       const absolute = TemporalAbsolute.from(stringIdent);
-      if (result.getOffsetFor(absolute) !== offset)
+      if (result.getOffsetFor(absolute) !== offset) {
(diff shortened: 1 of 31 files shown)


polyfill/lib/ecmascript.mjs
```diff
@@ -427,8 +429,7 @@ export const ES = ObjectAssign(ObjectAssign({}, ES2019), {
   GetEpochFromParts: (year, month, day, hour, minute, second, millisecond, microsecond, nanosecond) => {
     let ms = Date.UTC(year, month - 1, day, hour, minute, second, millisecond);
     // Date.UTC interprets one and two-digit years as being in the 20th century
-    if (year >= 0 && year < 100)
-      ms = new Date(ms).setUTCFullYear(year);
+    if (year >= 0 && year < 100) ms = new Date(ms).setUTCFullYear(year);
```
Ms2ger: I'm not sure I like this change, but please land either way.

ptomato: I'm personally not a fan of this one-line style either but I found it to be predominant in the existing code ... I'll land it, but we can certainly change the config file to mandate a different style and then have eslint auto-fix it.

PR #399, opened 2020-02-24
Specify Temporal.now.

Ms2ger:


polyfill/lib/ecmascript.mjs
```diff
@@ -920,15 +920,9 @@ export const ES = ObjectAssign(ObjectAssign({}, ES2019), {
...
-    let ns = Date.now() % 1e6;
-    return () => {
-      const ms = Date.now();
-      const result = bigInt(ms).multiply(1e6).plus(ns);
-      ns = ms % 1e6;
-      return result;
-    };
-  })(),
+  SystemUTCEpochNanoSeconds: () => {
+    return bigInt(Date.now()).multiply(1e6);
```
ptomato: In theory I'd say 0 is better than bogus nanoseconds, but I wonder if it was done this way intentionally so that two immediately consecutive calls would be less likely to return the same value if done within the same millisecond? Definitely when I do `a = Date.now() % 1e6; b = Date.now() % 1e6; console.log(a, b)` in the Node REPL it most often shows the same result.

spec/temporal.html
```diff
@@ -4,4 +4,126 @@
...
+        This property has the attributes { [[Writable]]: false, [[Enumerable]]: false, [[Configurable]]: false }.
+      </p>
+    </emu-clause>
+  </emu-clause>
+</emu-clause>
+
+<emu-clause id="sec-temporal-now">
+  <h1>The Temporal.now Object</h1>
+  <p>
+    The Temporal.now object is the %Temporal.now% intrinsic object.
```
ptomato: Should this also say that it is an ordinary object and its prototype is Object.prototype, as above with the Temporal object?

spec/temporal.html
```diff
@@ -4,4 +4,126 @@
...
+
+  <emu-clause id="sec-temporal-properties">
+    <h1>Properties of the Temporal Object</h1>
+    <emu-clause id="sec-temporal.now">
+      <h1>Temporal.now</h1>
+      <p>
+        The initial value of `Temporal.now` is %Temporal.now%.
+      </p>
+      <p>
+        This property has the attributes { [[Writable]]: false, [[Enumerable]]: false, [[Configurable]]: false }.
```
sffc: Should this be writable so that you can overwrite Temporal.now to hide the system info?  Basically, what do we need for SES compatibility?

littledan: I think we should make this both writable and configurable, both for allowing mocks, for SES compatibility and just to fit with normal JS conventions. (See also: [Unforgeable] being considered deprecated in WebIDL.)

pipobscure: Yes! It’s critical that it be a normal property

pipobscure: Yes! It’s critical that it be a normal property

Ms2ger: Thanks for pointing this out; I accidentally copied this line.

polyfill/lib/ecmascript.mjs
Ms2ger: @pipobscure thoughts?

Ms2ger:
Hearing no objections, I'll merge this one of the coming days.

polyfill/lib/ecmascript.mjs
pipobscure: That's why I originally added *bogus* nanoseconds that are based on the previous iterations milliseconds. That way there is a guarantee that the full nanoseconds are always going to be increasing at least. The issue being that using *0* makes it easy to write tests that never deal with below ms accuracy. This instead will pretty much ensure that the microsecond and nanosecond fields are likely to actually have a value.

ptomato: Maybe keep the old code and add a comment explaining why this was done?

Although also note that `performance.now()` rounds to 1 ms anyway in order to prevent Spectre/Meltdown attacks.

PR #400, opened 2020-02-24
Date interop fixes #299

jasonwilliams:
Should help with both https://github.com/tc39/proposal-temporal/issues/299 and https://github.com/tc39/proposal-temporal/issues/370

### DIFF
--- a/docs/cookbook/dateTimeFromLegacyDate.mjs
+++ b/docs/cookbook/dateTimeFromLegacyDate.mjs
@@ -1,0 +1,20 @@
+import assert from 'assert';
+
+/**
+ * Converts a Date instance into a Temporal.DateTime instance
+ * DateTime accepts an ISO String in its from() method, so if you need to convert a date to a datetime
+ * you're best off passing the result of that
+ *
+ * @param {Date} legacyDate This is a Date instance
+ * @returns {Temporal.DateTime} Temporal.DateTime instance
+ */
+function getDateTimeFromLegacyDate(legacyDate) {
+  return Temporal.DateTime.from(legacyDate.toISOString());
+}
+
+const date = new Date('1970-01-01T00:00:01Z');
+const result = getDateTimeFromLegacyDate(date);
+assert.equal(result.day, 1);
+assert.equal(result.month, 1);
+assert.equal(result.year, 1970);
+assert.equal(result.toString(), '1970-01-01T00:00:01');
--- a/docs/cookbook/legacyDateFromDateTime.mjs
+++ b/docs/cookbook/legacyDateFromDateTime.mjs
@@ -1,0 +1,29 @@
+import assert from 'assert';
+
+/**
+ * Converts a Temporal.DateTime instance into a Date.
+ * This is the inverse of dateTimeFrom legacy Date example
+ *
+ * @param {Temporal.DateTime} dateTime This is a DateTime instance
+ * @returns {Date} legacy Date instance
+ */
+function getLegacyDateInUTCFromDateTime(dateTime) {
+  return new Date(
+    Date.UTC(
+      dateTime.year,
+      dateTime.month - 1,
+      dateTime.day,
+      dateTime.hour,
+      dateTime.minute,
+      dateTime.second,
+      dateTime.millisecond
+    )
+  );
+}
+
+const dateTime = Temporal.DateTime.from('2020-01-01T00:00:01Z');
+const result = getLegacyDateInUTCFromDateTime(dateTime);
+assert.equal(result.getDate(), 1); // Day of the month
+assert.equal(result.getFullYear(), 2020);
+assert.equal(result.getMonth(), 0); // the month (zero-indexed)
+assert.equal(result.toISOString(), '2020-01-01T00:00:01.000Z');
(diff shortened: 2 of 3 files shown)


docs/cookbook/dateTimeFromLegacyDate.mjs
```diff
@@ -0,0 +1,19 @@
...
+ */
+function getDateTimeFromDate(esDate) {
+    return Temporal.DateTime.from(esDate.toISOString());
+}
+
+const date = new Date("1970-01-01T00:00:01Z");
+const result = getDateTimeFromDate(date);
+assert.equal(result.day, 1);
+assert.equal(result.month, 1);
+assert.equal(result.year, 1970);
```
gibson042: I think it would be nice to add a concise assertion of all the data, e.g. `assert.equal(result.toString(), "1970-01-01T00:00:01")`.

docs/cookbook/legacyDateFromDateTime.mjs
```diff
@@ -0,0 +1,26 @@
...
+        dateTime.second,
+        dateTime.millisecond
+    );
+}
+
+const dateTime = Temporal.DateTime.from("1970-01-01T00:00:01Z");
+const result = getDateFromDateTime(dateTime);
+assert.equal(result.getDate(), 1); // Day of the month
+assert.equal(result.getFullYear(), 1970);
+assert.equal(result.getMonth(), 0); // the month (zero-indexed)
```
gibson042: I think it would be nice to add a concise assertion of all the data, e.g. `assert.equal(result.toISOString(), "1970-01-01T10:00:01.000Z")`.

jasonwilliams:
I’ve been using and prefer LegacyDate, I don’t like es2019date, it’s confusing, makes it sound new and something else.

docs/cookbook/legacyDateFromDateTime.mjs
jasonwilliams: This fails :/

jasonwilliams:
``` 
 actual: '1969-12-31T23:00:01.000Z',
  expected: '1970-01-01T10:00:01.000Z',
  operator: '=='
```

ptomato: Good point. This will have a different result based on your system's current time zone. We should rename the function to something like `getLegacyDateInLocalTimeZoneFromDateTime()`. I think the only other consistent thing we could do instead, is use `Date.UTC()` instead of `new Date()` and rename the function to something like `getLegacyDateInUTCFromDateTime()`.

docs/cookbook/legacyDateFromDateTime.mjs
```diff
@@ -0,0 +1,27 @@
+import assert from "assert";
+
+/**
+ * Converts a Temporal.DateTime instance into a Date.
+ * This is the inverse of dateTimeFromLegacyDate example
+ *
+ * @param {Temporal.DateTime} dateTime This is a Date instance
```
ptomato: Date → DateTime

docs/cookbook/legacyDateFromDateTime.mjs
```diff
@@ -0,0 +1,27 @@
...
+ * @returns {Date} LegacyDate instance
```
ptomato: Small nitpick, but if we go with legacy as the preferred term, I'd say "legacy Date" here instead of "LegacyDate". LegacyDate makes it sound like there's a class named that.

docs/cookbook/dateTimeFromLegacyDate.mjs
```diff
@@ -0,0 +1,20 @@
+import assert from "assert";
+
+/**
+ * Converts a Date instance into a Temporal.DateTime instance
+ * DateTime accepts an ISO String in its from() method, so if you need to convert a date to a datetime
+ * you're best off passing the result of that
+ *
+ * @param {Date} LegacyDate This is a Date instance
```
ptomato: For consistency with other examples, I'd make this parameter camel case, `legacyDate`.

jasonwilliams:
Done @Ms2ger @ptomato

docs/cookbook/dateTimeFromLegacyDate.mjs
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * Converts a Date instance into a Temporal.DateTime instance
+ * DateTime accepts an ISO String in its from() method, so if you need to convert a date to a datetime
+ * you're best off passing the result of that
+ *
+ * @param {Date} legacyDate This is a Date instance
+ * @returns {Temporal.DateTime} Temporal.DateTime instance
+ */
+function getDateTimeFromDate(legacyDate) {
+  return Temporal.DateTime.from(legacyDate.toISOString());
```
ptomato: Small nitpick, I'd name this `getDateTimeFromLegacyDate`

ptomato:
@jasonwilliams This needs another rebase, hopefully we can merge it straightaway after that.

ptomato:
I realized I could do the rebase in the web UI, so that's done now. @Ms2ger could you approve, squash and merge please?

PR #402, opened 2020-02-24
Min/max values for all types

ptomato:
Closes: #24


ptomato:
This includes a commit with a proposed resolution for #396, so we should make sure that is the consensus before merging this.

polyfill/lib/ecmascript.mjs
```diff
@@ -743,14 +744,61 @@ export const ES = ObjectAssign(ObjectAssign({}, ES2019), {
...
+    if (year === YEAR_MIN) {
+      try {
+        ES.GetEpochFromParts(year, month, day + 1, hour, minute, second, millisecond, microsecond, nanosecond - 1);
+      } catch (e) {
+        if (!(e instanceof RangeError)) throw e;
+        month = 4;
+        day = 19;
+        hour = minute = second = millisecond = microsecond = 0;
+        nanosecond = 1;
+      }
```
Ms2ger: I feel like this implementation relying on the exception type is somewhat brittle; could you review for alternatives?

ptomato: Err, of course -1 is not a good out of range value :face_with_head_bandage:
`null` maybe?

ptomato:
It looked like there was general agreement in #24 and #198 about the limits, though since this depends on the fix from #396 I would nonetheless not want to merge this yet. I'll merge the misc. bugfixes to master, but keep the limit commits in this pull request until #396 is resolved.

ptomato:
Rebased on #396.

Issue #71, opened 2018-07-24
Write spec text for accepted ISO formats on a per object basis


maggiepint:

pipobscure:
Didn't we say we'd explicitly describe the `toString()` methods and only accept that exact format in `.fromString()` and then leave parsing out of scope? Or is this already a result of the next meeting?

pipobscure:
Is this https://github.com/tc39/proposal-temporal/pull/65/files#diff-935dfcf8322fcb97745317a026a9c7b8R72 what you were referring to?

maggiepint:
Ya. Just have to make sure it's in Sasha's work and reflected in the readme.

ryzokuken:
@maggiepint are you still working on this?

PR #404, opened 2020-02-26
Properly spec ISO 8601 parsing

ptomato:
Reflects the various discussion we've had on this topic over the past weeks.

Closes: #312 
Closes: #198 
Closes: #71


polyfill/test/absolute.mjs
```diff
@@ -273,6 +273,11 @@ describe('Absolute', () => {
       equal(`${Absolute.from('+0019761118T152330.1+00:00')}`, '1976-11-18T15:23:30.100Z');
       equal(`${Absolute.from('+0019761118T152330.1+0000')}`, '1976-11-18T15:23:30.100Z');
     });
+    it('24:00 to mean next day midnight', () => {
```
Ms2ger: Let's also test that "24:01" doesn't work.

polyfill/test/monthday.mjs
```diff
@@ -73,6 +73,13 @@ describe('MonthDay', () => {
         equal(`${MonthDay.from('+0019761118T152330.1+00:00')}`, '11-18');
         equal(`${MonthDay.from('+0019761118T152330.1+0000')}`, '11-18');
       });
+      it('24:00 to mean next day midnight', () => {
+        equal(`${MonthDay.from('1976-11-18T24:00:00Z')}`, '11-19');
+        equal(`${MonthDay.from('1976-11-30T24:00:00Z')}`, '12-01');
+        equal(`${MonthDay.from('1976-12-31T24:00:00Z')}`, '01-01');
+        equal(`${MonthDay.from('1976-02-29T24:00:00Z')}`, '03-01');
```
Ms2ger: Following this, maybe also test "1976-02-28T24:00:00Z" → "02-29"

spec/timezone.html
```diff
@@ -61,7 +70,11 @@ <h1>Temporal.TimeZone.from ( _arg_ )</h1>
         The following steps are taken:
       </p>
       <emu-alg>
-        1. Return ? ToTimeZone(_arg_).
+        1. If Type(_arg_) is Object, then
+          1. Return ? ToTimeZone(_arg_).
+        1. If Type(_arg_) is String, then
+          1. Return ? TimeZoneFromString(_arg_).
+        1. Throw a new *TypeError* exception.
```
Ms2ger:
```suggestion
        1. Set _arg_ to ? ToString(_arg_).
        1. Return ? TimeZoneFromString(_arg_).
```

polyfill/test/validStrings.mjs
```diff
@@ -0,0 +1,340 @@
+// Generates 1000 valid ISO 8601 strings according to the EBNF grammar and
```
Ms2ger: I didn't review this at all.

polyfill/lib/ecmascript.mjs
```diff
@@ -196,9 +196,13 @@ export const ES = ObjectAssign(ObjectAssign({}, ES2019), {
   ToAbsolute: (item) => {
     if (ES.IsAbsolute(item)) return item;
     const isoString = ES.ToString(item);
-    const { zone, ...props } = ES.ParseAbsoluteString(isoString);
+    const { zone, offset, ianaName, ...props } = ES.ParseAbsoluteString(isoString);
     const datetime = ES.ToDateTime(props);
-    return datetime.inTimeZone(zone, 'reject');
+    const absolute = datetime.inTimeZone(zone, 'reject');
+    if (offset && ianaName && ES.GetTimeZoneOffsetString(absolute.getEpochNanoseconds(), zone) !== offset) {
+      throw new RangeError(`invalid offset ${offset}[${ianaName}]`);
```
Ms2ger: Do we want to allow this in the other parsing methods or check this in `ParseISODateTime`?

Ms2ger:
Do we want to fix #71 here too?

pipobscure:
I think #71 is not in scope here. We've moved away from the presumption in the issue which was that the only format permitted is the one emitted by `toString()` which is no longer the case because we're now accepting other ISO variations (space rather than T and so on)

polyfill/test/regex.mjs
```diff
@@ -33,7 +33,7 @@ describe('fromString regex', ()=>{
       test(`${dateTimeString}:30.123456789${zoneString}`, components);
     }
     // Time separators
-    ['T', ' '].forEach((timeSep) =>
+    ['T', 't', ' '].forEach((timeSep) =>
```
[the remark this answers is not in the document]
ptomato: I'd say based on https://github.com/tc39/proposal-temporal/issues/198#issuecomment-552404254 and https://github.com/tc39/proposal-temporal/issues/198#issuecomment-553225994 we should only accept a space and not `\s+`, since a single space is mentioned as an alternate character in RFC 3339. However, maybe multiple whitespaces are commonly output and that was the rationale for adding it ... @pipobscure?

polyfill/test/absolute.mjs
ptomato: Good call, that actually was accepted in cases where the time was ignored (Date, YearMonth, MonthDay). Fixed now.

polyfill/lib/ecmascript.mjs
ptomato: Good question, I'm leaning towards checking it in `ParseISODateTime` on the grounds that a string like `2000-01-01T00:00-5:00[Etc/UTC]` should always be considered invalid, not just in Absolute.from and TimeZone.from.

ptomato: This doesn't actually work properly at all, either before or after applying this branch, if you specify a time that happens twice during a DST transition. (e.g. `2019-02-16T23:45-02:00[America/Sao_Paulo]`.) So I think it would be better to remove this commit from the branch for now, and move the question to a follow up issue.

ptomato: → #415

pipobscure: That time doesn’t occur twice though; no time does. They are different in terms of their UTC offset.

What this us meant to prevent is 

1976-11-18T15:23:30.123-05:00[Europe/Vienna]

Which is never correct because -05:00 is never a valid offset in Vienna.

If is also meant to prevent 2019-08-05T15:34:23.234+01:00[Europe/Vienna] because +1:00 is the wrong offset for Vienna in August.

ptomato:
I think maybe #71 is already fixed?

polyfill/lib/ecmascript.mjs
ptomato: Indeed, I meant the wall time occurs twice, but providing the offset disambiguates it already, so it should not fail, but it does. See #415 for more details.

ptomato:
I think it's better not to land this until #415 is sorted out.

polyfill/test/regex.mjs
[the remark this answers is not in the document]
pipobscure: The rationale was actually a habit of mine that I always use `\s+` when looking for a space 😄  so nothing as well thought out as that.

So yes let’s replace that with just matching a single space character.

ptomato:
#428 is sorted out now, so I think this can be reviewed again.

It still does support 24:00 times. If preferred, I can remove that. However, I think that it being in http://www.ecma-international.org/ecma-262/5.1/#sec-15.9.1.15 is a strong argument for keeping it even if we think it was a mistake to put it in there in the first place. Reviewers, let me know what you prefer.

ptomato:
I'd really like to get this in, so I removed support for 24:00 from the branch as the discussion is likely to hold it up. We can defer that discussion to another issue.

polyfill/test/validStrings.mjs
ryzokuken: Wow, this test is quite a handful. I'd have to block some time out to understand it too. @ptomato I'm curious: why this as opposed to just verifying a bunch of hand-picked strings? We could make sure to include all edge-cases in the test cases...

ptomato: Don't worry, it doesn't run as part of the automated test suite. You can only run it manually. It's meant more as a development tool.

I found it very useful while developing the grammar and the regex, because on the one hand you can test and debug the grammar by inputting it in this EBNF-like framework and examining the "valid" strings it outputs to see if they are really valid. Once you are happy with the grammar, you can then feed the strings into from() and investigate when it breaks. Anything that I found to be worth testing for, I also added manually to the test cases. With this tool, both the grammar and the regex ended up better than I would have been able to do without it.

I would only expect this file to be used when making a change to either the grammar or the regex. It doesn't necessarily need to be checked in to the repo, but my reasoning was, better to check it in than leave it only on my machine.

ryzokuken:
This is a very long standing PR, and contains much required changes to the code and the spec. I'm merging this with @Ms2ger's consent, if anyone has any remaining issues with this, let's deal with that in a separate issue.

PR #405, opened 2020-02-27
Move disambiguation from constructors to from() methods

ptomato:
Closes: #260


polyfill/lib/absolute.mjs
```diff
@@ -97,13 +97,12 @@ export class Absolute {
...
     const diff = twons.minus(onens);
 
-    const ns = diff.mod(1e3);
-    const us = diff.divide(1e3).mod(1e3);
-    const ms = diff.divide(1e6).mod(1e3);
-    const ss = diff.divide(1e9);
-
-    const Duration = ES.GetIntrinsic('%Temporal.Duration%');
-    const duration = new Duration(0, 0, 0, 0, 0, ss, ms, us, ns, 'balance');
+    const duration = ES.ToDuration({
```
Ms2ger: Can author code mess with this by setting `Object.prototype.minutes`? `Object.create(null, {..})` would avoid that, I think. Test would be good as well.

polyfill/lib/date.mjs
```diff
@@ -146,13 +137,13 @@ export class Date {
     const MonthDay = ES.GetIntrinsic('%Temporal.MonthDay%');
     return new MonthDay(GetSlot(this, MONTH), GetSlot(this, DAY));
   }
-  static from(arg) {
-    let result = ES.ToDate(arg);
+  static from(arg, { disambiguation = 'constrain' } = {}) {
+    disambiguation = ES.ToString(disambiguation);
```
Ms2ger: Should check if `disambiguation` is one of the allowed values. (`ToDate` misses this check at least in the case that `arg` is a `Date` already.) Test.

polyfill/lib/datetime.mjs
```diff
@@ -296,8 +284,9 @@ export class DateTime {
     );
   }
 
-  static from(arg) {
-    let result = ES.ToDateTime(arg);
+  static from(arg, { disambiguation = 'constrain' } = {}) {
+    disambiguation = ES.ToString(disambiguation);
```
Ms2ger: Check for allowed values.

polyfill/lib/duration.mjs
```diff
@@ -147,8 +118,9 @@ export class Duration {
     if (!ES.IsDuration(this)) throw new TypeError('invalid receiver');
     return new Intl.DateTimeFormat(...args).format(this);
   }
-  static from(arg) {
-    let result = ES.ToDuration(arg);
+  static from(arg, { disambiguation = 'constrain' } = {}) {
+    disambiguation = ES.ToString(disambiguation);
```
Ms2ger: Check for allowed values.

polyfill/lib/monthday.mjs
```diff
@@ -62,14 +53,14 @@ export class MonthDay {
     const month = GetSlot(this, MONTH);
     const day = GetSlot(this, DAY);
     const Date = ES.GetIntrinsic('%Temporal.Date%');
-    return new Date(year, month, day, disambiguation);
+    return new Date(year, month, day);
   }
-  static from(arg) {
-    let result = ES.ToMonthDay(arg);
+  static from(arg, { disambiguation = 'constrain' } = {}) {
+    disambiguation = ES.ToString(disambiguation);
```
Ms2ger: Check for allowed values.

spec/abstractops.html
```diff
@@ -4,6 +4,28 @@
...
+    <h1>GetOption ( _options_, _property_, _values_, _fallback_ )</h1>
+
+    <p>
+      The abstract operation GetOption extracts the value of the property named _property_ from the provided _options_ object, converts it to a string, checks whether it is one of a List of allowed _values_, and fills in a _fallback_ value if necessary. If _values_ is *undefined*, there is no fixed set of values and any is permitted. If _options_ is *undefined*, then _fallback_ is returned.
+    </p>
+
+    <emu-alg>
+      1. If _options_ is *undefined*, then
+        1. Return _fallback_.
+      1. Set _options_ to ? ToObject(_options_).
```
Ms2ger: Do we want to throw on `null`? IDL-based APIs don't.

spec/date.html
```diff
@@ -414,9 +415,10 @@ <h1>CreateDate ( _y_, _m_, _d_ [ , _newTarget_ ] )</h1>
     </emu-clause>
 
     <emu-clause id="sec-temporal-datefromstring" aoid="DateFromString">
-      <h1>DateFromString ( _date_ )</h1>
+      <h1>DateFromString ( _date_, _disambiguation_ )</h1>
```
Ms2ger: Hmm, did we discuss this? I'm not sure if we want to support anything besides "reject" here.

spec/abstractops.html
ptomato: Throwing on `null` would be familiar to programmers using defaulting in JS:
```
> function foo({ disambiguation = 'bar' } = {}) { console.log(disambiguation); }
> foo()
bar
> foo(null)
TypeError: Cannot destructure property `disambiguation` of 'undefined' or 'null'.
```
This would be the first options bag API in ECMA-262, but ECMA-402 (where I copied this algorithm from) also seems to throw on null:
```
> new Intl.DateTimeFormat('en', null)
TypeError: Cannot convert undefined or null to object
```

spec/date.html
ptomato: See https://github.com/tc39/proposal-temporal/issues/260#issuecomment-591711184, this is an open question.

spec/abstractops.html
ljharb: The language most frequently treats undefined and null the same (default arguments does not follow this pattern); i'd prefer to stick to nullish coalescing/optional chaining/the more common pattern of not throwing on null (unless you also throw on undefined)

polyfill/lib/absolute.mjs
ptomato: I ended up writing this a different way after rebasing it this morning, but this is still something to test for, I think.

docs/date.md
```diff
@@ -80,6 +75,14 @@ date === Temporal.Date.from(date)  // => true
 date = Temporal.Date.from({year: 2006, month: 8, day: 24});  // => 2006-08-24
 date = Temporal.Date.from(Temporal.DateTime.from('2006-08-24T15:43:27'));
   // => same as above; Temporal.DateTime has year, month, and day properties
+
+// Different disambiguation modes
+date = Temporal.Date.from({ year: 2001, month: 13, day: 1 }, 'constrain')  // => 2001-12-01
+date = Temporal.Date.from({ year: 2001, month: -1, day: 1 }, 'constrain')  // => 2001-01-01
+date = Temporal.Date.from({ year: 2001, month: 13, day: 1 }, 'balance')  // => 2002-01-01
+date = Temporal.Date.from({ year: 2001, month: -1, day: 1 }, 'balance')  // => 2000-11-01
```
jasonwilliams: This one doesn't make sense to me, wouldn't the "next highest unit" be "2001-01-01" ? Why does it go to november 2000?

ptomato: Here's how I'd interpret it: month 0 would be one month before 2001-01, which would be December 2000, so month –1 would be two months before 2001-01, so November 2000.

jasonwilliams: I see @ptomato thanks

spec/monthday.html
```diff
@@ -191,18 +191,16 @@ <h1>Temporal.MonthDay.prototype.toJSON ( )</h1>
...
         The following steps are taken:
       </p>
       <emu-alg>
         1. Let _monthDay_ be the *this* value.
         1. Perform ? RequireInternalSlot(_monthDay_, [[InitializedTemporalMonthDay]]).
         1. Let _y_ be ? ToInteger(_year_).
-        1. Set _disambiguation_ to ? ToDisambiguation(_disambiguation_).
-        1. Let _result_ be ? RegulateDate(_y_, _monthDay_.[[Month]], _monthDay_.[[Day]], _disambiguation_).
-        1. Return ? CreateDate(_result_.[[Year]], _result_.[[Month]], _result_.[[Day]]).
+        1. Return ? CreateDate(_y_, _monthDay_.[[Month]], _monthDay_.[[Day]]).
```
Ms2ger: I think we'll need to throw for `MonthDay.from("02-29").withYear(2019)`.

spec/yearmonth.html
```diff
@@ -296,18 +296,16 @@ <h1>Temporal.YearMonth.prototype.toJSON ( )</h1>
...
         The following steps are taken:
       </p>
       <emu-alg>
         1. Let _yearMonth_ be the *this* value.
         1. Perform ? RequireInternalSlot(_yearMonth_, [[InitializedTemporalYearMonth]]).
         1. Let _d_ be ? ToInteger(_day_).
-        1. Set _disambiguation_ to ? ToDisambiguation(_disambiguation_).
-        1. Let _result_ be ? RegulateDate(_yearMonth_.[[Year]], _yearMonth_.[[Month]], _d_, _disambiguation_).
-        1. Return ? CreateDate(_result_.[[Year]], _result_.[[Month]], _result_.[[Day]]).
+        1. Return ? CreateDate(_yearMonth_.[[Year]], _yearMonth_.[[Month]], _d_).
```
Ms2ger: This needs to verify that the day is in range as well.

docs/date.md
Ms2ger: I also didn't get this immediately, so maybe add the zero case to the docs as well.

docs/datetime.md
```diff
@@ -101,6 +99,14 @@ dt = Temporal.DateTime.from({
...
 dt = Temporal.DateTime.from(Temporal.Date.from('1995-12-07T03:24:30'));
   // => same as above; Temporal.Date has year, month, and day properties
+
+// Different disambiguation modes
+dt = Temporal.DateTime.from({ year: 2001, month: 13, day: 1 }, 'constrain')  // => 2001-12-01T00:00
+dt = Temporal.DateTime.from({ year: 2001, month: -1, day: 1 }, 'constrain')  // => 2001-01-01T00:00
+dt = Temporal.DateTime.from({ year: 2001, month: 13, day: 1 }, 'balance')  // => 2002-01-01T00:00
+dt = Temporal.DateTime.from({ year: 2001, month: -1, day: 1 }, 'balance')  // => 2000-11-01T00:00
+dt = Temporal.DateTime.from({ year: 2001, month: 13, day: 1 }, 'reject')  // throws
+dt = Temporal.DateTime.from({ year: 2001, month: -1, day: 1 }, 'reject')  // throws
```
Ms2ger: Consider adding some cases that show the effect on time arguments as well

docs/yearmonth.md
```diff
@@ -75,6 +70,14 @@ ym === Temporal.YearMonth.from(ym)  // => true
 ym = Temporal.YearMonth.from({year: 2019, month: 6});  // => 2019-06
 ym = Temporal.YearMonth.from(Temporal.Date.from('2019-06-24'));
   // => same as above; Temporal.Date has year and month properties
+
+// Different disambiguation modes
+ym = Temporal.YearMonth.from({ year: 2001, month: 13 }, 'constrain')  // => 2001-12
+ym = Temporal.YearMonth.from({ year: 2001, month: -1 }, 'constrain')  // => 2001-01
+ym = Temporal.YearMonth.from({ year: 2001, month: 13 }, 'balance')  // => 2002-01
+ym = Temporal.YearMonth.from({ year: 2001, month: -1 }, 'balance')  // => 2000-11
```
Ms2ger: Same comment about showing `month:0`

polyfill/lib/ecmascript.mjs
```diff
@@ -204,105 +204,226 @@ export const ES = ObjectAssign(ObjectAssign({}, ES2019), {
...
     }
-    const isoString = ES.ToString(item);
-    const {month, day} = ES.ParseMonthDayString(isoString);
-    return new TemporalMonthDay(month, day, 'reject');
+    let {
+      day,
+      month,
+    } = props;
+
+    const leapYear = 1972; // XXX #261 leap year
```
Ms2ger:
```suggestion
    const leapYear = 1972;
```

polyfill/lib/ecmascript.mjs
```diff
@@ -313,32 +434,59 @@ export const ES = ObjectAssign(ObjectAssign({}, ES2019), {
...
+      microseconds = 0,
+      milliseconds = 0,
+      minutes = 0,
+      months = 0,
+      nanoseconds = 0,
+      seconds = 0,
+      years = 0,
+    } = props;
+
+    switch (ES.ToString(disambiguation)) {
```
Ms2ger: I think there's a bug in the caller if this isn't a string yet.

spec/date.html
```diff
@@ -46,16 +46,17 @@ <h1>Temporal.Date.prototype</h1>
...
     <emu-clause id="sec-temporal.date.from">
-      <h1>Temporal.Date.from ( _arg_ )</h1>
+      <h1>Temporal.Date.from ( _arg_ [ , _options_ ] )</h1>
       <p>
-        The `from` method takes one argument _arg_.
+        The `from` method takes two arguments, _arg_ and _options_.
         The following steps are taken:
       </p>
       <emu-alg>
+        1. Let _disambiguation_ be ? GetOption(_options_, *"disambiguation"*, « "*constrain"*, *"balance"*, *"reject"* », *"constrain"*).
```
Ms2ger:
```suggestion
        1. Let _disambiguation_ be ? GetOption(_options_, *"disambiguation"*, « *"constrain"*, *"balance"*, *"reject"* », *"constrain"*).
```

spec/datetime.html
```diff
@@ -54,16 +54,17 @@ <h1>Temporal.DateTime.prototype</h1>
...
     <emu-clause id="sec-temporal.datetime.from">
-      <h1>Temporal.DateTime.from ( _arg_ )</h1>
+      <h1>Temporal.DateTime.from ( _arg_ [ , _options_ ] )</h1>
       <p>
-        The `from` method takes one argument _arg_.
+        The `from` method takes two arguments, _arg_ and _options_.
         The following steps are taken:
       </p>
       <emu-alg>
+        1. Let _disambiguation_ be ? GetOption(_options_, *"disambiguation"*, « "*constrain"*, *"balance"*, *"reject"* », *"constrain"*).
```
Ms2ger:
```suggestion
        1. Let _disambiguation_ be ? GetOption(_options_, *"disambiguation"*, « *"constrain"*, *"balance"*, *"reject"* », *"constrain"*).
```

spec/monthday.html
ptomato: To address this and some of the following comments, I changed the "Assert: ValidateX(...) is *true*" inside all the CreateX abstract operations to instead be the "If ValidateX(...) is *false*, throw a RangeError" that we were previously doing in the constructors. So CreateDate should now throw in the case you mentioned.

polyfill/lib/absolute.mjs
ptomato: My conclusion is that while `Object.prototype.minutes = 5489755` will badly mess with `Duration.from({...})` (and Duration.with if we add it), it will not affect this code the way it's currently written. I'm not sure what approach we should generally take in adding tests for this sort of thing to the test suite.

PR #409, opened 2020-02-28
First Tuesday Of Month Fixes #240

jasonwilliams:
An implementation of getting the first Tuesday of the month.
Have at it.

### DIFF
--- a/docs/cookbook/getFirstTuesdayOfMonth.mjs
+++ b/docs/cookbook/getFirstTuesdayOfMonth.mjs
@@ -4,11 +4,11 @@ import assert from 'assert';
  * Gets the first Tuesday of the month and returns its date
  *
  * @param {Temporal.YearMonth} queriedMonth YearMonth instance to query
- * @returns {Temporal.DateTime} DateTime Instance which gives first tuesday
+ * @returns {Temporal.Date} Temporal.Date Instance which gives first tuesday
  */
 function getFirstTuesday(queriedMonth) {
   // We first need to convert to a date
-  let date = Temporal.DateTime.from({
+  let date = Temporal.Date.from({
     year: queriedMonth.year,
     month: queriedMonth.month,
     day: 1
@@ -25,5 +25,5 @@ function getFirstTuesday(queriedMonth) {
 
 const myMonth = Temporal.YearMonth.from('2020-02');
 const firstTuesdayOfMonth = getFirstTuesday(myMonth);
-assert(firstTuesdayOfMonth.toString(), '2020-02-04T00:00');
-assert(firstTuesdayOfMonth.dayOfWeek, 2);
+assert.equal(firstTuesdayOfMonth.toString(), '2020-02-04');
+assert.equal(firstTuesdayOfMonth.dayOfWeek, 2);
--- a/docs/cookbook/getFirstTuesdayOfMonth.mjs
+++ b/docs/cookbook/getFirstTuesdayOfMonth.mjs
@@ -6,11 +6,11 @@ import assert from "assert";
  * @param {Temporal.YearMonth} queriedMonth YearMonth instance to query
  * @returns {Temporal.DateTime} DateTime Instance which gives first tuesday
  */
-function getFirstTuesday(queriedDate) {
+function getFirstTuesday(queriedMonth) {
     // We first need to convert to a date
     let date = Temporal.DateTime.from({
-        year: queriedDate.year,
-        month: queriedDate.month,
+        year: queriedMonth.year,
+        month: queriedMonth.month,
         day: 1
     });
 
--- a/docs/cookbook/getFirstTuesdayOfMonth.mjs
+++ b/docs/cookbook/getFirstTuesdayOfMonth.mjs
@@ -1,30 +1,29 @@
-import assert from "assert";
+import assert from 'assert';
 
 /**
- * Gets the first tuesday of the month and returns its date
+ * Gets the first Tuesday of the month and returns its date
  *
  * @param {Temporal.YearMonth} queriedMonth YearMonth instance to query
  * @returns {Temporal.DateTime} DateTime Instance which gives first tuesday
  */
 function getFirstTuesday(queriedMonth) {
-    // We first need to convert to a date
-    let date = Temporal.DateTime.from({
-        year: queriedMonth.year,
-        month: queriedMonth.month,
-        day: 1
-    });
+  // We first need to convert to a date
+  let date = Temporal.DateTime.from({
+    year: queriedMonth.year,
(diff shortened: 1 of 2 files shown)


docs/cookbook/getFirstTuesdayOfMonth.mjs
```diff
@@ -0,0 +1,30 @@
...
+    let date = Temporal.DateTime.from({
+        year: queriedDate.year,
+        month: queriedDate.month,
+        day: 1
+    });
+
+    // Start at the beginning of the month then stop at the first tuesday
+    while (date.dayOfWeek !== 2) {
+        date = date.plus({ days: 1 });
+    }
```
mikemccaughan: I think this can just be `date = date.plus({ days: 7 - date.dayOfWeek + 2 })`? Or something similar?

docs/cookbook/getFirstTuesdayOfMonth.mjs
```diff
@@ -0,0 +1,30 @@
...
+ * @param {Temporal.YearMonth} queriedMonth YearMonth instance to query
+ * @returns {Temporal.DateTime} DateTime Instance which gives first tuesday
+ */
+function getFirstTuesday(queriedMonth) {
+    // We first need to convert to a date
+    let date = Temporal.DateTime.from({
+        year: queriedMonth.year,
+        month: queriedMonth.month,
+        day: 1
+    });
```
Ms2ger:
```suggestion
    let date = queriedMonth.withDay(1);
```

docs/cookbook/getFirstTuesdayOfMonth.mjs
jasonwilliams: could you enlighten me here, i don't see how this would work if say the first day of the month Monday was the 1st. So then we're looking for the 2nd (Tuesday) but would end up skipping over it.
1st + (7 - 1 + 2)

mikemccaughan: I had something like that for `Date`; I can't remember now. It might have worked because of the numbering `Date` uses for the result of `getDay()` though, so it might not work for `Temporal`... Maybe `Math.abs(7 - date.dayOfWeek) + 3`? You know what? What you have works, pay no attention to me :).

docs/cookbook/getFirstTuesdayOfMonth.mjs
```diff
@@ -0,0 +1,26 @@
+import assert from "assert";
+
+/**
+ * Gets the first tuesday of the month and returns its date
+ *
+ * @param {Temporal.YearMonth} queriedMonth YearMonth instance to query
+ * @returns {Temporal.DateTime} DateTime Instance which gives first tuesday
```
ptomato: The code as currently written returns a Temporal.Date, and I think that's correct because we don't care about the time of day on Tuesday.

docs/cookbook/getFirstTuesdayOfMonth.mjs
```diff
@@ -0,0 +1,26 @@
+import assert from "assert";
+
+/**
+ * Gets the first tuesday of the month and returns its date
```
ptomato: tuesday → Tuesday

docs/cookbook/getFirstTuesdayOfMonth.mjs
ptomato: I think `date.plus({ days: (9 - date.dayOfWeek) % 7 })` ought to work. We have Monday = 1, Sunday = 7, and we want to add a positive number smaller than 7 to get to the first Tuesday. If we're already on a Tuesday (2) then we want to add 0. So for the first of the month being a Monday through Sunday the additions are: 1, 0, 6, 5, 4, 3, 2 which is given by that formula.

Alternatively making it more explicit via a lookup table might look less mystical to the reader: `[1, 0, 6, 5, 4, 3, 2][date.dayOfWeek - 1]`

docs/cookbook/getFirstTuesdayOfMonth.mjs
```diff
@@ -0,0 +1,26 @@
...
+        date = date.plus({ days: 1 });
+    }
+
+    return date;
+}
+
+const myMonth = Temporal.YearMonth.from("2020-02");
+const firstTuesdayOfMonth = getFirstTuesday(myMonth);
+assert.equal(
+    firstTuesdayOfMonth.toLocaleString("en-UK", { weekday: "long" }),
```
ptomato: I'd assert an ISO string but maybe `console.log` the nicely formatted string, or something like that. (This assertion fails for me anyway, I get `'Tuesday, 2/4/2020'`.)

Consider also asserting that `firstTuesdayOfMonth.dayOfWeek` equals 2.

ptomato:
Re-reading the original use case in #240:

> Get the "first Tuesday", "second Tuesday", or "last Tuesday" of a month, and also, given a day, identify whether it's the first, second, etc. of that day in the month.

I'd suggest adding two more cookbook functions here, or maybe in a follow up PR. Something like, `getNthWeekday(yearMonth, dayOfWeek, n)` where, say, `getNthWeekday(ym, 6, 3)` would get you the third Saturday and `getNthWeekday(ym, 1, -1)` the last Monday? And then another function `isNthWeekday(date)` where for example `isNthWeekday(Temporal.Date.from('2020-02-28'))` returns 4 because today is the 4th Friday.

docs/cookbook/getFirstTuesdayOfMonth.mjs
Ms2ger: Did you mean "en-GB"?

docs/cookbook/getFirstTuesdayOfMonth.mjs
jasonwilliams: Im not sure this is correct, it seems to be returning a DateTime instance for me

docs/cookbook/getFirstTuesdayOfMonth.mjs
```diff
@@ -0,0 +1,29 @@
...
+  //    1, 0, 6, 5, 4, 3, 2 which is given by that formula.
+  date = date.plus({ days: [1, 0, 6, 5, 4, 3, 2][date.dayOfWeek - 1] });
+
+  return date;
+}
+
+const myMonth = Temporal.YearMonth.from('2020-02');
+const firstTuesdayOfMonth = getFirstTuesday(myMonth);
+assert(firstTuesdayOfMonth.toString(), '2020-02-04T00:00');
+assert(firstTuesdayOfMonth.dayOfWeek, 2);
```
Ms2ger: Should these not be `assert.equal`?

PR #412, opened 2020-02-28
Run the cookbook examples in CI.

Ms2ger:


polyfill/test/resolve.cookbook.mjs
```diff
@@ -0,0 +1,25 @@
...
+import fs from 'fs';
+const PKG = JSON.parse(fs.readFileSync('package.json', { encoding: 'utf-8' }));
+export function resolve(specifier, parent, defaultResolve) {
+  if (specifier === PKG.name) {
+    specifier = new URL('../lib/temporal.mjs', import.meta.url).toString();
+  }
+  return defaultResolve(specifier, parent);
+}
+
+export async function transformSource(
```
ptomato: This wasn't working for me until I figured out that this hook was only added in Node 12.16.0. So we should require that version somewhere. It probably needs to be both in package.json and in the CI configuration.

ljharb: be sure to add it to the "engines" field, which is advisory but usually sufficient

Issue #388, opened 2020-02-19
Temporal.Duration.from and disambiguation = "balance"


sffc:
*Split from #324*

From @sffc:

*Use case:* I want to be able to say, "this video is 100 seconds long", and have Temporal automatically fix that to be 1 minute and 40 seconds.

My expectation is that `disambiguation: "balance"` would do this, but it seems it doesn't, based on the polyfill:

```javascript
> Temporal.Duration.from({ seconds: 100, disambiguation: "balance" }).toString()
'PT100S'
```

My expected output would be "PT1M40S" instead.

-----

From @ptomato:

Agreed, this kind of ties in to the discussion about the largest-element cutoff parameter in #307, are Duration objects with too-high values valid? As implemented, currently the polyfill only constrains or balances negative numbers.

-----

From @sffc:

I think it's *valid* to have a Duration object with "too-high" values; but Temporal should also give a way to balance the fields if I specifically request that behavior.

ptomato:
Meeting Feb 20: Balance on Duration.from() should convert things to the largest possible unit, up to hours. (Not days, months, or years.)

PR #413, opened 2020-02-28
Duration balance

ptomato:
Closes: #388

### DIFF
--- a/polyfill/lib/duration.mjs
+++ b/polyfill/lib/duration.mjs
@@ -57,17 +57,14 @@ export class Duration {
         break;
       }
       case 'balance': {
-        let deltaDays;
         ({
-          deltaDays,
-          hour: hours,
-          minute: minutes,
-          second: seconds,
-          millisecond: milliseconds,
-          microsecond: microseconds,
-          nanosecond: nanoseconds
-        } = ES.BalanceTime(hours, minutes, seconds, milliseconds, microseconds, nanoseconds));
-        days += deltaDays;
+          hours,
+          minutes,
+          seconds,
+          milliseconds,
+          microseconds,
+          nanoseconds,
+        } = ES.BalanceDuration(0, hours, minutes, seconds, milliseconds, microseconds, nanoseconds, 'hours'));
         for (const prop of [years, months, days, hours, minutes, seconds, milliseconds, microseconds, nanoseconds]) {
           if (!Number.isFinite(prop)) throw new RangeError('infinite values not allowed as duration fields');
         }
--- a/spec/duration.html
+++ b/spec/duration.html
@@ -364,15 +364,28 @@
     </emu-clause>
 
     <emu-clause id="sec-temporal-balanceduration" aoid="BalanceDuration">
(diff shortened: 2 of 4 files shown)


spec/duration.html
```diff
@@ -364,15 +364,28 @@ <h1>CreateDuration ( _years_, _months_, _days_, _hours_, _minutes_, _seconds_, _
...
+        1. Set _seconds_ to _bt_.[[Second]].
+        1. If _largestUnit_ is *"hours"*, *"minutes"*, or *"seconds"*, then
+          1. Increment _hours_ by 24 &times; _days_.
+          1. Set _days_ to 0.
+        1. If _largestUnit_ is *"minutes"* or *"seconds"*, then
+          1. Increment _minutes_ by 60 &times; _hours_.
+          1. Set _hours_ to 0.
+        1. If _largestUnit_ is *"seconds"*, then
+          1. Increment _seconds_ by 60 &times; _minutes_.
+          1. Set _minutes_ to 0.
```
sffc: Also put [milli,micro,nano] seconds in [0,999], right?

ptomato: That should have already been done by BalanceTime. (We don't support largest-unit for sub-second units (yet) so they have to be in [0,999] after balancing no matter what.)

sffc: So, was the spec already doing the #388 behavior?

polyfill/lib/duration.mjs
```diff
@@ -57,17 +57,14 @@ export class Duration {
...
-          nanosecond: nanoseconds
-        } = ES.BalanceTime(hours, minutes, seconds, milliseconds, microseconds, nanoseconds));
-        days += deltaDays;
+          hours,
+          minutes,
+          seconds,
+          milliseconds,
+          microseconds,
+          nanoseconds,
+        } = ES.BalanceDuration(0, hours, minutes, seconds, milliseconds, microseconds, nanoseconds, 'hours'));
```
sffc: Where is `ES.BalanceDuration` defined?

spec/duration.html
ptomato: Sort of. The constructor was balancing up to days instead of hours, so that's what this PR changes. #388 is talking about Duration.from() which this PR doesn't make possible either; we only get that after we merge #405. (This applies to both spec and polyfill.)

polyfill/lib/duration.mjs
ptomato: Currently, https://github.com/tc39/proposal-temporal/blob/main/polyfill/lib/ecmascript.mjs#L799-L834

sffc: ok.  I'm slightly confused because last I checked, the behavior of the polyfill did not balance between duration fields.  I don't see how this PR concretely changes the PR behavior.

ptomato: My guess is that when you checked, you probably used Duration.from(), which currently doesn't balance, instead of `new Duration(..., 'balance')`. Duration.from() is _supposed_ to balance, if requested to, which is implemented in #405 for both the spec and the polyfill. If it makes things less confusing, I'm happy to put this on hold and revisit it after #405 is reviewed and merged.

sffc: Aha! Yep, that makes sense.

PR #417, opened 2020-03-04
Define the Symbol.species properties.

Ms2ger:
This commit does not use the property yet; this will happen in a future commit.

### DIFF
--- a/polyfill/test/Absolute/constructor/species/getter-length.js
+++ b/polyfill/test/Absolute/constructor/species/getter-length.js
@@ -1,0 +1,26 @@
+// Copyright (C) 2020 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-get-temporal.absolute-@@species
+info: |
+    Every built-in function object, including constructors, has a "length" property whose value is
+    an integer. Unless otherwise specified, this value is equal to the largest number of named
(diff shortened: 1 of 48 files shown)


polyfill/test/Absolute/constructor/species/getter-length.js
```diff
@@ -0,0 +1,26 @@
...
+    Every built-in function object, including constructors, has a "length" property whose value is
+    an integer. Unless otherwise specified, this value is equal to the largest number of named
+    arguments shown in the subclause headings for the function description. Optional parameters
+    (which are indicated with brackets: [ ]) or rest parameters (which are shown using the form
+    «...name») are not included in the default argument count.
+
+    Unless otherwise specified, the "length" property of a built-in function object has the
+    attributes { [[Writable]]: false, [[Enumerable]]: false, [[Configurable]]: true }.
+includes: [propertyHelper.js]
+features: [Symbol.species]
```
ptomato: This is my first time reviewing test262 tests, so I had to consult the documentation to find out about this metadata format, please feel free to disregard if I'm not understanding the conventions correctly. It looks like this feature can probably be omitted since it has been published in the standard and is shipping widely?

Ms2ger: It probably could be omitted, but it doesn't hurt anything to keep it either, I think, so I'd just leave it in.

PR #418, opened 2020-03-06
Balance out of scope duration fields

ptomato:
Closes: #324


docs/absolute.md
```diff
@@ -242,6 +242,10 @@ The `duration` argument can be any value that could be passed to `Temporal.Durat
 - any object with properties denoting a duration, such as `{ hours: 5, minutes: 30 }`;
 - a string in ISO 8601 duration format, such as `PT5H30M`.
 
+The `years` and `months` fields of `duration` must be zero, because adding a year or a month to a `Temporal.Absolute` is invalid and will throw a `RangeError`.
```
sffc: Can you remind me where we discussed this behavior?

polyfill/lib/yearmonth.mjs
```diff
@@ -62,10 +62,11 @@ export class YearMonth {
...
     if (!ES.IsYearMonth(this)) throw new TypeError('invalid receiver');
     disambiguation = ES.ToArithmeticDisambiguation(disambiguation);
-    const duration = ES.ToLimitedDuration(durationLike, [DAYS, HOURS, MINUTES, SECONDS, MILLISECONDS, MICROSECONDS, NANOSECONDS]);
+    const duration = ES.ToLimitedDuration(durationLike);
     let { year, month } = this;
-    const { years, months } = duration;
-    ({ year, month } = ES.AddDate(year, month, 1, years, months, 0, disambiguation));
+    const { years, months, hours, minutes, seconds, milliseconds, microseconds, nanoseconds } = duration;
+    const { days } = ES.BalanceDuration(duration.days, hours, minutes, seconds, milliseconds, microseconds, nanoseconds, 'days');
+    ({ year, month } = ES.AddDate(year, month, 1, years, months, days, disambiguation));
```
sffc: We forgot to discuss YearMonth this morning, but your solution here seems fine on first glance.

polyfill/lib/yearmonth.mjs
```diff
@@ -77,10 +78,11 @@ export class YearMonth {
...
     if (!ES.IsYearMonth(this)) throw new TypeError('invalid receiver');
     disambiguation = ES.ToArithmeticDisambiguation(disambiguation);
-    const duration = ES.ToLimitedDuration(durationLike, [DAYS, HOURS, MINUTES, SECONDS, MILLISECONDS, MICROSECONDS, NANOSECONDS]);
+    const duration = ES.ToLimitedDuration(durationLike);
     let { year, month } = this;
-    const { years, months } = duration;
-    ({ year, month } = ES.SubtractDate(year, month, 1, years, months, 0, disambiguation));
+    const { years, months, hours, minutes, seconds, milliseconds, microseconds, nanoseconds } = duration;
+    const { days } = ES.BalanceDuration(duration.days, hours, minutes, seconds, milliseconds, microseconds, nanoseconds, 'days');
+    ({ year, month } = ES.SubtractDate(year, month, 1, years, months, days, disambiguation));
```
sffc: There's one potential problem here.  Let's say you have a Duration that balances to 5 days.  When you add it to a YearMonth, you get the same YearMonth (assuming no month is < 5 days long), but when you subtract it, you get the previous YearMonth, meaning that arithmetic is not symmetric.

Not sure about the best solution.  Probably it's an edge case we can just ignore.

docs/absolute.md
ptomato: It's been the case in the polyfill since before I started working on this project, so maybe @pipobscure knows?

In https://github.com/tc39/proposal-temporal/issues/307#issuecomment-588521748 and the following thread we discussed that Absolute.difference() never returns days or months, so in any case this makes sense to me accordingly.

polyfill/lib/yearmonth.mjs
ptomato: Hmm. It doesn't seem all that important but then again it should be consistent with other types, and we have e.g. `Temporal.Date.from('2020-03-12').minus({hours: 5})` equals 2020-03-12, so it's not currently consistent.

Maybe for subtraction we should subtract the date from the last day of the month. I'll revise this PR to do that.

ptomato:
@sffc Sorry, I somehow missed a notification for your review last week, probably clicked it away by accident! I've updated the branch.

spec/yearmonth.html
```diff
@@ -228,11 +227,11 @@ <h1>Temporal.YearMonth.prototype.minus ( _temporalDurationLike_ [ , _options_ ]
...
         1. Let _yearMonth_ be the *this* value.
         1. Perform ? RequireInternalSlot(_yearMonth_, [[InitializedTemporalYearMonth]]).
-        1. Let _duration_ be ? ToLimitedTemporalDuration(_temporalDurationLike_, « *"days"*, *"hours"*, *"minutes"*, *"seconds"*, *"milliseconds"*, *"microseconds"*, *"nanoseconds"* »).
+        1. Let _duration_ be ? ToLimitedTemporalDuration(_temporalDurationLike_, « »).
+        1. Let _balanceResult_ be ? BalanceDuration(_duration_.[[Days]], _duration_.[[Hours]], _duration_.[[Minutes]], _duration_.[[Seconds]], _duration_.[[Milliseconds]], _duration_.[[Microseconds]], _duration_.[[Nanoseconds]], *"days"*).
         1. Let _disambiguation_ be ? ToArithmeticTemporalDisambiguation(_options_).
-        1. Let _y_ be _yearMonth_.[[Year]] - _duration_.[[Years]].
-        1. Let _m_ be _yearMonth_.[[Month]] - _duration_.[[Months]].
-        1. Let _result_ be ! BalanceYearMonth(_y_, _m_).
+        1. Let _lastDay_ be ? DaysInMonth(_yearMonth_.[[Year]], _yearMonth_.[[Month]]).
```
Ms2ger:
```suggestion
        1. Let _lastDay_ be ! DaysInMonth(_yearMonth_.[[Year]], _yearMonth_.[[Month]]).
```

PR #424, opened 2020-03-10
Define subclassing support for Absolute static methods.

Ms2ger:


polyfill/lib/absolute.mjs
```diff
@@ -142,20 +142,38 @@ export class Absolute {
...
+    ES.RejectAbsolute(epochNanoseconds);
+    const result = new this(epochNanoseconds.value);
+    if (!ES.IsAbsolute(result)) throw new TypeError('invalid result');
+    return result;
   }
   static from(arg, zone) {
-    let result = ES.ToAbsolute(arg, zone);
-    return this === Absolute ? result : new this(GetSlot(result, EPOCHNANOSECONDS));
+    const absolute = ES.ToAbsolute(arg, zone);
+    const result = this === Absolute ? absolute : new this(GetSlot(absolute, EPOCHNANOSECONDS).value);
```
ptomato: nitpick: I have a local branch that I haven't pushed yet where I replaced all of this kind of expression, because I find this easier to read. (I think we don't need to call IsAbsolute on the return value of ToAbsolute, either.)
```suggestion
    if (this === Absolute) return absolute;
    const result = new this(GetSlot(absolute, EPOCHNANOSECONDS).value);
```

polyfill/test/Absolute/constructor/from/subclass-invalid-receiver.js
```diff
@@ -0,0 +1,15 @@
+// Copyright (C) 2020 Igalia, S.L. All rights reserved.
```
ptomato: Is the name of this test intentional? In some of the other tests it's called "branding", and this one doesn't seem to have to do with subclasses (although it does test for invalid receivers).

(Ditto for several other tests below)

polyfill/test/Absolute/constructor/from/subclass-out-of-range.js
```diff
@@ -0,0 +1,19 @@
...
+class MyAbsolute extends Temporal.Absolute {
+  constructor(ns) {
+    called = true;
+    super(ns);
+  }
+}
+
+assert.throws(RangeError, () => MyAbsolute.from("-271821-04-19T23:59:59.999999999Z"));
+assert.throws(RangeError, () => MyAbsolute.from("+275760-09-13T00:00:00.000000001Z"));
+assert.sameValue(called, false);
```
ptomato: Is it correct that we want to mandate this? It seems like an implementation detail whether the out-of-range value is rejected in the constructor or in from().

spec/absolute.html
```diff
@@ -76,7 +80,9 @@ <h1>Temporal.Absolute.fromEpochSeconds ( _arg_ )</h1>
         1. Let _arg_ be ? ToNumber(_arg_).
         1. Let _arg_ be ? NumberToBigInt(_arg_).
         1. Let _epochNanoseconds_ be _arg_ × 1,000,000,000.
-        1. Return ? CreateAbsolute(_epochNanoseconds_).
+        1. Perform ? RejectAbsolute(_epochNanoseconds_).
```
ptomato: Similar question as above, do we want to mandate this here?

Issue #238, opened 2019-10-30
Editorial suggestion: Use TemporalDate, temporalDateLike , etc internally


littledan:
Algorithms like `ToDate` are a bit confusing n name--it's very close to existing spec algorithms like `ToDateString`. I'd suggest we try to never use "Date" or "Time" alone, and always use `TemporalDate`, `TemporalTime`, including in algorithm names and local variables (`temporalDate` or `tDate`), to reduce potential ambiguity.

In some places, the variable name `date` or `otherDate` is used for a partial date, which is a very different concept. I'd suggest that we use the variable name `temporalDateLike` or `partialTemporalDate`. This goes for partial time as well.

PR #426, opened 2020-03-12
Define subclassing support for Absolute prototype methods.

Ms2ger:


polyfill/lib/absolute.mjs
```diff
@@ -62,8 +62,12 @@ export class Absolute {
     add = add.plus(bigInt(days).multiply(24 * 60 * 60 * 1e9));
 
     const ns = bigInt(GetSlot(this, EPOCHNANOSECONDS)).plus(add);
+    ES.RejectAbsolute(ns);
```
ljharb: side note: it’s pretty weird to have “reject absolute” rejecting a different thing than what “is absolute” predicates.

ptomato: FWIW, one interpretation of #238 would be to change "IsAbsolute" to "IsTemporalAbsolute" along with the other types, I think that helps a bit with this.

PR #429, opened 2020-03-12
Use TemporalDate, temporalDateLike, etc. internally

ptomato:
Closes: #238


ptomato:
It's definitely possible that this PR either goes too far or not far enough to do what is described in the issue. Certainly open to revising it.

ptomato:
Rebased and gave it one more look over, merging now.

PR #432, opened 2020-03-14
meta: add pnpm log and lockfile to gitignore

ryzokuken:
Just a small meta commit that'd make my life easier. Will merge after someone +1s.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@ polyfill/script.js.map
 polyfill/test262/
 out/
 .vscode/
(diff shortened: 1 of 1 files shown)


ljharb:
Why not put these in your global gitignore instead of in each repo you work in?

ryzokuken:
@ljharb fair.

ptomato:
Agreed, I tend to avoid putting any setup-specific files in .gitignore, I think even .vscode shouldn't be in there.

PR #435, opened 2020-03-16
Define subclassing support for Date prototype methods.

Ms2ger:


polyfill/test/Date/prototype/minus/subclass-invalid-result.js
```diff
@@ -0,0 +1,25 @@
+// Copyright (C) 2020 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.date.minus
```
ptomato: Needs Symbol.species feature for consistency with the other tests where we specify it? (We discussed on a previous PR that it was optional, but it should be consistent.) Seems to apply to a few other tests as well.

spec/date.html
```diff
@@ -420,6 +422,18 @@ <h1>CreateDate ( _y_, _m_, _d_ [ , _newTarget_ ] )</h1>
       </emu-alg>
     </emu-clause>
 
+    <emu-clause id="sec-temporal-createdatefrominstance" aoid="CreateDateFromInstance">
```
ptomato: If we decide to go with https://github.com/tc39/proposal-temporal/pull/429, then this should be named `CreateTemporalDateFromInstance`.

PR #438, opened 2020-03-17
Define subclassing support for DateTime prototype methods.

Ms2ger:


spec/datetime.html
```diff
@@ -698,6 +701,18 @@ <h1>CreateDateTime ( _year_, _month_, _day_, _hour_, _minute_, _second_, _millis
       </emu-alg>
     </emu-clause>
 
+    <emu-clause id="sec-temporal-createdatetimefrominstance" aoid="CreateDateTimeFromInstance">
+      <h1>CreateDateTimeFromInstance ( _dateTime_, _year_, _month_, _day_, _hour_, _minute_, _second_, _millisecond_, _microsecond_, _nanosecond_ )</h1>
```
ptomato: I'd prefer this to be CreateTemporalDateTimeFromInstance, but since we already have a mix I'll merge this anyway and rebase #429 to cover this.

PR #439, opened 2020-03-17
Define subclassing support for Time prototype methods.

Ms2ger:


Ms2ger:
> Does the spec still need to be changed on this one or are you planning that for a later pull request?

Juggling a few too many things. Fixed.

PR #442, opened 2020-03-18
Misc fixes

ptomato:
Best reviewed commit-by-commit, here's a bunch of miscellaneous small changes I've made while working on other things. Should be fairly uncontroversial.

### DIFF
--- a/polyfill/lib/date.mjs
+++ b/polyfill/lib/date.mjs
@@ -146,11 +146,8 @@ export class Date {
   static from(arg, options = undefined) {
     const disambiguation = ES.ToDisambiguation(options);
     let result = ES.ToDate(arg, disambiguation);
-    return this === Date ? result : new this(
-      GetSlot(result, YEAR),
-      GetSlot(result, MONTH),
-      GetSlot(result, DAY),
-    );
+    if (this === Date) return result;
+    return new this(GetSlot(result, YEAR), GetSlot(result, MONTH), GetSlot(result, DAY));
   }
   static compare(one, two) {
     if (!ES.IsDate(one) || !ES.IsDate(two)) throw new TypeError('invalid Date object');
--- a/polyfill/lib/date.mjs
+++ b/polyfill/lib/date.mjs
@@ -4,6 +4,12 @@ import {
   YEAR,
   MONTH,
   DAY,
+  HOUR,
+  MINUTE,
+  SECOND,
+  MILLISECOND,
+  MICROSECOND,
+  NANOSECOND,
   CreateSlots,
   GetSlot,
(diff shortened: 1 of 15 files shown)


spec/absolute.html
```diff
@@ -329,9 +329,8 @@ <h1>Temporal.Absolute.prototype.toLocaleString ( [ _locales_ [ , _options_ ] ] )
     </emu-clause>
 
     <emu-clause id="sec-temporal.absolute.prototype.tojson">
-      <h1>Temporal.Absolute.prototype.toJSON ( [ _timeZone_ ] )</h1>
+      <h1>Temporal.Absolute.prototype.toJSON ( )</h1>
```
Ms2ger: Hmm, I'm guessing it should be used, though.

spec/abstractops.html
```diff
@@ -142,6 +142,8 @@ <h1>FormatSecondsStringPart ( _second_, _millisecond_, _microsecond_, _nanosecon
     <h1>NonNegativeModulo ( _x_, _y_ )</h1>
     <emu-alg>
       1. Let _result_ be _x_ modulo _y_.
+      1. If _result_ is *&minus;0*, then
+        1. Return 0.
```
Ms2ger: Good catch.

polyfill/lib/date.mjs
```diff
@@ -123,13 +129,18 @@ export class Date {
     if (!ES.IsDate(this)) throw new TypeError('invalid receiver');
     return new Intl.DateTimeFormat(...args).format(this);
   }
-  withTime(timeLike) {
+  withTime(time) {
     if (!ES.IsDate(this)) throw new TypeError('invalid receiver');
+    if (!ES.IsTime(time)) throw new TypeError('invalid Temporal.Time object');
```
Ms2ger: There was already a test for this in `date.mjs`, but it was using `{ hours: 11, minutes: 30, seconds: 23 }`. Let's add a test that would have failed before as well. (Same for `withDate`.)

spec/absolute.html
ptomato: I'll just remove this commit from this PR for now, and open an issue to figure it out.

polyfill/lib/date.mjs
ptomato: The existing test was actually a bug, since it called withTime on the DateTime object that resulted from a previous call to withTime, and therefore always threw an exception because the method didn't exist. I've fixed that up and added a similar test to Time.withDate.

spec/absolute.html
ptomato: → #446

PR #447, opened 2020-03-18
Linter rules

ptomato:


package.json
```diff
@@ -25,12 +28,22 @@
...
     "build": "npm run build:javascript && npm run build:docs && npm run build:spec"
   },
   "repository": {
     "type": "git",
     "url": "git+https://github.com/tc39/proposal-temporal.git"
   },
   "author": "Ujjwal Sharma <usharma@igalia.com>",
-  "license": "SEE LICENSE IN https://tc39.github.io/ecma262/#sec-copyright-and-software-license"
+  "license": "SEE LICENSE IN https://tc39.github.io/ecma262/#sec-copyright-and-software-license",
+  "prettier": {
```
ljharb: I’d suggest running prettier only via eslint, that way you also turn off conflicting eslint rules.

polyfill/lib/timezone.mjs
```diff
@@ -69,23 +69,16 @@ export class TimeZone {
...
     const after = ES.GetTimeZoneOffsetNanoseconds(utcns.plus(bigInt(86400 * 1e9)), GetSlot(this, IDENTIFIER));
-    const diff = ES.ToDuration({
-      nanoseconds: after.minus(before)
-    }, 'reject');
+    const diff = ES.ToDuration(
+      {
+        nanoseconds: after.minus(before)
+      },
+      'reject'
+    );
```
Ms2ger:
```suggestion
    const nanoseconds = after.minus(before);
    const diff = ES.ToDuration({ nanoseconds }, 'reject');
```

kaizhu256: admit opinion is more political than technical (and unlikely to change minds),
but prefer switching enforcement from single-quote to double-quote:

#### pros:
1. better interoperability with sql-database having queries with single-quotes, e.g.:
```js
sqljsWorker.postMessage({
    action: "exec",
    sql: "INSERT INTO myTable (username, user_level) VALUES ('johndoe', 'admin');"
});
```

2. easier to inline multi-line javascript-code in [linux] shell-scripts like this real-world example:
```bash
#!/bin/sh
shImageToDataUri () {
# this shell-function will convert the image $1 to a data-uri string
    node -e '
/* jslint utility2:true */
(function () {
"use strict";
console.log(
    "data:image/"
    + require("path").extname(process.argv[1]).slice(1)
    + ";base64,"
    + require("fs").readFileSync(process.argv[1]).toString("base64")
);
}());
' "$1"
}

shImageToDataUri favicon.png
# stdout
# data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAADAAAAAwBAMAAAClLOS0AA...
```

3. jslint forbids single-quote strings (yea, this ones political but cant help myself ;)



#### cons:
1. disliked by new-users coming from python-background (which i once was)

2. awkward escaping of double-quotes when manipulating serialized JSON data, e.g.
```js
constantData = "{\"foo\":\"bar\"}";
```

ljharb: Single quotes for JS strings is by an exceedingly large margin the prevailing common style in the community.

SQL queries are a vanishingly small use case for JS compared to literally everything else that uses strings, inlining JS code like that is a bad practice anyways, and in any case you can use backticks if you want to avoid escaping.

kaizhu256: sqlite-queries (with isostrings) is the future of state-management in browsers.  the indexeddb/mongodb experiments to do table-joins in userland-javascript have ended in failure.

ptomato: The eslint rule proposed in this PR is for single quotes with an exception to allow double quotes when the string contains single quotes. Prettier does the same thing by default. So you can still quote SQL queries with double quotes, not that we would use any in this polyfill...

ptomato:
> Risking being overly opinionated… I'd like to remove the enforcement of single-quotes and no-trailing-commas and exclude the test262 test from the line length restriction.

Normally I'd prefer to have linter rules only on things that really do need to be consistent, such as indentation and semicolons, but in this case I figured prettier will rewrite those lines anyway, so we may as well have eslint rules for them. Prettier describes itself as "opinionated", so you can't really have it not care about a particular rule like you can with eslint. I'd also be happy to remove prettier if we find it too restrictive.

Alternatively, if the long lines and double quotes are needed in order to match the style of the rest of the test262 tests where these will be copied into, then I'm happy to exclude those files from prettier and make a separate eslint config for the test262 tests.

ljharb:
@ptomato if you include `eslint-plugin-prettier` and enable it (and inline the prettier configuration into the eslint config, rather than in package.json), and extend `eslint-config-pretter`, then all the eslint rules that overlap with prettier are disabled, and `eslint --fix .` will be all you need to do to format and lint everything.

ptomato:
I added another commit to run prettier via eslint. They do recommend keeping your prettier config outside of eslintrc so that external tools can continue to read it.

ptomato:
Let's hear from @Ms2ger on https://github.com/tc39/proposal-temporal/pull/447#issuecomment-601359494 and then decide whether to merge this or exclude the test262 dir from prettier.

ljharb:
@ptomato if prettier is only used with JS, no other tools would be using that config.

Ms2ger:
Please exclude test262, at least for now.

ptomato:
@Ms2ger I've excluded the test262 tests from prettier, and disabled the rules that you mentioned. There are still other rules that apply to test262 tests, e.g. semicolons, would you prefer those turned off as well or are you okay with this as-is?

@ljharb According to the advice in https://github.com/prettier/eslint-plugin-prettier#options, the Atom and VSCode plugins won't read prettier options inside eslintrc. It seems friendlier not to move the options inside eslintrc for that reason.

ptomato:
I don't know in what case one would use those plugins, I don't use them myself.

If you feel strongly about putting the prettier config in eslintrc, then we can do it, I'm just mentioning it because the documentation says "this is not recommended".

PR #454, opened 2020-03-19
Remove time zone argument from Absolute.toJSON()

ptomato:
Closes: #446

### DIFF
--- a/polyfill/lib/absolute.mjs
+++ b/polyfill/lib/absolute.mjs
@@ -128,16 +128,13 @@ export class Absolute {
   toString(timeZoneParam = 'UTC') {
     if (!ES.IsAbsolute(this)) throw new TypeError('invalid receiver');
     let timeZone = ES.ToTimeZone(timeZoneParam);
-    let dateTime = timeZone.getDateTimeFor(this);
-    let year = ES.ISOYearString(dateTime.year);
-    let month = ES.ISODateTimePartString(dateTime.month);
-    let day = ES.ISODateTimePartString(dateTime.day);
-    let hour = ES.ISODateTimePartString(dateTime.hour);
-    let minute = ES.ISODateTimePartString(dateTime.minute);
-    let seconds = ES.ISOSecondsString(dateTime.second, dateTime.millisecond, dateTime.microsecond, dateTime.nanosecond);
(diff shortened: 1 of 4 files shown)


polyfill/lib/absolute.mjs
```diff
@@ -139,6 +139,10 @@ export class Absolute {
     let resultString = `${year}-${month}-${day}T${hour}:${minute}${seconds ? `:${seconds}` : ''}${timeZoneString}`;
     return resultString;
   }
+  toJSON() {
+    if (!ES.IsAbsolute(this)) throw new TypeError('invalid receiver');
+    return this.toString('UTC');
```
Ms2ger: We should not call other methods on `this`; please move the formatting parts into an abstract op.

polyfill/lib/ecmascript.mjs
```diff
@@ -604,6 +604,18 @@ export const ES = ObjectAssign(ObjectAssign({}, ES2019), {
     let post = parts.length ? `.${parts.join('')}` : '';
     return `${secs}${post}`;
   },
+  AbsoluteToString: (absolute, timeZone) => {
+    const dateTime = timeZone.getDateTimeFor(absolute);
```
ljharb: should this be observably calling `getDateTimeFor` on a `timeZone`, or should it be calling an abstract operation?

ptomato: Good catch, that was something I didn't consider when writing this, but I propose leaving it for now. I'm working on a draft of custom TimeZone implementations (see #300) so it will have to call _something_ observably on a TimeZone, whether here or in another abstract operation, but this code will likely change anyway.

ljharb: What I would hope for is that it does a symbol lookup, for which the default is a method that defers entirely to a slot-reading abstract operation.

pipobscure:
I missed that this was a thing. The reason we have to JSON accept a timezone argument was to have a way to serialise a point in time relating to a specific timezone. This is the ONLY way to do that. So I disagree with the purpose of this PR!

Ms2ger:
> This is the ONLY way to do that.

It is not - `toString` keeps the `timeZone` argument.

PR #455, opened 2020-03-20
Fix Absolute.from() disambiguation

ptomato:
Closes: #415

### DIFF
--- a/polyfill/lib/ecmascript.mjs
+++ b/polyfill/lib/ecmascript.mjs
@@ -181,9 +181,38 @@ export const ES = ObjectAssign(ObjectAssign({}, ES2019), {
   ToTemporalAbsolute: (item) => {
     if (ES.IsTemporalAbsolute(item)) return item;
     const isoString = ES.ToString(item);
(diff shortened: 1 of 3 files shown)


ptomato:
The abstract operations that we use in the polyfill are not yet specified, so I didn't write the full specification, just an outline with TODOs.

polyfill/lib/ecmascript.mjs
```diff
@@ -186,9 +186,38 @@ export const ES = ObjectAssign(ObjectAssign({}, ES2019), {
...
+      month,
+      day,
+      hour,
+      minute,
+      second,
+      millisecond,
+      microsecond,
+      nanosecond
+    );
+    if (possibleEpochNs.length === 1) return new TemporalAbsolute(possibleEpochNs[0]);
```
Ms2ger: Hmm, now I wonder if we should drop this special case and make sure we verify the offset even if there's only a single result. Or am I missing something?

ptomato: I did it this way to preserve the current behaviour and make only the minimal changes necessary to fix the bug. Maybe this question is part of #428?

PR #459, opened 2020-03-21
Pin @babel/plugin-transform-for-of version

ptomato:
This seems to be a regression with Babel 7.9.0 that wasn't yet caught
in any of their emergency releases today.

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -29,9 +29,13 @@
     "@rollup/plugin-commonjs": "^11.0.2",
     "@rollup/plugin-node-resolve": "^7.1.1",
     "marked": "^0.7.0",
+    "npm-force-resolutions": "0.0.3",
     "rollup": "^1.23.1",
     "rollup-plugin-babel": "^4.3.3",
     "rollup-plugin-node-builtins": "^2.1.2",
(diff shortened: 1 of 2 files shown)


docs/package.json
```diff
@@ -29,9 +29,13 @@
...
     "@rollup/plugin-node-resolve": "^7.1.1",
     "marked": "^0.7.0",
+    "npm-force-resolutions": "0.0.3",
     "rollup": "^1.23.1",
     "rollup-plugin-babel": "^4.3.3",
     "rollup-plugin-node-builtins": "^2.1.2",
     "rollup-plugin-uglify": "^6.0.4"
+  },
+  "resolutions": {
+    "@babel/plugin-transform-for-of": "7.8.6"
```
ljharb: i think just making this a dep would work, without the force resolutions thing?

ptomato: `@babel/preset-env` would pull in 7.9.0, so I'm not sure if npm would dedupe it. Anyway, hopefully this will shortly be moot.

ljharb: ah, fair enough

ptomato:
Not necessary anymore.

PR #469, opened 2020-03-24
Disallow string arguments to plus/minus.

Ms2ger:


Ms2ger:
todo: add some more tests with non-string, non-object values. Will wait a bit in case someone objects to the change entirely, though.

PR #473, opened 2020-03-25
Add lint-fix command to fix lint errors easily.

Ms2ger:


package.json
```diff
@@ -24,6 +24,7 @@
     "cookbook": "cd polyfill && npm install && npm run cookbook && cd ..",
     "test262": "cd polyfill && npm install && npm run test262",
     "lint": "eslint . --ext js,mjs",
+    "lint-fix": "eslint . --ext js,mjs --fix",
```
ljharb:
```suggestion
    "lint-fix": "npm run lint -- --fix”,
```
but I’d also suggest just running that locally instead of adding a script :-)

ptomato:
It occurred to me that I already changed the `npm run pretty` script to do this ... you could just rename that to lint-fix or remove it altogether.

PR #474, opened 2020-03-26
Fix some specification build errors and warnings.

Ms2ger:


Ms2ger:
This wasn't caught due to https://github.com/bterlson/ecmarkup/issues/184.

PR #478, opened 2020-03-26
Editorial: Introduce a ParseTemporalAbsolute abstract operation.

Ms2ger:


spec/absolute.html
```diff
@@ -422,13 +422,21 @@ <h1>ToTemporalAbsolute ( _absolute_ )</h1>
       <emu-alg>
         1. If Type(_absolute_) is Object and _absolute_ has an [[InitializedTemporalAbsolute]] internal slot, then
           1. Return _absolute_.
-        1. Let _absolute_ be ? ToString(_absolute_).
-        1. Let _valid_ be *true* if _absolute_ matches <a href="https://github.com/tc39/proposal-temporal/blob/main/polyfill/lib/regex.mjs">the "absolute" regular expression in the polyfill</a> and *false* otherwise.
+        1. Let _string_ be ? ToString(_absolute_).
+        1. Let _result_ be ? ParseTemporalAbsolute(_string_).
+        1. Return ? CreateTemporalAbsolute(_result_).
```
ptomato: For consistency with the other types, I'd suggest:
```suggestion
        1. Return ? TemporalAbsoluteFromString(_string_).
```

TemporalAbsoluteFromString could then call a separate ParseTemporalAbsolute when we've specified how that works.

Ms2ger: I'm planning to inline all the Temporal*FromString ops into from() Soon:tm:, so will avoid the back and forth here :)

PR #479, opened 2020-03-27
Remove shims

ptomato:
Removes ad-hoc shims and workarounds for Symbol, WeakMap, and Object.assign.

Closes: #449.


polyfill/lib/ecmascript.mjs
```diff
@@ -1354,7 +1353,18 @@ function bisect(getState, left, right, lstate = getState(left), rstate = getStat
   }
   return right;
 }
-
+function unique(arr) {
```
ljharb: This whole function could be `return [...new Set(arr)]`

polyfill/lib/intl.mjs
```diff
@@ -118,7 +117,7 @@ function amend(options = {}, amended = {}) {
 function timeAmend(options) {
   options = amend(options, { year: false, month: false, day: false });
   if (!hasTimeOptions(options)) {
-    options = ObjectAssign(options, {
+    options = Object.assign(options, {
```
ljharb: Note that if you want these to be robust against runtime env modification, you still have to cache the function at module level.

polyfill/lib/ecmascript.mjs
ptomato: Unfortunately, that doesn't seem to work with the wrappers from the big-integer library.

ljharb: ah, because they don't use `===`, gotcha.

polyfill/lib/ecmascript.mjs
```diff
@@ -1355,7 +1354,18 @@ function bisect(getState, left, right, lstate = getState(left), rstate = getStat
...
+  for (var i = 0; i < arr.length; i++) {
+    var v = arr[i];
+    obj[v] = v;
+  }
+  var res = [];
+  for (var p in obj) {
+    res.push(obj[p]);
+  }
+  return res;
+}
```
Ms2ger: This is fine, but pretty heavy-handed for one call with an array of two elements. Followup to remove entirely?

ptomato: :+1:, see #488.

PR #481, opened 2020-03-27
BigInt fixes

ptomato:
- implement ECMA-262 ToBigInt in terms of big-integer library so that our conversions are more compliant
- only get the `bigInt.value` property if it is actually a native BigInt, otherwise return the big-integer wrapper

### DIFF
--- a/polyfill/lib/absolute.mjs
+++ b/polyfill/lib/absolute.mjs
@@ -24,12 +24,12 @@ export class Absolute {
   }
   getEpochMicroseconds() {
     if (!ES.IsTemporalAbsolute(this)) throw new TypeError('invalid receiver');
-    const value = bigInt(GetSlot(this, EPOCHNANOSECONDS));
-    return value.divide(1e3).value;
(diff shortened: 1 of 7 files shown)


polyfill/lib/absolute.mjs
```diff
@@ -27,12 +24,14 @@ export class Absolute {
...
-    return value.divide(1e3).value;
+    const value = GetSlot(this, EPOCHNANOSECONDS);
+    const result = value.divide(1e3);
+    return typeof BigInt === 'undefined' ? result : result.value;
   }
   getEpochNanoseconds() {
     if (!ES.IsTemporalAbsolute(this)) throw new TypeError('invalid receiver');
-    return bigInt(GetSlot(this, EPOCHNANOSECONDS)).value;
+    const result = GetSlot(this, EPOCHNANOSECONDS);
+    return typeof BigInt === 'undefined' ? result : result.value;
```
Ms2ger: This needs to be a helper function, IMO

Issue #328, opened 2020-01-25
Cookbook code for getParseableZonedStringAtInstant is incorrect


gibson042:
The code at [cookbook/getParseableZonedStringAtInstant.mjs](https://github.com/tc39/proposal-temporal/blob/ca126f887e765e222977ff4a4a8234d897045264/cookbook/getParseableZonedStringAtInstant.mjs) merged by ca126f887e765e222977ff4a4a8234d897045264 returns strings that lack UTC offset and time zone information required by the [use case](https://github.com/tc39/proposal-temporal/issues/240#user-content-zoned-instant-from-instant-and-time-zone).
```js
getParseableIanaZonedStringAtInstant(
    Temporal.Absolute.from("2020-01-09T00:00Z"),
    "Europe/Paris"
).replace(/(T\d+:\d+).*?([+-])/, "$1$2") ===
"2020-01-09T01:00+01:00[Europe/Paris]";
```

PR #482, opened 2020-03-27
Cookbook use-case fix and a few minor improvements

ptomato:
Closes: #328

### DIFF
--- a/docs/cookbook/absoluteFromLegacyDate.mjs
+++ b/docs/cookbook/absoluteFromLegacyDate.mjs
@@ -1,16 +1,8 @@
 import assert from 'assert';
 
-/**
- * Converts a Date instance into a Temporal.Absolute instance
- *
- * @param {Date} esDate This is a Date instance
- * @returns {Temporal.Absolute} Temporal.Absolute instance
- */
-function getAbsoluteFromDate(esDate) {
-  const date = Temporal.Absolute.fromEpochMilliseconds(esDate.getTime());
-  return date;
-}
+const legacyDate = new Date('1970-01-01T00:00:01Z');
 
-const date = new Date('1970-01-01T00:00:01Z');
-const absolute = getAbsoluteFromDate(date);
+// Convert legacy Date to Temporal.Absolute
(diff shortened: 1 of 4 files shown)


docs/cookbook/absoluteFromLegacyDate.mjs
```diff
@@ -1,16 +1,8 @@
...
- * Converts a Date instance into a Temporal.Absolute instance
- *
- * @param {Date} esDate This is a Date instance
- * @returns {Temporal.Absolute} Temporal.Absolute instance
- */
-function getAbsoluteFromDate(esDate) {
-  const date = Temporal.Absolute.fromEpochMilliseconds(esDate.getTime());
-  return date;
-}
+const legacyDate = new Date('1970-01-01T00:00:01Z');
```
gibson042: The function-based approach is more general and (perhaps more importantly) testable... why was it abandoned here?

docs/cookbook/getParseableZonedStringAtInstant.mjs
```diff
@@ -1,20 +1,42 @@
...
+  const timeZoneObject = Temporal.TimeZone.from(ianaTimeZoneName);
+  return getParseableZonedStringAtInstant(absolute, timeZoneObject);
+}
+
+/**
+ * Converts an absolute point in time to a string with the time in the specified
+ * time zone and a time zone designation.
+ *
+ * @param {Temporal.Absolute} absolute An absolute point in time
+ * @param {string} absolute An absolute point in time
```
gibson042: This parameter is misdocumented.
```suggestion
 * @param {Temporal.TimeZone} timeZone The target time zone
```

docs/cookbook/absoluteFromLegacyDate.mjs
Ms2ger: Generality isn't really a goal; this is just showing how the API can be used. The code is being tested in CI, so not sure what that's about.

docs/cookbook/getParseableZonedStringAtInstant.mjs
ptomato: Thanks, opened #489 as a follow up for this.

PR #483, opened 2020-03-27
meta: update meeting notes/agenda

ryzokuken:
Pretty self-explanatory.


littledan:
Would it be too much work to split up `weekly.md` into separate documents for each week? (Also, I don't know why you set up the "old" folder--those are just as much minutes; you can leave them as is.)

PR #495, opened 2020-04-03
Improved handling of option bags.

Ms2ger:
This should match the specification.

### DIFF
--- a/polyfill/test/Absolute/prototype/minus/order-of-operations.js
+++ b/polyfill/test/Absolute/prototype/minus/order-of-operations.js
@@ -1,0 +1,59 @@
+// Copyright (C) 2020 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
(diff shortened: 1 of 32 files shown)


polyfill/test/Absolute/prototype/minus/order-of-operations.js
```diff
@@ -0,0 +1,59 @@
...
+  hours: 1.7,
+  minutes: 1.7,
+  seconds: 1.7,
+  milliseconds: 1.7,
+  microseconds: 1.7,
+  nanoseconds: 1.7,
+};
+const argument = new Proxy(fields, {
+  get(target, key) {
+    actual.push(`get ${String(key)}`);
```
ptomato, line 40: nitpick: Throughout the tests, I think `String()` is redundant here.

Ms2ger: I thought I'd hit a `Symbol` here during development, but I cannot reproduce it now. Changed.

Issue #448, opened 2020-03-19
Improve assert imports across tests


ryzokuken:
Across tests, improve the imports for the Node.js assert module, as suggested by @ptomato over at https://github.com/tc39/proposal-temporal/pull/433#discussion_r393222870

PR #496, opened 2020-04-06
intl: reorganize tests, add coverage

ryzokuken:
some more intl fixes, better test organization and increase coverage

### DIFF
--- a/polyfill/test/intl.mjs
+++ b/polyfill/test/intl.mjs
@@ -4,14 +4,16 @@ const { describe, it, report } = Demitasse;
 import Pretty from '@pipobscure/demitasse-pretty';
 const { reporter } = Pretty;
 
-import { strict as assert } from 'assert';
-const { deepEqual, equal, throws } = assert;
+import { strict as Assert } from 'assert';
+const { deepEqual, equal, ok: assert, throws } = Assert;
 
 import { DateTimeFormat } from '../lib/intl.mjs';
 Intl.DateTimeFormat = DateTimeFormat;
 import * as Temporal from 'tc39-temporal';
 
 describe('Intl', () => {
+  // TODO: move these to their respective test files.
+
   describe('absolute.toLocaleString()', () => {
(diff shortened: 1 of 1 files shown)


polyfill/test/intl.mjs
```diff
@@ -4,14 +4,16 @@ const { describe, it, report } = Demitasse;
 import Pretty from '@pipobscure/demitasse-pretty';
 const { reporter } = Pretty;
 
-import { strict as assert } from 'assert';
-const { deepEqual, equal, throws } = assert;
+import { strict as Assert } from 'assert';
+const { deepEqual, equal, ok: assert, throws } = Assert;
```
Ms2ger, line 8: I think you'll want to revert this.

ryzokuken:
Converted to draft PR so that someone doesn't accidentally merge this, it is rebased on top of #496 in order to get a better idea of the actual coverage changes.

polyfill/test/intl.mjs
ryzokuken: I don't understand, what's wrong here?

ryzokuken:
Now this PR is ready for review/merge again. There's no change in coverage, because intl.mjs is already at 100% coverage (yay :tada:).

I did remove half the tests (I'm still testing for both English and German, just not together) for efficiency and readability, since it didn't affect coverage. If any of you insist, I'd test twice for each method again, but you may merge it as-is.

ptomato:
I don't mind this as-is, but I'd prefer continuing to test both English and German, for completeness that isn't expressed by the coverage number.

polyfill/test/intl.mjs
Ms2ger: We just made them all consistent in #448, why should this one be inconsistent?

Ms2ger, line 8:
```suggestion
import { strict as assert } from 'assert';
const { deepEqual, equal, throws } = assert;
```

PR #498, opened 2020-04-06
Draft of custom time zone API

ptomato:
See also #300

### DIFF
--- a/docs/timezone-draft.md
+++ b/docs/timezone-draft.md
@@ -1,0 +1,184 @@
+# Draft design of custom time zone API
+
+## Why?
+
+Currently, `Temporal.TimeZone` objects can only be obtained by passing an identifier to the constructor or to `Temporal.TimeZone.from()`.
+These identifiers can be:
+
+- an offset string such as `-07:00`, in which case a time zone object is returned with a constant offset, with no daylight saving changes ever;
+- an IANA time zone name such as `America/New_York`, in which case a time zone object is returned that will give a different offset depending on the date that it is queried with, which attempts to be historically accurate (within reason) as to the actual time that clocks showed there.
+
+No other identifiers than the ones deemed valid by one of the above two rules (`±HH:MM` or [IsValidTimeZoneName](https://tc39.es/ecma402/#sec-isvalidtimezonename)), are currently permitted.
+
+Custom time zones are a desirable feature because implementations may want to prevent leaking information about what historical revision of the time zone data is present on the host system, or control what information about the host system's current time zone is exposed.
+See [issue #273](https://github.com/tc39/proposal-temporal/issues/273).
+
+API users may also want to perform time zone calculations with older, or newer, or more accurate versions of the time zone data than what is present in their implementation.
+
+As an example, the `America/Chicago` time zone in the tzdata database is based on the historical time in Chicago.
+Before the year 1883, when the US introduced time zones due to the influence of railroads, Chicago observed the local mean time of GMT-05:50:36.
+Other cities in what is now the `America/Chicago` time zone would have observed their own local mean times.
+If a user wanted to do accurate calculations for times occurring before 1883 in places in the `America/Chicago` time zone other than Chicago, a custom time zone would be required.
+For example, St. Louis's local mean time is GMT-06:00:49.11 ([Byrd, Mary E., 1899, _A Laboratory Manual In Astronomy_, p. 63. Athenaeum Press: Boston.](https://books.google.ca/books?id=Xfg3AAAAMAAJ&pg=PA63))
+
+## Usage of a custom time zone
+
+Like built-in time zones, custom time zones have an `[[Identifier]]`.
+In these examples we assume a custom time zone class `StLouisTime` with the identifier `America/St_Louis`.
+(See the following section for how such a time zone object would be obtained.)
+
+When parsing an ISO 8601 string, the only places the time zone identifier is taken into account are `Temporal.Absolute.from()` and `Temporal.TimeZone.from()`.
+
+These functions gain an `options` parameter with an `idToTimeZone` option, whose value is a function returning the appropriate time zone object for the `id` passed in, or `null` to indicate that the time zone doesn't exist.
+
+If the option is not given, `Temporal.TimeZone.fromId()` is called.
+This is a new static method of `Temporal.TimeZone`.
+
+`Temporal.TimeZone.fromId()` can be monkeypatched by time zone implementors if it is necessary to make new time zones available globally.
+The expectation is that it would rarely be necessary to do so, because if you have implemented a custom time zone for a particular calculation, you probably don't need it to be available globally.
+
+> **FIXME:** Note that there is an issue open regarding this behaviour for calendars: [#294](https://github.com/tc39/proposal-temporal/issues/294)
+
+```javascript
+function idToTimeZone(id) {
+  if (id === 'America/St_Louis')
+    return new StLouisTime();
+  return Temporal.TimeZone.fromId(id);
+}
+
+Temporal.Absolute.from('1820-04-01T18:16:25-06:00[America/St_Louis]', { idToTimeZone })
+  // returns the Absolute corresponding to 1820-04-02T00:17:14.110Z
+
+Temporal.TimeZone.from('1820-04-01T18:16:25-06:00[America/St_Louis]', { idToTimeZone })
+  // returns a new StLouisTime instance
+```
+
+> **FIXME:** Passing a custom time zone's identifier to the built-in `Temporal.TimeZone` constructor currently doesn't work: `new Temporal.TimeZone('America/St_Louis')` throws.
+> However, this must change, because implementations would need to call `super(id)` to set the _[[Identifier]]_ and _[[InitializedTemporalTimeZone]]_ internal slots.
+> Maybe we need to only throw if `new.target === Temporal.TimeZone`?
+
+In order to lock down any leakage of information about the host system's time zone database, one would monkeypatch the `Temporal.TimeZone.fromId()` function which performs the built-in mapping, change the list of allowed time zones that `Temporal.TimeZone` iterates through, and replace `Temporal.now.timeZone()` to avoid exposing the current time zone.
+Or just replace the `Temporal.TimeZone` class and `Temporal.now` object altogether:
+
+```javascript
+// For example, to allow only offset time zones:
+
+TemporalTimeZone_fromId = Temporal.TimeZone.fromId;
+Temporal.TimeZone.fromId = function (id) {
+  if (/^[+-]\d{2}:?\d{2}$/.test(id) || id === 'UTC')
+    return TemporalTimeZone_fromId(id);
+  return null;
+}
+Temporal.TimeZone[Symbol.iterator] = function* () { return null; }
+Temporal.now.timeZone = function () { return Temporal.TimeZone.fromId('UTC'); }
+
+// or, to replace the built-in implementation altogether:
+
+Temporal.TimeZone = LockedDownTimeZoneImplementation;
+Temporal.now = lockedDownNowObject;
+```
+
+## Implementation of a custom time zone
+
+### TimeZone interface
+
+```javascript
+class Temporal.TimeZone {
+  /** Sets the [[Identifier]] internal slot to @id, and creates the
+   * [[InitializedTemporalTimeZone]] internal slot. A subclassed custom
+   * time zone must chain up to this constructor. */
+  constructor(id : string) : Temporal.TimeZone;
+
+  // Methods that a subclassed custom time zone must implement
+
+  /** Given an absolute instant returns this time zone's corresponding
+   * UTC offset, in nanoseconds (signed). */
+  getOffsetAtInstant(absolute : Temporal.Absolute) : number;
+
+  /** Given the calendar/wall-clock time, returns an array of 0, 1, or
+   * 2 absolute instants that are possible points on the timeline
+   * corresponding to it. In getAbsoluteFor(), one of these will be
+   * selected, depending on the disambiguation option. */
+  possibleInstants(dateTime : Temporal.DateTime) : array<Temporal.Absolute>;
+
+  /** Returns an iterator of all following offset transitions, starting
+   * from @startingPoint. */
+  *getTransitions(startingPoint : Temporal.Absolute) : iterator<Temporal.Absolute>;
+
+  // API methods that a subclassed custom time zone doesn't need to touch
+
+  get name() : string;
+  getDateTimeFor(absolute : Temporal.Absolute) : Temporal.DateTime;
+  getAbsoluteFor(
+      dateTime : Temporal.DateTime,
+      options?: object
+  ) : Temporal.Absolute;
+  getOffsetFor(absolute : Temporal.Absolute) : string;
+  toString() : string;
+  toJSON() : string;
+
+  static from(item : any, options?: object) : Temporal.TimeZone;
+  static fromId(id: string) : Temporal.TimeZone;
+  static [Symbol.iterator]() : iterator<Temporal.TimeZone>;
+}
+```
+
+All the methods that custom time zones inherit from `Temporal.TimeZone` are implemented in terms of `getOffsetAtInstant()`, `possibleInstants()`, and the value of the _[[Identifier]]_ internal slot.
+For example, `getOffsetFor()` and `getDateTimeFor()` call `getOffsetAtInstant()`, and `getAbsoluteFor()` calls both.
+
+> **FIXME:** These names are not very good.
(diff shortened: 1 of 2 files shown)


docs/timezone-draft.md
```diff
@@ -0,0 +1,169 @@
...
+  // Methods that a subclassed custom time zone must implement
+
+  /** Given an instant (in epoch nanoseconds) returns this time zone's
+   * corresponding UTC offset, in nanoseconds (signed). */
+  getOffsetAtInstant(epochNs : bigint) : bigint;
+
+  /** Given the calendar/wall-clock time, returns an array of 0, 1, or
+   * 2 instants (in epoch nanoseconds) that are possible points on the
+   * timeline corresponding to it. */
+  possibleInstants(
```
sffc, line 88: Why not pass a Temporal.DateTime here?

docs/timezone-draft.md
```diff
@@ -0,0 +1,169 @@
...
+class Temporal.TimeZone {
+  /** Sets the [[Identifier]] internal slot to @id, and creates the
+   * [[InitializedTemporalTimeZone]] internal slot */
+  constructor(id : string) : Temporal.TimeZone;
+
+  // Methods that a subclassed custom time zone must implement
+
+  /** Given an instant (in epoch nanoseconds) returns this time zone's
+   * corresponding UTC offset, in nanoseconds (signed). */
+  getOffsetAtInstant(epochNs : bigint) : bigint;
```
sffc, line 83: Why not pass a Temporal.Absolute here?

docs/timezone-draft.md
```diff
@@ -0,0 +1,169 @@
...
+  toString() : string;
+  toJSON() : string;
+
+  static from(item : any, options?: object) : Temporal.TimeZone;
+  static idToTimeZone(id: string) : Temporal.TimeZone;
+  static [Symbol.iterator]() : iterator<Temporal.TimeZone>;
+}
+```
+
+The `getOffsetAtInstant()` and `possibleInstants()` methods are conceptually "private", meaning that they are not intended as API to be normally used in user code.
```
sffc, line 120: Why not make them public?  It seems like the feature they expose is potentially useful.

docs/timezone-draft.md
```diff
@@ -0,0 +1,169 @@
...
+
+> **FIXME:** These methods are named "instant" and not "absolute" to reflect that they don't take a `Temporal.Absolute` object.
+> Help is welcome in determining the color of this bike shed.
+
+Alternatively, a custom time zone doesn't have to be a subclass of `Temporal.TimeZone`.
+In this case, it can be a plain object, which must implement all of the non-static methods in the interface, except for `getOffsetAtInstant()` and `possibleInstants()`.
+
+> **FIXME:** This means we have to remove any checks for the _[[InitializedTemporalTimeZone]]_ slot in any APIs outside of `Temporal.TimeZone`.
+
+> **FIXME:** `Temporal.TimeZone` is supposed to be an iterable through all time zones known to the implementation, but what do we do about custom time zones there?
```
sffc, line 134: We should decide how to do this in #294 for calendars and apply the same solution here.

docs/timezone-draft.md
ptomato: You're quite right, come to think of it! `getOffsetForInstant()` would be a potential answer to #349, and `possibleInstants()` would be a potential answer to #318.

docs/timezone-draft.md
ptomato: Here I figured that the only thing the code would do with a DateTime, is get the ISO fields, so why not just pass the ISO fields in directly. But if we do consider this method public as per your other comment, then I'd say you're right and it's more convenient for the caller to pass in the DateTime.

docs/timezone-draft.md
ptomato: I could go either way on this. I guess if we make it public then it should take an Absolute.

One minor consideration for not passing Temporal objects was to stop implementors from creating infinite loops by calling Absolute.inTimeZone or DateTime.inTimeZone in these methods, but I agree that if we consider the methods public then the other reasons for passing Temporal objects outweigh this consideration. Anyway, someone who made that mistake would find out soon enough.

docs/timezone-draft.md
sffc: They should call `dateTime.getISOFields()`, like we decided in #354.

docs/timezone-draft.md
sffc: The infinite loop was the same issue I was raising as a potential concern in #354.  We decided there that it was safe and sufficient to have methods like `.getISOFields()` that don't call into the calendar.  I think the same line of thinking should apply here.

This is only relevant when implementing custom time zones, so it's a bit of an edge case.

docs/timezone-draft.md
```diff
@@ -0,0 +1,160 @@
...
+class Temporal.TimeZone {
+  /** Sets the [[Identifier]] internal slot to @id, and creates the
+   * [[InitializedTemporalTimeZone]] internal slot */
+  constructor(id : string) : Temporal.TimeZone;
+
+  // Methods that a subclassed custom time zone must implement
+
+  /** Given an absolute instant returns this time zone's corresponding
+   * UTC offset, in nanoseconds (signed). */
+  getOffsetAtInstant(absolute : Temporal.Absolute) : bigint;
```
sffc, line 83: It should return a Temporal.Absolute, right?

docs/timezone-draft.md
```diff
@@ -0,0 +1,160 @@
...
+
+  /** Given the calendar/wall-clock time, returns an array of 0, 1, or
+   * 2 instants (in epoch nanoseconds) that are possible points on the
+   * timeline corresponding to it. */
+  possibleInstants(dateTime : Temporal.DateTime) : array<bigint>;
```
sffc, line 88: It should return an array of Temporal.Absolute, right?

docs/timezone-draft.md
ptomato: I suppose it could, but this is conceptually more of a duration than an absolute point on the timeline. I only didn't use Temporal.Duration because it can't represent a sign, and then I'd have to return a pair of `{ sign, duration }`.

sffc: Oh, I see, it's an offset from UTC, not nanos since epoch.  Never mind then; we should use a numeric type.  It's an interesting question about whether to use bigint or number.  Maybe open a thread to answer that outside of this PR.

ptomato: It's fairly minor and easy to change if we change our minds — I'll just change it to a number for now. The allowed offsets range from -24:00 to +24:00, not including endpoints, and 86400e9 is less than `Number.MAX_SAFE_INTEGER`, so a bigint doesn't seem necessary.

docs/timezone-draft.md
```diff
@@ -0,0 +1,160 @@
...
+> **FIXME:** `Temporal.TimeZone` is supposed to be an iterable through all time zones known to the implementation, but what do we do about custom time zones there?
+
+## Show Me The Code
+
+Here's what it could look like to implement the built-in offset-based time zones as custom time zones.
+The `MakeDate`, `MakeDay`, and `MakeTime` functions are as in the [ECMA-262 specification](https://tc39.es/ecma262/#sec-overview-of-date-objects-and-definitions-of-abstract-operations), except that instead of milliseconds, `MakeDate` and `MakeDay` return BigInt-typed nanoseconds, and `MakeTime` returns Number-typed nanoseconds.
+(This example leaves out all the type checking, range checking, and error handling, just to show the bare bones.)
+
+```javascript
+class OffsetTimeZone extends Temporal.TimeZone {
```
Ms2ger, line 133: Maybe add an `idToTimeZone` implementation as well.

docs/timezone-draft.md
```diff
@@ -0,0 +1,160 @@
...
+All the methods that custom time zones inherit from `Temporal.TimeZone` are implemented in terms of `getOffsetAtInstant()`, `possibleInstants()`, and the value of the _[[Identifier]]_ internal slot.
+For example, `getOffsetFor()` and `getDateTimeFor()` call `getOffsetAtInstant()`, and `getAbsoluteFor()` calls both.
+
+> **FIXME:** These names are not very good.
+> Help is welcome in determining the color of this bike shed.
+
+Alternatively, a custom time zone doesn't have to be a subclass of `Temporal.TimeZone`.
+In this case, it can be a plain object, which should (must?) implement all of the non-static methods in the interface.
+
+> **FIXME:** This means we have to remove any checks for the _[[InitializedTemporalTimeZone]]_ slot in any APIs outside of `Temporal.TimeZone`.
```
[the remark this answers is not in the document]
ptomato: When I wrote this I didn't intend to remove checks in the methods on `Temporal.TimeZone.prototype`, but I see in your example how that could be useful. If we did `Temporal.TimeZone.prototype.getOffsetFor.call(possiblyPlainObject, absolute)` internally then plain objects wouldn't have to implement all the methods. They'd only have to implement the methods that subclasses would have to implement. So I think we should support that.

docs/timezone-draft.md
```diff
@@ -0,0 +1,160 @@
...
+```
+
+All the methods that custom time zones inherit from `Temporal.TimeZone` are implemented in terms of `getOffsetAtInstant()`, `possibleInstants()`, and the value of the _[[Identifier]]_ internal slot.
+For example, `getOffsetFor()` and `getDateTimeFor()` call `getOffsetAtInstant()`, and `getAbsoluteFor()` calls both.
+
+> **FIXME:** These names are not very good.
+> Help is welcome in determining the color of this bike shed.
+
+Alternatively, a custom time zone doesn't have to be a subclass of `Temporal.TimeZone`.
+In this case, it can be a plain object, which should (must?) implement all of the non-static methods in the interface.
```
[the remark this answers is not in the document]
ptomato: I based that on the discussion in #300.

docs/timezone-draft.md
```diff
@@ -0,0 +1,160 @@
...
+  constructor(id : string) : Temporal.TimeZone;
+
+  // Methods that a subclassed custom time zone must implement
+
+  /** Given an absolute instant returns this time zone's corresponding
+   * UTC offset, in nanoseconds (signed). */
+  getOffsetAtInstant(absolute : Temporal.Absolute) : number;
+
+  /** Given the calendar/wall-clock time, returns an array of 0, 1, or
+   * 2 absolute instants that are possible points on the timeline
```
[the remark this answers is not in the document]
ptomato: > Do we want to enforce the maximum of two elements?

I don't know, I was wondering if it would theoretically be possible for a time zone to decide to have two DST transitions back-to-back so that a DateTime was repeated three times. That wouldn't be handled by getAbsoluteFor, since the disambiguation parameter only takes "earlier" or "later" into account, but possibleInstants could be used by user code to deal with that case.

If we did want to enforce the maximum then it might make more sense to return an object `{ earlier: time1, later: time2 }` and just have getAbsoluteFor take the key that corresponds to the disambiguation parameter, or throw if the two are different in "reject" mode. That would move the responsibility for deciding what to do in the case of 0 absolutes found, from getAbsoluteFor into possibleInstants, and therefore into user code in the case of custom time zones.

> Do we want to enforce this returns an array, rather than an iterable?

In the current form I'm not sure it makes sense to return an iterable, since the first thing we do is check its length.

> Should this comment say something about how we'd pick one of them?

I'll add this.

docs/timezone-draft.md
```diff
@@ -0,0 +1,160 @@
...
+
+Temporal.TimeZone.from('2020-04-01T18:16+00:00[Mars/Olympus_Mons]', { idToTimeZone })
+  // returns a new MarsTime instance
+```
+
+> **FIXME:** Passing a custom time zone's identifier to the built-in `Temporal.TimeZone` constructor currently doesn't work: `new Temporal.TimeZone('Mars/Olympus_Mons')` throws.
+> However, this must change, because implementations would need to call `super(id)` to set the _[[Identifier]]_ and _[[InitializedTemporalTimeZone]]_ internal slots.
+> Maybe we need to make `Temporal.TimeZone` not directly constructable?
+
+In order to perform attenuation, an implementation would monkeypatch the `Temporal.TimeZone.idToTimeZone()` function which performs the built-in mapping, and change the list of allowed time zones that `Temporal.TimeZone` iterates through.
```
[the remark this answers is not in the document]
ptomato: It comes from https://github.com/tc39/proposal-temporal/issues/273, but I agree it should be changed.

docs/timezone-draft.md
```diff
@@ -0,0 +1,160 @@
...
+
+No other identifiers than the ones deemed valid by one of the above two rules (`±HH:MM` or [IsValidTimeZoneName](https://tc39.es/ecma402/#sec-isvalidtimezonename)), are currently permitted.
+
+Custom time zones are a desirable feature because implementations may want to prevent leaking information about what historical revision of the time zone data is present.
+API users may also want to perform time zone calculations with older (or newer!) versions of the time zone data than what is present in their implementation.
+
+## Usage of a custom time zone
+
+Like built-in time zones, custom time zones have an `[[Identifier]]`.
+In these examples we assume a custom time zone class `MarsTime` with the identifier `Mars/Olympus_Mons`.
```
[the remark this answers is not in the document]
ptomato: Mars time was meant as a toy example, but I'm not sure I can come up with a more compelling use case that's not locking down information exposed from the host system. Maybe someone else has a good idea?

ryzokuken:

docs/timezone-draft.md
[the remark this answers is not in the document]
sffc: A use could be if you want to express the time in a location predating the railroad; for example, maybe you want "St. Louis Time in 1800", which would have been calibrated to the sun rather than to GMT.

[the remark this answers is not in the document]
ptomato: I managed to discover in a 120-year-old astronomy manual that St. Louis local mean time is GMT-06:00:49.11... isn't the internet great?! :smile:

docs/timezone-draft.md
```diff
@@ -0,0 +1,180 @@
...
+
+Like built-in time zones, custom time zones have an `[[Identifier]]`.
+In these examples we assume a custom time zone class `StLouisTime` with the identifier `America/St_Louis`.
+(See the following section for how such a time zone object would be obtained.)
+
+When parsing an ISO 8601 string, the only places the time zone identifier is taken into account are `Temporal.Absolute.from()` and `Temporal.TimeZone.from()`.
+
+These functions gain an `options` parameter with an `idToTimeZone` option, whose value is a function returning the appropriate time zone object for the `id` passed in, or `null` to indicate that the time zone doesn't exist.
+
+If the option is not given, `Temporal.TimeZone.idToTimeZone()` is called.
```
Ms2ger, line 34: If it wasn't for the fact that it makes sense to match the `idToTimeZone` option, I'd say `Temporal.TimeZone.fromId()` would make more sense. Now I don't know. On the other hand, do we need this given that `Temporal.TimeZone.from()` exists?

docs/timezone-draft.md
```diff
@@ -0,0 +1,180 @@
...
+}
+```
+
+All the methods that custom time zones inherit from `Temporal.TimeZone` are implemented in terms of `getOffsetAtInstant()`, `possibleInstants()`, and the value of the _[[Identifier]]_ internal slot.
+For example, `getOffsetFor()` and `getDateTimeFor()` call `getOffsetAtInstant()`, and `getAbsoluteFor()` calls both.
+
+> **FIXME:** These names are not very good.
+> Help is welcome in determining the color of this bike shed.
+
+Alternatively, a custom time zone doesn't have to be a subclass of `Temporal.TimeZone`.
```
Ms2ger, line 129: Another question: what happens if you call `Temporal.TimeZone.from()` with such a plain object?

docs/timezone-draft.md
ptomato: I think we can hash this out in #294.

docs/timezone-draft.md
ptomato: Since the plain object must implement the `name` property, it would be the same result as calling `Temporal.TimeZone.from(obj.name)`.

Ms2ger: Hmm, can we require `toString()`? Then `from()` will call it by default. But then again, you'll still need the `idToTimeZone`…

ptomato: Yes, that would be fine, implementing `name` would be optional then. I'll change this.

In #294 various ways of getting around the `idToTimeZone` requirement are suggested.

ptomato:
@Ms2ger Although #294 is still under discussion, I've changed this to use the `fromId()` method that we had suggested. I believe the comments are resolved as much as possible without #294, are there any blockers left in your opinion?

Ms2ger:
Let's merge it as a draft, we can iterate from there.

PR #499, opened 2020-04-07
Some cookbook changes

ptomato:
Implementing @littledan's suggestion to not use functions for one-liner cookbook examples.

See #240.

### DIFF
--- a/docs/cookbook/getTimeZoneObjectFromIanaName.mjs
+++ b/docs/cookbook/getTimeZoneObjectFromIanaName.mjs
@@ -1,29 +1,6 @@
-import assert from 'assert';
+// Construct a Temporal.TimeZone from an IANA name:
+const tz = Temporal.TimeZone.from('Europe/London');
 
-/**
- *
- * @param ianaTimeZoneName {string} TimeZone String
- * @returns {Temporal.Timezone}
- * @example getTimeZoneObjectFromIanaName("Europe/London")
- */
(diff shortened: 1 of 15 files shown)


docs/cookbook/getTimeZoneObjectFromIanaName.mjs
```diff
@@ -1,29 +1,6 @@
...
-// Same as above but during but non-DST
-const tzOffset2 = tz.getOffsetFor(Temporal.Absolute.from('2020-11-06T01:00Z'));
-assert.equal(tzOffset2, '+00:00');
-
-// Get a dateTime object using the timezone as an offset
-const dateTimeOffset = tz.getDateTimeFor(Temporal.Absolute.from('2020-08-06T15:00Z'));
-assert.equal(dateTimeOffset.toString(), '2020-08-06T16:00');
+// Cast the timezone back to an IANA name, two ways:
+tz.toString(); // Europe/London
+tz.name; // Europe/London
```
Ms2ger, line 6: Why not assert those?

ptomato: Readability, mostly... I figured these are tested elsewhere.

What I would _really_ like is to autogenerate assertions out of these comments when running these files, because I think the assertions decrease the readability of all the examples, but I don't know if I should spend a lot of time on that.

ptomato:
I added a few more commits to this.

ryzokuken:

PR #501, opened 2020-04-07
Fix Duration.toLocaleString

ptomato:
Define the abstract operation that is expected here, and add a test that
exercises toLocaleString without caring about the contents of the
returned string.

### DIFF
--- a/polyfill/lib/ecmascript.mjs
+++ b/polyfill/lib/ecmascript.mjs
@@ -640,6 +640,37 @@ export const ES = ObjectAssign({}, ES2019, {
     const timeZoneString = ES.ISOTimeZoneString(timeZone, absolute);
     return `${year}-${month}-${day}T${hour}:${minute}${seconds ? `:${seconds}` : ''}${timeZoneString}`;
   },
+  TemporalDurationToString: (duration) => {
+    function formatNumber(num) {
+      if (num <= Number.MAX_SAFE_INTEGER) return num.toString(10);
+      return bigInt(num).toString();
+    }
+    const dateParts = [];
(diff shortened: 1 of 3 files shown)


polyfill/lib/ecmascript.mjs
```diff
@@ -640,6 +640,37 @@ export const ES = ObjectAssign({}, ES2019, {
...
+    if (GetSlot(duration, MONTHS)) dateParts.push(`${formatNumber(GetSlot(duration, MONTHS))}M`);
+    if (GetSlot(duration, DAYS)) dateParts.push(`${formatNumber(GetSlot(duration, DAYS))}D`);
+
+    const timeParts = [];
+    if (GetSlot(duration, HOURS)) timeParts.push(`${formatNumber(GetSlot(duration, HOURS))}H`);
+    if (GetSlot(duration, MINUTES)) timeParts.push(`${formatNumber(GetSlot(duration, MINUTES))}M`);
+
+    const secondParts = [];
+    let ms = GetSlot(duration, MILLISECONDS);
+    let µs = GetSlot(duration, MICROSECONDS);
```
ryzokuken, line 659: While I appreciate the sentiment (and also realize that you didn't introduce this to begin with) do you think it's a good idea to change the name of this variable to make sure it's strictly a subset of Latin-1?

ptomato: I think I did introduce it to begin with, just not in this particular pull request. I'm in favour of this rather than using `us`, as I'm not aware of any downsides except on 15 years old browsers. It's quite possible I'm wrong about that, in that case I can make a separate PR to remove all occurrences of `µs`.

ptomato:
Merging as per discussion after fixing the linter error

PR #503, opened 2020-04-08
WIP: Add proposals for resolving questions with from()/Temporal.parse()

ptomato:
See: #428

### DIFF
--- a/docs/parse-draft.md
+++ b/docs/parse-draft.md
@@ -1,0 +1,389 @@
+parse
+=====
+
+Use cases we identified:
+
+1. **"Parsing ISO"** — Parse an ISO 8601 string and get the exact information that is mandated by ISO 8601, which means that the whole string is relevant. (24:00 is midnight the following day, :60 seconds are allowed although we follow POSIX in coercing them to :59).
+2. **"Parsing parts"** — Parse something that might look like an ISO 8601 string but might have been generated incorrectly by some other program. In this case it's important to ignore irrelevant parts of the string.
+
+## Pros and cons
+
+|   | 👍️ | 👎️ |
+|---|---|---|
+| Proposal 1. Always ISO 8601 semantics in from() | 👍️ **Parsing ISO** easy<br>👍️ Prioritizes common use case? | 👎️ **Parsing parts** needs workaround via Temporal.parse()<br>👎️ Inconsistency in not validating time zone offset/name mismatch |
+| Proposal 2. No irrelevant parts in from()       | 👍️ Always unambiguous about what parts of the string are parsed | 👎️ **Parsing ISO** needs workaround via Absolute.from() or Temporal.parse()<br>👎️ **Parsing parts** no easier than in proposal 1<br>👎️ Prioritizes less common use case?<br>  |
+| Proposal 3. Irrelevant parts ignored in from()  | 👍️ **Parsing parts** easy<br>👍️ Temporal.parse() not needed<br>👍️ from() less likely to throw | 👎️ from() is a mix of **parsing ISO** and **parsing parts**<br>👎️ Can parse an ISO string and not get the result specified by ISO<br>👎️ Not clear what API to use for each use case |
+
+## Details
+
+Here are some strings of interest, used in the examples below:
+```javascript
+// Valid string
+valid = '2001-09-08T18:46:40-07:00[America/Vancouver]';
+
+// String where the time affects the date
+time24 = '2020-04-07T24:00:00-07:00[America/Vancouver]';
+
+// Invalid time zone (although valid ISO 8601 string)
+invalidZone = '2001-09-08T18:46:40+01:00[America/Vancouver]';
+
+// Invalid date part, also invalid ISO 8601
+invalidDate = '2020-99-99T18:46:40-07:00[America/Vancouver]';
+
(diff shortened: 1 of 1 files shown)


docs/parse-draft.md
```diff
@@ -0,0 +1,331 @@
...
+Here are some strings of interest, used in the examples below:
+```javascript
+// Valid string
+valid = '2001-09-08T18:46:40-07:00[America/Vancouver]';
+
+// String where the time affects the date
+time24 = '2020-04-07T24:00:00-07:00[America/Vancouver]';
+
+// Invalid time zone (although valid ISO 8601 string)
+invalidZone = '2001-09-08T18:46:40+01:00[America/Vancouver]';
```
gibson042, line 18: This case is particularly interesting, because there are cases where you can't know if the offset is valid or invalid for the time zone (or where you can "know" but be wrong), e.g. `2019-12-15T10:23-02:00[America/Sao_Paulo]` was valid per 2019a tz data (announced March 2019), but not valid per 2019b (announced July 2019) because Brazil stopped observing Daylight Saving Time. Being strict without an escape hatch could also stymie well-intentioned code.

docs/parse-draft.md
```diff
@@ -0,0 +1,331 @@
...
+// Only date/time
+onlyDateTime = '2001-09-09T01:46:40';
+
+// Only date
+onlyDate = '2001-09-09';
+
+// In any of the proposals, these would only be accepted by YearMonth.from,
+// MonthDay.from, or Time.from, respectively
+onlyYearMonth = '2001-09';
+onlyMonthDay = '09-09';  // valid RFC 3339, I think technically not ISO 8601
```
gibson042, line 35: I don't think this comment is accurate; "MM-DD" is only ever a _part_ of a representation in either specification, not a full representation in its own right. The closest approximation I can find is in RFC 3339's [attempt to formalize ISO 8601 (1988)](https://tools.ietf.org/html/rfc3339#appendix-A), which includes `date = datespec-mday` and `datespec-mday = "--" dateopt-month date-mday`, allowing for values like "--09-09".

ryzokuken:

docs/parse-draft.md
```diff
@@ -0,0 +1,331 @@
+parse
+=====
+
+Use cases we identified:
+
+1. Parse an ISO 8601 string and get the exact information that is mandated by ISO 8601, which means that the whole string is relevant. (24:00 is midnight the following day, :60 seconds are coerced to :59).
```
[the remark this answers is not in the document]
ptomato: FWIW, hours 24 is valid ISO 8601 (and was added to the RFC 3339 grammar in an erratum.) I should elaborate here to say that what I mean by "parse an ISO 8601 string" is really "parse a string that conforms to our particular subset and interpretation of ISO 8601". In another pull request (#404, which I think can't be merged until this issue is resolved) I've made an [attempt at specifying that format exactly](https://github.com/tc39/proposal-temporal/pull/404/commits/20d4c4eba1f09961f707cd1f416b40fd55b36293#diff-9a345f12050ecc4961fb4fd6cec600ef).

docs/parse-draft.md
ptomato: I originally thought that the escape hatch would be Temporal.parse. But I agree with you, that is too cumbersome when parsing serialized strings that may have been invalidated by political changes since the time when they were serialized. I'll change this.

Temporal.Absolute would still need to consider both the offset and the IANA name, to be able to disambiguate in the case of a wall-clock time that is repeated in a DST transition.

docs/parse-draft.md
ptomato: You're right, I had misremembered this from when I was researching #404.

docs/parse-draft.md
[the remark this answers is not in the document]
gibson042: > FWIW, hours 24 is valid ISO 8601 (and was added to the RFC 3339 grammar in an erratum.)

My understanding is that it's only valid in time _intervals_.
> The end of day representation, where [hh] has a value of [24], shall not be used for a single time point.

> I should elaborate here to say that what I mean by "parse an ISO 8601 string" is really "parse a string that conforms to our particular subset and interpretation of ISO 8601". In another pull request (#404, which I think can't be merged until this issue is resolved)

That makes sense, but I'm pointing out that there can be problems when the ECMAScript interpretation differs from author requirements in a way that is difficult or impossible to correct.

docs/parse-draft.md
pipobscure: That was the motivation for the disambiguation parameter in the original `Temporal.Absolute.fromString`.

Because there are three valid parsing scenarios:

 1. IANA & Offset have to match (default)
 1. Use the IANA-Zone and only use offset to disambiguate
 1. Use the Offset and ignore IANA

ptomato: Hmm, (3) is not covered in this proposal for how `Temporal.parse()` works ... I'll revise it to make sure the offset is accessible separately.

ptomato:
Thanks for the review. I think the remaining open comments are on the topic which of these alternatives we should choose, not disagreeing on the contents of this document, so I'll merge this now. The issue remains open, because we still have to pick an alternative :smile:

PR #509, opened 2020-04-15
Add table of contents to cookbook

ptomato:
The document is getting quite long now, and would benefit from a table
of contents. (This is autogenerated by a plugin in my editor,
unfortunately. Since markedjs doesn't support generating tables of
contents, it's either this, or maintain it manually, or generate it in
JavaScript when loading the page.)


ptomato:
Another option would be to switch to a different markdown processor which can do this.

PR #512, opened 2020-04-17
Custom Markdown renderer

ptomato:
We now have enough junk tacked on to our Markdown rendering command in
package.json, and more junk generated at page load time, that it makes
sense to move it into a build script with a custom Markdown renderer.
This generates as much as possible statically, removing scripts from the
page where possible. (For Mermaid, it doesn't seem to be convenient.)


ptomato:
Thanks for the review. I did browse through the rendered documentation and verified that it still looked the same, so I'll merge this after the rebase completes CI.

Issue #510, opened 2020-04-16
Polyfill implementation of TimeZone.getTransitions can give incorrect information


ptomato:
It seems the bisect implementation that we have now for TimeZone.getTransitions has a few limitations, as it gives wrong information in some cases.

```js
> const nyc = Temporal.TimeZone.from('America/New_York');
> const a1 = Temporal.Absolute.from('2020-04-16T21:01Z');
> nyc.getTransitions(a1).next().value.toString()
'2021-11-07T06:00Z'  // should be 2020-11-01T06:00Z
> const a2 = Temporal.Absolute.from('1800-01-01T00:00Z');
> nyc.getTransitions(a2).next().value
null  // should be a time in 1883 when time zones were introduced in the US
```

ryzokuken:
I'll investigate.

PR #513, opened 2020-04-20
timezone: fix issues with getTransitions

ryzokuken:
Fixes: #510 

1. Change window size to 2 weeks.
2. Bump up upper search limit from lower start + 2y to now + 1y.

### DIFF
--- a/polyfill/lib/ecmascript.mjs
+++ b/polyfill/lib/ecmascript.mjs
@@ -797,12 +797,12 @@ export const ES = ObjectAssign({}, ES2019, {
     let rightOffset = leftOffset;
     let weeks = 0;
     while (leftOffset === rightOffset && weeks < 104) {
-      rightNanos = bigInt(leftNanos).plus(7 * 24 * DAYMILLIS * 1e6);
+      rightNanos = bigInt(leftNanos).plus(2 * 7 * DAYMILLIS * 1e6);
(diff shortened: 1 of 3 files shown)


pipobscure:
LGTM from phone. Will do full review when I’m back on my big screen

polyfill/lib/ecmascript.mjs
```diff
@@ -791,18 +791,18 @@ export const ES = ObjectAssign({}, ES2019, {
       return null;
     }
 
+    const current = new TemporalAbsolute(ES.SystemUTCEpochNanoSeconds());
```
pipobscure, line 794: I don’t think we need to construct an extra temporal object here. Do we?

This entire function works off raw bigint epochNanoSeconds

ryzokuken: I literally just construct this function to use the fancy `Absolute` math (see! It works!)

I didn't think it would be a big deal, but if you feel strongly about it, I'd do the nanosecond math manually.

pipobscure: I don’t feel strongly. Just thinking it adds unnecessary complexity 😀 but it really doesn’t matter much.

ryzokuken: @pipobscure what about now? :smile:

Issue #403, opened 2020-02-24
Should field properties be enumerable?


gibson042:
It would be nice to have e.g. `Object.assign({}, Duration.from("PT42S")).seconds === 42`.

gibson042:
That section says that built-in data properties are not enumerable "unless otherwise specified", but I'm suggesting that we _do_ specify otherwise because instances of the Temporal types are basically `struct`-like records very similar to objects with well-known property names and it would be convenient to preserve that similarity.

Just like the index properties of strings and arrays are enumerable, so too should be the properties of Temporal type instances that hold their structured data.

littledan:
About matching existing conventions, these feel pretty similar to RegExp properties like `global` to me. (I believe these are non-enumerable.) Or maybe we could consider that a mistake, and try to establish new conventions?

ptomato:
If we go in the direction of considering objects with well-known property names to be equivalent to Temporal objects, should we go so far as to remove some of the checks that we have in the spec for the [[InitializedTemporalX]] slot? e.g. allow things like `date.compare({year: 2020, month: 2, day: 27})` after all? or even `Temporal.DateTime.inTimeZone.call({year: 2020, month: 2, day: 27, hour: 10, minute: 27}, 'UTC')`?

Ms2ger:
> It would be nice to have e.g. `Object.assign({}, Duration.from("PT42S")).seconds === 42`.

Note that this will not work even if we make the fields enumerable; `Object.assign` only takes *own* properties into account, and these must be on the prototype.

gibson042:
I don't know if it's fair to say that they _must_ be on the prototype, but that is nonetheless a good point. :\

ptomato:
This would seem to be a good argument in favour of adding a Duration.with method; because `Duration.from({...otherDuration, seconds: 42})` wouldn't work.

ljharb:
Are these own data properties? I'd expect them to be accessors on the prototype, which would make their enumerability irrelevant for most use cases (object spread, Object.assign, Object.keys/entries/values, etc).

ptomato:
We did discuss that they would need to be own properties in order to be enumerable, but that deviates from an existing ES convention which is why we wanted input from the plenary on this.

ljharb:
They can be enumerable without being own, but they have to be both to be included in assign/spread.

Is this on the agenda yet?

ptomato:
I believe @pipobscure already put it on the agenda.

[...]

PR #520, opened 2020-04-21
Add getFields() method to types with fields

ptomato:
getFields() is how to get an object with enumerable own data properties.

Closes: #403

### DIFF
--- a/docs/time.md
+++ b/docs/time.md
@@ -357,3 +357,17 @@ time = Temporal.Time.from('15:23:30.003');
 date = Temporal.Date.from('2006-08-24');
 time.withDate(date)  // => 2006-08-24T15:23:30.003
 ```
+
+### time.**getFields**() : { hour: number, minute: number, second: number, millisecond: number, microsecond: number, nanosecond: number }
+
+**Returns:** a plain object with properties equal to the fields of `time`.
+
+This method can be used to convert a `Temporal.Time` into a record-like data structure.
+It returns a new plain JavaScript object, with all the fields as enumerable, writable, own data properties.
+
+Usage example:
+```javascript
+time = Temporal.Time.from('15:23:30.003');
+Object.assign({}, time).minute  // => undefined
+Object.assign({}, time.getFields()).minute  // => 23
+```
--- a/polyfill/test/Date/prototype/getFields/branding.js
+++ b/polyfill/test/Date/prototype/getFields/branding.js
@@ -1,0 +1,13 @@
+// Copyright (C) 2020 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+const getFields = Temporal.Date.prototype.getFields;
+
+assert.sameValue(typeof getFields, "function");
+
+assert.throws(TypeError, () => getFields.call(undefined), "undefined");
+assert.throws(TypeError, () => getFields.call(null), "null");
+assert.throws(TypeError, () => getFields.call(true), "true");
+assert.throws(TypeError, () => getFields.call(""), "empty string");
+assert.throws(TypeError, () => getFields.call(Symbol()), "symbol");
+assert.throws(TypeError, () => getFields.call(1), "1");
--- a/polyfill/test/Date/prototype/getFields/order-of-operations.js
+++ b/polyfill/test/Date/prototype/getFields/order-of-operations.js
@@ -1,0 +1,47 @@
+// Copyright (C) 2020 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.date.prototype.getfields
+includes: [compareArray.js]
+features: [Reflect]
+---*/
+
+const expected = [
+  "get day",
+  "valueOf day",
+  "get month",
+  "valueOf month",
+  "get year",
+  "valueOf year"
+];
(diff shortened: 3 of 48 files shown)


docs/time.md
```diff
@@ -356,3 +356,17 @@ time = Temporal.Time.from('15:23:30.003');
...
 time.withDate(date)  // => 2006-08-24T15:23:30.003
 ```
+
+### time.**getFields**() : { hour: number, minute: number, second: number, millisecond: number, microsecond: number, nanosecond: number }
+
+**Returns:** a plain object with properties equal to the fields of `time`.
+
+This method can be used to convert a `Temporal.Time` into a record-like data structure.
+It returns a new plain JavaScript object, with all the fields as enumerable, writable, own data properties.
+
```
sffc, line 366: I think we're allowing time to be calendar-dependent, too.

polyfill/test/Date/prototype/getFields/branding.js
```diff
@@ -0,0 +1,16 @@
+// Copyright (C) 2020 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+const getFields = Temporal.Date.prototype.getFields;
+
+assert.sameValue(typeof getFields, "function");
+
+assert.throws(TypeError, () => getFields.call(undefined), "undefined");
```
sffc, line 8: Why does getFields need to be branded?  It just delegates down to the calendar.  I think getISOFields maybe should be branded, but I see no reason to put that requirement on getFields.

polyfill/test/Date/prototype/getFields/order-of-operations.js
```diff
@@ -0,0 +1,47 @@
+// Copyright (C) 2020 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.date.prototype.getfields
```
sffc, line 5: The operations don't have to delegate back to the Temporal.Date.  They could go straight to the calendar methods:

```javascript
Temporal.Date.prototype.getFields = function() {
  return this.calendar.getFields(this);
}
Temporal.Calendar.prototype.getFields = function(input) {
  return {
    year: this.year(input),
    month: this.month(input),
    day: this.day(input)
  };
};
```

Going back to Temporal.Date would be more like:

```javascript
Temporal.Date.prototype.getFields = function() {
  return this.calendar.getFields(this);
}
Temporal.Calendar.prototype.getFields = function(input) {
  return {
    year: input.year,
    month: input.month,
    day: input.day
  };
};
```

Another option would be to make getFields return a string list:

```javascript
Temporal.Date.prototype.getFields = function() {
  let result = {};
  for (let field of this.calendar.getFields()) {
    result[field] = this[field];  // or this.calendar[field](this);
  }
  return result;
}
Temporal.Calendar.prototype.getFields = function(input) {
  return ["year", "month", "day"];
};
```

polyfill/test/Date/prototype/getFields/branding.js
ljharb: it should throw on a nullish receiver, whether it's branded or not

docs/time.md
ptomato: I think the last word on that was in #389, that days are always defined as solar days?

sffc: Yeah, days are solar days, but do we allow calendars to customize how they tell time?  I've heard of calendars that have 4 day periods (morning, afternoon, evening, night) instead of 2 day periods (AM and PM).  The days are still solar days, but within the days, they don't use the same clock system for telling time.

Probably a topic for another issue.

sffc: #522

ptomato: OK, thanks, I see what you meant now. I think it's indeed best to defer that to the issue you opened.

polyfill/test/Date/prototype/getFields/order-of-operations.js
ptomato: After working a bit more on the calendar, I think the last option (`Calendar.getFields` returning a list of property keys) is the best one, since it can be used for other things in the calendar implementation as well.

However, I believe that's orthogonal to this pull request; calendar or no calendar, the spec would still require the getters to be called in a particular order.

sffc: The question is whether the methods get called on the Temporal.Date.prototype, or whether they get called on the Temporal.Calendar.prototype directly.  I'm not sure which is better.

Can you file a follow-up issue?

ptomato: → #528

ptomato:
Added a `[propName: string]: any` to the signatures of the objects returned from getFields, since I realized that calendars can stick any extra properties they choose onto these.

ljharb:
`unknown` > `any` :-)

polyfill/test/Date/prototype/getFields/branding.js
ptomato: OK, since this method indeed does not read the internal slots, I've amended this to throw only on a null or non-object receiver.

FWIW, it really doesn't matter that much; if it delegates to the calendar, then the calendar will read the internal slots (built-in calendar) or call getISOFields (userland calendar) which will be branded.

PR #525, opened 2020-04-23
Add nearest UTC offset transition example to cookbook

ptomato:
This one works now thanks to #513 :smile: 

See: #240

### DIFF
--- a/docs/cookbook/getInstantOfNearestOffsetTransitionToInstant.mjs
+++ b/docs/cookbook/getInstantOfNearestOffsetTransitionToInstant.mjs
@@ -1,0 +1,34 @@
+/**
+ * Get the nearest following instant that the given time zone transitions to
+ * another UTC offset, inclusive or exclusive.
+ *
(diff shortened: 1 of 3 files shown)


docs/cookbook/getInstantOfNearestOffsetTransitionToInstant.mjs
```diff
@@ -0,0 +1,34 @@
...
+    [nearest] = timeZone.getTransitions(absolute);
+  }
+  return nearest;
+}
+
+const absolute = Temporal.Absolute.from('2020-04-16T21:01Z');
+
+const nyc = Temporal.TimeZone.from('America/New_York');
+const nextTransition = getInstantOfNearestOffsetTransitionToInstant(absolute, nyc, false);
+assert.equal(nextTransition.toString(), '2020-11-01T06:00Z');
```
Ms2ger, line 26: Nit: prefer not to use dates in the future... Who knows if NYC will still exist in November?

Issue #493, opened 2020-04-03
Review ToInteger calls


Ms2ger:
In particular, those can return ±∞, and I'm not quite sure we handle that correctly in all cases. This certainly needs tests, too.

littledan:
Hmm, should we go to 0 in these cases? Or throw an exception?

Ms2ger:
Probably depends on the case; if we're applying a disambiguation option, we could probably throw for "reject" and "balance", and clamp for "constrain", for example.

PR #536, opened 2020-04-29
Review ToInteger usage in the spec

ptomato:
Closes: #493


spec/monthday.html
```diff
@@ -294,18 +294,6 @@ <h1>ValidateMonthDay ( _month_, _day_ )</h1>
       </emu-alg>
     </emu-clause>
 
-    <emu-clause id="sec-temporal-balancemonthday" aoid="BalanceMonthDay">
-      <h1>BalanceMonthDay ( _month_, _day_ )</h1>
```
Ms2ger, line 298: Looks like we should add `RegulateTemporalMonthDay` that will call it.

ptomato: ...which is a problem you identified as issue #444 :smile: 

I guess the best course of action here is to just merge this pull request but drop the commit that removes BalanceMonthDay, and fix that in #444.

PR #537, opened 2020-04-30
parse: add Temporal.parse

ryzokuken:
Fixes #244 
Fixes #428

### DIFF
--- a/docs/parse.md
+++ b/docs/parse.md
@@ -1,0 +1,47 @@
+# `Temporal.parse`
+
+The `Temporal.parse` function parses a full ISO string and divides it into its
+respective components, which can then be used to produce the corresponding
+Temporal objects.
+
+This function can be used to parse full valid ISO strings that represent invalid
+absolutes and salvage valid components from the string.
+
+## Syntax
+
+```javascript
+const parseResult = Temporal.parse(input);
+```
+
+### Parameters
+
+- `input` (`string`): The input string to be parsed. It should be a full valid
+ISO string, but it may or may not represent a valid absolute.
+
+### Returns
+
+The returned value is an `object` that contains all the components of the parsed
+ISO string. It has a structure described by the following TypeScript interface:
+
+```typescript
+interface parseResult {
+    absolute: string;
+    dateTime: string;
+    date: string;
+    time: string;
+    yearMonth: string;
+    monthDay: string;
+    zone: {
+        ianaName?: string;
+        offset?: string;
+    }
+}
+```
+
+## Example Usage
+
+```javascript
+const parsed = Temporal.parse('1937-01-01T12:00:27.870+00:20');
+const time = Temporal.Time.from(parsed.time); // 12:00:27.870
+const date = Temporal.Date.from(parsed.date); // 1937-01-01
+```
--- a/docs/parse.md
+++ b/docs/parse.md
@@ -15,12 +15,12 @@ const parseResult = Temporal.parse(input);
 
 ### Parameters
 
-- `input` (`string`): The input string to be parsed. It should be a full valid
+- `input` (string): The input string to be parsed. It should be a full valid
 ISO string, but it may or may not represent a valid absolute.
 
 ### Returns
 
-The returned value is an `object` that contains all the components of the parsed
+The returned value is an object that contains all the components of the parsed
 ISO string. It has a structure described by the following TypeScript interface:
 
 ```typescript
--- a/polyfill/lib/parse.mjs
+++ b/polyfill/lib/parse.mjs
@@ -1,15 +1,14 @@
-import { absolute as regex } from './regex.mjs';
+import { ES } from './ecmascript.mjs';
 
 export function parse(isoString) {
-  const match = regex.exec(isoString);
-  if (!match) throw new RangeError(`Invalid ISO 8601 string: ${isoString}`);
+  const { absolute, dateTime, date, time, yearMonth, monthDay, ianaName, offset } = ES.ParseFullISOString(isoString);
   return {
-    absolute: match[0],
-    dateTime: match[1],
-    date: match[2],
-    yearMonth: `${match[3]}-${match[4]}`,
-    monthDay: `${match[4]}-${match[5]}`,
-    time: match[6],
-    zone: match[13]
+    absolute,
+    dateTime,
+    date,
+    time,
+    yearMonth,
+    monthDay,
+    zone: { ianaName, offset }
   };
 }
--- a/polyfill/lib/parse.mjs
+++ b/polyfill/lib/parse.mjs
@@ -1,0 +1,15 @@
+import { absolute as regex } from './regex.mjs';
+
+export function parse(isoString) {
+  const match = regex.exec(isoString);
+  if (!match) throw new RangeError(`Invalid ISO 8601 string: ${isoString}`);
+  return {
+    absolute: match[0],
+    dateTime: match[1],
+    date: match[2],
+    yearMonth: `${match[3]}-${match[4]}`,
+    monthDay: `${match[4]}-${match[5]}`,
+    time: match[6],
+    zone: match[13]
+  };
+}
--- a/polyfill/test/parse.mjs
+++ b/polyfill/test/parse.mjs
@@ -20,9 +20,8 @@ describe('Temporal.parse', () => {
   it('does not fail for valid ISO strings', () => assert(() => Temporal.parse('1937-01-01T12:00:27.870+00:20')));
 
   it('can be used to salvage valid parts of invalid ISO strings', () => {
-    const date = Temporal.Date.from('1922-12-30');
     const salvagedDate = Temporal.Date.from(Temporal.parse('1922-12-30T99:99:99Z').date);
-    equal(date.toString(), salvagedDate.toString());
+    equal(salvagedDate.toString(), '1922-12-30');
   });
 
   const parsed = Temporal.parse('1937-01-01T12:00:27.870+00:20');
--- a/polyfill/test/parse.mjs
+++ b/polyfill/test/parse.mjs
@@ -1,0 +1,43 @@
+#! /usr/bin/env -S node --experimental-modules
+
+// Copyright (C) 2020 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+import Demitasse from '@pipobscure/demitasse';
+const { describe, it, report } = Demitasse;
+
+import Pretty from '@pipobscure/demitasse-pretty';
+const { reporter } = Pretty;
+
+import { strict as assert } from 'assert';
+const { equal, throws } = assert;
+
+import * as Temporal from 'tc39-temporal';
+
+describe('Temporal.parse', () => {
+  it('is a function', () => equal(typeof Temporal.parse, 'function'));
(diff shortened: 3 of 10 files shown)


polyfill/lib/parse.mjs
```diff
@@ -0,0 +1,15 @@
...
+  const match = regex.exec(isoString);
+  if (!match) throw new RangeError(`Invalid ISO 8601 string: ${isoString}`);
+  return {
+    absolute: match[0],
+    dateTime: match[1],
+    date: match[2],
+    yearMonth: `${match[3]}-${match[4]}`,
+    monthDay: `${match[4]}-${match[5]}`,
+    time: match[6],
+    zone: match[13]
```
ptomato, line 13: I think it would be better to have the offset and the IANA name available as separate properties on this object, rather than one `zone: "+01:00[Europe/Berlin]"` property. That way, you can more easily make an intelligent decision on what to do if they don't agree. (With just one property, you'd have to parse the square brackets yourself.)

docs/parse.md
```diff
@@ -0,0 +1,47 @@
+# `Temporal.parse`
+
+The `Temporal.parse` function parses a full ISO string and divides it into its
+respective components, which can then be used to produce the corresponding
+Temporal objects.
+
+This function can be used to parse full valid ISO strings that represent invalid
+absolutes and salvage valid components from the string.
+
+## Syntax
```
ptomato, line 10: I'd prefer to keep the same style as the rest of the docs, putting the syntax in a h3:
```md
### Temporal.**parse**(input: string) : object
```
or
```md
### Temporal.**parse**(input: string) : { absolute: string, dateTime: string, etc... }
```

docs/parse.md
```diff
@@ -0,0 +1,47 @@
...
+
+```javascript
+const parseResult = Temporal.parse(input);
+```
+
+### Parameters
+
```
ptomato, line 17:
```suggestion
**Parameters:**

```

docs/parse.md
```diff
@@ -0,0 +1,47 @@
...
+- `input` (`string`): The input string to be parsed. It should be a full valid
```
ptomato, line 18:
```suggestion
- `input` (string): The input string to be parsed. It should be a full valid
```

docs/parse.md
```diff
@@ -0,0 +1,47 @@
...
+ISO string, but it may or may not represent a valid absolute.
+
+### Returns
+
+The returned value is an `object` that contains all the components of the parsed
```
ptomato, line 23:
```suggestion
The returned value is an object that contains all the components of the parsed
```

docs/parse.md
```diff
@@ -0,0 +1,47 @@
...
+ISO string. It has a structure described by the following TypeScript interface:
+
+```typescript
+interface parseResult {
```
ptomato, line 27: I have mixed feelings about this. On the one hand, it is very clear and concise and expresses exactly what is meant. On the other hand, before I knew any TypeScript, I found it annoying when documentation did this; after all, I'm writing JavaScript, not TypeScript.

If you decide to put the signature of the returned object in the h3 as suggested above, then I'd say this isn't necessary.

That said, I'd either make these all optional, or make them `string | null` (my preference), because we need a way to represent what happens when you are parsing a partial string.

docs/parse.md
```diff
@@ -0,0 +1,47 @@
...
+    monthDay: string;
+    zone: {
+        ianaName?: string;
+        offset?: string;
+    }
+}
+```
+
+## Example Usage
+
```
ptomato, line 42:
```suggestion
Example usage:
```

polyfill/test/parse.mjs
```diff
@@ -0,0 +1,43 @@
...
+
+describe('Temporal.parse', () => {
+  it('is a function', () => equal(typeof Temporal.parse, 'function'));
+  it('fails for invalid ISO strings', () => throws(() => Temporal.parse('this is an invalid ISO string'), RangeError));
+  it('does not fail for valid ISO strings', () => assert(() => Temporal.parse('1937-01-01T12:00:27.870+00:20')));
+
+  it('can be used to salvage valid parts of invalid ISO strings', () => {
+    const date = Temporal.Date.from('1922-12-30');
+    const salvagedDate = Temporal.Date.from(Temporal.parse('1922-12-30T99:99:99Z').date);
+    equal(date.toString(), salvagedDate.toString());
```
ptomato, line 25: I think it's fine to just put `'1922-12-30'` here instead of constructing `date`, because we just construct it from the string and then call toString on it.

docs/parse.md
ryzokuken: So, regarding this and all other related comments, I noted that I was deviating from the rest of the documentation, but I figured it would be okay, since all other pages described a class and then further used the format that you're suggesting to describe a method *on that class*, while this is a first-class method on the `Temporal` global. Since I found nothing similar to it (first-class methods, should be treated like `Date`, for example, not like `Date.from`) in the current documentation, and since we wanted to go with MDN-style documentation anyway, I went ahead and borrowed this format from MDN. If you still feel that I should use the same format here, then let me know.

docs/parse.md
ryzokuken: https://github.com/tc39/proposal-temporal/pull/537#discussion_r419914410

docs/parse.md
ryzokuken: Will do.

docs/parse.md
ryzokuken: Will do.

docs/parse.md
ryzokuken: https://github.com/tc39/proposal-temporal/pull/537#discussion_r419914410

polyfill/test/parse.mjs
ryzokuken: Got it.

docs/parse.md
ryzokuken: > I have mixed feelings about this.

I share your sentiments exactly. I tried to come up with a different way to document this which was just as clear yet concise, but couldn't. I'd understand if you'd ask me to do it through text instead.

> because we need a way to represent what happens when you are parsing a partial string

As of right now, we're not accepting partial ISO strings, nor was I planning to change that, since I didn't get that sentiment so far. Do you think we should support partial ISO strings?

docs/parse.md
ptomato: That sounds fine.

Although in the processing script for this Markdown there is some magic in h3's that generates a permalink icon, a human-readable anchor, and bolds the function name to make it stand out. I'd still recommend putting the function signature in a h3 if possible, to take advantage of that. So the anchor name would become `#parse` instead of `#Temporal-parse-input-string-object`.

Alternatively, you could make that magic smarter so that it only does the magic if the heading looks like a function signature, and processes h2's and other headings as well. (I didn't get around to doing this yet.)

docs/parse.md
ptomato: Oh, I hadn't realized that yet. I do think we need this. For example, DateTime strings without a time zone can just as well have invalid parts and need to be split up.

In the [draft doc](http://tc39.es/proposal-temporal/docs/parse-draft.html) I suggested something like this:
```js
Temporal.parse('2001-09-09T01:46:40')
// {
//   absolute: null,
//   dateTime: '2001-09-09T01:46:40',
//   date: '2001-09-09',
//   yearMonth: '2001-09',
//   monthDay: '09-09',
//   time: '01:46:40',
//   offset: null,
//   timeZone: null,
// }
```

ryzokuken: apart from the optional time zone, do you think there's any cases you'd want to support? Like parsing just a `Date` to get a `Date`, `YearMonth` and `MonthDay`? Asking this since I'd need to go back to my regex changes and undo some of the refactoring I did around `ES.ParseISODateTime`.

docs/parse.md
ryzokuken: I don't think we need markers in this tiny document, since it's for a single function anyway. Plus the whole link is just `parse.html`, which seems fine.

I added some code to make sure the special linking only applies to function signatures, though.

docs/parse.md
ptomato: > parsing just a `Date` to get a `Date`, `YearMonth` and `MonthDay`?

That was one of the use cases that I mentioned in http://tc39.es/proposal-temporal/docs/parse-draft.html, as well. Supporting it does make sense to me, but I'm open to being convinced otherwise.

ryzokuken: I'm sorry I missed that case. No, I don't want to convince you otherwise, I just wanted to ask your preference since my implementation would depend on that.

docs/parse.md
```diff
@@ -0,0 +1,47 @@
...
+    absolute: string;
+    dateTime: string;
+    date: string;
+    time: string;
+    yearMonth: string;
+    monthDay: string;
+    zone: {
+        ianaName?: string;
+        offset?: string;
+    }
```
justingrant, line 37: I have a bunch of concerns with the `zone` type here:
* there's no other Temporal API that accepts this shape. I worry that users would logically assume they could construct a TimeZone instance using this object, but they can't.
*  no other Temporal API calls it `zone`... it's always `timeZone`
* `ianaName` is called `timeZoneIdentifier` [elsewhere](https://tc39.es/proposal-temporal/docs/timezone.html#new-Temporal-TimeZone). Many users won't know what IANA is.
* `offset`'s name, unlike elsewhere in Temporal  (e.g. `TimeZone.prototype.getOffsetStringFor` vs. `TimeZone.prototype.getOffsetNanosecondsFor`), doesn't clarify whether it's a string or a number, or if numeric what unit it's using.
* IMHO it's a little weird to have only one nested type. Why nest `ianaName` under zone, but not day under `monthDay`?  Not suggesting we do this this, just pointing out the inconsistency. 

My suggestion would be to instead have two top-level properties, e.g. `timeZoneIdentifier` and `timeZoneOffset` (or maybe `timeZoneOffsetString`, although that may be implicit if all fields are strings?)

justingrant:
Some feedback on naming (let me know if there's a different issue I should comment on):
 
* Many developers will expect that a `parse()` method will return typed JS objects like [JSON.parse](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse) does.  Some users will be surprised to get strings back.
* Because it returns strings not objects, this is likely an API that only relatively advanced users or unusual use-cases should require. For basic usage, we should be directing users to `from()` methods. But I think there's a mismatch between the appealing, short name `parse` and the fact that we won't want most developers to use it.
* Its placement on the Temporal object means that this method will be *very* discoverable for IDE and browser devtools users via autocomplete, unlike `from` methods that are one level deeper and not obviously named for parsing. This doesn't mean we need to rename `from`, but having `parse` may make it harder for users to discover `from` because they might think that `parse` is the only place in Temporal where parsing happens. (One thing I see a lot in usability testing is users who find one "destination" and unfortunately stop hunting for more ways to solve their problem.)

For these reasons, I'd recommend calling this API something longer that: a) makes it clearer that only strings not objects are returned and b) makes the API a little more complicated-sounding so noobs will be encouraged to skip over it and keep hunting around until they find `from`.

Some ideas are below, but honestly I don't care too much about the name as long as it's longer and scarier.
* `parseIsoToStringParts` - currently my favorite
* `parseIsoStringParts`
* `parseIsoStringToParts`
* `parseStringToParts`

ryzokuken:
Closing due to inactivity.

PR #538, opened 2020-04-30
Add constructor parameter to calendar methods

ptomato:
As discussed in #387.

### DIFF
--- a/docs/calendar-draft.md
+++ b/docs/calendar-draft.md
@@ -58,27 +58,32 @@ All of the following methods return new Temporal objects.
 class Temporal.Calendar {
 	/** Constructs a Temporal.Date from a free-form option bag */
 	dateFromFields(
-		fields: object
+		fields: object,
+		constructor: function
 	) : Temporal.Date;
 
 	/** Constructs a Temporal.DateTime from a free-form option bag */
 	dateTimeFromFields(
-		fields: object
+		fields: object,
+		constructor: function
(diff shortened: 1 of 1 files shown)


docs/calendar-draft.md
```diff
@@ -352,8 +359,9 @@ Temporal.Date.from = function(thing: string | object, options: object) {
 As discussed earlier, Temporal.Date will defer to Temporal.Calendar methods wherever necessary.  Example implementation of selected Temporal.Date methods:
 
 ```javascript
-Temporal.Date.prototype.plus = function(duration) {
-	return this.calendar.plus(this, duration);
+Temporal.Date.prototype.plus = function(duration, options) {
+	const constructor = this.constructor[Symbol.species] || Temporal.Date;
```
littledan, line 363: Note, this isn't *exactly* what SpeciesConstructor does, but it's close enough for now.

ptomato: Yeah, this was my attempt at doing "approximately SpeciesConstructor" in a pseudocode one-liner, I considered actually putting `ES.SpeciesConstructor(this, Temporal.Date)` but felt that would be more confusing :smile:

littledan: If you have ES.SpeciesConstructor available, I'd use it.

ptomato: "available" is maybe not the right word here, as this code is for illustration and will never be run, but I've made this change.

PR #545, opened 2020-05-05
Untangle dependencies

ptomato:
This removes the warning about circular dependencies that is printed when bundling the JS sources.

### DIFF
--- a/polyfill/lib/temporal.mjs
+++ b/polyfill/lib/temporal.mjs
@@ -1,12 +1,9 @@
-import { ES } from './ecmascript.mjs';
-
-export const Absolute = ES.GetIntrinsic('%Temporal.Absolute%');
-export const TimeZone = ES.GetIntrinsic('%Temporal.TimeZone%');
-export const DateTime = ES.GetIntrinsic('%Temporal.DateTime%');
-export const Date = ES.GetIntrinsic('%Temporal.Date%');
-export const YearMonth = ES.GetIntrinsic('%Temporal.YearMonth%');
-export const MonthDay = ES.GetIntrinsic('%Temporal.MonthDay%');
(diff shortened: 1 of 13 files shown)


polyfill/test/all.mjs
```diff
@@ -9,39 +9,24 @@ import Demitasse from '@pipobscure/demitasse';
...
-void now;
-void timezone;
-void absolute;
-void date;
-void time;
-void datetime;
-void duration;
-void yearmonth;
-void monthday;
-void intl;
```
ryzokuken, line 44: :100: this had been bugging me since forever.

polyfill/lib/temporal.mjs
```diff
@@ -1,12 +1,9 @@
...
-
+export { Absolute } from './absolute.mjs';
+export { Date } from './date.mjs';
+export { DateTime } from './datetime.mjs';
+export { Duration } from './duration.mjs';
+export { MonthDay } from './monthday.mjs';
 export { now } from './now.mjs';
+export { Time } from './time.mjs';
+export { TimeZone } from './timezone.mjs';
+export { YearMonth } from './yearmonth.mjs';
```
ryzokuken, line 9: Just to be the devil's advocate here: is this any different from the original code? Is the only benefit of this that we don't need to import `intrinsticclass.mjs` anymore?

ptomato: To be honest, I'm not sure of the answer to that. That's why I was hoping @pipobscure would know why this file was using GetIntrinsic in the first place :smile: 

That said, I haven't been able to figure out a way in which it's _observably_ different; the polyfill still works, and it still works in bundled form in the browser console as well.

PR #547, opened 2020-05-07
Add storage tank example to cookbook

ptomato:
See: #240

### DIFF
--- a/docs/cookbook/storageTank.js
+++ b/docs/cookbook/storageTank.js
@@ -1,0 +1,52 @@
+// tankDataX and tankDataY are X and Y-axis data for the last 24 hours,
+// obtained from elsewhere, e.g. const [tankDataX, tankDataY] = fetchData();
+// tankDataX is an array of Temporal.Absolute, and tankDataY is an array of numbers.
+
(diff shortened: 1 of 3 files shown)


docs/cookbook/storageTank.js
```diff
@@ -0,0 +1,52 @@
...
+  .getDate()
+  .withTime(Temporal.Time.from('00:00'))
+  .inTimeZone(tankTimeZone);
+const atOrAfterMidnight = (x) => Temporal.Absolute.compare(x, tankMidnight) >= 0;
+const dataStartIndex = tankDataX.findIndex(atOrAfterMidnight);
+const labelFormatter = new Intl.DateTimeFormat(undefined, {
+  weekday: 'short',
+  hour: 'numeric',
+  minute: 'numeric',
+  timeZone: Temporal.now.timeZone().name
```
littledan, line 19: In general, you could just pass the timezone here, and toString would be called, giving the right result, right?

ptomato: Yes, that works too — I'll add a fixup in the next cookbook pull request.

PR #550, opened 2020-05-11
withYear() object argument

ptomato:
Closes: #490

### DIFF
--- a/docs/monthday.md
+++ b/docs/monthday.md
@@ -229,7 +229,7 @@ function reviver(key, value) {
 JSON.parse(str, reviver);
 ```
 
-### monthDay.**withYear**(_year_: number) : Temporal.Date
+### monthDay.**withYear**(_year_: number | object) : Temporal.Date
 
 **Parameters:**
 - `year` (number): A year, which must have a day corresponding to `monthDay`.
@@ -243,8 +243,11 @@ Usage example:
 ```javascript
 md = Temporal.MonthDay.from('08-24');
 md.withYear(2017)  // => 2017-08-24
+md.withYear({ year: 2017 })  // equivalent to above
 
 md = Temporal.MonthDay.from('02-29');
 md.withYear(2017)  // throws
 md.withYear(2020)  // => 2020-02-29
 ```
+
+In calendars where more information than just the year is needed to convert a `Temporal.MonthDay` to a `Temporal.Date`, you can pass an object to `withYear()` that contains the necessary properties.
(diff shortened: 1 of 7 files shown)


docs/monthday.md
```diff
@@ -248,3 +248,5 @@ md = Temporal.MonthDay.from('02-29');
 md.withYear(2017)  // throws
 md.withYear(2020)  // => 2020-02-29
 ```
+
+In calendars where more information than just the year is needed to convert a `Temporal.MonthDay` to a `Temporal.Date`, you can pass an object to `withYear()` that contains the necessary properties.
```
sffc, line 252: It's worth giving an example here.

```javascript
let md = Temporal.MonthDay.from({
  calendar: "japanese",
  month: 1,
  day: 1
});

let date = md.withYear({ era: "reiwa", year: 2 });
```

docs/monthday.md
```diff
@@ -248,3 +248,5 @@ md = Temporal.MonthDay.from('02-29');
 md.withYear(2017)  // throws
 md.withYear(2020)  // => 2020-02-29
```
sffc, line 249: I would also add

```javascript
md.withYear({ year: 2020 })  // equivalent to above
```

polyfill/test/monthday.mjs
```diff
@@ -132,6 +132,22 @@ describe('MonthDay', () => {
...
   });
+  describe('MonthDay.withYear()', () => {
+    const md = MonthDay.from('01-22');
+    it('takes a number argument', () => {
+      equal(`${md.withYear(2002)}`, '2002-01-22');
+    });
+    it('takes an object argument with year property', () => {
+      equal(`${md.withYear({ year: 2002 })}`, '2002-01-22');
+    });
+    it('needs at least a year property on the object', () => {
```
sffc, line 143: it('needs at least a year property on the object *in the ISO calendar*', ...

docs/monthday.md
ptomato: I tried to avoid doing that here, intending to postpone it until later, for two reasons: it won't currently work with the polyfill if someone tries it out, whereas all the rest of the examples do; and because of #526 the `era: 'reiwa'` is only going to work in the hacky version of the Japanese calendar implementation in my branch, and we may have to pick a different way to specify eras in the end.

I'll make sure to add this example to the calendar branch though, so that we make sure it gets merged into here in whatever form it ultimately takes.

sffc: OK.

ryzokuken:
Thanks @ptomato for the PR and @sffc for the review. I'll merge this in now.

PR #570, opened 2020-05-14
Add more examples of with() to the cookbook

ptomato:
Use cases from #561 suggested by @justingrant

Closes: #561
See also: #240

### DIFF
--- a/docs/cookbook/adjustDayOfMonth.mjs
+++ b/docs/cookbook/adjustDayOfMonth.mjs
@@ -1,0 +1,19 @@
+const date = Temporal.Date.from('2020-04-14');
+
+// Third day of next month:
+
+const thirdOfNextMonth = date.plus({ months: 1 }).with({ day: 3 });
+
+assert.equal(thirdOfNextMonth.toString(), '2020-05-03');
+
+// Last day of this month:
+
+const lastOfThisMonth = date.with({ day: date.daysInMonth });
+
+assert.equal(lastOfThisMonth.toString(), '2020-04-30');
+
+// On the 18th of this month at 8 PM:
+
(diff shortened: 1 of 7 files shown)


docs/cookbook/adjustDayOfMonth.mjs
```diff
@@ -0,0 +1,21 @@
...
+
+// Third day of next month:
+
+const thirdOfNextMonth = date.plus({ months: 1 }).with({ day: 3 });
+
+assert.equal(thirdOfNextMonth.toString(), '2020-05-03');
+
+// Last day of this month:
+
+const lastOfThisMonth = date.with({ day: 31 });
```
sffc, line 11: Does `date.with({ day: Infinity })` work?  (Months could in theory have more than 31 days)

Another option:

```javascript
const lastOfThisMonth = date.plus({ months: 1 }).minus({ days: 1 });
```

ptomato: > Does `date.with({ day: Infinity })` work?

Yes, that does work in the polyfill currently, although I haven't checked the spec to see if it's intended to work.

> ```js
> const lastOfThisMonth = date.plus({ months: 1 }).minus({ days: 1 });
> ```

For an input of May 15, that will give June 14. Maybe this?

```js
date.plus({ months: 1 }).with({ day: 1 }).minus({ days: 1 })
```

sffc: Sounds good.  Use either form, which ever you like more, or put them both into the cookbook

ptomato: I forgot about `day: date.daysInMonth` which would be more readable than either of those two. I'll go with that.

PR #571, opened 2020-05-15
Time zone offset in nanoseconds

ptomato:
Method discussed in the custom time zone API draft.

Closes: #349

### DIFF
--- a/docs/timezone.md
+++ b/docs/timezone.md
@@ -109,7 +109,7 @@ Effectively, this is the canonicalized version of whatever `timeZoneIdentifier` 
 
 ## Methods
 
-### timeZone.**getOffsetFor**(_absolute_: Temporal.Absolute) : string
+### timeZone.**getOffsetStringFor**(_absolute_: Temporal.Absolute) : string
 
 **Parameters:**
 - `absolute` (`Temporal.Absolute`): The time for which to compute the time zone's UTC offset.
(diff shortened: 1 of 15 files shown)


docs/timezone.md
```diff
@@ -127,36 +127,59 @@ Effectively, this is the canonicalized version of whatever `timeZoneIdentifier`
 
 ## Methods
 
-### timeZone.**getOffsetFor**(_absolute_: Temporal.Absolute) : string
+### timeZone.**getOffsetFor**(_absolute_: Temporal.Absolute) : number
```
sffc, line 130: Did we discuss whether this should return Number or BigInt?

ptomato: We didn't, but I put Number in the draft because the extreme values are ±86399999999999 (1 ns less than ±24:00) which is a safe integer.

sffc: @pdunkel @littledan opinions on this?

justingrant:
If `Duration` ends up supporting negative durations (#558), then should this method return a `Duration` instead of a number?

pipobscure:
I’d say no, since an offset is NOT a duration

justingrant:
AFAIK this is the only API in all of Temporal that returns a nanoseconds value without having the word "nanoseconds" in the name of the method, or accepts a nanoseconds value without having "nanoseconds" in the parameter name.

Should this method have "nanoseconds" in it too? Especially because comparable APIs *don't* return nanoseconds (e.g. [`Date.timeZoneOffset`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTimezoneOffset) returns a number of minutes, as does moment.js's [TimeZone](https://momentjs.com/timezone/docs/#/zone-object/) API).

That said, figuring out that the result is 10e6 larger than expected will be a pretty easy thing for a developer to find because (unlike the DST issues I keep complaining about) this will repro 100% of the time and is not a subtle difference.  Also this API is almost certainly used by experts (who're more likely to read the docs). So naming is less of a big deal here than other places.

ptomato:
That's true, elsewhere we have `getEpochNanoseconds` and `fromEpochNanoseconds`, so `getOffsetNanosecondsFor` seems consistent, and would contrast with `getOffsetStringFor`.

ptomato:
Revised to rename the method to `getOffsetNanosecondsFor`.

docs/timezone.md
sffc: Temporal.DateTime's [[Nanoseconds]] slot is a Number, so it's fine for this one to also be a Number.  I'll resolve this thread.

PR #582, opened 2020-05-18
docs: fix issue in monthday withYear documentation

ryzokuken:
cc @ptomato @justingrant

Fixes: #578

### DIFF
--- a/docs/monthday.md
+++ b/docs/monthday.md
@@ -245,7 +245,7 @@ JSON.parse(str, reviver);
 ### monthDay.**withYear**(_year_: number | object) : Temporal.Date
 
 **Parameters:**
-- `year` (number): A year, which must have a day corresponding to `monthDay`.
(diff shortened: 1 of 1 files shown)


docs/monthday.md
```diff
@@ -245,7 +245,7 @@ JSON.parse(str, reviver);
 ### monthDay.**withYear**(_year_: number | object) : Temporal.Date
 
 **Parameters:**
-- `year` (number): A year, which must have a day corresponding to `monthDay`.
+- `year` (number | object): A year, which must have a day corresponding to `monthDay`. Additionally, an object with a `'year'` field is also accepted.
```
ljharb, line 248:
```suggestion
- `year` (number | object): A year, which must have a day corresponding to `monthDay`. Additionally, an object with a `'year'` property is also accepted.
```
fields are class fields, which may or may not produce a property :-)

PR #588, opened 2020-05-19
Remove MonthDay.compare and add equals() methods

ptomato:
Closes: #523


docs/absolute.md
```diff
@@ -334,6 +334,25 @@ epoch.inTimeZone(utc).difference(billion.inTimeZone(utc), { largestUnit: 'years'
...
 ```
 
+### absolute.**equal**(_other_: Temporal.Absolute) : boolean
+
+**Parameters:**
+- `other` (`Temporal.Absolute`): Another time to compare.
+
+**Returns:** `true` if `absolute` and `other` are equal, or `false` if not.
+
+Compares two `Temporal.Absolute` objects for equality.
```
sffc, line 344: Here and elsewhere: it would be good, via either description text or the examples, to explain why you need `.equals()` and can't just use `===`

docs/yearmonth.md
```diff
@@ -313,6 +313,26 @@ ym.difference(other, { largestUnit: 'months' })  // => P154M
...
+### yearMonth.**equal**(_other_: Temporal.YearMonth) : boolean
+
+**Parameters:**
+- `other` (`Temporal.YearMonth`): Another month to compare.
+
+**Returns:** `true` if `yearMonth` and `other` are equal, or `false` if not.
+
+Compares two `Temporal.YearMonth` objects for equality.
+
+The difference with `Temporal.YearMonth.compare` is that this function will return `false` if the two months are expressed in different calendar systems.
```
sffc, line 325: Temporal.YearMonth.prototype.compare isn't as much of an analogy here because of the refDay field.  (I assume that this just performs a straight equality on the data members.)

docs/absolute.md
ptomato: Very good point, will do.

ljharb:
I'm confused; why would `compare() === 0` not always be identical to `.equals`? `.equals` is supposed to always be derived from compare, in languages with those as protocols.

ptomato:
By the way, `equal()` or `equals()`?

> I also realized that when this comes to the calendar, we may need another .equals() method there, because you can't `===` calendars since they aren't singletons any more.

I didn't add one to TimeZone yet because you can do `tz1.name === tz2.name`, and intuition-wise, unlike the other Temporal types, I think I wouldn't expect to be able to do `tz1 === tz2` on some random object and have it work.

> I'm confused; why would `compare() === 0` not always be identical to `.equals`? `.equals` is supposed to always be derived from compare, in languages with those as protocols.

In the absence of those precedents, I'd say this behaviour seems useful; in `compare()` you can compare the order of two dates, regardless of what calendar system they are in; in `equal()` you can tell whether two objects have the same internal state and are therefore identical. I don't think we can change `compare()` since a calendar can't be greater or lesser than another calendar, so if those precedents are important, then we have to change `equal()` to disregard the calendar.

ljharb:
I think that the calendar system shouldn't be disregarded for `compare`; how else can you have a consistent comparison function?

ptomato:
> I think that the calendar system shouldn't be disregarded for `compare`; how else can you have a consistent comparison function?

What do you think the result of this code should be?
```js
const today = Temporal.now.date();
Temporal.Date.compare(today.withCalendar('japanese'), today.withCalendar('hebrew'))
```

ljharb:
I don't have an intuition - but `compare` absolutely must pick one, or else it violates the spec's requirement that sort comparators be consistent.

sffc:
Due to the data model (#391), comparisons always take place in the ISO calendar space, no matter what is in the calendar slot.  So, in `.compare()`, "25 Iyar 5780" == "May 19, 2 Reiwa" == "2020-05-19".

ljharb:
In that case, `compare` is not useful for a consistent comparison function, which is the *sole* purpose of a compare method.

In other words, anything for which `.compare` returns 0 *must be conceptually equal* - `.equals()` must match it, in other words.

sffc:
The way to fix that issue would be to return false when the calendars differ. I don't really understand why that's important, though.

We could alternatively do string comparison on the calendar IDs to break ties.

ljharb:
When sorting, two elements are either conceptually identical (0), or need to be sorted (negative or positive). It's fine if two conceptually identical things are observably different! However, `.equals` is something that's supposed to be algebraically derivable from `compare`.

ptomato:
Where is this precedent coming from? There are no `equal()` or `equals()` methods anywhere in ECMAScript currently and it's not clear to me that the sole `compare()` method (Intl.Collator) adheres to it, since you can explicitly specify that accented letters compare equal with the `sensitivity` option:
```js
new Intl.Collator('en', { sensitivity: 'base' }).compare('e', 'é')
// => 0
```

ljharb:
@ptomato from other languages that have a "compare" protocol. see also [this proposal](https://github.com/hemanth/generic-comparison) on the next agenda.

sffc:
If we're comparing to other languages, Rust uses [PartialOrd](https://doc.rust-lang.org/std/cmp/trait.PartialOrd.html) and [PartialEq](https://doc.rust-lang.org/std/cmp/trait.PartialEq.html) for its `==` and `<` operators, without the requirement that the types are a perfect equivalence relation (NaN is an example they use).  Rust additionally has the Ord and Eq traits for types that are full equivalence relations.  The statement "`a.cmp(b) == Ordering::Equal` if and only if `a == b` and `Some(a.cmp(b)) == a.partial_cmp(b)` for all `a` and `b`" appears only when using `Ord` (equivalence relation).

sffc:
I can see the argument that what we do here sets precedent, since as @ptomato pointed out, there aren't many other `compare` methods.  I recall similar questions coming up when we were talking about the semantics of `===` on value types.  If ECMA-262 wants to declare that `.compare()` and `.equals()` should follow a strict equivalence relation, then that's fine and we can make it work here (e.g., by comparing calendar IDs in addition to the ISO fields).

ptomato:
The thing is, I find `compare()` less useful if it _doesn't_ compare ISO fields, and I find `equals()` less useful if it _does_. Would renaming `equals()` to `identical()` help?

ljharb:
I think that if there's utility in distinguishing things based on X, then all methods that compare two things should take those into account.

ptomato:
Are you saying that because it's _sometimes_ useful to check if two Temporal.Dates are identical including the calendar, I can _never_ compare if two Temporal.Dates with different calendars are the same date? If so, I don't agree.

ljharb:
Of course, both comparisons should be achievable! However, by default I think it's very important everything agrees with each other.

If `compare` would ignore the calendar, why not let `.equals` do so as well, and those who care about the calendar being different can do `a.equals(b) && a.calendar.equals(b.calender)` ?

sffc:
I guess you can always do `a.withCalendar('iso').equals(b.withCalendar('iso'))`.

If we need to check the same set of fields each time, as @ljharb is saying, then I propose we just perform a string comparison on the calendar's `.id` field and move on.

ptomato:
If we have to pick one, then I think it would be better to prioritize the ignore-calendar use case and have both `compare()` and `equals()` only read the ISO fields. For the record, I do think this is a mistake, but this is blocking further implementation of calendars so let's just go with it.

sffc:
I think the comparison order should be first ISO fields, and then calendar, such that calendar is only checked if two date[time]s are otherwise identical.  Probably make ISO always sort first (or last) in a list of calendars.  I think this should basically cover the "ignore calendar" use case, since you get dates in their natural order, with calendar merely as a tiebreaker.

For example, an ordering would be:

1. `1900-05-20`
1. `2020-05-20`
1. `2020-05-20[c=hebrew]`
1. `2020-05-20[c=japanese]`
2. `2020-05-22`
3. `2020-05-22[c=hebrew]`
4. `2020-05-22[c=japanese]`
5. `2020-06-01`

ptomato:
I agree calendar should be checked last, _if_ we check it at all, but in my estimation the most common use case is going to be checking if two dates are the same date, regardless of calendar, and I think the above suggestion makes that less ergonomic: `date1.withCalendar('iso8601').equals(date2.withCalendar('iso8601'))`.

PS. After the latest revision, this discussion no longer blocks the PR from being merged, since there is no calendar slot in the polyfill yet anyway ... :wink:

ljharb:
How can they be the same date if they're not using the same calendar?

ptomato:
These are all the same date, not using the same calendar:
- `Temporal.Date.from({ year: 2020, month: 5, day: 22, calendar: 'iso8601' })`
- `Temporal.Date.from({ year: 2020, month: 5, day: 22, calendar: 'gregory' })`
- `Temporal.Date.from({ year: 5780, month: 8, day: 28, calendar: 'hebrew' })`
- `Temporal.Date.from({ year: 1441, month: 9, day: 29, calendar: 'islamic' })`
- `Temporal.Date.from({ era: 'reiwa', year: 2, month: 5, day: 22, calendar: 'japanese' })`

ljharb:
@ptomato is that generally true? or is that just true in those instances? what i mean is, if they're the same because their ISO fields are the same, then great, that's what `.equals` and means. If they're not the same because their calendars are different, then `.compare` shouldn't ignore the calendar when deciding to produce `0`.

ptomato:
They're the same date because they're all showing on the calendars on people's walls at the same time :smile:

I've already given up on having compare() and equals() treat the calendar differently. So now the question is whether equals() and compare() _both_ compare the calendar after comparing the ISO fields, or _both_ ignore the calendar. If I understand correctly, either of these will fulfill your requirement.

If both methods compare the calendar after comparing the ISO fields, but you want to compare the actual dates, i.e. the dates I listed in my previous comment all compare equal, then you need to work around it like this:
- `date1.withCalendar('iso8601').equals(date2.withCalendar('iso8601'))`
- `dates.sort((d1, d2) => Temporal.Date.compare(d1.withCalendar('iso8601'), d2.withCalendar('iso8601'))`

If both methods ignore the calendar, but you want object identity, then you need to work around it like this:
- `date1.equals(date2) && date1.calendar.id === date2.calendar.id`
- `dates.sort((d1, d2) => Temporal.Date.compare(d1, d2) || d1.calendar.id <=> d2.calendar.id)` (I know there's no spaceship operator, but just to avoid cluttering it up with all the `?:`)

I estimate that comparison of actual dates is the more common use case, so I'd prefer the latter workarounds to the former ones.

ljharb:
Fair enough; I'll certainly defer to the champion group here on which alternative is better.

<sub>(ᕕ( ᐛ )ᕗ  [`<=>`](https://github.com/hemanth/generic-comparison#how-we-could-do-it))</sub>

sffc:
Can we make `.compare()` return -2 or +2 if the dates differ, -1 or +1 if the dates equal but the calendars differ, and 0 if the dates and calendars are both equal?

FWIW, I don't find the equality workaround to be painful.

sffc:
Just an observation about https://github.com/tc39/proposal-temporal/pull/588#issuecomment-632865851: that example also has the correct string ordering.  If we order calendar IDs alphabetically, then the ISO strings are also in lexicographic order, which seems nice.

ptomato:
> Can we make `.compare()` return -2 or +2 if the dates differ, -1 or +1 if the dates equal but the calendars differ, and 0 if the dates and calendars are both equal?

Sure, but I'd recommend moving that discussion to another issue. I can't implement it in this pull request because there is no calendar slot yet :smile:

Is there a precedent for using specific codes in the return value of a compare-like function?

ljharb:
`.sort` only cares about negative, positive, or zero. Any comparator that uses subtraction is technically providing "codes"; there's no problem with doing that, but it's not super clear to me what the value would be.

ptomato:
I opened #625 to summarize the discussion so far, and as a place to continue it.

PR #589, opened 2020-05-19
TimeZone.getPossibleAbsolutesFor()

ptomato:
See time zone draft document.

Can be used to implement #317, #318.

### DIFF
--- a/docs/cookbook/getInstantWithLocalTimeInZone.mjs
+++ b/docs/cookbook/getInstantWithLocalTimeInZone.mjs
@@ -1,0 +1,92 @@
+/**
+ * Get an absolute time corresponding with a calendar date / wall-clock time in
+ * a particular time zone, the same as Temporal.TimeZone.getAbsoluteFor() or
+ * Temporal.DateTime.inTimeZone(), but with more disambiguation options.
+ *
+ * As well as the default Temporal disambiguation options 'earlier', 'later',
+ * and 'reject', there are additional options possible:
+ *
+ * - 'earlierLater': Same as what the Moment Timezone and Luxon libraries do;
+ *   equivalent to 'earlier' when turning the clock back, and 'later' when
+ *   setting the clock forward.
+ * - 'clipEarlier': Equivalent to 'earlier' when turning the clock back, and
+ *   when setting the clock forward returns the time just before the clock
+ *   changes.
+ * - 'clipLater': Equivalent to 'later' when turning the clock back, and when
+ *   setting the clock forward returns the exact time of the clock change.
+ *
+ * @param {Temporal.DateTime} dateTime - Calendar date and wall-clock time to
+ *   convert
+ * @param {Temporal.TimeZone} timeZone - Time zone in which to consider the
+ *   wall-clock time
+ * @param {string} disambiguation - Disambiguation mode, see description.
+ * @returns {Temporal.Absolute} Absolute time in timeZone at the time of the
+ *   calendar date and wall-clock time from dateTime
+ */
+function getInstantWithLocalTimeInZone(dateTime, timeZone, disambiguation = 'earlier') {
+  // Handle the built-in modes first
+  if (['earlier', 'later', 'reject'].includes(disambiguation)) {
+    return timeZone.getAbsoluteFor(dateTime, { disambiguation });
+  }
+
+  const possible = timeZone.getPossibleAbsolutesFor(dateTime);
+
+  // Return only possibility if no disambiguation needed
+  if (possible.length === 1) return possible[0];
+
+  switch (disambiguation) {
+    case 'earlierLater':
+      if (possible.length === 0) return timeZone.getAbsoluteFor(dateTime, { disambiguation: 'later' });
+      return possible[0];
+    case 'clipEarlier':
+      if (possible.length === 0) {
+        const before = dateTime.inTimeZone(timeZone, { disambiguation: 'earlier' });
+        return timeZone
+          .getTransitions(before)
+          .next()
+          .value.minus({ nanoseconds: 1 });
+      }
+      return possible[0];
+    case 'clipLater':
+      if (possible.length === 0) {
+        const before = dateTime.inTimeZone(timeZone, { disambiguation: 'earlier' });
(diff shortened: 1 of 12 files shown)


ptomato:
> Is the term "instant" use anywhere else? We used to have Temporal.Instant before that was changed to Temporal.Absolute, and I wonder whether this is leftover from that time.

Indeed, I was not fond of the name, I was hoping this PR would trigger some name bikeshedding :grin: 

I think I originally came up with `possibleInstants` when I originally had the method returning an array of epoch-nanoseconds bigints, and then never changed the name when that changed to array of Absolute. Maybe `getPossibleAbsolutesFor` would be a better name, analogous to `getAbsoluteFor`?

docs/cookbook/getInstantWithLocalTimeInZone.mjs
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  const possible = timeZone.getPossibleAbsolutesFor(dateTime);
+
+  // Return only possibility if no disambiguation needed
+  if (possible.length === 1) return possible[0];
+
+  switch (disambiguation) {
+    case 'earlierLater':
+      if (possible.length === 0) return dateTime.plus({ hours: 1 }).inTimeZone(timeZone);
```
sffc, line 39: Should you handle time zones that potentially switch by something other than `{ hours: 1 }`?

docs/timezone.md
```diff
@@ -211,6 +211,21 @@ For usage examples and a more complete explanation of how this disambiguation wo
 
 If the result is outside the range that `Temporal.Absolute` can represent, then a `RangeError` will be thrown, no matter the value of `disambiguation`.
 
+### timeZone.**getPossibleAbsolutesFor**(_dateTime_: Temporal.DateTime) : array&lt;Temporal.Absolute&gt;
+
+**Parameters:**
+- `dateTime` (`Temporal.DateTime`): A calendar date and wall-clock time to convert.
+
+**Returns:** An array of 0, 1, or 2 `Temporal.Absolute` objects.
```
sffc, line 219: Is it guaranteed to be 0, 1, or 2?

docs/cookbook/getInstantWithLocalTimeInZone.mjs
ptomato: True, maybe adding a day would be safer.

docs/timezone.md
ptomato: @Ms2ger asked the a similar question in https://github.com/tc39/proposal-temporal/pull/498#discussion_r406020942 and I'm not sure we really resolved at the time whether to enforce that or not.

I doubt there is a triply-occurring wall-clock time in any of the supported IANA time zones. You could implement a custom time zone that had one, but you wouldn't be able to use `DateTime.inTimeZone()` to access all three Absolutes anyway, so you'd have to have a function like in the cookbook example to get your own `earlier`/`middle`/`later` disambiguation behaviour.

docs/cookbook/getInstantWithLocalTimeInZone.mjs
ptomato: ...or I could just let Temporal do it.

docs/cookbook/getInstantWithLocalTimeInZone.mjs
```diff
@@ -0,0 +1,92 @@
...
+  // Return only possibility if no disambiguation needed
+  if (possible.length === 1) return possible[0];
+
+  switch (disambiguation) {
+    case 'earlierLater':
+      if (possible.length === 0) return timeZone.getAbsoluteFor(dateTime, { disambiguation: 'later' });
+      return possible[0];
+    case 'clipEarlier':
+      if (possible.length === 0) {
+        const before = dateTime.minus({ hours: 1 }).inTimeZone(timeZone);
```
sffc, line 43: The change above looks good, but you still have `{ hours: 1 }` hiding down here.  Is there a way to do this more generally?

gilmoreorless: From testing this locally, it looks like the `.minus({ hours: 1 })` isn't even needed. The subsequent `.inTimeZone(timeZone)` call is using the default `earlier` disambiguation, so it always generates an absolute _before_ the transition.
I've tested it with the most extreme ends of transition ranges I could think of off the top of my head:

* `Australia/Lord_Howe` only changing clocks by 30 minutes for DST (using `2019-10-06T02:15`).
* `Pacific/Apia` skipping the entirety of December 30, 2011 by moving the International Date Line (using `2011-12-30T12:00`).

Both cases worked correctly with and without the `.minus({ hours: 1 })`.

ptomato: Ah, thanks, that's true — I've updated this (although spelling out the `disambiguation: 'earlier'` explicitly).

gilmoreorless: > although spelling out the disambiguation: 'earlier' explicitly

Good move, it's now much easier to understand what it's doing at a glance.

docs/timezone.md
sffc: What's the *benefit* of not allowing more than 2?  What if WackoTimeZone decided that instead of repeating the hour one time at a DST transition, it would repeat the same minute 60 times in a row?  Or what if it decided to make the clocks stand still for the whole hour, leading to many thousands of overlaps (one per nanosecond in an hour)?

If there's a benefit to capping this, fine.  I agree that it's not common.  But if there isn't a good reason, I don't see why to impose that limitation.

docs/cookbook/getInstantWithLocalTimeInZone.mjs
sffc: Much better!  Thanks!

docs/timezone.md
ptomato: This code doesn't currently impose any limitation, so to make sure I understand, the only change you are suggesting is in the wording of the docs? (I guess I should change the cookbook example as well to use `possible[possible.length - 1]` instead of `possible[1]` as well)

sffc:
Thank you!

PR #596, opened 2020-05-20
[Do Not Merge] Annotating cookbook with calendar examples

sffc:
This is split from #590 to illustrate how cookbook examples would be affected by default calendar choices.


ptomato:
I'm assuming this can be closed since we made the appropriate changes to the cookbook long ago.

PR #619, opened 2020-05-25
Custom time zones, phase 1

ptomato:
Prep work for implementing custom time zones. The next pull request will be stacked on top of this one.

### DIFF
--- a/polyfill/lib/timezone.mjs
+++ b/polyfill/lib/timezone.mjs
@@ -82,11 +82,12 @@ export class TimeZone {
     switch (disambiguation) {
       case 'earlier': {
         const earlier = dateTime.minus(diff);
-        return this.getAbsoluteFor(earlier, disambiguation);
+        return this.getPossibleAbsolutesFor(earlier)[0];
       }
       case 'later': {
         const later = dateTime.plus(diff);
-        return this.getAbsoluteFor(later, disambiguation);
+        const possible = this.getPossibleAbsolutesFor(later);
+        return possible[possible.length - 1];
       }
       case 'reject': {
         throw new RangeError('no such absolute found');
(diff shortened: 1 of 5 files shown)


spec/datetime.html
```diff
@@ -816,6 +805,25 @@ <h1>ConstrainDateTime ( _year_, _month_, _day_, _hour_, _minute_, _second_, _mil
       </emu-alg>
     </emu-clause>
 
+    <emu-clause id="sec-temporal-balancedatetime" aoid="BalanceDateTime">
+      <h1>BalanceDateTime</h1>
```
ryzokuken, line 809: It'll be nice to add a `<p>` talking a little about `BalanceDateTime`. No hurries though, I could add one later while scrubbing the spec for Stage 3.

ryzokuken: What about arguments though? I'd really like if you added arguments to this line.

polyfill/lib/timezone.mjs
```diff
@@ -73,20 +87,30 @@ export class TimeZone {
...
     switch (disambiguation) {
       case 'earlier': {
         const earlier = dateTime.minus(diff);
-        return this.getAbsoluteFor(earlier, disambiguation);
+        return this.getPossibleAbsolutesFor(earlier)[0];
       }
       case 'later': {
         const later = dateTime.plus(diff);
-        return this.getAbsoluteFor(later, disambiguation);
+        return this.getPossibleAbsolutesFor(later)[0];
```
ryzokuken, line 113: Does this really match the previous behavior?

ptomato: Hmm, I thought it didn't matter, but I see now it could make a difference if a time zone had two transitions immediately adjacent to each other.

spec/datetime.html
ptomato: Oops! I've added the arguments, but I skipped the `<p>` for now because none of the other operations have one and I'm trying to land this quickly :sweat_smile:

polyfill/lib/timezone.mjs
ptomato: Fixed this case, I think.

ptomato:
Thanks for the review!

PR #631, opened 2020-05-29
Custom time zones, phase 3

ptomato:
Stacked on top of #620, for the time being, and marked as a draft until that is merged. This is the rest of the support needed for custom time zones. Adds tests to make sure that passing in a subclass of Temporal.TimeZone and a plain object with the required methods both work.


ptomato:
This is ready for review now.

PR #647, opened 2020-06-04
Add additional i18n correctness example involving getMonthDay()

sffc:


docs/calendar-draft.md
```diff
@@ -334,6 +334,7 @@ The following table describes these semantics.  Option 5 is not shown because th
 | T.now.isoDate() | N/A | N/A | N/A | N/A | Full ISO |
 | absolute.inTimeZone() | Full ISO | Explicit | Partial ISO | Environ. | Explicit |
 | absolute.inZoneISO() | N/A | N/A | N/A | N/A | Full ISO |
+| date.getMonthDay() | Full ISO | Inherit | Explicit | Environ. | Inherit |
```
ptomato, line 337: I think in options 1 and 4 this is "Inherit" as well, no? I can only imagine that we'd want to do something different than "Inherit" in options 3 and 5.

docs/calendar-draft.md
```diff
@@ -350,7 +351,39 @@ As compared to option 1 (always default to Full ISO), the following are examples
...
+/// Options 2 and 6: calendar in factory method
+const calendar = navigator.locales[0].getLikelyCalendar();
+const today = Temporal.now.date(calendar);
+const monthDay = today.getMonthDay();
+console.log("Today is:", monthDay.toLocaleString());
+
+/// Options 3 and 5: calendar only when needed
+const today = Temporal.now.date();
+const calendar = navigator.locales[0].getLikelyCalendar();
+const monthDay = today.getMonthDay(calendar);
```
ptomato, line 384: I'm not convinced this is a different use case than the below one, why would we add a calendar argument to `getMonthDay()` rather than calling `today.withCalendar(calendar).getMonthDay()` to convert `today` from partial to full ISO (3) or to get the appropriate calendared type (5)? Fundamentally this problem lies with `Temporal.now.date()`, not `getMonthDay()`.

sffc: Makes sense.  I split this into two different explicit options.

docs/calendar-draft.md
sffc: Clarified (See below)

sffc:
Thanks.  I feel it is distinct from the birthday example because this example is about formatting, not business logic, and MonthDay is a tempting mechanism for programmers to use when they want to render a localized string with the month and day but not year (something we know programmers want to do).

PR #681, opened 2020-06-16
timezone: call toString in toJSON

ryzokuken:
/cc @ptomato @rbuckton

### DIFF
--- a/polyfill/lib/timezone.mjs
+++ b/polyfill/lib/timezone.mjs
@@ -222,6 +222,9 @@ export class TimeZone {
     if (!ES.IsTemporalTimeZone(this)) throw new TypeError('invalid receiver');
     return this.name;
   }
+  toJSON() {
+    return this.toString();
+  }
   static from(item) {
     if (ES.IsTemporalTimeZone(item) || (typeof item === 'object' && item)) return item;
     const timeZone = ES.TemporalTimeZoneFromString(ES.ToString(item));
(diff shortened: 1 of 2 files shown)


polyfill/lib/timezone.mjs
```diff
@@ -222,6 +222,9 @@ export class TimeZone {
     if (!ES.IsTemporalTimeZone(this)) throw new TypeError('invalid receiver');
     return this.name;
   }
+  toJSON() {
+    return this.toString();
```
ljharb, line 226: why would this observably look up `toString` as opposed to just using the same algorithm?

ljharb: This is very much against the design of JS imo, and I don't think it's a good change.

ptomato: TimeZones can be subclassed, or provided as plain objects with function properties, and need only implement `getPossibleAbsolutesFor()`, `getDateTimeFor()`, and `toString()`. The other methods are defined in terms of these three.

ljharb: Why is that a desirable property? I think a subclass should have to override `toJSON`, just like Arrays and Objects.

ptomato: The subclassable API was designed such that you'd only have to implement the essential parts of a time zone (one-to-many datetime-to-absolute, one-to-one absolute-to-datetime, ID string, and optionally offset transitions) and could chain up or call `Temporal.TimeZone.prototype` methods for everything else. I think that's desirable.

But, since we are now in the stage of looking for feedback on the API, please open a new issue and tag it with "feedback"! :smile:

Issue #698, opened 2020-06-24
DST bug in "opening/closing times" cookbook sample


justingrant:
As part of prototyping changes to cookbook samples for a ZonedAbsolute proof-of-concept type I found a DST bug in one of the cookbook samples.

https://github.com/tc39/proposal-temporal/blob/4481a79de33e06321da5b2a7316e5ee04b61cf18/docs/cookbook/getBusinessOpenStateText.mjs#L59-L66
Problem: `soonWindow` is an absolute or hybrid duration (for time-only durations there's no difference between absolute vs hybrid). But this sample uses `DateTime.prototype.plus`.  This will give an inaccurate result if closing time is near a DST transition.  

To do this properly, the `DateTime` would have to be converted to an Absolute and only then should the `plus` should be applied on top.

PR #701, opened 2020-06-24
Fix DST bug in cookbook

ptomato:
The result will be wrong if the closing time is near a DST transition
because of adding soonWindow to a Temporal.DateTime instead of a
Temporal.Absolute.

Closes: #698


ptomato:
@justingrant For some reason I can't request a review from you in the GitHub UI, but would you like to take a look?

justingrant:
Sorry @ptomato I didn't get to this today. I'm sure it's fine but I'll take a look tomorrow just in case.

PR #705, opened 2020-06-25
Rename conversion methods to toFoo

ryzokuken:
Fixes: #574.

/cc @ptomato @justingrant


ptomato:
This looks fine, is it marked draft because you still intend to rename other methods as well?

ryzokuken:
@ptomato precisely.

ryzokuken:
Made all expected changes, rebased and now this is ready for review.

I'd like to note that it introduces a new ambiguity (now three methods, not two) between `Absolute.prototype.toDateTime`, `Date.prototype.toDateTime` and `Time.prototype.toDateTime`. However, I realized that it's much easier to keep track of things if the method is named after the return type, especially for long chained calls (which is something the current API tries to cater to). This was very evident after editing the cookbook examples. Things like `Temporal.now.absolute().toDateTime(...).getDate().toDateTime(...).toAbsolute(...)` just make so much more sense.

justingrant:
Changes look good, but I was under the impression that we were also going to rename all `getXxx` methods to `toXxx`, e.g. `DateTime.prototype.getDate`->`toDate`. Are those changes coming later, or did we decide not to rename those "downgrade" methods?

(copying from https://github.com/tc39/proposal-temporal/issues/574#issuecomment-646211421) 

> Meeting, June 18: We'll rename everything to `toTheTypeThatItConvertsTo`. (`inTimeZone` → `toAbsolute` or `toDateTime`; `withYear` → `toDate`; `getMonthDay` → `toMonthDay`), and do this fairly quickly while still in the early stages of gathering feedback.

ryzokuken:
@justingrant oops, I totally forgot about those. Will do them today.

docs/cookbook/birthdayIn2030.mjs
```diff
@@ -1,6 +1,6 @@
 const birthday = Temporal.MonthDay.from('12-15');
 
-const birthdayIn2030 = birthday.withYear(2030);
+const birthdayIn2030 = birthday.toDate(2030);
```
ptomato, line 3: Hmm, here is an example of where I really prefer the old names ... `toDate(year)` as below is OK, but `toDate(2030)` with a number literal is more mystifying to me. I know I was in the minority on that viewpoint so I definitely don't want to push for bringing the old names back; but maybe there is a third way? How about calling it toDateIn()? e.g. `toDateIn(2030)`, `toDateIn({ era: 'reiwa',  year: 12 })`

docs/cookbook/getFirstTuesdayOfMonth.mjs
```diff
@@ -6,7 +6,7 @@
  */
 function getFirstTuesday(queriedMonth) {
   // We first need to convert to a date
-  const firstOfMonth = queriedMonth.withDay(1);
+  const firstOfMonth = queriedMonth.toDate(1);
```
ptomato, line 9: Same here, maybe `toDateOn(1)`? (although that reads to me as less obvious than `toDateIn(2030)`)

docs/cookbook/meetingPlanner.js
```diff
@@ -9,11 +9,11 @@ const timeZones = [
 ];
 
 // Start the table at midnight local time
-const calendarNow = now.inTimeZone(here);
+const calendarNow = now.toDateTime(here);
 const startTime = calendarNow
   .getDate()
-  .withTime(Temporal.Time.from('00:00')) // midnight
-  .inTimeZone(here);
+  .toDateTime(Temporal.Time.from('00:00')) // midnight
```
ptomato, line 15: After reading #700 I thought this might be clearer as `calendarNow.with(Temporal.Time.from('00:00'))` anyway?

docs/cookbook/meetingPlanner.js
```diff
@@ -9,11 +9,11 @@ const timeZones = [
...
+  .toDateTime(here);
```
ptomato, line 16: I think this should be `toAbsolute()`?

docs/cookbook/meetingPlanner.js
ryzokuken: @ptomato didn't you update it? I'll take care of it while rebasing.

docs/cookbook/meetingPlanner.js
ryzokuken: Oops. Will fix.

docs/cookbook/meetingPlanner.js
ptomato: Yes, it should be taken care of if you rebase.

ryzokuken:
@ptomato that rebase was a little sad, with 29 commits and all, but it's finally done! Please take another look?

ptomato:
This looks good, but I would really like to consider `MonthDay.toDateIn(2020)` and some similar rename for `YearMonth.toDate()`.

justingrant:
One pattern we could consider would be `toXxxWith` for conversions that require another argument, e.g. `toDateWith({year: 2020})` or `toDateTimeWith({timeZone: 'Europe/Paris'})`.

ryzokuken:
@ptomato I have no strong preferences here, I think we're much better now in that:

1. `X.p.withFoo`/`X.p.with` => `X + Foo -> X`.
2. `X.p.toBar` => `X (+ Y) -> Bar`.

Feel free to make any suggestion for the methods, I have no strong preference here.

@justingrant's suggestion of separating `toBar` like:

1. `X.toBar` => `X -> Bar`
2. `X.toBarWith` => `X + Y -> Bar`

However, as he points it, it tends to get too verbose. Besides, I'm atleast slightly uncomfortable reusing `with` here.

justingrant:
@ptomato - Given the scope of this PR and the challenge of rebasing it each time, would it make sense to merge as-is and then do a smaller follow-up PR to address naming for the two methods that you're concerned about?

ryzokuken:
@justingrant @ptomato squashed and rebased :innocent:

justingrant:
@ptomato - your concerns about the YearMonth/MonthDay methods prompted me to propose #720 which doesn't fix the problems you found but might provide an alternative way to accomplish the same thing with clearer code.

PR #721, opened 2020-07-02
Clean up and update spec

ryzokuken:
Start the process of updating the spec. Draft and rebased on top of #705.

/cc @ptomato

### DIFF
--- a/spec/temporal.html
+++ b/spec/temporal.html
@@ -4,7 +4,7 @@
   <ul>
     <li>is the intrinsic object <dfn>%Temporal%</dfn>.</li>
     <li>is the initial value of the *"Temporal"* property of the global object.</li>
-    <li>is a single ordinary object.</li>
+    <li>is an ordinary object.</li>
     <li>has a [[Prototype]] internal slot whose value is %Object.prototype%.</li>
     <li>is not a function object.</li>
     <li>does not have a [[Construct]] internal method; it cannot be used as a constructor with the `new` operator.</li>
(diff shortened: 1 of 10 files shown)


ptomato:
Can you rebase this now that #705 is merged? Does it still need to be draft?

spec/temporal.html
```diff
@@ -1,51 +1,77 @@
...
-  <p>
-    The Temporal object is not a function object.
-    It does not have a [[Construct]] internal method; it is not possible to use the Temporal object as a constructor with the *new* operator.
-    The Temporal object does not have a [[Call]] internal method; it is not possible to invoke the Temporal object as a function.
-  </p>
+  <p>The Temporal object:</p>
+  <ul>
+    <li>is the intrinsic object <dfn>%Temporal%</dfn>.</li>
+    <li>is the initial value of the *"Temporal"* property of the global object.</li>
+    <li>is a single ordinary object.</li>
```
ptomato, line 7: Pre-existing, so feel free to punt to a follow-up if you like: what does "single" mean here? I searched ECMA-262 for the phrase "single ordinary object" and didn't find it anywhere else.

Also, I noticed that ECMA-262 has nifty links to the definition of 'ordinary object', I managed to get the same by adding this to the biblio file:
```json
    {
      "type": "term",
      "term": "ordinary object",
      "refId": "ordinary-object",
      "id": "ordinary-object"
    }
```
But again this is pre-existing so feel free to ignore it for now.

ryzokuken: Yeah, I didn't find "single" anywhere either, but I decided not to change it. I'll make these improvements in a PR now.

PR #723, opened 2020-07-02
Rename MonthDay.toDate() and YearMonth.toDate()

ptomato:
Follow up to #574, I think these two methods in particular are special because they can be used with integer literals.

Closes: #722


ptomato:
Yes, I'd wait for more comments.

justingrant:
LGTM. I like how the method names include both the destination type and a hint about the parameter needed.  Kills 2 birds with one stone.

sffc:
I don't really like the vague hybrid names.  I would rather go all-in with the original names `.withYear()`, or keep a clean and consistent `.toDate()`.  My third choice would be something like `.toDateInYear()`.

justingrant:
> My third choice would be something like `.toDateInYear()`.

I like the specificity of this option.  It clears up any ambiguity about what the parameter will be. `.toDateOnDay()` would presumably be the other one?

ptomato:
I personally prefer `withDay()` but I think we already established that that would be inconsistent unless we reverted the whole rename. But `toDate(1)` is just too cryptic to me. `toDateOnDay()`/`toDateInYear()` seem fine to me. I will change this PR to implement those and leave it up to see if there are any more comments.

ptomato:
This one has been up as `toDateOnDay`/`toDateInYear` for a few days, so I'll merge it. There is still a larger discussion going on at #747.

PR #736, opened 2020-07-07
Fix `Time.from` when param is another `Time`

### LINKED ISSUE #735: `Temporal.Time.from` fails when another `Time` object is supplied
The "same object" form of Temporal.Time.from doesn't work. Specifically, the hours are missing in the result.

There were no test cases that checked whether all units were the same between source and result.

Repro: 
```js
Temporal.Time.from(Temporal.Time.from('12:00'))
```

Expected: 12:00
Actual: 00:00

Cause is missing "HOUR" in the code below. PR coming shortly. Also I'll fix that inaccurate comment about "alphabetical".

https://github.com/tc39/proposal-temporal/blob/15601b16af4c907b790a18a30bec6d465f403d1f/polyfill/lib/time.mjs#L256-L266

justingrant:
- Fixes #735.
- Adds tests for `Time.from` when the source is `Time` or `DateTime`.
- Also corrects an inaccurate comment.

### DIFF
--- a/polyfill/lib/time.mjs
+++ b/polyfill/lib/time.mjs
@@ -255,13 +255,14 @@ export class Time {
     let hour, minute, second, millisecond, microsecond, nanosecond;
     if (typeof item === 'object' && item) {
       if (ES.IsTemporalTime(item)) {
+        hour = GetSlot(item, HOUR);
         minute = GetSlot(item, MINUTE);
         second = GetSlot(item, SECOND);
         millisecond = GetSlot(item, MILLISECOND);
         microsecond = GetSlot(item, MICROSECOND);
         nanosecond = GetSlot(item, NANOSECOND);
       } else {
-        // Intentionally alphabetical
+        // Intentionally largest to smallest units
         ({ hour, minute, second, millisecond, microsecond, nanosecond } = ES.ToTemporalTimeRecord(item));
       }
     } else {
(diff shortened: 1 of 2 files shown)


polyfill/lib/time.mjs
```diff
@@ -255,13 +255,14 @@ export class Time {
...
       if (ES.IsTemporalTime(item)) {
+        hour = GetSlot(item, HOUR);
         minute = GetSlot(item, MINUTE);
         second = GetSlot(item, SECOND);
         millisecond = GetSlot(item, MILLISECOND);
         microsecond = GetSlot(item, MICROSECOND);
         nanosecond = GetSlot(item, NANOSECOND);
       } else {
-        // Intentionally alphabetical
+        // Intentionally largest to smallest units
```
ptomato, line 265: For future reference, we can just delete this comment, but no need to block the PR on it. (Calling the property getters on a passed-in object in alphabetical order is mandated by the spec, which is now implemented in `ES.ToTemporalTimeRecord`. Here, they can be in any old order.

justingrant: > Calling the property getters on a passed-in object in alphabetical order is mandated by the spec

OK, got it. Out of curiosity, why is alpha order required?

ptomato: Alphabetical doesn't really matter, but some order is required to be specified, since the calls are observable by client code if you pass in a Proxy, for example. I'm not sure why alphabetical was chosen originally as the arbitrary order, but I'd guess it's because there's some precedent elsewhere.

Issue #733, opened 2020-07-07
toLocaleString issue


thojanssens:
I run the following code to get the formatted localized day of week from a date time:
```
Temporal.now.dateTime().toLocaleString('en-US', { weekday: 'long' })
```
But I get more than the day of week:

> "Tuesday, 7/7/2020, 11:33:30 AM"

I expect `toLocaleString` to work like it would for `Date`:

```
new Date().toLocaleString('en-US', { weekday: 'long' })
```
> "Tuesday"

ptomato:
You found a bug, thanks!

PR #738, opened 2020-07-07
Fix localized weekdays

ptomato:
If given, 'weekday' should be included for types where the day of the
week is defined, and excluded from types that don't have that concept.

Closes: #733


ptomato:
I believe that section of the spec is still under construction, so I wasn't too worried about it. But it does look like it is correct, `weekday` is listed as allowed for Date, DateTime, and Absolute, and not for YearMonth, MonthDay, and Time.

PR #739, opened 2020-07-07
Add minutes from the last three weeks

ptomato:
Tagging everyone for review who was present at one of the meetings, also @younies @leobalter


ryzokuken:
This has been around for enough time, I suppose.

PR #740, opened 2020-07-08
Various TimeZone-related changes.

Ms2ger:


polyfill/ci_test.sh
```diff
@@ -36,7 +38,7 @@ test262-harness \
   --test262Dir ../test262 \
   --prelude "../$PRELUDE" \
   --transformer ./transform.test262.js \
-  "./*/**/*.js" \
+  $TESTS \
```
ptomato, line 41: :pray:

polyfill/lib/ecmascript.mjs
```diff
@@ -708,15 +708,15 @@ export const ES = ObjectAssign({}, ES2019, {
...
-    let nanosecond = Math.floor(nanos / 1) % 1e3;
+    const { quotient, remainder } = bigInt(epochNanoseconds).divmod(1e6);
+    let epochMilliseconds = +quotient;
+    let nanos = +remainder;
+    if (nanos < 0) {
+      nanos += 1e6;
+      epochMilliseconds -= 1;
+    }
+    const microsecond = Math.floor(nanos / 1e3) % 1e3;
+    const nanosecond = Math.floor(nanos / 1) % 1e3;
```
ptomato, line 719: optional nitpick, maybe change to `const nanosecond = nanos % 1e3` while you're at it?

spec/timezone.html
```diff
@@ -327,6 +327,14 @@ <h1>FormatTimeZoneOffsetString ( _offsetNanoseconds_ )</h1>
       </emu-alg>
     </emu-clause>
 
+    <emu-clause id="sec-temporal-timezonefrom" aoid="TimeZoneFrom">
+      <h1>TimeZoneFrom ( _identifier_ )</h1>
+      <emu-alg>
+        1. Let _from_ be ? Get(%Temporal.TimeZone%, *"from"*).
```
ptomato, line 323: Can this be %Temporal.TimeZone%.from or does it have to be %Temporal%.TimeZone.from for full monkeypatching purposes?

ryzokuken: Wait, we allow monkey-patching `Temporal.TimeZone`?

docs/timezone-draft.md
```diff
@@ -23,9 +23,9 @@ For example, St. Louis's local mean time is GMT-06:00:49.11 ([Byrd, Mary E., 189
 
 ## Usage of a custom time zone
 
-Like built-in time zones, custom time zones have an `[[Identifier]]`.
+Like built-in time zones, custom time zones have an identifier.
 In these examples we assume a custom time zone class `StLouisTime` with the identifier `America/St_Louis`.
-(See the following section for how such a time zone object would be obtained.)
+(See the following section for how such a time zone object would be implemented.)
```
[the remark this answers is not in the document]
ptomato: That's true, it could be removed or moved to an 'old-proposals/' folder or something.

spec/timezone.html
ptomato: I didn't mean to suggest that we allow it or disallow it, but you certainly _can_ replace the `Temporal.TimeZone` constructor; it's a question of whether we want this to get the `from` property from the original or the replaced one.

ryzokuken: I think allowing monkey-patching for top level classes is a can of worms we don't really want to open.

ljharb: This definitely has to be `%Temporal.TimeZone.from%` and not do an observable lookup.

ptomato: No, monkeypatching `from` is how SES locks down the time zone database, so it does need to do an observable lookup in any case.

ljharb: oof, that's something we'll want to discuss in committee. I thought the purpose of the `now` namespace was to make that be the only place SES had to monkeypatch, and that there were no places with an observable lookup.

ptomato: OK. Previous discussion is [here](https://github.com/tc39/proposal-temporal/blob/main/meetings/agenda-minutes-2020-05-07.md#294-idtocalendar--idtotimezone-follow-up-after-discussing-with-ses).

PR #744, opened 2020-07-09
Fix year zero leap day

ptomato:
Found by attempting to run the date-fns test suite with Temporal


ptomato:
Yes, that was exactly the problem.

PR #748, opened 2020-07-10
Add disambiguation to MonthDay.toDateInYear

ptomato:
Closes: #746


ptomato:
Note, this does change the behaviour; previously `reject` was always implied, now `constrain` is the default. I don't know if `reject` was there for a reason. I think there is a reasonable case to be made for `reject` being the default, but I think consistency with the other disambiguation parameters weighs stronger.

ptomato:
(forgot to update the types file)

PR #749, opened 2020-07-10
Correct minor problem with script examples which results in an error when run

cwadrupldijjit:
modified "const" to "let" since the code sample would fail in the reassignment on the next line

### DIFF
--- a/docs/duration.md
+++ b/docs/duration.md
@@ -309,7 +309,7 @@ oneAndAHalfYear = Temporal.Duration.from({ years: 1, months: 6 });
 threeYears.minus(oneAndAHalfYear)  // throws; months are negative and cannot be balanced
 // Example of a custom conversion using ISO calendar rules:
 function yearsToMonths(duration) {
-    const { years, months } = duration;
+    let { years, months } = duration;
     months += years * 12;
     return duration.with({ years: 0, months });
 }
--- a/docs/duration.md
+++ b/docs/duration.md
@@ -256,7 +256,7 @@ result = oneAndAHalfYear.plus(oneAndAHalfYear)  // => P2Y12M
 result.with(result, { disambiguation: 'balance' })  // => P2Y12M
 // Example of custom conversion using ISO calendar rules:
 function monthsToYears(duration) {
-    const { years, months } = duration;
+    let { years, months } = duration;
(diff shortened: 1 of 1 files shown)


cwadrupldijjit:
Wait, how in the world does changing a .md file (technically decreasing the size, too) actually decrease the code coverage?  That doesn't make any sense!  There _has_ to be more to that...

cwadrupldijjit:
Okay, never mind.  It just updated to saying the code coverage won't change...

ryzokuken:
> Wait, how in the world does changing a .md file (technically decreasing the size, too) actually decrease the code coverage? That doesn't make any sense! There has to be more to that...

Sorry about that, `codecov` takes a while to calculate and updates asynchronously.

justingrant:
Does [coveralls](https://coveralls.io/) work better? I admit that `codecov`'s spam was confusing to me too.

ptomato:
It should work fine again when we have fully ported the demitasse tests over to test262. The problem is that codecov gets confused about the two test suites when one finishes earlier than the other. I haven't investigated if coveralls does handle this.

ryzokuken:
I think it's a better idea to put more effort into porting the tests than spend time trying to find an alternative to codecov, to be honest.

justingrant:
Sounds reasonable. re: porting the tests, once very useful thing about the demitasse tests is that they all run in <5s, but the Test262 tests take 5+ minutes to run. This makes iterative testing really challenging. Is there a way to speed them up?

ryzokuken:
The test262 harness isn't the fastest or the most feature-rich, but any additions/improvements would directly help all other tc39 efforts, so it seems worth looking into.

justingrant:
Agreed.  My main suggestion would be to wait a few months to port our existing tests until the rate of change slows down, because having tests run 50x-100x slower would make it much more time-consuming to make changes to Temporal.

ptomato:
I agree, that was my intention as well. (FWIW, when I was investigating this, I found the test262 harness isn't inherently slow. The slowdown comes from eval'ing the entirety of the bundled Temporal polyfill that gets prepended to every test because there isn't a better way to install Temporal into the test262 environment. That's why, when I reduced the size of the bundle, the test262 ran faster by several minutes.)

PR #756, opened 2020-07-10
Fix docs navigation

flouc001:
For a bit of background on this PR please see this issue: #754 

Essentially a call to `slugger.slug` in the `buildDocs.js` file was causing marked's slugger to increment an internal accumulator and subsequently suffix our headings with a `-1` even though there was only one occurrence of them. This meant items in our TOC were referencing for example `frequently-asked-questions`, but marked was outputting the header with an id set to `frequently-asked-questions-1`.

This PR reverses the accumulator action erroneously caused.


ptomato:
Thanks for digging into the problem!

Issue #763, opened 2020-07-12
ES.ToBigInt is incorrect


ryzokuken:
`ES.ToBigInt` in the polyfill is incorrect.

It does not adhere to the standard (https://tc39.es/ecma262/#sec-tobigint) for a reason I don't recall, so I am assuming that it was a mistake but I might be incorrect and it may be intentional.

It calls the `bigInt` method exported by the `"big-integer"` npm module which is *also* not compliant to the `BigInt` object as spec'd in 262, so that's a bit weird too, but let's make another issue to reconsider the choice of polyfill there.

For now, we should fix up `ES.ToBigInt` to fix the bug in the `Absolute` constructor referenced in #761.

PR #767, opened 2020-07-13
polyfill: fix ES.ToBigInt

ryzokuken:
Fixes: https://github.com/tc39/proposal-temporal/issues/763

### DIFF
--- a/polyfill/lib/ecmascript.mjs
+++ b/polyfill/lib/ecmascript.mjs
@@ -1423,14 +1423,35 @@ export const ES = ObjectAssign({}, ES2019, {
     return result;
   },
   ToBigInt: (arg) => {
-    if (bigInt.isInstance(arg)) return arg;
+    if (bigInt.isInstance(arg)) {
+      return arg;
+    }
+
     const prim = ES.ToPrimitive(arg, Number);
-    if (typeof prim === 'number') throw new TypeError('Use BigInt() to convert Number to BigInt');
-    try {
-      return bigInt(prim);
-    } catch (e) {
-      if (e instanceof Error && e.message.startsWith('Invalid integer')) throw new SyntaxError(e.message);
-      throw e;
+    switch (typeof prim) {
+      case 'undefined':
+      case 'object':
+      case 'number':
+      case 'symbol':
+        throw new TypeError(`cannot convert ${typeof arg} to bigint`);
+      case 'string':
+        if (!prim.match(/^\s*(?:[+-]?\d+\s*)?$/)) {
+          throw new SyntaxError('invalid BigInt syntax');
+        }
+      // eslint: no-fallthrough: false
+      case 'bigint':
+        try {
+          return bigInt(prim);
+        } catch (e) {
+          if (e instanceof Error && e.message.startsWith('Invalid integer')) throw new SyntaxError(e.message);
+          throw e;
+        }
+      case 'boolean':
+        if (prim) {
+          return bigInt(1);
+        } else {
+          return bigInt(0);
+        }
     }
   },
 
--- a/polyfill/test/Absolute/constructor/argument.js
+++ b/polyfill/test/Absolute/constructor/argument.js
@@ -1,0 +1,12 @@
+// Copyright (C) 2020 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.absolute
+features: [Symbol]
+---*/
+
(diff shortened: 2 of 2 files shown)


polyfill/lib/ecmascript.mjs
```diff
@@ -1423,14 +1423,26 @@ export const ES = ObjectAssign({}, ES2019, {
...
-      throw e;
+    switch (typeof prim) {
+      case 'undefined':
+      case 'object':
+      case 'number':
+      case 'symbol':
+        throw new TypeError(`cannot convert ${typeof arg} to bigint`);
+      case 'bigint':
+      case 'string':
+        return bigInt(prim);
```
ljharb, line 1439: what's the expected result for the empty string?

ryzokuken: @ljharb I assumed here that `bigInt`'s string parser is compliant to the standard (so equivalent of https://tc39.es/ecma262/#sec-stringtobigint) but on second thought, that assumption might be incorrect, and now I'm questioning why we are using this polyfill in the first place? It has a lot of compliance issues...

ryzokuken: (as opposed to something like https://github.com/GoogleChromeLabs/jsbi)

ptomato: `bigInt(string)` seems to work almost like the spec, except that it accepts strings like `1e3` which `BigInt('1e3')` does not.

After a chat discussion between me and @ryzokuken, I think rather than changing bigint libraries at this point, we should wait for Safari 14 to become mainstream and then use no library at all, possibly adding a hack in the meantime to disallow string input like `'1e3'`.

ptomato:
Seems there is a test failure, by the way: https://travis-ci.com/github/tc39/proposal-temporal/jobs/360548634#L5579

ryzokuken:
It's because of the string issue. The string parser is completely non-compliant (it throws an `Error` on invalid strings instead of `SyntaxError`). Makes me wonder if we should actually bite the bullet and switch to JSBI.

That said, I'm glad we have such good coverage.

ptomato:
I see. In that case maybe preserve the try block from the previous code, after all?

ryzokuken:
That plus a condition to throw on `e` notation strings. And that's just what we know so far...

ryzokuken:
@ptomato @ljharb PTAL. I added two hacks to make it compatible for the time-being, but it still smells weird to me.

polyfill/lib/ecmascript.mjs
```diff
@@ -1423,14 +1423,35 @@ export const ES = ObjectAssign({}, ES2019, {
...
-      if (e instanceof Error && e.message.startsWith('Invalid integer')) throw new SyntaxError(e.message);
-      throw e;
+    switch (typeof prim) {
+      case 'undefined':
+      case 'object':
+      case 'number':
+      case 'symbol':
+        throw new TypeError(`cannot convert ${typeof arg} to bigint`);
+      case 'string':
+        if (!prim.match(/\d*/)) {
```
ptomato, line 1438: This doesn't rule out `1e3`.

ryzokuken: Ah wait, my bad I missed something.

polyfill/lib/ecmascript.mjs
```diff
@@ -1423,14 +1423,35 @@ export const ES = ObjectAssign({}, ES2019, {
...
-      if (e instanceof Error && e.message.startsWith('Invalid integer')) throw new SyntaxError(e.message);
-      throw e;
+    switch (typeof prim) {
+      case 'undefined':
+      case 'object':
+      case 'number':
+      case 'symbol':
+        throw new TypeError(`cannot convert ${typeof arg} to bigint`);
+      case 'string':
+        if (!prim.match(/^\d*$/)) {
```
ptomato, line 1438: Now this rules out ` ` and `-1`. I would suggest putting the entirety of the allowed syntax from ToBigInt in the regexp:
```suggestion
        if (!prim.match(/^\s*(?:[+-]\d+\s*)?$/)) {
```

ryzokuken: Sorry, I just noticed https://tc39.es/ecma262/#sec-tonumber-applied-to-the-string-type. Thanks!

ptomato: My bad, it should be `/^s*(?:[+-]?\d+\s*)?$/`

ryzokuken: I think you mean `/^\s*(?:[+-]?\d+\s*)?$/` :D

PR #768, opened 2020-07-13
More spec improvements

ryzokuken:


spec/date.html
```diff
@@ -61,7 +63,7 @@ <h1>Temporal.Date.from ( _item_ [ , _options_ ] )</h1>
       </p>
       <emu-alg>
         1. Let _disambiguation_ be ? ToTemporalDisambiguation(_options_).
-        1. If Type(_item_) is Object, then
+        1. If Type(_item_) is Object and _item_ is not *null*, then
```
Ms2ger, line 66: Type(null) is Null, unlike `typeof`

spec/date.html
```diff
@@ -468,7 +470,7 @@ <h1>Properties of Temporal.Date Instances</h1>
   </emu-clause>
 
   <emu-clause id="sec-temporal-date-abstract-ops">
-    <h1>Abstract operations</h1>
+    <h1>Abstract Operations for Temporal.Date Objects</h1>
```
Ms2ger, line 473: None of the other abstract ops sections make this explicit; do we want to be consistent?

spec/date.html
ryzokuken: Oops, will undo this.

spec/date.html
ryzokuken: That's because I didn't push changes to those! I'll update these everywhere to ultimately make the spec consistent with ECMA 262.

spec/date.html
ryzokuken: Done!

PR #774, opened 2020-07-14
Clarify Temporal.Absolute largest unit

ptomato:
Closes: #771

### DIFF
--- a/docs/absolute.md
+++ b/docs/absolute.md
@@ -308,9 +308,12 @@ A difference of two hours will become 7200 seconds when `largestUnit` is `"secon
 However, a difference of 30 seconds will still be 30 seconds even if `largestUnit` is `"hours"`.
 
 By default, the largest unit in the result is seconds.
-Unlike other Temporal types, months and years are not allowed.
-This is because in the ISO calendar, months and years can be different lengths depending on which month is meant and whether the year is a leap year.
-`Temporal.Absolute` is intended to be calendar-independent and therefore free of these ambiguities.
+Weeks, months and years are not allowed, unlike the difference methods of the other Temporal types.
+This is because months and years can be different lengths depending on which month is meant, and whether the year is a leap year, which all depends on the start and end date of the difference.
+You cannot determine the start and end date of a difference between `Temporal.Absolute`s, because `Temporal.Absolute` has no time zone or calendar.
+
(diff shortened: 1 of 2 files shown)


docs/absolute.md
```diff
@@ -308,10 +308,13 @@ A difference of two hours will become 7200 seconds when `largestUnit` is `"secon
 However, a difference of 30 seconds will still be 30 seconds even if `largestUnit` is `"hours"`.
 
 By default, the largest unit in the result is seconds.
-Unlike other Temporal types, months and years are not allowed.
-This is because in the ISO calendar, months and years can be different lengths depending on which month is meant and whether the year is a leap year.
+Weeks, months and years are not allowed, unlike the difference methods of the other Temporal types.
+This is because months and years can be different lengths depending on the calendar in use, which month is meant, and whether the year is a leap year.
```
sffc, line 312: Most importantly, the length of a month depends on the start date, and you don't know the start date of an Absolute unless you know the time zone.  The calendar system is secondary.

docs/absolute.md
```diff
@@ -308,10 +308,13 @@ A difference of two hours will become 7200 seconds when `largestUnit` is `"secon
...
 `Temporal.Absolute` is intended to be calendar-independent and therefore free of these ambiguities.
 
+If you do need to calculate the difference between two `Temporal.Absolute`s in years, months, or weeks, then you can eliminate this ambiguity by choosing your starting point by converting to a `Temporal.DateTime`.
+For example, you might decide to base the calculation on your user's current time zone, or on UTC, or on your user's preferred calendar.
```
sffc, line 316: You need both.  You can't set a calendar system without setting a time zone and have this work.

ptomato: Yeah, this was badly worded — I intended (user time zone OR utc) AND (user non-iso calendar OR iso) but condensed it wrong :stuck_out_tongue:

ptomato:
I updated the branch, but it doesn't seem to be reflected in the GitHub UI.

justingrant:
FYI: #759 (see bullet point 2 in that issue) may cause us to add "days" to the illegal units list for Absolute math. I don't think this possibility should block this PR, but just wanted to give a heads up that we might need to modify some of the same docs and code pretty soon.

ptomato:
Agreed.

PR #776, opened 2020-07-15
[polyfill] Refactor using FormatSecondsStringPart.

Ms2ger:
This matches the specification, and leads to more readable code overall.

### DIFF
--- a/polyfill/lib/ecmascript.mjs
+++ b/polyfill/lib/ecmascript.mjs
@@ -591,7 +591,7 @@ export const ES = ObjectAssign({}, ES2019, {
     return yearString;
   },
   ISODateTimePartString: (part) => `00${part}`.slice(-2),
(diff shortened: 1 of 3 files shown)


polyfill/lib/ecmascript.mjs
```diff
@@ -600,7 +600,7 @@ export const ES = ObjectAssign({}, ES2019, {
     if (millis || parts.length) parts.unshift(`000${millis || 0}`.slice(-3));
     let secs = `00${seconds}`.slice(-2);
     let post = parts.length ? `.${parts.join('')}` : '';
-    return `${secs}${post}`;
+    return `:${secs}${post}`;
```
ryzokuken, line 603: I'm not quite sure about this one. On one hand, sure, it's common to literally every place this operation is called. On another, shouldn't the string interpolator in the calling function take care of the `:`?

PR #779, opened 2020-07-16
Document all default values in JSDoc. Fixes #765

Ryuno-Ki:
I found one more instance while working on #765.

Also, on a clean clone, the `npm run build:docs` command errors because of missing target directory structure with

```sh
[Error: ENOENT: no such file or directory, open 'PROJECT_ROOT/out/docs/README.html'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '$PROJECT_ROOT/out/docs/README.html'
}
```

(multiple places, I replaced the actual value with `$PROJECT_ROOT` here).

### DIFF
--- a/docs/buildDocs.js
+++ b/docs/buildDocs.js
@@ -1,5 +1,6 @@
 const fs = require('fs').promises;
 const marked = require('marked');
+const mkdirp = require('mkdirp');
 const path = require('path');
 const Prism = require('prismjs');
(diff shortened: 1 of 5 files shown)


Ryuno-Ki:
Would be shocking, if editing Doc strings could wreak havoc test coverage :scream_cat:

PR #795, opened 2020-07-28
spec: add initial spec text for calendar

ryzokuken:
Added initial spec text for calendar. Wdyt? Should I add the stubs to the spec too?

### DIFF
--- a/spec/calendar.html
+++ b/spec/calendar.html
@@ -1,0 +1,153 @@
+<emu-clause id="sec-temporal-calendar-objects">
+  <h1>Temporal.Calendar Objects</h1>
+  <p>A Temporal.Calendar object is an immutable Object representing a calendar.</p>
+
+  <emu-clause id="sec-temporal-calendar-abstract-ops">
+    <h1>Abstract Operations for Temporal.Calendar Objects</h1>
+
+    <emu-clause id="sec-temporal-getbuiltincalendar" aoid="GetBuiltinCalendar">
+      <h1>GetBuiltinCalendar ( _id_ )</h1>
+      <p>An ECMAScript implementation that includes the ECMA-402 Internationalization API must implement the *GetBuiltinCalendar* abstract operation as specified in the ECMA-402 specification. If an ECMAScript implementation does not include the ECMA-402 API the following specification of the *GetBuiltinCalendar* abstract operation is used.</p>
+      <emu-alg>
+        1. If _id_ is not *"iso8601"*, throw a *RangeError* exception.
+        1. Return ? Construct(%ISO8601Calendar%, « »).
+      </emu-alg>
+    </emu-clause>
+
+    <emu-clause id="sec-temporal-getdefaultcalendar" aoid="GetDefaultCalendar">
+      <h1>GetDefaultCalendar ( )</h1>
+      <emu-alg>
+        1. Return ? GetBuiltinCalendar(*"iso8601"*).
+      </emu-alg>
+    </emu-clause>
+
+    <emu-clause id="sec-temporal-calendartostring" aoid="CalendarToString">
+      <h1>CalendarToString ( _calendar_ )</h1>
+      <emu-alg>
+        1. Let _toString_ be ? Get(_calendar_, *"toString"*).
+        1. If _toString_ is *undefined*, then
+          1. Set _toString_ to %Temporal.Calendar.prototype.toString%.
+        1. Return ? ToString(? Call(_toString_ , _calendar, « »)).
+      </emu-alg>
+    </emu-clause>
+  </emu-clause>
+
+  <emu-clause id="sec-temporal-calendar-constructor">
+    <h1>The Temporal.Calendar Constructor</h1>
+    <p>The Temporal.Calendar constructor:</p>
+    <ul>
+      <li>is the intrinsic object <dfn>%Temporal.Calendar%</dfn>.</li>
+      <li>creates and initializes a new Temporal.Calendar object when called as a constructor.</li>
+      <li>is designed to be subclassable. It may be used as the value of an `extends` clause of a class definition. Subclass constructors that intend to inherit the specified Temporal.Calendar behaviour must include a super call to the %Temporal.Calendar% constructor to create and initialize subclass instances with the necessary internal slots.</li>
+      <li>has a *"length"* property whose value is 1.</li>
+    </ul>
+
+    <emu-clause id="sec-temporal.calendar">
+      <h1>Temporal.Calendar ( _id_ )</h1>
+      <p>When the `Temporal.Calendar` function is called, the following steps are taken:</p>
+      <emu-alg>
+        1. If NewTarget is *undefined*, then
+          1. Throw a *TypeError* exception.
+        1. Let _calendar_ be ? OrdinaryCreateFromConstructor(NewTarget, *"%Temporal.Calendar.prototype%"*, « [[InitializedTemporalCalendar]], [[Identifier]] »).
+        1. Set _calendar_.[[Identifier]] to _id_.
+        1. Return _calendar_.
+      </emu-alg>
+    </emu-clause>
+  </emu-clause>
+
+  <emu-clause id="sec-properties-of-the-temporal-calendar-constructor">
+    <h1>Properties of the Temporal.Calendar Constructor</h1>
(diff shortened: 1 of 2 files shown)


spec/calendar.html
```diff
@@ -0,0 +1,153 @@
...
+          1. Return _item_.
+        1. Let _stringIdent_ be ? ToString(_item_).
+        1. Return ? GetBuiltinCalendar(_stringIdent_).
+      </emu-alg>
+    </emu-clause>
+
+    <emu-clause id="sec-temporal-calendar-internal-slots">
+      <h1>Internal Slots</h1>
+
+      <p>The value of the [[BuiltinCalendars]] internal slot must be a record with keys corresponding to each supported calendar system and their corresponding values should be constructor objects that construct a Temporal.Calendar object corresponding to the calendar system.</p>
```
Ms2ger, line 92: Do you really want to use "should" in this sentence?

spec/calendar.html
```diff
@@ -0,0 +1,153 @@
...
+      <h1>Temporal.Calendar.prototype.toString ( )</h1>
+      <p>
+        The following steps are taken:
+      </p>
+      <emu-alg>
+        1. Let _calendar_ be the *this* value.
+        1. Perform ? RequireInternalSlot(_calendar_, [[InitializedTemporalDate]]).
+        1. Return _calendar_.[[ID]].
+      </emu-alg>
+    </emu-clause>
```
Ms2ger, line 145: Do we want to define one in terms of the other, so custom calendars can implement just that one? I have a wip change to that effect for TimeZone in <https://github.com/tc39/proposal-temporal/compare/timezone-name>. I don't think it's super useful, but consistency is nice.

spec/calendar.html
```diff
@@ -0,0 +1,153 @@
+<emu-clause id="sec-temporal-calendar-objects">
+  <h1>Temporal.Calendar Objects</h1>
+  <p>A Temporal.Calendar object is an immutable Object referencing a calendar.</p>
```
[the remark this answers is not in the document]
ryzokuken: copypasta

spec/calendar.html
```diff
@@ -0,0 +1,153 @@
...
+      <p>The value of the *"name"* property of this function is *"get [Symbol.species]"*.</p>
+    </emu-clause>
+
+    <emu-clause id="sec-temporal.calendar.from">
+      <h1>Temporal.Calendar.from ( _item_ )</h1>
+      <p>The `from` method takes one argument: item. The following steps are taken:</p>
+      <emu-alg>
+        1. If _item_ has an [[InitializedTemporalCalendar]] internal slot, then
+          1. Return _item_.
+        1. If Type(_item_) is Object and _item_ is not *null*, then
```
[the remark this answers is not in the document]
ryzokuken: My bad.

spec/calendar.html
ryzokuken: Changing to "must".

spec/calendar.html
```diff
@@ -0,0 +1,151 @@
...
+      </emu-alg>
+
+      <p>The value of the *"name"* property of this function is *"get [Symbol.species]"*.</p>
+    </emu-clause>
+
+    <emu-clause id="sec-temporal.calendar.from">
+      <h1>Temporal.Calendar.from ( _item_ )</h1>
+      <p>The `from` method takes one argument: item. The following steps are taken:</p>
+      <emu-alg>
+        1. If _item_ has an [[InitializedTemporalCalendar]] internal slot or Type(_item_) is Object, then
```
ptomato, line 80:
```suggestion
        1. If Type(_item_) is Object, then
```

spec/calendar.html
```diff
@@ -0,0 +1,151 @@
+<emu-clause id="sec-temporal-calendar-objects">
+  <h1>Temporal.Calendar Objects</h1>
+  <p>A Temporal.Calendar object is an immutable Object representing a calendar.</p>
+
+  <emu-clause id="sec-temporal-calendar-abstract-ops">
+    <h1>Abstract Operations for Temporal.Calendar Objects</h1>
+
+    <emu-clause id="sec-temporal-getbuiltincalendar" aoid="GetBuiltinCalendar">
```
ptomato, line 8: I wonder if the builtin calendars should be in ECMA-402 instead. The only calendar that an implementation is required to have, if not supporting Intl, is the ISO calendar. That would mean we don't have to have a `%Temporal.Calendar%.[[BuiltinCalendars]]`, we only need an intrinsic `%ISO8601Calendar%` object. We could have a prose description of what Temporal.Calendar.from does if the implementation doesn't have ECMA-402 similar to https://tc39.es/ecma262/#sec-number.prototype.tolocalestring

spec/calendar.html
```diff
@@ -0,0 +1,151 @@
...
+    </ul>
+
+    <emu-clause id="sec-temporal.calendar">
+      <h1>Temporal.Calendar ( _id_ )</h1>
+      <p>When the `Temporal.Calendar` function is called, the following steps are taken:</p>
+      <emu-alg>
+        1. If NewTarget is *undefined*, then
+          1. Throw a *TypeError* exception.
+        1. Let _calendar_ be ? OrdinaryCreateFromConstructor(NewTarget, *"%Temporal.Calendar.prototype%"*, « [[InitializedTemporalCalendar]], [[ID]] »).
+        1. Set _calendar_.[[ID]] to _id_.
```
ptomato, line 43: For TimeZone we use [[Identifier]], I think it should be consistent.

ryzokuken: Done.

spec/calendar.html
ryzokuken: Done.

spec/calendar.html
ryzokuken: I thought we could set/describe the exact values stored in the `%Temporal.Calendar%.[[BuiltinCalendars]]` in 402. This is because there isn't any precedent for overloading an abstract operation in 262, only for public methods. Perhaps we could add a static `Temporal.Calendar.getBuiltinCalendars` method that can be overloaded in 402 and `getDefaultCalendar` could call that method instead.

ryzokuken: Or perhaps I could make `Temporal.Calendar.from` overload-able in 402 and make `getDefaultCalendar` call `from` instead.

littledan: I don't see any particular problem with overloading an abstract operation in 262, as long as we're clear in the spec text about what we're doing.

ryzokuken: @littledan alright then.

ryzokuken: @ptomato @littledan done, PTAL.

ryzokuken:
@ptomato addressed all your comments, lmk what you think.

Still TODO: address @Ms2ger's suggestion, preferably in a follow-up.

ryzokuken:
@Ms2ger this should look better now.

spec/calendar.html
```diff
@@ -0,0 +1,153 @@
...
+      <emu-alg>
+        1. Return the *this* value.
+      </emu-alg>
+
+      <p>The value of the *"name"* property of this function is *"get [Symbol.species]"*.</p>
+    </emu-clause>
+
+    <emu-clause id="sec-temporal.calendar.from">
+      <h1>Temporal.Calendar.from ( _item_ )</h1>
+      <p>The `from` method takes one argument: item. The following steps are taken:</p>
```
Ms2ger, line 87:
```suggestion
      <p>The `from` method takes one argument _item_. The following steps are taken:</p>
```

spec/calendar.html
```diff
@@ -0,0 +1,153 @@
...
+    </emu-clause>
+
+    <emu-clause id="sec-temporal.calendar.prototype.tostring">
+      <h1>Temporal.Calendar.prototype.toString ( )</h1>
+      <p>
+        The following steps are taken:
+      </p>
+      <emu-alg>
+        1. Let _calendar_ be the *this* value.
+        1. Perform ? RequireInternalSlot(_calendar_, [[InitializedTemporalDate]]).
```
Ms2ger, line 142:
```suggestion
        1. Perform ? RequireInternalSlot(_calendar_, [[InitializedTemporalCalendar]]).
```

ryzokuken: sorry for the copypasta!

Issue #800, opened 2020-07-30
Absolute doesnt have 'with' method?


henryw374:
The cookbook suggests all temporal types do have this: https://tc39.es/proposal-temporal/docs/cookbook.html#round-a-time-down-to-whole-hours "Use the with() method of each Temporal type if you want to round or balance the fields."

ptomato:
Good catch, thanks. Temporal.Absolute doesn't have it because it doesn't have any fields, only a timestamp in nanoseconds. If you want to round an Absolute then you have to convert it to a calendar-based type such as Temporal.DateTime. I will clarify this in the documentation.

PR #801, opened 2020-07-30
Clarify that rounding cookbook example doesn't apply to Absolute

ptomato:
Temporal.Absolute can't be rounded to a certain field because it doesn't
have any fields.

Closes: #800

### DIFF
--- a/docs/cookbook.md
+++ b/docs/cookbook.md
@@ -123,12 +123,15 @@ Sort a list of ISO 8601 date/time strings, for example to place log entries in o
 
 ### Round a time down to whole hours
 
-Use the `with()` method of each Temporal type if you want to round or balance the fields.
+Use the `with()` method of each Temporal type (except `Temporal.Absolute`) if you want to round or balance the fields.
 Here's an example of rounding a time _down_ to the previously occurring whole hour:
 
 ```javascript
 {{cookbook/roundDownToWholeHours.mjs}}
 ```
+
+`Temporal.Absolute` is an absolute timestamp and doesn't have any concept of a calendar or wall clock, so it can't be rounded to a certain field value.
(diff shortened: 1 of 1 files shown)


docs/cookbook.md
```diff
@@ -123,13 +123,16 @@ Sort a list of ISO 8601 date/time strings, for example to place log entries in o
 
 ### Round a time down to whole hours
 
-Use the `with()` method of each Temporal type if you want to round or balance the fields.
+Use the `with()` method of each calendar-based Temporal type if you want to round or balance the fields.
```
justingrant, line 126: Time isn't calendar based, but does have a `with` method. Should this be "non-`Absolute`" instead of "calendar-based" ?

docs/cookbook.md
```diff
@@ -123,13 +123,16 @@ Sort a list of ISO 8601 date/time strings, for example to place log entries in o
...
 Here's an example of rounding a time _down_ to the previously occurring whole hour:
 
 ```javascript
 {{cookbook/roundDownToWholeHours.mjs}}
 ```
 
+`Temporal.Absolute` is an absolute timestamp and doesn't have any concept of a calendar, so it can't be rounded to a certain field value.
+If you need to round a `Temporal.Absolute` instance, convert it to a calendar-aware type such as `Temporal.DateTime`.
```
justingrant, line 134: Same here. Suggest removing "calendar-aware".

docs/cookbook.md
ptomato: You're right, this is a confusing shorthand — I really wish there was a word for "calendar and/or wall-clock based"! "Civil" would do nicely except it was decided long ago that that brought in its own confusion.

PR #807, opened 2020-08-08
Duration fixes

ptomato:
Minor bugs that I have run into while working on negative durations.

### DIFF
--- a/polyfill/test/regex.mjs
+++ b/polyfill/test/regex.mjs
@@ -487,12 +487,11 @@ describe('fromString regex', () => {
       ['0.123456789S', { ms: 123, µs: 456, ns: 789 }],
       ['0,123456789S', { ms: 123, µs: 456, ns: 789 }]
     ];
-    const tim = sec.reduce(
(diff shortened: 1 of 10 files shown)


polyfill/test/regex.mjs
```diff
@@ -492,18 +492,19 @@ describe('fromString regex', () => {
...
     });
-    tim.forEach(([p, expect]) => {
+    tim.slice(1).forEach(([p, expect]) => {
       test(`PT${p}`, expect);
       test(`Pt${p}`, expect);
       test(`pt${p.toLowerCase()}`, expect);
     });
     for (let [d, dexpect] of day) {
       for (let [t, texpect] of tim) {
+        if (d === '' && t === '') continue;
```
gibson042, line 507: This allows through empty `t` and then uses it to build input like "P3DT", which should instead be "P3D".

ryzokuken: +1 https://tools.ietf.org/html/rfc3339#appendix-A

PR #808, opened 2020-08-10
Formalize details for DateTime.toAbsolute.

Ms2ger:


spec/timezone.html
```diff
@@ -371,6 +371,20 @@ <h1>GetTemporalDateTimeFor ( _timeZone_, _absolute_ )</h1>
       </emu-alg>
     </emu-clause>
 
+    <emu-clause id="sec-temporal-getabsolutefor" aoid="GetAbsoluteFor">
```
ptomato, line 374: I'd suggest calling this GetTemporalAbsoluteFor.

PR #816, opened 2020-08-12
Add explainer for LocalDateTime

ptomato:
I intend to eventually adapt this into the documentation for
LocalDateTime, but given the draft status of the API this seems more
appropriate.

### DIFF
--- a/docs/localdatetime-draft.md
+++ b/docs/localdatetime-draft.md
@@ -1,0 +1,101 @@
+# Zoned Date/Time Type
+
+**Name:** It's currently being referred to as `Temporal.LocalDateTime`, but we already know this is not a good name for reasons discussed years ago.
+This name is not intended to stick.
+
+**What it is:** The LocalDateTime type is intended to refer to an event, that happened (or will happen) in a particular place.
+
+**Motivation:**
+- "Event that happened in a particular place" is a very common use case, perhaps the most common.
+  So we want a type that does the right thing for that use case.
+- Both Absolute and DateTime have cases where they do the right thing for this use case, and cases where they do the wrong thing.
+  (ex.: "Move a meeting three days later" is correct with DateTime and wrong with Absolute; "Postpone a reminder by one hour" is correct with Absolute and wrong with DateTime.)
+- The cases where Absolute and DateTime do the wrong thing will only produce the wrong result when used across a DST transition, so in many places the bugs will only pop up twice per year, if at all.
+- When writing the cookbook, to deal with this use case, we found that it was actually necessary a lot of the time to pass around a record consisting of `{ absolute, timeZone }`.
+  If that is so common, then we may as well have a type for it.
+- Even worse, there is a temptation to pass around a serialized string instead of such a record, which is wrong because the time zone offset may have changed when you read in the string again with `Temporal.Absolute.from()`.
+
+**API**: The API looks something like this:
+
+```typescript
+class Temporal.LocalDateTime {
+  // creation
+  constructor(absolute: Temporal.Absolute, timeZone: Temporal.TimeZone, calendar: Temporal.Calendar);
+  static from(item: string | object, options?: object) : Temporal.LocalDateTime;
+
+  // immutable 'mutation'
+  with(localDateTimeLike: object, options?: object) : Temporal.LocalDateTime;
+  withCalendar(calendar: Temporal.Calendar) : Temporal.LocalDateTime;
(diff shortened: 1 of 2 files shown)


docs/localdatetime-draft.md
```diff
@@ -0,0 +1,95 @@
...
+
+It basically has all the API of DateTime, but with a time zone as well.
+
+To create one, you use `Temporal.LocalDateTime.from`.
+- From a string: `from(string)`
+- From an Absolute: `from({ absolute, timeZone })`
+- From a DateTime, which potentially needs disambiguation: `from({ dateTime, timeZone }, { disambiguation })` or possibly `from({ ...dateTime.getFields(), timeZone }, { disambiguation })`
+- From raw DateTime fields: `from({ year, month, day, etc., timeZone })`
+
+As in the other Temporal types, the constructor is more low-level and takes an Absolute, a TimeZone, and an optional Calendar.
```
sffc, line 90: I strongly disagree with any notion of the time zone being required but the calendar being optional.  They should either both have defaults (UTC/ISO) or neither have defaults.

docs/localdatetime-draft.md
```diff
@@ -0,0 +1,95 @@
...
+  toString() : string;
+  toJSON() : string;
+  toLocaleString(locales?: string | Array<string>, options?: object): string;
+}
+```
+
+(This is not a fully functional description, it's made concise for readability.
+For example, the details of the options bags are omitted.)
+
+It basically has all the API of DateTime, but with a time zone as well.
```
sffc, line 82: I think it's important to emphasize the fact that although the API looks like DateTime+TimeZone, the data model is Absolute+TimeZone+Calendar.

pipobscure: I’m not sure about this: fundamentally the “data-model” to my mind is an implementation detail. So the DateTime + TimeZone is the relevant bit.

In consequence the Absolute+TimeZone+Calendar Tuple is an implementation detail.

docs/localdatetime-draft.md
pipobscure: Because the Absolute+TimeZone tuple is an implementation detail, and the calendar is really only present as a part of the DateTime I really disagree with this statement prima facie.

ptomato: I'm skeptical that insisting on this will change anyone's mind about the default calendar. Time zones and calendars are somewhat analogous, but not 100%. But I'm fine to write this very high-level document such that it doesn't indicate one way or the other :smile:

docs/localdatetime-draft.md
ptomato: I think it may be still relevant to mention, because DateTime does not unambiguously indicate a single event, and LocalDateTime does. I've added something about this.

docs/localdatetime-draft.md
```diff
@@ -0,0 +1,96 @@
...
+But since it represents an unambiguous moment in time (like Absolute, and unlike DateTime), the internal model is more like Absolute with a time zone and calendar.
+
+To create one, you use `Temporal.LocalDateTime.from`.
+- From a string: `from(string)`
+- From an Absolute: `from({ absolute, timeZone })`
```
justingrant, line 87: AFAIK this syntax has not been approved. Per our 2020-07-31 meeting I removed the `absolute` field from LocalDateTime in favor of `timeZoneOffsetNanoseconds`. 

Instead, the syntax that seems to have consensus for these use cases is `toLocalDateTime()` on Absolute, DateTime, Date, and Time. #700's poc.d.ts has the signatures of these methods.

docs/localdatetime-draft.md
justingrant: The public-facing API is most like DateTime+offset+TimeZone. So how about this:

> It basically has all the API of DateTime and adds a time zone, a time zone offset (for disambiguating near time zone offset transitions), and several convenience properties that depend on the time zone.

justingrant:
This is great. I noticed a few missing things: 
* The new `startOfDay` property (see #737) is now checked into #700. It should probably be in this content. 
* The `isTimeZoneOffsetTransition` property isn't mentioned.

docs/localdatetime-draft.md
ptomato: You're right; I started writing this a while ago and didn't update it. Should be fixed now.

ptomato:
Added highlighting to Prism in order to fix the docs build.

PR #821, opened 2020-08-13
Specify GetTimeZoneOffsetNanoseconds.

Ms2ger:


spec/timezone.html
```diff
@@ -134,7 +134,18 @@ <h1>Temporal.TimeZone.prototype.getOffsetNanosecondsFor ( _absolute_ )</h1>
         1. Let _timeZone_ be the *this* value.
         1. Perform ? RequireInternalSlot(_timeZone_, [[InitializedTemporalTimeZone]]).
         1. Perform ? RequireInternalSlot(_absolute_, [[InitializedTemporalAbsolute]]).
-        1. Return ? GetTimeZoneOffsetNanoseconds(_absolute_.[[Nanoseconds]], _timeZone_.[[Identifier]]).
+        1. Let _offset_ be ! ParseOffsetNanoseconds(_timeZone_.[[Identifier]]).
+        1. If _offset_ is not *undefined*, then
+          1. Return _offset_.
+        1. Let _epochNs_ be _absolute_.[[Nanoseconds]].
+        <!-- Can this be BalanceDateTime(1970, 1, 1, 0, 0, 0, 0, 0, _epochNs_)? -->
```
ptomato, line 141: That might disregard DST shifts? I'm not 100% sure though.

spec/timezone.html
```diff
@@ -134,7 +134,18 @@ <h1>Temporal.TimeZone.prototype.getOffsetNanosecondsFor ( _absolute_ )</h1>
         1. Let _timeZone_ be the *this* value.
         1. Perform ? RequireInternalSlot(_timeZone_, [[InitializedTemporalTimeZone]]).
         1. Perform ? RequireInternalSlot(_absolute_, [[InitializedTemporalAbsolute]]).
-        1. Return ? GetTimeZoneOffsetNanoseconds(_absolute_.[[Nanoseconds]], _timeZone_.[[Identifier]]).
+        1. Let _offset_ be ! ParseOffsetNanoseconds(_timeZone_.[[Identifier]]).
```
ptomato, line 137: This is the way it's implemented in the polyfill, but I wonder if we could consider having an internal slot in the spec to distinguish between whether a time zone object represents an offset or named time zone.

spec/timezone.html
```diff
@@ -408,5 +419,16 @@ <h1>TimeZoneToString ( _timeZone_ )</h1>
...
     </emu-clause>
+
+    <emu-clause id="sec-temporal-parseoffsetnanoseconds" aoid="ParseOffsetNanoseconds">
+      <h1>ParseOffsetNanoseconds ( _identifier_ )</h1>
+      <emu-alg>
+        1. Assert: Type(_identifier_) is String.
+        1. If _identifier_ is not an offset string, then
+          1. Return *undefined*.
+        1. Let _sign_, _hours_, _minutes_ be <mark>the result of parsing _identifier_</mark>.
+        1. Return _sign_ × (_hours_ × 60 + _minutes_) * 60 * 1,000,000,000.
```
ptomato, line 430: Nitpick: be consistent with using `×` or `*`.

spec/timezone.html
Ms2ger: Let's wait and see if we end up keeping offset time zones.

spec/timezone.html
```diff
@@ -134,7 +134,18 @@ <h1>Temporal.TimeZone.prototype.getOffsetNanosecondsFor ( _absolute_ )</h1>
...
+        1. Let _offset_ be ! ParseOffsetNanoseconds(_timeZone_.[[Identifier]]).
+        1. If _offset_ is not *undefined*, then
+          1. Return _offset_.
+        1. Let _epochNs_ be _absolute_.[[Nanoseconds]].
+        <!-- Can this be BalanceDateTime(1970, 1, 1, 0, 0, 0, 0, 0, _epochNs_)? -->
+        1. Let _ms_ be RoundTowardsZero(_epochNs_ / 1,000,000<sub>ℝ</sub>).
+        1. Let _mcs_ be ! NonNegativeModulo(RoundTowardsZero(_epochNs_ / 1000<sub>ℝ</sub>), 1000<sub>ℝ</sub>).
+        1. Let _ns_ be ! NonNegativeModulo(_epochNs_, 1000<sub>ℝ</sub>).
+        1. Let _tm_ be ! ToLocalTime(_ms_, *"gregory"*, _timeZone_.[[Identifier]]).
+        1. Let _result_ be ? BalanceDateTime(_tm_.[[Year]], _tm_.[[Month]], _tm_.[[Day]], _tm_.[[Hour]], _tm_.[[Minute]], _tm_.[[Second]], _ms_, _mcs_, _ns_).
```
justingrant, line 146: I'm late to the review party on this PR, but I didn't understand this line. Won't the `BalanceDateTime` operation end up almost doubling the distance from epoch in milliseconds?  Is there a missing step where we need to calculate the modulo `_ms_` use that remainder in `BalanceDateTime` instead of `_ms_`?

Ms2ger: Good catch, thank you: https://github.com/tc39/proposal-temporal/pull/828

PR #836, opened 2020-08-20
Consolidate getOffsetNanosecondsFor calls into ES.GetOffsetNanosecondsFor().

Ms2ger:
Also add tests for the result checks.

### DIFF
--- a/polyfill/lib/ecmascript.mjs
+++ b/polyfill/lib/ecmascript.mjs
@@ -607,6 +607,16 @@ export const ES = ObjectAssign({}, ES2019, {
       GetSlot(dateTime, NANOSECOND)
     );
   },
+  GetOffsetNanosecondsFor: (timeZone, absolute) => {
+    const offsetNs = timeZone.getOffsetNanosecondsFor(absolute);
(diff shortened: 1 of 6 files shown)


polyfill/lib/ecmascript.mjs
```diff
@@ -607,6 +607,16 @@ export const ES = ObjectAssign({}, ES2019, {
       GetSlot(dateTime, NANOSECOND)
     );
   },
+  GetOffsetNanosecondsFor: (timeZone, absolute) => {
+    const offsetNs = timeZone.getOffsetNanosecondsFor(absolute);
```
ryzokuken, line 611: Does this mean I have to update my PR once rebased on top of this to change this, and add new tests to handle `getOffsetNanosecondsFor = undefined`?

ryzokuken (approved):
LGTM, thanks.

polyfill/lib/ecmascript.mjs
Ms2ger: Might be easier if I deal with it.

PR #843, opened 2020-08-24
[spec] Move valueOf definitions into their sections.

Ms2ger:


ryzokuken (approved):
I already did this for `Date` in #806, but yes, this makes more sense.

PR #848, opened 2020-08-25
Fixes

ptomato:
Miscellaneous fixes.


spec/abstractops.html
```diff
@@ -662,9 +662,9 @@ <h1>ParseTemporalTimeZoneString ( _isoString_ )</h1>
...
         1. Return the new Record: {
-          [[OffsetSign]]: *undefined*,
-          [[OffsetHour]]: *undefined*,
-          [[OffsetMinute]]: *undefined*,
+          [[OffsetSign]]: 1,
+          [[OffsetHour]]: +0,
+          [[OffsetMinute]]: +0,
           [[Name]]: `"UTC"`
           }.
       1. If _hour_ is not *undefined*, then
```
Ms2ger, line 671: Can you clarify the invariant you're aiming for? `ParseTemporalTimeZone` expects that some of the values can be undefined, still (though the result is ignored). Not sure what `ParseTemporalAbsolute` expects.

polyfill/lib/timezone.mjs
```diff
@@ -175,6 +175,7 @@ export class TimeZone {
         GetSlot(dateTime, MICROSECOND),
         GetSlot(dateTime, NANOSECOND)
       );
+      if (epochNs === null) throw new RangeError('DateTime outside of supported range');
```
Ms2ger, line 178: This looks like it deserves a test.

Ms2ger (changes requested):

ryzokuken (approved):
LGTM with @Ms2ger's comments addressed.

spec/abstractops.html
ptomato: I have updated the callers and attempted to describe the invariant better in the commit message.

ptomato:
@Ms2ger Can you check this one again? It can't be merged until you mark that the requested changes have been addressed.

ptomato:
Oh, never mind, you can "dismiss review" if the comments have been addressed ... which strikes me as a bit rude though!

PR #851, opened 2020-08-26
Don't polyfill Temporal and Intl onto the global object

ptomato:
In order to avoid breaking the web when Temporal is shipped, don't
polyfill onto the global object. We don't want any `if (typeof Temporal
=== undefined)` checks in the wild until browsers actually start
shipping Temporal.

We do still need it on the global object for the browser playground and
the test262 tests, so we now have two separate entry points to the
library: index.mjs and polyfill.mjs.

Closes: #778


ryzokuken (approved):
Perfect, thanks a ton.

Issue #760, opened 2020-07-11
can't use toLocaleString() of Temporal.Duration in cookbook


hontas:
Running `Temporal.Duration.from('PT0S').toLocaleString()` in console produces a warning:

`duration.mjs:325 Temporal.Duration.prototype.toLocaleString() requires Intl.DurationFormat.`

ryzokuken:
@hontas that is correct. Localization for `Duration` instances is no simple task, and requires special data and routines of its own. To that effect, we've been working on a companion-proposal to introduce an `Intl.DurationFormat` class here: https://github.com/tc39/proposal-intl-duration-format. Unfortunately, we'd need to polyfill the said `Intl.DurationFormat` class in order to make `toLocaleString` work on `Duration` objects.

Since
1. No such polyfill exists so far.
2. This isn't a production-ready polyfill.

We print out this friendly warning to the console. Perhaps we could keep this issue open in order to add a note (or even a code comment) in the cookbook?

PR #852, opened 2020-08-26
Note that Duration.toLocaleString requires Intl.DurationFormat

ptomato:
Make a note of this in the documentation and the cookbook.

Closes: #760


ryzokuken (approved):
Thanks

PR #832, opened 2020-08-19
spec: misc normative timezone fixes

ryzokuken:

PR #863, opened 2020-09-01
polyfill: add intrinsic check

ryzokuken:
Add an additional check to ensure that intrinsics are registered once.

Refs: https://github.com/tc39/proposal-temporal/pull/832#discussion_r478316185

/cc @Ms2ger


Ms2ger (approved):
Thanks!

PR #865, opened 2020-09-01
timezonetojson: improve spec and add tests

ryzokuken:
Refs: https://github.com/tc39/proposal-temporal/pull/832#discussion_r478282533

/cc @Ms2ger @ptomato

### DIFF
--- a/polyfill/lib/timezone.mjs
+++ b/polyfill/lib/timezone.mjs
@@ -227,7 +227,7 @@ export class TimeZone {
     return String(GetSlot(this, TIMEZONE_ID));
   }
   toJSON() {
-    return this.toString();
+    return ES.TimeZoneToString(this);
   }
   static from(item) {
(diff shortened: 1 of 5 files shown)


polyfill/lib/timezone.mjs
```diff
@@ -227,7 +227,7 @@ export class TimeZone {
     return String(GetSlot(this, TIMEZONE_ID));
   }
   toJSON() {
-    return this.toString();
+    return ES.TimeZoneToString(this);
```
ptomato, line 230: At first it's a bit confusing that ES.TimeZoneToString isn't used in toString above, but if I think about it, I believe it's correct.

ryzokuken: It's the other way round (`ES.TimeZoneToString` calls `this.toString` if it exists)

Ms2ger (approved):

PR #868, opened 2020-09-02
[editorial] Fix formatting of an assertion.

Ms2ger:


Ms2ger:
Seems like this got stuck in "pending" on #848 somehow.

ryzokuken (approved):
Thanks

PR #872, opened 2020-09-03
polyfill: fix inconsistent method on ES object

ryzokuken:
/cc @ptomato @Ms2ger


Ms2ger (approved):
I actually like this syntax more - I wonder why we don't use it throughout

ryzokuken:
@Ms2ger the one I switched to now? I did it here just for consistency, because the inconsistency irritated me as well as my editor, but I agree with you that the syntax looks nicer.

PR #876, opened 2020-09-04
Split disambiguation and overflow

ptomato:
The big rename! Overflow is for out-of-range, disambiguation is for time zones.

Closes: #607


Ms2ger (approved):

ptomato:
Thanks for the review, I put back the if-clauses!

PR #879, opened 2020-09-08
Specify ISO8601Calendar.

Ms2ger:
This leaves the actual algorithms undefined, but it serves as a skeleton for
the work that needs to be done.

### DIFF
--- a/polyfill/lib/calendar.mjs
+++ b/polyfill/lib/calendar.mjs
@@ -114,7 +114,7 @@ export class Calendar {
 MakeIntrinsicClass(Calendar, 'Temporal.Calendar');
 DefineIntrinsic('Temporal.Calendar.from', Calendar.from);
 
-class ISO8601 extends Calendar {
+class ISO8601Calendar extends Calendar {
   constructor(id = 'iso8601') {
     // Needs to be subclassable, that's why the ID is a default argument
     super(id);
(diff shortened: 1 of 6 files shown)


ptomato (approved):
LGTM! Up to you whether you prefer to merge it in advance of the discussion being reopened on #847 or not. I think it's fine to merge it since it does reflect what's currently in the polyfill.

polyfill/lib/calendar.mjs
```diff
@@ -114,7 +114,7 @@ export class Calendar {
 MakeIntrinsicClass(Calendar, 'Temporal.Calendar');
 DefineIntrinsic('Temporal.Calendar.from', Calendar.from);
 
-class ISO8601 extends Calendar {
+class ISO8601Calendar extends Calendar {
```
ryzokuken, line 117: Why this change when it's `Japanese` and not `JapaneseCalendar`?

ryzokuken (approved):
LGTM, thanks.

polyfill/lib/calendar.mjs
Ms2ger: I haven't looked at the Japanese calendar yet.

ryzokuken: Basically, my question is, do you plan to suffix each class by `Calendar`? It looks weird when it's inconsistent, but if you plan to eventually rename all of them, then it's all good.

Ms2ger: If we end up going with subclassing, I think suffixing them all makes sense, yes.

ryzokuken:
Going ahead with the merge because https://github.com/tc39/proposal-temporal/pull/879#discussion_r485441277

PR #881, opened 2020-09-09
Object handling around Temporal.Calendar.from().

Ms2ger:


polyfill/lib/ecmascript.mjs
```diff
@@ -576,7 +576,11 @@ export const ES = ObjectAssign({}, ES2019, {
     if (from === undefined) {
       from = GetIntrinsic('%Temporal.Calendar.from%');
     }
-    return ES.Call(from, TemporalCalendar, [calendarLike]);
+    const calendar = ES.Call(from, TemporalCalendar, [calendarLike]);
+    if (ES.Type(calendar) !== 'Object') {
+      throw new TypeError('Temporal.from should return an object');
```
ryzokuken, line 581:
```suggestion
      throw new TypeError('Temporal.Calendar.from should return an object');
```

ryzokuken (approved):
Just one comment, thanks.

polyfill/lib/ecmascript.mjs
```diff
@@ -57,7 +58,8 @@ const ES2019 = {
   ToNumber,
   ToObject,
   ToPrimitive,
-  ToString
+  ToString,
+  Type
```
ptomato, line 62: This is also in https://github.com/tc39/proposal-temporal/pull/855, if you would like to review that

ptomato (approved):
:+1:

PR #883, opened 2020-09-09
Rename getISOCalendarFields to getISOFields

ptomato:
Result of a naming discussion. Also renames the properties in the
returned object, and adds 'calendar' to the object as well. This way,
the properties on the returned object exactly match what is passed to
each low-level constructor.

Closes: #630


sffc (approved):
Thanks!

Ms2ger (approved):

PR #885, opened 2020-09-10
Start defining Date/DateTime attributes in terms of Calendar.

Ms2ger:


spec/date.html
```diff
@@ -126,7 +126,9 @@ <h1>get Temporal.Date.prototype.year</h1>
       <emu-alg>
         1. Let _temporalDate_ be the *this* value.
         1. Perform ? RequireInternalSlot(_temporalDate_, [[InitializedTemporalDate]]).
-        1. Return _temporalDate_.[[ISOYear]].
+        1. Let _calendar_ be _temporalDate_.[[Calendar]].
+        1. Let _method_ be ? Get(_calendar_, *"year"*).
+        1. Return ? Call(_calendar_, _method_, « _temporalDate_ »).
```
ptomato, line 144: I think these are backwards, in https://tc39.es/ecma262/#sec-call the function object comes before the `this` value.

spec/isocalendar.html
```diff
@@ -151,112 +151,153 @@ <h1>Temporal.ISO8601Calendar.prototype.dateDifference ( _one_, _two_, _options_
     </emu-clause>
 
     <emu-clause id="sec-temporal.iso8601calendar.prototype.year">
-      <h1>Temporal.ISO8601Calendar.prototype.year ( _date_ )</h1>
+      <h1>Temporal.ISO8601Calendar.prototype.year ( _dateOrDateTime_ )</h1>
```
ptomato, line 154: It could also be a YearMonth, should we call it something like `_objectWithYear_`? (Same goes for below: era(), daysInMonth(), daysInYear(), monthsInYear(), and isLeapYear() are the same; the argument to month() could be a YearMonth or MonthDay; and the argument to day() could be a MonthDay)

spec/isocalendar.html
```diff
@@ -151,112 +151,153 @@ <h1>Temporal.ISO8601Calendar.prototype.dateDifference ( _one_, _two_, _options_
...
+    </emu-clause>
+
+    <emu-clause id="sec-temporal.iso8601calendar.prototype.daysinweek">
+      <h1>Temporal.ISO8601Calendar.prototype.daysInWeek ( _dateOrDateTime_ )</h1>
+      <p>
+        The `daysInWeek` method takes one argument _dateOrDateTime_.
+        The following steps are taken:
+      </p>
+      <emu-alg>
+        1. Perform ? RequireInternalSlot(_dateOrDateTime_, [[ISOYear]]).
```
ptomato, line 250: Should it require ISOMonth and ISODay as well? Otherwise it wouldn't throw if given a YearMonth.

ptomato (approved):
This is great. Just minor comments. Feel free to punt the YearMonth/MonthDay considerations until after we actually spec calendar on those types.

Ms2ger:
I punted on other types explicitly.

spec/isocalendar.html
```diff
@@ -151,112 +151,146 @@ <h1>Temporal.ISO8601Calendar.prototype.dateDifference ( _one_, _two_, _options_
...
-      <h1>Temporal.ISO8601Calendar.prototype.year ( _date_ )</h1>
+      <h1>Temporal.ISO8601Calendar.prototype.year ( _dateOrDateTime_ )</h1>
       <p>
-        The `year` method takes one argument _date_.
+        The `year` method takes one argument _dateOrDateTime_.
         The following steps are taken:
       </p>
       <emu-alg>
-        1. <mark>TODO.</mark>
+        1. Perform ? RequireOneOfInternalSlots(_dateOrDateTime_, « [[InitializedTemporalDate]], [[InitializedTemporalDateTime]] »).
```
ptomato, line 160: FWIW, I thought it was fine to have RequireInternalSlot(..., [[ISOYear]]) ... since all the types for which this is allowed have an [[ISOYear]] slot. But I think this is fine too. We can always change it if it gets too cumbersome when we add YearMonth, MonthDay, and LocalDateTime.

ptomato (approved):
Looks good!

spec/isocalendar.html
ljharb: fwiw i don't think we'll want to land a `RequireOneOfInternalSlots` abstract operation in the spec itself - a type should have *all* of the required internal slots, every time.

Ms2ger: @ljharb I don't understand. These methods are supposed to support several different types.

ljharb: @Ms2ger i'm just saying that in the spec text, we have RequireInternalSlot, but we don't have "RequireInternalSlots" - it's fine that the methods support several different types, but that shouldn't be done by conflating cross-type internal slots. If you want to test for multiple types, then do that with multiple abstract operations.

PR #907, opened 2020-09-15
Rounding, Part 2

ptomato:
This adds rounding options to Temporal.Absolute.prototype.difference, Temporal.DateTime.prototype.difference, and Temporal.Time.prototype.difference.

See: #827.

### DIFF
--- a/docs/absolute.md
+++ b/docs/absolute.md
@@ -302,10 +302,18 @@ Temporal.now.absolute().minus(oneHour);
   - `largestUnit` (string): The largest unit of time to allow in the resulting `Temporal.Duration` object.
     Valid values are `'hours'`, `'minutes'`, `'seconds'`, `'milliseconds'`, `'microseconds'`, and `'nanoseconds'`.
     The default is `"seconds"`.
+  - `smallestUnit` (string): The smallest unit of time to round to in the resulting `Temporal.Duration` object.
+    Valid values are the same as for `largestUnit`.
+    The default is `'nanoseconds'`, i.e., no rounding.
+  - `roundingIncrement` (number): The granularity to round to, of the unit given by `smallestUnit`.
+    The default is 1.
+  - `roundingMode` (string): How to handle the remainder, if rounding.
+    Valid values are `'ceil'`, `'floor'`, `'trunc'`, and `'nearest'`.
+    The default is `'nearest'`.
 
 **Returns:** a `Temporal.Duration` representing the difference between `absolute` and `other`.
 
-This method computes the difference between the two times represented by `absolute` and `other`, and returns it as a `Temporal.Duration` object.
+This method computes the difference between the two times represented by `absolute` and `other`, optionally rounds it, and returns it as a `Temporal.Duration` object.
 If `other` is later than `absolute` then the resulting duration will be negative.
 
 The `largestUnit` option controls how the resulting duration is expressed.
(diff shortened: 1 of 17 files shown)


docs/absolute.md
```diff
@@ -319,6 +327,10 @@ This is because months and years can be different lengths depending on which mon
 You cannot determine the start and end date of a difference between `Temporal.Absolute`s, because `Temporal.Absolute` has no time zone or calendar.
 In addition, weeks can be different lengths in different calendars, and days can be different lengths when the time zone has a daylight saving transition.
 
+You can round the result using the `smallestUnit`, `roundingIncrement`, and `roundingMode` options.
+These behave as in the `Temporal.Duration.round()` method, but increments of days and larger are counted relative to `absolute`.
```
justingrant, line 331: Could you explain this sentence?  I don't fully understand what you're trying to convey.

ptomato: Oops, this didn't apply to Absolute rounding anyway. In DateTime I'll change it to "These behave as in the `Temporal.Duration.round()` method, but increments of days and larger are allowed.
Since rounding to an increment expressed in days or larger units requires a reference point, `dateTime` is used as the reference point in that case."

justingrant: Cool. 

> `dateTime` is used as the reference point in that case."

Shouldn't it be this instead?

>  `this` is used as the reference point in that case

ptomato: I guess that's a matter of writing style; elsewhere in the docs we refer to the `this` object as whatever it's called in the `<h3>`, as in "**Returns:** a `Temporal.Duration` representing the difference between `datetime` and `other`." (rather than "...`this` and `other`")

justingrant: Personally I find `this` to be clearer (and easier for copy/pasting into different .md files!) but I don't have a strong opinion one way or the other.  Either way is fine.

ptomato: I'm not sure how it got started this way, tbh - I checked on MDN and it seems like they actually avoid using either way, which actually seems worse to me.

polyfill/lib/absolute.mjs
```diff
@@ -103,16 +103,53 @@ export class Absolute {
   difference(other, options) {
     if (!ES.IsTemporalAbsolute(this)) throw new TypeError('invalid receiver');
     if (!ES.IsTemporalAbsolute(other)) throw new TypeError('invalid Absolute object');
-    const largestUnit = ES.ToLargestTemporalUnit(options, 'seconds', ['years', 'months', 'weeks', 'days']);
+    const disallowedUnits = ['years', 'months', 'weeks', 'days'];
+    const smallestUnit = ES.ToSmallestTemporalDurationUnit(options, 'nanoseconds', disallowedUnits);
+    let defaultLargestUnit = 'seconds';
```
Ms2ger, line 108:
```suggestion
    const defaultLargestUnit = 'seconds';
```

spec/absolute.html
```diff
@@ -285,9 +285,37 @@ <h1>Temporal.Absolute.prototype.difference ( _other_ [ , _options_ ] )</h1>
...
+          1. Let _defaultLargestUnit_ be *"seconds"*.
+        1. Let _largestUnit_ be ? ToLargestTemporalUnit(_options_, « *"years"*, *"months"*, *"weeks"*, *"days"* », _defaultLargestUnit_).
+        1. Perform ? ValidateTemporalDifferenceUnits(_largestUnit_, _smallestUnit_).
+        1. Let _roundingMode_ be ? ToTemporalRoundingMode(_options_).
+        1. If _smallestUnit_ is *"hours"*, then
+          1. Let _maximum_ be 24.
+        1. Else if _smallestUnit_ is *"minutes"* or *"seconds"*, then
+          1. Let _maximum_ be 60.
+        1. Else,
+          1. Let _maximum_ be 1000.
```
Ms2ger, line 301: This text is repeated at least twice already. Let's make an abstract op.

docs/time.md
```diff
@@ -245,23 +245,38 @@ time.minus({ minutes: 5, nanoseconds: 800 })  // => 19:34:09.068345405
   - `largestUnit` (string): The largest unit of time to allow in the resulting `Temporal.Duration` object.
     Valid values are `'hours'`, `'minutes'`, `'seconds'`, `'milliseconds'`, `'microseconds'`, and `'nanoseconds'`.
     The default is `'hours'`.
+  - `smallestUnit` (string): The smallest unit of time to round to in the resulting `Temporal.Duration` object.
+    Valid values are the same as for `largestUnit`.
+    The default is `'nanoseconds'`, i.e. no rounding.
```
Ms2ger, line 250:
```suggestion
    The default is `'nanoseconds'`, i.e., no rounding.
```

Ms2ger (approved):

polyfill/lib/absolute.mjs
ptomato: It's reassigned in the next line if `smallestUnit` is larger than `defaultLargestUnit`.

PR #942, opened 2020-09-25
Update references to Absolute in test metadata.

Ms2ger:


ryzokuken (approved):
Some files went from `*/getPossibleAbsolutesFor/*` to `*/getPossibleInstantsFor/getPossibleAbsolutesFor/*`. Is that intentional?

PR #943, opened 2020-09-25
Fix and test the length of methods.

Ms2ger:


ptomato (approved):
I didn't check every test file, to be honest.

PR #954, opened 2020-10-02
Replace ÷ with /

aayushmau5:


ryzokuken:
@aayushmau5 you ran prettier on the spec text. Undo formatting and please only replace the required symbols.

justingrant:
> you ran prettier on the spec text. 

@ryzokuken  is there a prettier setting to avoid the HTML files on the spec directory? I've made the same mistake a few times myself.

ljharb:
Duplicate of #956, this should be closed.

PR #957, opened 2020-10-02
Legacy date conversion

ptomato:
Closes: #515 

Includes a few stragglers from the Absolute→Instant rename and from the addition of round(), that I noticed.


ryzokuken (approved):
LGTM, thanks. Perhaps we also need to have a diffed version of the 262 and 402 specs somewhere at some point.

ryzokuken:
There's a linter error that needs to be fixed. Also, should we rename `initialise` to `initialize` since we tend to prefer American spellings in TC39? :see_no_evil: Not sure, perhaps `init` is a more natural alternative.

PR #969, opened 2020-10-07
Fix documentation of era property in from() methods

ptomato:
Closes: #902


ptomato:
The corresponding polyfill code and spec text is already being done in #806.

ryzokuken (approved):
LGTM, thanks.

Issue #905, opened 2020-09-15
Should the fields object passed to calendar.*FromFields methods have a 'calendar' property?


Ms2ger:

sffc:
Do we pass the user-supplied fields object to calendar.\*FromFields, or do we make a new object?  If we pass the user-supplied one directly, we shouldn't bother with removing the calendar.  If we re-create a new object, then I think it makes sense to remove the calendar field.

Ms2ger:
From a quick `grep`, it seems like only the `from` methods pass through a user-supplied object (and I'd probably prefer to change those as well for consistency).

PR #970, opened 2020-10-07
Fix fields object in calendar.{date,yearMonth,monthDay}FromFields()

ptomato:
Closes: #905

### DIFF
--- a/polyfill/lib/ecmascript.mjs
+++ b/polyfill/lib/ecmascript.mjs
@@ -571,11 +571,8 @@ export const ES = ObjectAssign({}, ES2019, {
       const value = bag[property];
       if (value !== undefined) {
         any = any || {};
-        if (property === 'calendar') {
+        if (property === 'era') {
           // FIXME: this is terrible
-          const TemporalCalendar = GetIntrinsic('%Temporal.Calendar%');
-          any.calendar = TemporalCalendar.from(value);
-        } else if (property === 'era') {
           any.era = value;
         } else {
           any[property] = ES.ToInteger(value);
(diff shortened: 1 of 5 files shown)


ptomato:
No accompanying spec text change, because the fromFields methods aren't in the spec yet. When we get to that point, we can refer to the polyfill for the logic.

polyfill/lib/ecmascript.mjs
```diff
@@ -571,11 +571,8 @@ export const ES = ObjectAssign({}, ES2019, {
       const value = bag[property];
       if (value !== undefined) {
         any = any || {};
-        if (property === 'calendar') {
+        if (property === 'era') {
           // FIXME: this is terrible
```
ryzokuken, line 575: Still terrible? :smile:

polyfill/lib/ecmascript.mjs
```diff
@@ -596,11 +593,8 @@ export const ES = ObjectAssign({}, ES2019, {
         }
         value = defaultValue;
       }
-      if (property === 'calendar') {
+      if (property === 'era') {
         // FIXME: this is terrible
```
ryzokuken, line 597: Same.

ryzokuken (approved):
LGTM, thanks. Please remove those two comments too?

> When we get to that point, we can refer to the polyfill for the logic.

Will do.

polyfill/lib/ecmascript.mjs
ptomato: I still think it's terrible to special-case any property here, only slightly less so than before...

ryzokuken: Fair enough, will merge in that case.

PR #983, opened 2020-10-14
Add Duration.isZero

ptomato:
A convenience method that can be used instead of the less discoverable
duration.sign === 0.

Closes: #924


justingrant (approved):
Looks great! It'll be one more helper function I can remove from #700 once this lands! ;-)

ryzokuken (approved):
Perfect, thanks!

PR #992, opened 2020-10-14
Fix TypeScript bindings

ptomato:
I had forgotten to update the TypeScript bindings to include some of the
recently merged features.

### DIFF
--- a/polyfill/index.d.ts
+++ b/polyfill/index.d.ts
@@ -245,6 +245,7 @@ export namespace Temporal {
     readonly nanoseconds: number;
     negated(): Temporal.Duration;
     abs(): Temporal.Duration;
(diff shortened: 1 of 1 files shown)


justingrant (approved):
LGTM

justingrant:
Actually, what about TimeZone.id and TimeZoneProtocol.id ?

justingrant:
Also, why is `TimeZoneProtocol.id` optional?  Shouldn't it be required?

Ms2ger (approved):

ptomato:
You're right about `name` → `id`, I'll sneak that one into my next PR.
`id` is optional, `toString()` is required. If you don't provide an `id` then the output of `toString()` is used.

PR #996, opened 2020-10-15
Various editorial changes.

Ms2ger:


spec/temporal.html
```diff
@@ -160,14 +160,14 @@ <h1>Abstract operations</h1>
...
-        1. Return ? CreateTimeZone(_identifier_).
+        1. Return ? CreateTemporalTimeZone(_identifier_).
       </emu-alg>
     </emu-clause>
 
     <emu-clause id="sec-temporal-systeminstant" aoid="SystemInstant">
       <h1>SystemInstant ( )</h1>
       <emu-alg>
-        1. Let _ns_ be the current UTC date and time, in nanoseconds since the epoch.
+        1. Let _ns_ be the approximate current UTC date and time, in nanoseconds since the epoch.
```
ryzokuken, line 170: Any specific reason for this change?

ryzokuken (approved):
LGTM

spec/temporal.html
Ms2ger: #977, but maybe there's a better way to clarify that you're not necessarily getting the most precise result possible.

PR #1004, opened 2020-10-16
Various improvements

Ms2ger:


spec/isocalendar.html
```diff
@@ -23,6 +23,7 @@ <h1>Temporal.ISO8601Calendar ( _id_ )</h1>
       <emu-alg>
         1. If NewTarget is *undefined*, then
           1. Throw a *TypeError* exception.
+        1. Set _id_ to ? ToString(_id_).
```
Ms2ger, line 26: I want to fold this into the right commit, so please don't merge yet.

polyfill/lib/date.mjs
```diff
@@ -329,6 +336,5 @@ export class Date {
     return ES.CalendarCompare(GetSlot(one, CALENDAR), GetSlot(two, CALENDAR));
   }
 }
-Date.prototype.toJSON = Date.prototype.toString;
```
ptomato, line 336: I was wondering why this was necessary but I guess it's because you could otherwise say `Temporal.Date.prototype.toString.mydata = 3` and have it show up on `toJSON`?

polyfill/lib/date.mjs
```diff
@@ -288,10 +288,10 @@ export class Date {
   getISOFields() {
     if (!ES.IsTemporalDate(this)) throw new TypeError('invalid receiver');
     return {
-      isoYear: GetSlot(this, ISO_YEAR),
```
ptomato, line 283: Is the order observable somehow?

ptomato (approved):
Looks good, feel free to merge when you have rearranged the commits as you mentioned.

polyfill/lib/date.mjs
Ms2ger: Yeah, with `Object.keys()`

polyfill/lib/date.mjs
Ms2ger: Indeed, or just running `Temporal.Date.prototype.toString === Temporal.Date.prototype.toJSON`

PR #1006, opened 2020-10-16
toString() precision

ptomato:
The bulk of this commit is adding `:00` to expected test output; it's probably safe to ignore those parts when reviewing this.

Closes: #329


Ms2ger (changes requested):
Note that I rebased and fixed the failing test262 tests

spec/abstractops.html
```diff
@@ -142,6 +142,84 @@ <h1>ToTemporalRoundingIncrement ( _normalizedOptions_, _dividend_, _inclusive_ )
...
+            [[Increment]]: 1
+          }.
+      1. If _smallestUnit_ is *"nanosecond"*, then
+        1. Return the new Record {
+            [[Precision]]: 9,
+            [[Unit]]: *"nanosecond"*,
+            [[Increment]]: 1
+          }.
+      1. Let _digits_ be ? Get(_normalizedOptions_, *"fractionalSecondDigits"*).
+      1. If _digits_ is *undefined* or *"auto"*, then
```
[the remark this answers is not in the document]
ptomato: Intl.NumberFormat v3 has a [`useGrouping` which can be either a boolean or a string](https://github.com/tc39/ecma402/issues/367), but the only spec text that's been written so far is

> Let useGrouping be ? GetStringOrBooleanOption(options, "useGrouping", « "min2", "auto", "always" », "always", "auto").

GetStringOrBooleanOption hasn't been written yet AFAIK. It seems good to add a GetStringOrNumberOption here for this case.

polyfill/lib/ecmascript.mjs
```diff
@@ -482,6 +482,53 @@ export const ES = ObjectAssign({}, ES2019, {
...
+        return { precision: 'minute', unit: 'minute', increment: 1 };
+      case 'second':
+        return { precision: 0, unit: 'second', increment: 1 };
+      case 'millisecond':
+        return { precision: 3, unit: 'millisecond', increment: 1 };
+      case 'microsecond':
+        return { precision: 6, unit: 'microsecond', increment: 1 };
+      case 'nanosecond':
+        return { precision: 9, unit: 'nanosecond', increment: 1 };
+    }
```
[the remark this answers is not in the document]
ptomato: It isn't, `smallestUnit` can be `undefined` as well. I've added a default case with a comment to clarify this.

spec/abstractops.html
```diff
@@ -346,24 +442,22 @@ <h1>MaximumTemporalDurationRoundingIncrement ( _unit_ )</h1>
...
-      1. If _decimal_ is not empty, then
-        1. Set _result_ to the string-concatenation of _result_, the code unit 0x002E (FULL STOP), and _decimal_.
-      1. Return the string-concatenation of the code unit 0x003A (COLON) and _result_.
+    <h1>FormatSecondsStringPart ( _second_, _millisecond_, _microsecond_, _nanosecond_, _precision_ )</h1>
+    <emu-alg>
+      1. If _precision_ is *"minute"*, return *""*.
+      1. Let _secondsString_ be the string-concatenation of the code unit 0x003A (COLON) and _second_ formatted as a two-digit decimal number, padded to the left with zeroes if necessary.
+      1. Let _fraction_ be _millisecond_ × 10<sup>6</sup> + _microsecond_ × 10<sup>3</sup> + _nanosecond_.
+      1. If _precision_ is *"auto"*, then
+        1. If _fraction_ = 0, return _secondsString_.
```
Ms2ger, line 451: "is"

Ms2ger (changes requested):

polyfill/lib/ecmascript.mjs
```diff
@@ -483,6 +483,54 @@ export const ES = ObjectAssign({}, ES2020, {
...
+        return { precision: 9, unit: 'nanosecond', increment: 1 };
+      default: // fall through if option not given
+    }
+    let digits = options.fractionalSecondDigits;
+    if (digits === undefined || digits === 'auto') return { precision: 'auto', unit: 'nanosecond', increment: 1 };
+    digits = ES.ToNumber(digits);
+    if (NumberIsNaN(digits) || digits < 0 || digits > 9) {
+      throw new RangeError(`fractionalSecondDigits must be 'auto' or 0 through 9, not ${digits}`);
+    }
+    const precision = MathFloor(digits);
```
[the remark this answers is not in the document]
ptomato: This is what GetNumberOption from ECMA-402 does, which we've copied into this spec as well, and I think it's better to keep the behaviour consistent.

spec/abstractops.html
```diff
@@ -86,6 +86,28 @@ <h1>GetNumberOption ( _options_, _property_, _minimum_, _maximum_, _fallback_ )<
...
+    </p>
+    <emu-alg>
+      1. Assert: Type(_options_) is Object.
+      1. Let _value_ be ? Get(_options_, _property_).
+      1. If _value_ is *undefined*, return _fallback_.
+      1. Set _value_ to ? ToPrimitive(_value_, ~number~).
+      1. If Type(_value_) is Number, then
+        1. If _value_ is *NaN*, or _value_ &lt; _minimum_, or _value_ &gt; _maximum_, throw a *RangeError* exception.
+        1. Return floor(ℝ(_value_)).
+      1. Set _value_ to ? ToString(_value_).
```
Ms2ger, line 105: Calling ToString on the result of ToPrimitive seems like a bad idea to me. If we don't have precedent on the TC39 side, I'd like to follow WebIDL precedent, which comes down to just dropping to ToPrimitive call (so use the number if Type is number, and ToString otherwise).

spec/abstractops.html
```diff
@@ -148,6 +170,80 @@ <h1>ToTemporalRoundingIncrement ( _normalizedOptions_, _dividend_, _inclusive_ )
...
+            [[Unit]]: *"microsecond"*,
+            [[Increment]]: 1
+          }.
+      1. If _smallestUnit_ is *"nanosecond"*, then
+        1. Return the new Record {
+            [[Precision]]: 9,
+            [[Unit]]: *"nanosecond"*,
+            [[Increment]]: 1
+          }.
+      1. Let _digits_ be ? GetStringOrNumberOption(_normalizedOptions_, *"fractionalSecondDigits"*, « *"auto"* », 0, 9, *"auto"*).
```
Ms2ger, line 213: Maybe it would have reduced my earlier confusion if there was a step above this saying "Assert: _smallestUnit_ is **undefined**."

Ms2ger (approved):
LGTM modulo the one comment

spec/abstractops.html
ptomato: There is no precedent that I'm aware of of the TC39 side, so that sounds fine to me.

ptomato:
Addressed the review comments, so merging as per the approval

PR #1007, opened 2020-10-16
Add note to ECMA-402 diffs about host environment tzdata updates

ptomato:
Closes: #629

### DIFF
--- a/spec/intl.html
+++ b/spec/intl.html
@@ -460,6 +460,93 @@
       </emu-alg>
     </emu-clause>
     </ins>
+
+    <emu-clause id="sec-temporal-tolocaltime" aoid="ToLocalTime">
+      <h1>ToLocalTime ( _t_, _calendar_, _timeZone_ )</h1>
+
+      <p>
+        When the ToLocalTime abstract operation is called with arguments _t_, _calendar_, and _timeZone_, the following steps are taken:
+      </p>
+
+      <emu-alg>
+        1. Assert: Type(_t_) is Number.
+        1. If _calendar_ is *"gregory"*, then
+          1. Let _timeZoneOffset_ be the value calculated according to <emu-xref href="#sec-local-time-zone-adjustment">LocalTZA(_t_, *true*)</emu-xref> where the local time zone is replaced with timezone _timeZone_.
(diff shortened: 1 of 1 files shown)


sffc:
I don't get it.  Isn't ToLocalTime in ECMA-262?  Why are you adding it to intl.html?

ptomato:
It is in ECMA-402: https://tc39.es/ecma402/#sec-tolocaltime

spec/intl.html
```diff
@@ -460,6 +460,93 @@ <h1>GetDateTimeFormatPattern ( _opt_, _formats_, _hc_ )</h1>
...
+          1. Let _timeZoneOffset_ be the value calculated according to <emu-xref href="#sec-local-time-zone-adjustment">LocalTZA(_t_, *true*)</emu-xref> where the local time zone is replaced with timezone _timeZone_.
+          <ins>Given the same values of _t_ and _timeZone_, the result must be the same for the lifetime of the surrounding agent</ins>.
+          1. Let _tz_ be the time value _t_ + _timeZoneOffset_.
+          1. Return a record with fields calculated from _tz_ according to <emu-xref href="#table-temporal-datetimeformat-tolocaltime-record"></emu-xref>.
+        1. Else,
+          1. Return a record with the fields of Column 1 of <emu-xref href="#table-temporal-datetimeformat-tolocaltime-record"></emu-xref> calculated from _t_ for the given _calendar_ and _timeZone_. The calculations should use best available information about the specified _calendar_ and _timeZone_, including current and historical information about time zone offsets from UTC and daylight saving time rules.
+          <ins>Given the same values of _t_, _calendar_, and _timeZone_, the result must be the same for the lifetime of the surrounding agent</ins>.
```
justingrant, line 480: Why is the `<ins>` about caching not present in the Else branch?

ptomato: It's there in line 475 (after the call to LocalTZA)

sffc (approved):

spec/intl.html
justingrant: Right, but doesn't that `<ins>` refer to the `If` branch, not the `Else` branch?

justingrant: Oh, never mind- it's on line 280. No idea why I didn't see that earlier.

justingrant (approved):

ptomato:
Thanks for the reviews!

PR #1009, opened 2020-10-16
Remove "expected" from test names

justingrant:
The only way I've found to identify demitasse test failures is to search the output for the word "expected". This PR makes this grepping easier by removing "expected" from the names of tests.


ptomato (approved):
:+1:

Issue #665, opened 2020-06-11
Define legal userland time zone and calendar identifiers


ptomato:
Follow up from #638. Previously the [definition of what should be parsed as a time zone name](https://tc39.es/proposal-temporal/index.html#prod-TimeZoneIANAName) was "any string for which IsValidTimeZoneName returns true".

Now that we allow custom time zones, what names should be allowed? Is `Etc/🐈` a valid name for a time zone, making `2020-06-11T16:06:23.123+01:23[Etc/🐈]` a valid ISO string?
IANA names adhere to the following rules:

> Use only valid POSIX file name components (i.e., the parts of names other than `/`). Do not use the file name components `.` and `..`. Within a file name component, use only ASCII letters, `.`, `-` and `_`. Do not use digits, as that might create an ambiguity with POSIX TZ strings. A file name component must not exceed 14 characters or start with `-`.

Same question for calendar names.

ptomato:
(Calendar names consist of components of between 2 and 8 ASCII alphanumeric characters each, separated by dashes: https://tools.ietf.org/html/bcp47#section-2.1)

PR #1010, opened 2020-10-16
Define valid calendar and time zone IDs

ptomato:
Closes: #665

### DIFF
--- a/docs/calendar.md
+++ b/docs/calendar.md
@@ -41,6 +41,8 @@ For specialized applications where you need to do calculations in a calendar sys
 To do this, create a class inheriting from `Temporal.Calendar`, call `super()` in the constructor with a calendar identifier, and implement all the methods.
 Any subclass of `Temporal.Calendar` will be accepted in Temporal APIs where a built-in `Temporal.Calendar` would work.
 
+The identifier of a custom calendar must consist of one or more components of between 3 and 8 ASCII alphanumeric characters each, separated by dashes, as described in [Unicode Technical Standard 35](https://unicode.org/reports/tr35/tr35.html#Unicode_locale_identifier).
+
 ### Protocol
 
 It's also possible for a plain object to be a custom calendar, without subclassing.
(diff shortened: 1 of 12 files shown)


docs/calendar.md
```diff
@@ -41,6 +41,8 @@ For specialized applications where you need to do calculations in a calendar sys
 To do this, create a class inheriting from `Temporal.Calendar`, call `super()` in the constructor with a calendar identifier, and implement all the methods.
 Any subclass of `Temporal.Calendar` will be accepted in Temporal APIs where a built-in `Temporal.Calendar` would work.
 
+The identifier of a custom calendar must not be any of the built-in calendar identifiers, and must consist of one or more components of between 2 and 8 ASCII alphanumeric characters each, separated by dashes, as described in [BCP 47](https://tools.ietf.org/html/bcp47#section-2.1).
```
sffc, line 44: Unicode extension value subtags are 3-8, not 2-8

https://unicode.org/reports/tr35/tr35.html#Unicode_locale_identifier

sffc, line 44: The exact syntax is likely to depend on the outcome of the RFC 3339 discussions.

sffc, line 44: Why is it explicitly illegal for a custom calendar to have the same ID as a built-in calendar?

ptomato: Ah, so https://tools.ietf.org/html/bcp47#section-2.1 doesn't contain the full information? Will change this.

ptomato: Now that I think about it, I guess we do want to allow this, for overriding the built-in ones.

ptomato: I'll add an editorial note to that effect.

ptomato: (Added in IsValidCalendarId in the spec)

sffc (approved):

docs/calendar.md
sffc: Yeah; in ECMA-402 we use the UTS 35 version of locale identifiers, which is a subset of IETF.

ptomato:
Thanks for the review!

PR #1012, opened 2020-10-17
Change getEpochSeconds() and friends to getters

ptomato:
Closes: #750


docs/instant.md
```diff
@@ -178,52 +178,46 @@ sorted.join(' ');
...
+instant.epochSeconds; // => 1554000300
 ```
 
-### instant.**getEpochMilliseconds**() : number
-
-**Returns:** an integer number of milliseconds.
+### instant.**epochMilliseconds** : number
 
-Same as `getEpochSeconds()`, but with millisecond (10<sup>&minus;3</sup> second) precision.
+Same as `epochSeconds`, but with millisecond (10<sup>&minus;3</sup> second) precision.
```
justingrant, line 200: I think it'd be helpful to repeat the content about rounding towards zero here and in the property below. This is implied from "same as `epochSeconds`" but the rounding behavior is an important detail to understand so it's worth repeating. 

Also, minor nit: shouldn't we call it "truncated towards zero" instead of "rounded towards zero" to be more clear about what it does?  To me "rounding" implies that the larger unit might sometimes change while "truncation" forecloses that possibility.

justingrant (approved):

Ms2ger (approved):

Issue #666, opened 2020-06-11
Fields operation on calendar


ptomato:
Follow up from #638. The calendar needs a way to say which fields are relevant for a particular operation. A `getFields()` method was previously suggested in https://github.com/tc39/proposal-temporal/pull/520#discussion_r412520319.

> I wonder if this should be a method on the calendar, or maybe we should get the list of fields from the calendar and then *remove* 'day'.

_Originally posted by @sffc in https://github.com/tc39/proposal-temporal/pull/638_

One way to do it would be to pass a list of fields into the method and the calendar would return a modified list of fields. E.g.
```js
JapaneseCalendar.prototype.getFields = function(fields) {
  if ('year' in fields)
    fields.push('era')
  return fields;
}
```
(only without modifying the original array)

PR #1014, opened 2020-10-19
Calendar.fields method

ptomato:
Closes: #666

### DIFF
--- a/docs/calendar.md
+++ b/docs/calendar.md
@@ -293,6 +293,30 @@ Temporal.Calendar.from('chinese').dateDifference(
 )  // => P1M2D
 ```
 
+### calendar.**fields**(fields: array<string>) : array<string>
+
+**Parameters:**
+
+- `fields` (array of strings): A list of field names.
+
+**Returns:** a new list of field names.
+
+This method does not need to be called directly except in specialized code.
+It is called indirectly when using the `from()` static methods and `with()` methods of `Temporal.DateTime`, `Temporal.Date`, and `Temporal.YearMonth`.
+
+Custom calendars should override this method if they require more fields with which to denote the date than the standard `era`, `year`, `month`, and `day`.
+The input array contains the field names that are necessary for a particular operation (for example, `'month'` and `'day'` for `Temporal.MonthDay.prototype.with()`), and the method should make a copy of the array and add whichever extra fields are necessary.
+
+The default implementation of this method returns a copy of `fields`.
+
+Usage example:
+
+```js
+// In built-in calendars, this method just makes a copy of the input array
+Temporal.Calendar.from('iso8601').fields(['month', 'day']);
+  // => ['month', 'day']
+```
+
 ### calendar.**toString**() : string
 
 **Returns:** The string given by `calendar.id`.
--- a/polyfill/lib/date.mjs
+++ b/polyfill/lib/date.mjs
@@ -124,11 +124,12 @@ export class Date {
       calendar = GetSlot(this, CALENDAR);
       source = this;
     }
-    const props = ES.ToPartialRecord(temporalDateLike, ['day', 'era', 'month', 'year']);
+    const fieldNames = ES.CalendarFields(calendar, ['day', 'era', 'month', 'year']);
+    const props = ES.ToPartialRecord(temporalDateLike, fieldNames);
     if (!props) {
       throw new TypeError('invalid date-like');
     }
-    const fields = ES.ToTemporalDateRecord(source);
+    const fields = ES.ToTemporalDateFields(source, fieldNames);
     ObjectAssign(fields, props);
     const Construct = ES.SpeciesConstructor(this, Date);
     const result = calendar.dateFromFields(fields, options, Construct);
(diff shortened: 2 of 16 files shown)


ptomato:
Sorry for the noise. This seems to mysteriously fail the test suite with no message (even though all the tests pass). I will investigate this before re-requesting review.

ptomato:
demitasse seems to swallow exceptions in `after()` blocks, they still fail the test suite, but print no message. I've fixed the exception now, it's ready for review again.

@pipobscure I don't have time right now to investigate whether this still happens in demitasse 2.x, so I'm inclined to leave it, but if you prefer I can open an issue in demitasse so it doesn't get lost.

docs/calendar.md
```diff
@@ -291,6 +291,28 @@ Temporal.Calendar.from('chinese').dateDifference(
...
+**Parameters:**
+
+- `fields` (array of strings): A list of field names.
+
+**Returns:** a new list of field names.
+
+This method does not need to be called directly except in specialized code.
+It is called indirectly when using the `from()` static methods and `with()` methods of `Temporal.DateTime`, `Temporal.Date`, and `Temporal.YearMonth`.
+
+Custom calendars should override this method if they require more fields with which to denote the date than the standard `era`, `year`, `month`, and `day`.
```
sffc, line 305: Shouldn't the Japanese calendar add `"era"`?  The ISO-8601 calendar has no concept of an era.

pipobscure:
> @pipobscure I don't have time right now to investigate whether this still happens in demitasse 2.x, so I'm inclined to leave it, but if you prefer I can open an issue in demitasse so it doesn't get lost.


Please do. I have to go over that in a few weeks; so I’ll certainly have forgotten this by then.

docs/calendar.md
ptomato: I'll be addressing that in #901.

polyfill/lib/date.mjs
```diff
@@ -120,11 +120,12 @@ export class Date {
       calendar = GetSlot(this, CALENDAR);
       source = this;
     }
-    const props = ES.ToPartialRecord(temporalDateLike, ['day', 'era', 'month', 'year']);
+    const fieldNames = calendar.fields(['day', 'era', 'month', 'year']);
+    const props = ES.ToPartialRecord(temporalDateLike, fieldNames);
```
Ms2ger, line 124: This is too trusting of the calendar - should make sure we've got a clean array of strings. (Potentially for followup.)

polyfill/lib/date.mjs
```diff
@@ -120,11 +120,12 @@ export class Date {
...
     if (!props) {
       throw new TypeError('invalid date-like');
     }
-    const fields = ES.ToTemporalDateRecord(source);
+    const fields = ES.ToTemporalDateFields(source, calendar);
```
Ms2ger, line 128: I think we shouldn't call .fields twice.

polyfill/lib/date.mjs
```diff
@@ -269,20 +270,21 @@ export class Date {
...
     return calendar.monthDayFromFields(fields, {}, MonthDay);
   }
   getFields() {
-    const fields = ES.ToTemporalDateRecord(this);
-    if (!fields) throw new TypeError('invalid receiver');
-    fields.calendar = GetSlot(this, CALENDAR);
+    if (!ES.IsTemporalDate(this)) throw new TypeError('invalid receiver');
+    const calendar = GetSlot(this, CALENDAR);
+    const fields = ES.ToTemporalDateFields(this, calendar);
+    fields.calendar = calendar;
```
Ms2ger, line 287: What if fields() returns "calendar" as one of the fields?

polyfill/lib/ecmascript.mjs
```diff
@@ -2237,6 +2276,47 @@ export const ES = ObjectAssign({}, ES2020, {
       throw new RangeError(`${property} must be between ${minimum} and ${maximum}, not ${value}`);
     }
     return MathFloor(value);
+  },
+  // Overridden because the es-abstract version unconditionally uses util.inspect
+  Get: (O, P) => {
+    if (Type(O) !== 'Object') {
```
Ms2ger, line 2282: More consistent to use `ES.Type()`?

polyfill/lib/ecmascript.mjs
```diff
@@ -2237,6 +2276,47 @@ export const ES = ObjectAssign({}, ES2020, {
...
+  },
+  LengthOfArrayLike: (obj) => {
+    if (Type(obj) !== 'Object') {
+      throw new TypeError('Assertion failed: `obj` must be an Object');
+    }
+    return ES.ToLength(ES.Get(obj, 'length'));
+  },
+  CreateListFromArrayLike: (
+    obj,
+    elementTypes = ['Undefined', 'Null', 'Boolean', 'String', 'Symbol', 'Number', 'Object']
```
Ms2ger, line 2298: I wouldn't mind specializing to only String for the polyfill.

spec/date.html
```diff
@@ -708,76 +712,46 @@ <h1>DifferenceDate ( _y1_, _m1_, _d1_, _y2_, _m2_, _d2_, _largestUnit_ )</h1>
...
+        1. Let _result_ be ? OrdinaryObjectCreate(%Object.prototype%).
+        1. For each value _property_ of _fields_, do
           1. Let _value_ be ? Get(_temporalDateLike_, _property_).
-          1. If _value_ is *undefined*, then
+          1. If _property_ is one of *"day"*, *"month"*, or *"year"*, and _value_ is *undefined*, then
             1. Throw a *TypeError* exception.
-          1. Let _value_ be ? ToInteger(_value_).
-          1. Set _result_'s internal slot whose name is the Internal Slot value of the current row to _value_.
+          1. If _property_ is not *"era"*, then
+            1. Let _value_ be ? ToInteger(_value_).
```
Ms2ger, line 728: This makes no sense to me: this would convert any custom property to an integer?

Ms2ger (changes requested):

polyfill/lib/ecmascript.mjs
ptomato: I just removed the default argument since it's not called anywhere in the polyfill without the second argument.

polyfill/lib/date.mjs
ptomato: I think fixing the next comment below (https://github.com/tc39/proposal-temporal/pull/1014#discussion_r509278096) will be less work if I also fix this, so I won't defer it to a followup. To be clear, I assume you are suggesting moving this into an abstract operation named something like CalendarFields, which does some sort of validation on the returned value?

polyfill/lib/date.mjs
ptomato: I'm not sure what should be done in that case; I guess one option is to throw if `calendar` is returned in the validation step I mentioned in https://github.com/tc39/proposal-temporal/pull/1014#discussion_r510403224. The other option is to :shrug: and say "well, don't do that then." I'm mildly in favour of the latter since I think it's in line with what we do elsewhere for custom calendars and time zones.

spec/date.html
ptomato: That is what I intended. The alternative is to treat any custom property just like `era` and leave it unchanged. My reasoning was that we wouldn't want to encourage more non-numerical properties like era, but I don't have a strong opinion on which is correct.

Ms2ger (approved):

spec/date.html
sffc: We shouldn't assert a particular type on custom properties, except *perhaps* requiring that it is a primitive.

Ms2ger: I don't have a clear idea of the use case of this feature, but this seems to make it impossible to use non-integer properties, rather than just discourage. I'll merge now in the interest of moving forward, but filed #1045.

polyfill/lib/date.mjs
ptomato: → #1047

PR #1016, opened 2020-10-19
Calendar.from should take ISO string

ptomato:
Closes: #862


sffc (approved):

ptomato:
Thanks for the quick review!

PR #1017, opened 2020-10-19
Allow property bags in TimeZone.from and Calendar.from

ptomato:
Closes: #925

### DIFF
--- a/polyfill/lib/calendar.mjs
+++ b/polyfill/lib/calendar.mjs
@@ -117,7 +117,11 @@ export class Calendar {
     return GetSlot(this, CALENDAR_ID);
   }
   static from(item) {
-    if (ES.Type(item) === 'Object') return item;
+    if (ES.Type(item) === 'Object') {
+      if (!('calendar' in item)) return item;
+      item = item.calendar;
+      if (ES.Type(item) === 'Object' && !('calendar' in item)) return item;
+    }
     const stringIdent = ES.ToString(item);
     if (IsBuiltinCalendar(stringIdent)) return GetBuiltinCalendar(stringIdent);
     let calendar;
--- a/polyfill/lib/timezone.mjs
+++ b/polyfill/lib/timezone.mjs
@@ -236,7 +236,11 @@ export class TimeZone {
     return ES.TimeZoneToString(this);
   }
   static from(item) {
-    if (ES.Type(item) === 'Object') return item;
+    if (ES.Type(item) === 'Object') {
+      if (!('timeZone' in item)) return item;
+      item = item.timeZone;
+      if (ES.Type(item) === 'Object' && !('timeZone' in item)) return item;
+    }
     const timeZone = ES.TemporalTimeZoneFromString(ES.ToString(item));
     const result = new this(timeZone);
     if (!ES.IsTemporalTimeZone(result)) throw new TypeError('invalid result');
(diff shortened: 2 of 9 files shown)


polyfill/lib/calendar.mjs
```diff
@@ -117,7 +117,11 @@ export class Calendar {
     return GetSlot(this, CALENDAR_ID);
   }
   static from(item) {
-    if (ES.Type(item) === 'Object') return item;
+    if (ES.Type(item) === 'Object') {
+      if (!('calendar' in item)) return item;
```
Ms2ger, line 121: Prefer getting the property once and checking against `undefined`.

polyfill/lib/calendar.mjs
```diff
@@ -117,7 +117,11 @@ export class Calendar {
...
+      item = item.calendar;
+      if (ES.Type(item) === 'Object' && !('calendar' in item)) return item;
```
Ms2ger, line 123: What's the use case for a calendar property that's not a calendar object yet?

polyfill/lib/timezone.mjs
```diff
@@ -236,7 +236,11 @@ export class TimeZone {
     return ES.TimeZoneToString(this);
   }
   static from(item) {
-    if (ES.Type(item) === 'Object') return item;
+    if (ES.Type(item) === 'Object') {
+      if (!('timeZone' in item)) return item;
+      item = item.timeZone;
+      if (ES.Type(item) === 'Object' && !('timeZone' in item)) return item;
```
Ms2ger, line 242: Ditto

Ms2ger (approved):
At this point, I'm wondering if we should just require subclassing `Temporal.Calendar`/`Temporal.TimeZone` for custom implementations. What's the value of allowing plain objects, and does it justify this?

ptomato:
> At this point, I'm wondering if we should just require subclassing `Temporal.Calendar`/`Temporal.TimeZone` for custom implementations. What's the value of allowing plain objects, and does it justify this?

I had this same question, but Richard answered it in the last Temporal meeting. Calendars and TimeZones from other realms will be considered plain objects, so that's the motivation for this functionality.

polyfill/lib/calendar.mjs
ptomato: I think that considering an object with `calendar: undefined` to signify a custom calendar would work, but it is different from the semantic that we landed on, where it's the shape of the object that matters: if a `calendar` property is present then it's a property bag, if absent then it's a custom calendar.

polyfill/lib/calendar.mjs
ptomato: The calendar could be given as a string in the property bag:
```js
const myPropertyBag = { year: 2020, month: 10, day: 20, calendar: 'gregory' };
Temporal.Calendar.from(myPropertyBag);
```

ljharb:
Not if they inherit from a base TimeZone class that has an internal slot.

ptomato:
As far as I understand, Richard felt that a brand check was not correct to use in this case - @gibson042 can you clarify?

ljharb:
Subclassing (with one weird exception in RegExp) generally always uses a brand check. The `Set.prototype.size` getter, for example, can be `.call`ed on Set instances from other realms, because it checks the internal slot. Similarly, `[arr].flat()` where `arr` is an array with a single item, will result in an array with that single item, as long as `arr` passes `IsArray` in the spec, which detects cross-realm arrays as well.

Ms2ger (approved):
I think we should reconsider plain objects, but that's for another bug.

PR #1029, opened 2020-10-21
Fix regex pattern for timezone offset parsing

akshgpt7:
Fixes #1002


Ms2ger (commented):
I think we want to keep support for the fancy unicode minus sign. Reordering to put the plain hyphen-minus last will probably fix the original bug.

akshgpt7:
The `yearpart` and `duration` patterns were also using the same sequence to match `+` and `-`.

Ms2ger (approved):

PR #1035, opened 2020-10-22
Sub-minute time zone offsets

ptomato:
Closes: #935

### DIFF
--- a/polyfill/test/timezone.mjs
+++ b/polyfill/test/timezone.mjs
@@ -50,6 +50,10 @@ describe('TimeZone', () => {
     test('\u221201:00');
     test('\u22120650');
     test('\u221208');
+    test('+01:00:00');
+    test('-010000');
+    test('+03:30:00.000000001');
+    test('-033000.1');
     test('Europe/Vienna');
     test('America/New_York');
     test('Africa/CAIRO'); // capitalization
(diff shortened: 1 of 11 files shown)


polyfill/test/timezone.mjs
```diff
@@ -197,6 +208,18 @@ describe('TimeZone', () => {
       }
     });
   });
+  describe('sub-minute offset', () => {
+    const zone = new Temporal.TimeZone('Europe/Amsterdam');
+    const inst = Temporal.Instant.from('1900-01-01T12:00Z');
+    const dtm = Temporal.DateTime.from('1900-01-01T12:00');
+    it(`${zone} has ID ${zone}`, () => equal(zone.id, `${zone}`));
```
ryzokuken, line 215: A feel a bit iffy about comparing two computed values, one of these should ideally be a constant, otherwise they could _both_ be the wrong yet the same and it would pass. That said, it's a very minor nit.

ryzokuken (approved):

polyfill/test/timezone.mjs
ptomato: This was already the case in the other tests surrounding this one, but maybe it's worth a follow up issue if you want to open one?

PR #1042, opened 2020-10-22
Document where the time zone comes from in Instant.toLocaleString

ptomato:
With a couple of incidental bug fixes.

Closes: #1021

### DIFF
--- a/docs/instant.md
+++ b/docs/instant.md
@@ -649,10 +649,8 @@ Example usage:
 instant = Temporal.Instant.from('2019-11-18T11:00:00.000Z');
 instant.toLocaleString(); // => example output: 2019-11-18, 3:00:00 a.m.
 instant.toLocaleString('de-DE'); // => example output: 18.11.2019, 03:00:00
-instant.toLocaleString('de-DE', {
-  timeZone: 'Europe/Berlin',
-  weekday: 'long'
-}); // => Montag, 18.11.2019, 12:00:00
+instant.toLocaleString('de-DE', { timeZone: 'Europe/Berlin' });
+  // => 18.11.2019, 12:00:00
 instant.toLocaleString('en-US-u-nu-fullwide-hc-h12', {
   timeZone: 'Asia/Kolkata'
 }); // => １１/１８/２０１９, ４:３０:００ PM
--- a/docs/instant.md
+++ b/docs/instant.md
@@ -637,6 +637,12 @@ This method overrides `Object.prototype.toLocaleString()` to provide a human-rea
 
 The `locales` and `options` arguments are the same as in the constructor to [`Intl.DateTimeFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DateTimeFormat).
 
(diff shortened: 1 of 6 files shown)


docs/instant.md
```diff
@@ -637,16 +637,18 @@ This method overrides `Object.prototype.toLocaleString()` to provide a human-rea
 
 The `locales` and `options` arguments are the same as in the constructor to [`Intl.DateTimeFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DateTimeFormat).
 
+Since `Temporal.Instant` does not carry a time zone, the time zone used for the output will be the `timeZone` property of `options`, if present; and otherwise, the current time zone from the environment.
```
justingrant, line 640: Minor nit for a time-focused library: use "Because" not "Since".

> current time zone from the environment

I'd recommend a little more detail, e.g. current time zone from the environment, which is usually the user's or system's default time zone.  NOTE: when running Temporal code in a server environment, the server's time zone (which is usually set to `UTC`) won't match the user's time zone. So using the environment default should be avoided in that case.

ryzokuken: I don't think the last clarification is necessary since `toLocaleString` should only really appear in client-side code anyway, right?

ryzokuken (approved):

docs/instant.md
justingrant: If you're building a web app and generating HTML on the server, you'll need some way to get localized data.

justingrant (approved):

docs/instant.md
ryzokuken: You're right, it would be unfortunately if people used this on SSR-ed webapps.

docs/instant.md
```diff
@@ -637,16 +637,18 @@ This method overrides `Object.prototype.toLocaleString()` to provide a human-rea
...
 
 ```js
 instant = Temporal.Instant.from('2019-11-18T11:00:00.000Z');
 instant.toLocaleString(); // => example output: 2019-11-18, 3:00:00 a.m.
 instant.toLocaleString('de-DE'); // => example output: 18.11.2019, 03:00:00
-instant.toLocaleString('de-DE', {
-  timeZone: 'Europe/Berlin',
-  weekday: 'long'
-}); // => Montag, 18.11.2019, 12:00:00
+instant.toLocaleString('de-DE', { timeZone: 'Europe/Berlin' });
```
sffc, line 650: It would be helpful for an example to show output with `timeZoneName` enabled.

sffc, line 650: Should we make `timeZoneName` default to be enabled when formatting an Instant?

ptomato: That's really an ECMA-402 question, so I'll defer to you on that; note that legacy Date doesn't have it enabled by default.

ptomato: Good call, because that was broken, so I fixed the bug as well.

ptomato: All the other comments are addressed so if you'd like me to follow up on that, a new issue would be best.

PR #1043, opened 2020-10-22
Time zone and calendar protocol best practices

ptomato:
Closes: #1040

### DIFF
--- a/docs/calendar.md
+++ b/docs/calendar.md
@@ -46,7 +46,7 @@ The identifier of a custom calendar must consist of one or more components of be
 ### Protocol
 
 It's also possible for a plain object to be a custom calendar, without subclassing.
-The object must implement the `Temporal.Calendar` methods and have an `id` property.
+The object must implement all of the `Temporal.Calendar` methods except for `fields()`.
 It must not have a `calendar` property, so that it can be distinguished in `Temporal.Calendar.from()` from other Temporal objects that have a calendar.
 It is possible to pass such an object into any Temporal API that would normally take a built-in `Temporal.Calendar`.
 
--- a/docs/calendar.md
+++ b/docs/calendar.md
@@ -37,18 +37,20 @@ date.withCalendar(calendar).add({ months: 1 })
 
 ### Custom calendars
 
-For specialized applications where you need to do calculations in a calendar system that is not supported by Intl, you can also implement your own `Temporal.Calendar` object.
-To do this, create a class inheriting from `Temporal.Calendar`, call `super()` in the constructor with a calendar identifier, and implement all the methods.
-Any subclass of `Temporal.Calendar` will be accepted in Temporal APIs where a built-in `Temporal.Calendar` would work.
+For specialized applications where you need to do calculations in a calendar system that is not supported by Intl, you can implement a custom calendar.
+There are two ways to do this.
+
+The recommended way is to create a class inheriting from `Temporal.Calendar`, call `super()` in the constructor with a calendar identifier string, and implement all the members except `id`, `toString()`, and `fields()`, which are optional.
+If you don't implement the optional members, then the base class's default implementations will be used.
+
+The other, more difficult, way to create a custom calendar is to create a plain object implementing the `Temporal.Calendar` protocol, without subclassing.
+The object must implement all of the `Temporal.Calendar` methods except for `fields()`.
+Any object with the required methods will return the correct output from any Temporal property or method.
+However, most other code will assume that custom calendars act like built-in `Temporal.Calendar` objects.
+To interoperate with libraries or other code that you didn't write, then you should implement the `id` property and the `fields()` method as well.
+Your object must not have a `calendar` property, so that it can be distinguished in `Temporal.Calendar.from()` from other Temporal objects that have a calendar.
 
 The identifier of a custom calendar must consist of one or more components of between 3 and 8 ASCII alphanumeric characters each, separated by dashes, as described in [Unicode Technical Standard 35](https://unicode.org/reports/tr35/tr35.html#Unicode_locale_identifier).
-
-### Protocol
-
-It's also possible for a plain object to be a custom calendar, without subclassing.
-The object must implement all of the `Temporal.Calendar` methods except for `fields()`.
-It must not have a `calendar` property, so that it can be distinguished in `Temporal.Calendar.from()` from other Temporal objects that have a calendar.
-It is possible to pass such an object into any Temporal API that would normally take a built-in `Temporal.Calendar`.
 
 ## Constructor
 
@@ -120,6 +122,8 @@ cal = Temporal.Calendar.from({id: 'mycalendar'});
 
 The `id` property gives an unambiguous identifier for the calendar.
 Effectively, this is whatever `calendarIdentifier` was passed as a parameter to the constructor.
+
+When subclassing `Temporal.Calendar`, this property doesn't need to be overridden because the default implementation gives the result of calling `toString()`.
 
 ## Methods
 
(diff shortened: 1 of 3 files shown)


docs/timezone.md
```diff
@@ -27,6 +27,12 @@ The object must have `getOffsetNanosecondsFor()`, `getPossibleInstantsFor()`, an
 It must not have a `timeZone` property, so that it can be distinguished in `Temporal.TimeZone.from()` from other Temporal objects that have a time zone.
 It is possible to pass such an object into any Temporal API that would normally take a built-in `Temporal.TimeZone`.
 
+Any implementation of the protocol with those three methods will return the correct output from any Temporal property or method.
```
justingrant, line 30: I'm confused about the content above what's changed in the PR. Why are required methods different between the subclass implementation and the custom-object implementation?  

Also, the TS types also list `getDateTimeFor` as required.  Which is correct?

docs/timezone.md
```diff
@@ -127,6 +133,8 @@ tz2 = Temporal.TimeZone.from(tz);
 The `id` property gives an unambiguous identifier for the time zone.
 Effectively, this is the canonicalized version of whatever `timeZoneIdentifier` was passed as a parameter to the constructor.
 
+The default implementation of this property gives the result of calling `toString()`.
```
justingrant, line 136: Minor concern: will talking about the default implementations inside each method's docs be confusing for the vast majority of developers who will never implement a custom time zone? 

Most developer documentation separates the "plugin authors" part of the docs from the "plugin users" part of the docs because there are 100x+ more of the latter and including info that only the former will use may be confusing. So IMHO, instead of combining custom time zone info into the "users" page, it seems better to have a separate page that's dedicated to people creating custom zones or calendars. This would simplify the docs for most users while giving the relevant info for the minority of advanced users who need it. 

I'd volunteer to do this split in a separate PR if you think this would be worthwhile-- I know you're super-busy on lots of things that I can't help with, but I can help with docs! ;-)

docs/timezone.md
ptomato: The subclass allows you to skip implementing both `id` and `toString()` by virtue of letting you set the [[Identifier]] internal slot, by calling `super(id)` in the constructor. But note that might go away depending on the outcome of #780 

You're right though, `getNextTransition()` and `getPreviousTransition()` should be documented to be optional in the subclass case as well. I'll fix this up tomorrow.

docs/timezone.md
ptomato: For the same reason I thought that the text you had earlier was too long, I don't think that we even want to extensively document the plain object time zones, definitely not on a separate page that makes them into A Thing that people might think they should do. Rather I'd consider them just a peculiarity of how protocols work in JS.

If this wording might be confusing, how about adding "When subclassing Temporal.TimeZone," to each of these sentences?

docs/timezone.md
```diff
@@ -27,6 +27,12 @@ The object must have `getOffsetNanosecondsFor()`, `getPossibleInstantsFor()`, an
 It must not have a `timeZone` property, so that it can be distinguished in `Temporal.TimeZone.from()` from other Temporal objects that have a time zone.
 It is possible to pass such an object into any Temporal API that would normally take a built-in `Temporal.TimeZone`.
 
+Any implementation of the protocol with those three methods will return the correct output from any Temporal property or method.
+However, most other code will assume that custom time zones act like built-in `Temporal.TimeZone` objects.
+If you want to interoperate with libraries or other code that you didn't write, then you should implement all the other `Temporal.TimeZone` methods and properties described below as well: `id`, `getOffsetStringFor()`, `getDateTimeFor()`, `getZonedDateTimeFor()`, `getInstantFor()`, `getNextTransition()`, `getPreviousTransition()`, and `toJSON()`.
+
+Creating a class that inherits from the built-in `Temporal.TimeZone` and overriding the methods you need to override, instead of using a plain object that implements the time zone protocol, is by far the easiest (and recommended) way to achieve this.
```
justingrant, line 34: We could probably simplify (and maybe shrink) the entire custom timezones section by re-ordering the content so that you explain which approach is easier at the top, instead of at the bottom.  Here's some suggested content to show what I mean.

## Custom Time Zones

For specialized applications where you need to do calculations in a time zone that is not built into Temporal, you can implement a custom time zone.
There are two ways to do this.

The easiest way is to create a class inheriting from `Temporal.TimeZone`, call `super()` in the constructor with a time zone identifier string, and implement  `getOffsetNanosecondsFor()` and `getPossibleInstantsFor()`.
You can optionally also implement `getNextTransition()` and `getPreviousTransition()`, or leave the default implementation which always returns `null` which means that no next or previous transitions are known.
You don't need to implement any other methods because other properties and methods will be automatically implemented via the base class calling the methods that you override.

The harder way to create a custom time zone is to create a plain object without subclassing.
The object must at least implement `getOffsetNanosecondsFor()`, `getPossibleInstantsFor()`, and `toString()`.
An object with those three methods will return correct output from any Temporal property or method.
However, most other code will assume that custom time zones act like built-in `Temporal.TimeZone` objects.
To interoperate with libraries or other code that you didn't write, then you should implement all other `Temporal.TimeZone` members as well: `id`, `getOffsetStringFor()`, `getDateTimeFor()`, `getZonedDateTimeFor()`, `getInstantFor()`, `getNextTransition()`, `getPreviousTransition()`, and `toJSON()`.
Your object must not have a `timeZone` property, so that `Temporal.TimeZone.from()` can distinguish your custom time zone from named-parameter input objects like `{ timeZone: 'Asia/Tokyo' }`.

The identifier of a custom time zone must consist of one or more components separated by slashes (`/`), as described in the [tzdata documentation](https://htmlpreview.github.io/?https://github.com/eggert/tz/blob/master/theory.html#naming).
Each component must consist of between one and 14 characters.
Valid characters are ASCII letters, `.`, `-`, and `_`.
`-` may not appear as the first character of a component, and a component may not be a single dot `.` or two dots `..`.

docs/timezone.md
justingrant: OK I'm convinced. 



> If this wording might be confusing, how about adding "When subclassing Temporal.TimeZone," to each of these sentences?

This seems like a great solution. Thanks!

justingrant: > "When subclassing Temporal.TimeZone,"

It might be good to clarify that the methods with this message don't need to be overridden, for example: 
* "When subclassing Temporal.TimeZone, this method does not need to be overridden because the default implementation..."
* (for the getXxxTransition methods) "When subclassing Temporal.TimeZone, this method should be overridden if the time zone changes offsets. Single-offset time zones can use the default implementation which returns `null`."

justingrant (approved):

docs/calendar.md
```diff
@@ -46,10 +46,16 @@ The identifier of a custom calendar must consist of one or more components of be
...
 
 It's also possible for a plain object to be a custom calendar, without subclassing.
-The object must implement the `Temporal.Calendar` methods and have an `id` property.
+The object must implement all of the `Temporal.Calendar` methods except for `fields()`.
 It must not have a `calendar` property, so that it can be distinguished in `Temporal.Calendar.from()` from other Temporal objects that have a calendar.
 It is possible to pass such an object into any Temporal API that would normally take a built-in `Temporal.Calendar`.
 
+Any implementation of the protocol with the required methods will return the correct output from any Temporal property or method.
+However, most other code will assume that custom calendars act like built-in `Temporal.Calendar` objects.
+If you want to interoperate with libraries or other code that you didn't write, then you should implement the `id` property and the `fields()` method as well.
```
sffc, line 55: I don't get why `id` and `fields()` are optional.  What happens if Temporal gets a calendar without those fields and some method that needs them is called (like toString for `.id` or getFields() for fields()?)

Ms2ger: Note that `id` defers to `toString`, not the other way around.

ptomato: If `fields` isn't present on the object, then %Temporal.Calendar.prototype.fields% is called, which just clones the input array. It's optional to override `fields` if you subclass, so it should be optional to provide it on a plain-object calendar. This is consistent with what we do for the TimeZone methods that are optional to override when you subclass TimeZone.

docs/timezone.md
ptomato: I've adopted a slightly edited version of this.

Ms2ger (approved):

PR #1044, opened 2020-10-23
Partial spec for ZonedDateTime

ptomato:
This has all the entries for ZonedDateTime, although many of them are marked TODO. Includes a few minor other fixes.

See: #569

### DIFF
--- a/docs/zoneddatetime.md
+++ b/docs/zoneddatetime.md
@@ -1218,9 +1218,9 @@ zdt.toMonthDay(); // => 12-07
 zdt.toTime(); // => 03:24:30
 ```
 
-### zonedDateTime.**getFields**() : { year: number, month: number, day: number, hour: number, minute: number, second: number, millisecond: number, microsecond: number, nanosecond: number, calendar: object, [propName: string]: unknown }
+### zonedDateTime.**getFields**() : { year: number, month: number, day: number, hour: number, minute: number, second: number, millisecond: number, microsecond: number, nanosecond: number, offset: string, timeZone: object, calendar: object, [propName: string]: unknown }
 
-**Returns:** a plain object with properties equal to the fields of `zonedDateTime`, including all date/time fields (expressed in the current calendar) as well as the `calendar`, `timeZone`, and `offsetNanoseconds` properties.
+**Returns:** a plain object with properties equal to the fields of `zonedDateTime`, including all date/time fields (expressed in the current calendar) as well as the `calendar`, `timeZone`, and `offset` properties.
 
 This method can be used to convert a `Temporal.ZonedDateTime` into a record-like data structure.
 It returns a new plain JavaScript object, with all the fields as enumerable, writable, own data properties.
@@ -1263,9 +1263,9 @@ JSON.stringify(thisWillWork, undefined, 2);
 // }"
 ```
 
-### zonedDateTime.**getISOFields**(): { isoYear: number, isoMonth: number, isoDay: number, hour: number, minute: number, second: number, millisecond: number, microsecond: number, nanosecond: number, calendar: object }
+### zonedDateTime.**getISOFields**(): { isoYear: number, isoMonth: number, isoDay: number, hour: number, minute: number, second: number, millisecond: number, microsecond: number, nanosecond: number, offset: string, timeZone: object, calendar: object }
 
-**Returns:** a plain object with properties expressing `zonedDateTime` in the ISO 8601 calendar, including all date/time fields as well as the `calendar`, `timeZone`, and `offsetNanoseconds` properties.
+**Returns:** a plain object with properties expressing `zonedDateTime` in the ISO 8601 calendar, including all date/time fields as well as the `calendar`, `timeZone`, and `offset` properties.
 Note that date/time properties have different names with an `iso` prefix to better differentiate from "normal" `getFields` results.
 
 This is an advanced method that's mainly useful if you are implementing a custom calendar.
--- a/docs/zoneddatetime.md
+++ b/docs/zoneddatetime.md
@@ -28,7 +28,7 @@ The `Temporal.ZonedDateTime` API is a superset of `Temporal.DateTime`, which mak
 
 ## Constructor
 
-### **new Temporal.ZonedDateTime**(_epochNanoseconds_: bigint, _timeZone_: string | object, _calendar_: string | object) : Temporal.ZonedDateTime
+### **new Temporal.ZonedDateTime**(_epochNanoseconds_: bigint, _timeZone_: string | object, _calendar_?: string | object) : Temporal.ZonedDateTime
 
 **Parameters:**
 
--- a/spec/zoneddatetime.html
+++ b/spec/zoneddatetime.html
@@ -1,0 +1,1048 @@
+<!doctype html>
+<meta charset="utf8">
+
+<emu-clause id="sec-temporal-zoneddatetime-objects">
+  <h1>Temporal.ZonedDateTime Objects</h1>
+  <p>
+    A Temporal.ZonedDateTime object is an immutable Object referencing a point in time with nanoseconds precision, and containing Object values corresponding to a particular time zone and calendar system.
+  </p>
+
+  <emu-clause id="sec-temporal-zoneddatetime-constructor">
+    <h1>The Temporal.ZonedDateTime Constructor</h1>
+    <p>
+      The Temporal.ZonedDateTime constructor is the %Temporal.ZonedDateTime% intrinsic object.
+      When called as a constructor, it creates and initializes a new Temporal.ZonedDateTime object.
+    </p>
+    <p>
+      The Temporal.ZonedDateTime constructor is designed to be subclassable.
+      It may be used as the value of an extends clause of a class definition.
+      Subclass constructors that intend to inherit the specified ZonedDateTime behaviour must include a super call to the %Temporal.ZonedDateTime% constructor to create and initialize subclass instances with the necessary internal slots.
+    </p>
+
+    <emu-clause id="sec-temporal.zoneddatetime">
+      <h1>Temporal.ZonedDateTime ( _epochNanoseconds_, _timeZoneLike_ [ , _calendarLike_ ] )</h1>
+      <p>
+        When the `Temporal.ZonedDateTime` function is called, the following steps are taken:
+      </p>
+      <emu-alg>
+        1. If NewTarget is *undefined*, then
+          1. Throw a *TypeError* exception.
+        1. Set _epochNanoseconds_ to ? ToBigInt(_epochNanoseconds_).
+        1. Perform ? RejectInstant(_epochNanoseconds_).
+        1. Let _timeZone_ be ? ToTemporalTimeZone(_timeZoneLike_).
+        1. If _calendarLike_ is *undefined*, then
+          1. Set _calendar_ to ? GetISO8601Calendar().
+        1. Else,
+          1. Set _calendar_ to ? ToTemporalCalendar(_calendarLike_).
+        1. Return ? CreateTemporalZonedDateTime(_epochNanoseconds_, _timeZone_, _calendar_, NewTarget).
+      </emu-alg>
+    </emu-clause>
+  </emu-clause>
+
+  <emu-clause id="sec-properties-of-the-temporal-zoneddatetime-constructor">
+    <h1>Properties of the Temporal.ZonedDateTime Constructor</h1>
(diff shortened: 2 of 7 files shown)


spec/zoneddatetime.html
```diff
@@ -0,0 +1,1049 @@
+<!doctype html>
+<meta charset="utf8">
+
+<emu-clause id="sec-temporal-zoneddatetime-objects">
+  <h1>Temporal.ZonedDateTime Objects</h1>
+  <p>
+    A Temporal.ZonedDateTime object is an immutable Object referencing a point in time with nanoseconds precision, and containing a String value corresponding to a particular time zone.
```
justingrant, line 7: What about the calendar?

ptomato: Oh yeah, not sure where I got the "String" time zone either

spec/zoneddatetime.html
```diff
@@ -0,0 +1,1049 @@
...
+      <p>
+        `Temporal.ZonedDateTime.prototype.era` is an accessor property whose set accessor function is *undefined*.
+        Its get accessor function performs the following steps:
+      </p>
+      <emu-alg>
+        1. Let _zonedDateTime_ be the *this* value.
+        1. Perform ? RequireInternalSlot(_zonedDateTime_, [[InitializedTemporalZonedDateTime]]).
+        1. Let _timeZone_ be _zonedDateTime_.[[TimeZone]].
+        1. Let _instant_ be ? CreateTemporalInstant(_zonedDateTime_.[[Nanoseconds]]).
+        1. Let _temporalDateTime_ be ? GetTemporalDateTimeFor(_timeZone_, _instant_).
```
justingrant, line 162: General question: it's possible that re-building the DateTime for each property access could get quite expensive, esp in cases where you pull multiple props from the same instance which is common.  Seems like an implementation may want to cache the DateTime the first time it's generated.  Is that something that ECMAScript specs address?  Or is that an implementation detail that the implementers worry about but the spec doesn't?

spec/zoneddatetime.html
```diff
@@ -0,0 +1,1049 @@
...
+      </p>
+      <emu-alg>
+        1. Let _zonedDateTime_ be the *this* value.
+        1. Perform ? RequireInternalSlot(_zonedDateTime_, [[InitializedTemporalZonedDateTime]]).
+        1. Return <mark>TODO</mark>.
+      </emu-alg>
+    </emu-clause>
+
+    <emu-clause id="sec-get-temporal.zoneddatetime.prototype.isoffsettransition">
+      <h1>get Temporal.ZonedDateTime.prototype.isOffsetTransition</h1>
```
justingrant, line 548: FYI, in #1019 I've proposed renaming this property to `atOffsetTransition`.

spec/zoneddatetime.html
```diff
@@ -0,0 +1,1049 @@
...
+    </emu-table>
+  </emu-clause>
+
+  <emu-clause id="sec-temporal-zoneddatetime-abstract-ops">
+    <h1>Abstract operations</h1>
+
+    <emu-clause id="sec-temporal-totemporalzoneddatetime" aoid="ToTemporalZonedDateTime">
+      <h1>ToTemporalZonedDateTime ( _item_ [ , _constructor_ [ , _overflow_, _disambiguation_, _offset_ ] ] )</h1>
+      <p>
+        The abstract operation ToTemporalZonedDateTime returns its argument _item_ if it is already a Temporal.ZonedDateTime instance, converts _item_ to a new Temporal.Date instance if possible, and throws otherwise.
```
justingrant, line 948: I thought that `from()` was always guaranteed to create a new instance, and to clone the instance if it's already an instance of the same type?

justingrant: Oh never mind. I think I understand: this AO is used for internal conversions, while a different AO (which does clone) is used for `from()`?

spec/zoneddatetime.html
```diff
@@ -0,0 +1,1049 @@
...
+
+    <emu-clause id="sec-temporal-temporalzoneddatetimetostring" aoid="TemporalZonedDateTimeToString">
+      <h1>TemporalZonedDateTimeToString ( _zonedDateTime_ )</h1>
+      <p>
+        The abstract operation TemporalZonedDateTimeToString returns an ISO 8601 string representation of its argument, including a time zone name annotation and calendar annotation, which are extensions to the ISO 8601 format.
+      </p>
+      <emu-alg>
+        1. Assert: Type(_zonedDateTime_) is Object and _zonedDateTime_ has an [[InitializedTemporalZonedDateTime]] internal slot.
+        1. Let _instant_ be ? CreateTemporalInstant(_zonedDateTime_.[[Nanoseconds]]).
+        1. Return ? TemporalInstantToString(_instant_, _zonedDateTime_.[[TimeZone]], _zonedDateTime_.[[Calendar]]).
```
justingrant, line 1045: Is this affected by the decision to remove the bracketed time zone from `Instant.prototype.toString()`?  Or is whether to show brackets or not determined by the presence of the calendar argument?

justingrant (approved):
Looks great. Your productivity is astounding! ;-)

spec/zoneddatetime.html
ptomato: That's right. In from() we check if the object is an instance, and clone and return it if so. If it's already not an instance, then we use this AO.

spec/zoneddatetime.html
ptomato: Oh right, this is not correct for the case of an offset time zone where the offset needs to be shown again in brackets. So it cannot share TemporalInstantToString after all. (And in #703 Instant is always shown with `Z` anyway)

spec/zoneddatetime.html
ptomato: I planned to find out later on what could be done to address this, so I will definitely follow up later, but my understanding is that it's not possible to cache the DateTime in the case of a custom time zone. For example, you could do this:
```
class MyTimeZone extends Temporal.TimeZone {
    constructor() { super('bad'); }
    getDateTimeFor() { return Temporal.DateTime.from('1970-01-01'); }
}
const tz = new MyTimeZone();
const zdt = new Temporal.ZonedDateTime(0n, tz);
zdt.year // => 1970
MyTimeZone.prototype.getDateTimeFor = function() { return Temporal.DateTime.from('2000-01-01'); }
zdt.year // => 2000
```
If we wrote in the spec that the DateTime had to be cached, then you could get into the situation where
the result of `zdt.year` is not equal to `zdt.timeZone.getDateTimeFor(new Instant(zdt.epochNanoseconds)).year`.

Implementations can cache the calls if the time zone is a built-in one because then the call to getDateTimeFor won't be observable.

docs/zoneddatetime.md
```diff
@@ -1200,7 +1200,7 @@ zdt.toMonthDay(); // => 12-07
 zdt.toTime(); // => 03:24:30
 ```
 
-### zonedDateTime.**getFields**() : { year: number, month: number, day: number, hour: number, minute: number, second: number, millisecond: number, microsecond: number, nanosecond: number, calendar: object, [propName: string]: unknown }
+### zonedDateTime.**getFields**() : { year: number, month: number, day: number, hour: number, minute: number, second: number, millisecond: number, microsecond: number, nanosecond: number, offsetNanoseconds: number, timeZone: object, calendar: object, [propName: string]: unknown }
```
justingrant, line 1203: This recently changed to have `offset` be the field and `offsetNanoseconds` be the convenience property, instead of the other way around. LMK if you need me to dig up the GH issue where we made this change, but the current code in #700 should already have this change included.

spec/zoneddatetime.html
justingrant: > in #703 Instant is always shown with `Z` anyway

Not if a time zone is supplied to `toString`, where it'd show an offset instead of `Z`.

spec/zoneddatetime.html
justingrant: Got it. As long as it's on your radar, that's good enough for me!

justingrant (approved):
I didn't go back and re-review the ZDT spec after my initial review, and it's a force-push so I'm not sure what changed since your initial checkin and my initial review, so I'm assuming that it's nothing but puppies and rainbows in there. ;-)  If there are significant changes in the ZDT spec that you want me to re-review, just LMK. Thanks!

spec/zoneddatetime.html
```diff
@@ -0,0 +1,1045 @@
...
+      </emu-alg>
+    </emu-clause>
+
+    <emu-clause id="sec-temporal-createtemporalzoneddatetime" aoid="CreateTemporalZonedDateTime">
+      <h1>CreateTemporalZonedDateTime ( _epochNanoseconds_, _timeZone_, _calendar_ [ , _newTarget_ ] )</h1>
+      <p>
+        The abstract operation CreateTemporalZonedDateTime is used to specify the creation of a new `Temporal.ZonedDateTime` object.
+      </p>
+      <emu-alg>
+        1. Assert: ValidateInstant(_epochNanoseconds_) is *true*.
```
Ms2ger, line 988: `! ValidateInstant(_epochNanoseconds_)`

spec/zoneddatetime.html
```diff
@@ -0,0 +1,1045 @@
...
+      <p>
+        `Temporal.ZonedDateTime.prototype.year` is an accessor property whose set accessor function is *undefined*.
+        Its get accessor function performs the following steps:
+      </p>
+      <emu-alg>
+        1. Let _zonedDateTime_ be the *this* value.
+        1. Perform ? RequireInternalSlot(_zonedDateTime_, [[InitializedTemporalZonedDateTime]]).
+        1. Let _timeZone_ be _zonedDateTime_.[[TimeZone]].
+        1. Let _instant_ be ? CreateTemporalInstant(_zonedDateTime_.[[Nanoseconds]]).
+        1. Let _temporalDateTime_ be ? GetTemporalDateTimeFor(_timeZone_, _instant_).
```
Ms2ger, line 162: I think it'll be better to create the DateTime in the constructor and stick it in an internal slot. Followup fine.

Ms2ger (approved):
Thanks! Feel free to argue that in favour of deferring any of the comments below to a followup

spec/zoneddatetime.html
```diff
@@ -0,0 +1,1045 @@
...
+        1. Let _smallestUnit_ be ? ToSmallestTemporalUnit(_options_, « »).
+        1. Let _roundingMode_ be ? ToTemporalRoundingMode(_options_).
+        1. If _smallestUnit_ is *"day"*, then
+          1. Let _maximum_ be 1.
+        1. Else if _smallestUnit_ is *"hour"*, then
+          1. Let _maximum_ be 24.
+        1. Else if _smallestUnit_ is *"minute"* or *"second"*, then
+          1. Let _maximum_ be 60.
+        1. Else,
+          1. Let _maximum_ be 1000.
```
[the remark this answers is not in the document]
ptomato: Only twice. They are different for every type (except ZonedDateTime and PlainDateTime, now.)

spec/zoneddatetime.html
ptomato: See explanation in https://github.com/tc39/proposal-temporal/pull/1044#discussion_r510968455 — the DateTime can't be cached unless the time zone is a built-in one and Temporal.TimeZone.prototype is %Temporal.TimeZone.prototype%.

spec/zoneddatetime.html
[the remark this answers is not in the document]
ptomato: (I will follow up on this.)

spec/zoneddatetime.html
```diff
@@ -0,0 +1,1045 @@
...
+        `Temporal.ZonedDateTime.prototype.hour` is an accessor property whose set accessor function is *undefined*.
+        Its get accessor function performs the following steps:
+      </p>
+      <emu-alg>
+        1. Let _zonedDateTime_ be the *this* value.
+        1. Perform ? RequireInternalSlot(_zonedDateTime_, [[InitializedTemporalZonedDateTime]]).
+        1. Let _timeZone_ be _zonedDateTime_.[[TimeZone]].
+        1. Let _instant_ be ? CreateTemporalInstant(_zonedDateTime_.[[Nanoseconds]]).
+        1. Let _temporalDateTime_ be ? GetTemporalDateTimeFor(_timeZone_, _instant_).
+        1. Return _temporalDateTime_.[[Hour]].
```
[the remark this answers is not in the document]
ptomato: I'd prefer to leave this and write the spec for time calendars all at once.

PR #1048, opened 2020-10-23
Don't use 'era' in iso8601 calendar

ptomato:
Closes: #901

### DIFF
--- a/polyfill/index.d.ts
+++ b/polyfill/index.d.ts
@@ -565,7 +565,7 @@ export namespace Temporal {
   };
 
   type DateFields = {
-    era: string | undefined;
+    era?: string | undefined;
     year: number;
     month: number;
     day: number;
@@ -647,7 +647,7 @@ export namespace Temporal {
   };
 
   type DateTimeFields = {
-    era: string | undefined;
+    era?: string | undefined;
     year: number;
     month: number;
     day: number;
@@ -967,7 +967,7 @@ export namespace Temporal {
   };
 
   type YearMonthFields = {
-    era: string | undefined;
+    era?: string | undefined;
     year: number;
     month: number;
     calendar: CalendarProtocol;
--- a/polyfill/index.d.ts
+++ b/polyfill/index.d.ts
@@ -435,7 +435,6 @@ export namespace Temporal {
     year(date: Temporal.Date): number;
     month(date: Temporal.Date): number;
     day(date: Temporal.Date): number;
-    era(date: Temporal.Date): string | undefined;
     dayOfWeek?(date: Temporal.Date): number;
     dayOfYear?(date: Temporal.Date): number;
     weekOfYear?(date: Temporal.Date): number;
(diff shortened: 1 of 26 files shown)


polyfill/test/calendar.mjs
```diff
@@ -150,6 +150,32 @@ describe('Calendar', () => {
...
+describe('Built-in calendars (not standardized yet)', () => {
+  describe('gregory', () => {
+    it('era AD', () => {
+      const date = Temporal.Date.from('1999-12-31[c=gregory]');
+      equal(date.era, 'ad');
+      equal(date.year, 1999);
+    });
+    it('era BC', () => {
+      const date = Temporal.Date.from('-000001-12-31[c=gregory]');
+      equal(date.era, 'bc');
```
justingrant, line 162: is lowercase "ad" and "bc" a convention in 402?  If not, seems unexpected because I've only seen those capitalized like "BC" or "AD"

polyfill/index.d.ts
```diff
@@ -564,7 +564,7 @@ export namespace Temporal {
   };
 
   type DateFields = {
-    era: string | undefined;
+    era?: string | undefined;
```
justingrant, line 567: I think we should remove `era` completely from the base types. After we release to reviewers, I'm happy to take a look at how we can improve TS types for non-ISO calendars, but in the meantime it doesn't make sense to have the all-calendars TS types include one property from non-ISO calendars but not other common ones like `yearType`.

polyfill/test/calendar.mjs
ptomato: These are string IDs - they aren't intended to be human-readable, I think of `era: 'bc'` more like `roundingMode: 'trunc'`. When printed with `toLocaleString()` in an English locale, they'd be printed as "BC" or "B.C" or "BCE" or whatever.

These string IDs are just for demonstration anyway - their exact values would have to be specified by ECMA-402, because the `gregory` calendar is in that domain.

polyfill/index.d.ts
ptomato: Added another commit removing these entirely from the TypeScript types.

polyfill/lib/monthday.mjs
```diff
@@ -104,17 +104,24 @@ export class MonthDay {
...
     if (ES.Type(item) === 'Object') {
-      ({ era, year } = ES.ToRecord(item, [['era', undefined], ['year']]));
+      const inputFieldNames = ES.CalendarFields(calendar, ['year']);
+      const entries = [['year']];
+      // Add extra fields from the calendar at the end
+      inputFieldNames.forEach((fieldName) => {
+        if (!entries.some(([name]) => name === fieldName)) {
+          entries.push([fieldName, undefined]);
+        }
+      });
```
Ms2ger, line 118: Followup: move into abstract op

Ms2ger (approved):

PR #1050, opened 2020-10-24
Allow property bags and strings in Calendar methods

ptomato:
This also clarifies which Temporal objects are accepted by which Calendar
methods.

Closes: #1030


ptomato:
@Ms2ger Would you mind taking an especially critical look at this one? I'm afraid it might be stomping over something that you were working on.

Ms2ger (approved):

PR #1060, opened 2020-10-26
until() methods

ptomato:
Implement until() methods to match the existing since() methods.

Closes: #998


docs/calendar.md
```diff
@@ -265,34 +265,36 @@ date.day         // => 7
...
+- `one` (`Temporal.Date`): A date.
+- `two` (`Temporal.Date`): Another date.
 - `options` (object): An object with properties representing options for the operation.
   The following options are recognized:
   - `largestUnit` (optional string): The largest unit of time to allow in the resulting `Temporal.Duration` object.
     Valid values are `'auto'`, `'years'`, `'months'`, and `'days'`.
     The default is `'auto'`.
 
-**Returns:** a `Temporal.Duration` representing the difference between `larger` and `smaller`.
+**Returns:** a `Temporal.Duration` representing the time elapsed after `one` and until `twor`.
```
justingrant, line 279: typo: `twor`

docs/calendar.md
```diff
@@ -265,34 +265,36 @@ date.day         // => 7
 date.toString()  // => 2020-06-28[c=islamic]
 ```
 
-### calendar.**dateDifference**(_smaller_: Temporal.Date, _larger_: Temporal.Date, _options_: object) : Temporal.Duration
+### calendar.**dateUntil**(_one_: Temporal.Date, _two_: Temporal.Date, _options_: object) : Temporal.Duration
```
justingrant, line 268: We have both `dateAdd` and `dateSubtract` but only one `dateUntil` (not `dateSince`). Why do we have to methods in one place but only one method here?  IMHO, one method in both places might be cleanest.

docs/date.md
```diff
@@ -470,7 +470,7 @@ This is because months and years can be different lengths depending on which mon
 
 You can round the result using the `smallestUnit`, `roundingIncrement`, and `roundingMode` options.
 These behave as in the `Temporal.Duration.round()` method, but increments of days and larger are allowed.
-Since rounding to calendar units requires a reference point, `date` is used as the reference point.
+Since rounding to calendar units requires a reference point, `date` is used as the starting point.
```
justingrant, line 473: "Since" => "Because".  For clarity, IMHO we should only use "since" in the docs to refer to relative time.

docs/date.md
```diff
@@ -482,22 +482,57 @@ Usage example:
...
+    Valid values are `'nearest'`, `'ceil'`, `'trunc'`, and `'floor'`.
+    The default is `'nearest'`.
+
+**Returns:** a `Temporal.Duration` representing the time elapsed before `date` and since `other`.
+
+This method computes the difference between the two dates represented by `date` and `other`, optionally rounds it, and returns it as a `Temporal.Duration` object.
+If `other` is later than `date` then the resulting duration will be negative.
+
+This method does the same thing as the `Temporal.Date.prototype.until()` method, but reversed, and rounding takes place relative to `date` as an ending point instead of a starting point.
+If no rounding takes place, then the outcome of `date1.since(date2)` is the same as `date1.until(date2).negated()`.
```
justingrant, line 526: > If no rounding takes place, then the outcome of `date1.since(date2)` is the same as `date1.until(date2).negated()`.

I believe that this statement is incorrect. AFAIK, the starting point controls how month-end calculations work for variable-length months. (And, in the future, for variable-length days with ZDT.)  This means that reversing the order can yield different results.

For example, with `largestUnit: 'months'` if I start from `2019-02-28` and want to get to `2019-04-01`, then I'll add one month to get to Mar 28, and 4 days to get to April 1, yielding `P1M4D`. But if I start from April 1, then I'll subtract one month to get to March 1, and then subtract one more day to get to Feb 28, or `-P1M1D`.

docs/datetime.md
```diff
@@ -555,30 +555,65 @@ Usage example:
...
+    Valid values are `'nearest'`, `'ceil'`, `'trunc'`, and `'floor'`.
+    The default is `'nearest'`.
+
+**Returns:** a `Temporal.Duration` representing the elapsed time before `datetime` and since `other`.
+
+This method computes the difference between the two times represented by `datetime` and `other`, optionally rounds it, and returns it as a `Temporal.Duration` object.
+If `other` is later than `datetime` then the resulting duration will be negative.
+
+This method does the same thing as the `Temporal.DateTime.prototype.until()` method, but reversed, and rounding takes place relative to `datetime` as an ending point instead of a starting point.
+If no rounding takes place, then the outcome of `dt1.since(dt2)` is the same as `dt1.until(dt2).negated()`.
```
justingrant, line 607: Same issue as noted above.

docs/yearmonth.md
```diff
@@ -366,7 +366,7 @@ A value of `'auto'` means `'years'`.
 
 You can round the result using the `smallestUnit`, `roundingIncrement`, and `roundingMode` options.
 These behave as in the `Temporal.Duration.round()` method, but increments of months and larger are allowed.
-Since rounding to calendar units requires a reference point, the first day of `yearMonth` is used as the reference point.
+Since rounding to calendar units requires a reference point, the first day of `yearMonth` is used as the starting point.
```
justingrant, line 369: Since=>Because

docs/calendar.md
ptomato: I believe we could consolidate dateAdd() and dateSubtract() once #993 is completed, but not before.

docs/yearmonth.md
```diff
@@ -377,21 +377,56 @@ Usage example:
...
+    Valid values are `'nearest'`, `'ceil'`, `'trunc'`, and `'floor'`.
+    The default is `'nearest'`.
+
+**Returns:** a `Temporal.Duration` representing the elapsed time before `yearMonth` and since `other`.
+
+This method computes the difference between the two months represented by `yearMonth` and `other`, optionally rounds it, and returns it as a `Temporal.Duration` object.
+If `other` is later than `yearMonth` then the resulting duration will be negative.
+
+This method does the same thing as the `Temporal.YearMonth.prototype.until()` method, but reversed, and rounding takes place relative to `yearMonth` as an ending point instead of a starting point.
+If no rounding takes place, then the outcome of `ym1.since(ym2)` is the same as `ym1.until(ym2).negated()`.
```
justingrant, line 420: I think that the same reversibility issue as noted above applies here too, although I'm not 100% sure. It might only happen in calendars with variable numbers of months in each year, and with a starting or ending point inside a leap month.

polyfill/index.d.ts
```diff
@@ -602,6 +622,18 @@ export namespace Temporal {
...
     subtract(durationLike: Temporal.Duration | DurationLike | string, options?: ArithmeticOptions): Temporal.Date;
+    until(
+      other: Temporal.Date | DateLike | string,
+      options?: DifferenceOptions<
+        | 'years'
+        | 'months'
+        | 'weeks'
+        | 'days'
+        | /** @deprecated */ 'year'
+        | /** @deprecated */ 'month'
```
justingrant, line 633: `week` missing?  BTW, same in `since` below.

justingrant: oh, never mind-- there is no `week` unit. ;-)

polyfill/lib/datetime.mjs
```diff
@@ -418,7 +418,7 @@ export class DateTime {
     if (!ES.IsTemporalDateTime(result)) throw new TypeError('invalid result');
     return result;
   }
-  since(other, options = undefined) {
+  until(other, options = undefined) {
```
justingrant, line 421: If `since` is the same as `until` with the args reversed and the result negated, then why are both methods mostly repeated code?  Why not refactor into a single implementation?  Or is that best left for when we have more time?

justingrant (approved):
I focused mostly on docs and TS types, which LGTM other than the few minor issues that I noted and the one big issue about lack of reversibility. I only skimmed the polyfill and spec changes.

docs/datetime.md
```diff
@@ -539,7 +539,7 @@ This is because months and years can be different lengths depending on which mon
 
 You can round the result using the `smallestUnit`, `roundingIncrement`, and `roundingMode` options.
 These behave as in the `Temporal.Duration.round()` method, but increments of days and larger are allowed.
-Since rounding to an increment expressed in days or larger units requires a reference point, `datetime` is used as the reference point in that case.
+Since rounding to an increment expressed in days or larger units requires a reference point, `datetime` is used as the starting point in that case.
```
justingrant, line 542: Since=>Because

justingrant: Also, the same note about reversibility should be in the docs for all `since` methods of Date, DateTime, and (maybe) YearMonth as we have for `until`.

polyfill/lib/datetime.mjs
ptomato: I refactored the identical part of Instant.until() and Instant.since() because it was quite large, but the bulk of all the other until() and since() methods are already in common operations like DifferenceTime, DifferenceDate, and RoundDuration — it could still be made smaller but you are right that I don't have the extra time to spend on that right now. (It looks worse than it is because the lists of "year, month, day, hour, ..." are so long.)

docs/date.md
ptomato: Could it be that this is only true after #993 is completed? Currently, the polyfill does the following because of `constrain`:
```
> Temporal.Date.from('2019-04-01').since('2019-02-28', {largestUnit: 'months'})
Temporal.Duration <P1M4D> {
  months: 1,
  days: 4
}
> Temporal.Date.from('2019-02-28').since('2019-04-01', {largestUnit: 'months'})
Temporal.Duration <-P1M4D> {
  months: -1,
  days: -4
}
> Temporal.Date.from('2019-02-28').add('P1M4D')
Temporal.Date <2019-04-01>
> Temporal.Date.from('2019-04-01').subtract('P1M1D')
Temporal.Date <2019-02-28>
> Temporal.Date.from('2019-04-01').subtract('P1M4D')
Temporal.Date <2019-02-28>
```
(as it stands, the starting point controls nothing except rounding, and `largestUnit` isn't rounding)

justingrant: > Could it be that this is only true after #993 is completed?

Correct.  But the docs should reflect expected behavior, right?

ptomato: I'd rather the docs change with #993, otherwise I have to write docs for every open issue in order to land a pull request for only one. But, I think it would be correct in either case to say, "With the default options, then the outcome of `date1.since(date2)` is the same as `date2.since(date1).negated()`." I'll go with that for now.

justingrant: SGTM

docs/calendar.md
justingrant: SGTM. Filed #1066 to track this.

Ms2ger (approved):

PR #1062, opened 2020-10-27
Introduce calendar systems to time

ryzokuken:
Fixes: #522 

/cc @ptomato


polyfill/lib/date.mjs
```diff
@@ -256,18 +256,27 @@ export class Date {
...
+    const hour = GetSlot(temporalTime, ISO_HOUR);
+    const minute = GetSlot(temporalTime, ISO_MINUTE);
+    const second = GetSlot(temporalTime, ISO_SECOND);
+    const millisecond = GetSlot(temporalTime, ISO_MILLISECOND);
+    const microsecond = GetSlot(temporalTime, ISO_MICROSECOND);
+    const nanosecond = GetSlot(temporalTime, ISO_NANOSECOND);
+    const timeCalendar = GetSlot(temporalTime, CALENDAR);
+
+    let calendar;
+    if (dateCalendar.toString() !== timeCalendar.toString() && dateCalendar.toString() === 'iso8601') {
```
ptomato, line 274: Use `ES.CalendarEquals()` and `ES.CalendarToString()` here.

I'm also not sure the logic is correct - what if the calendars are different and neither of them are ISO? I think it should throw in that case.

polyfill/lib/datetime.mjs
```diff
@@ -650,15 +660,15 @@ export class DateTime {
     if (!ES.IsTemporalDateTime(this)) throw new TypeError('invalid receiver');
     return {
       calendar: GetSlot(this, CALENDAR),
-      hour: GetSlot(this, HOUR),
+      isoHour: GetSlot(this, ISO_HOUR),
```
ptomato, line 663: From what I understood from Ms2ger, these should be re-alphabetized.

polyfill/lib/time.mjs
```diff
@@ -1,32 +1,30 @@
...
 
+import { GetISO8601Calendar } from './calendar.mjs';
 import { ES } from './ecmascript.mjs';
 import { DateTimeFormat } from './intl.mjs';
 import { GetIntrinsic, MakeIntrinsicClass } from './intrinsicclass.mjs';
 
 import {
-  ISO_YEAR,
-  ISO_MONTH,
-  ISO_DAY,
```
ptomato, line 10: I think these three lines still need to be in here.

polyfill/lib/time.mjs
```diff
@@ -404,37 +412,62 @@ export class Time {
...
+    const hour = GetSlot(this, ISO_HOUR);
+    const minute = GetSlot(this, ISO_MINUTE);
+    const second = GetSlot(this, ISO_SECOND);
+    const millisecond = GetSlot(this, ISO_MILLISECOND);
+    const microsecond = GetSlot(this, ISO_MICROSECOND);
+    const nanosecond = GetSlot(this, ISO_NANOSECOND);
+    const timeCalendar = GetSlot(this, CALENDAR);
+
+    let calendar;
+    if (dateCalendar.toString() !== timeCalendar.toString() && timeCalendar.toString() === 'iso8601') {
```
ptomato, line 431: Same comment here as in Date.toDateTime.

polyfill/lib/time.mjs
```diff
@@ -404,37 +412,62 @@ export class Time {
...
     const fields = ES.ToTemporalTimeRecord(this);
     if (!fields) throw new TypeError('invalid receiver');
+    fields.calendar = GetSlot(this, CALENDAR);
     return fields;
   }
+  getISOFields() {
+    if (!ES.IsTemporalTime(this)) throw new TypeError('invalid receiver');
+    return {
+      calendar: GetSlot(this, CALENDAR),
+      isoHour: GetSlot(this, ISO_HOUR),
```
ptomato, line 450: These should be alphabetized.

ptomato (commented):
This is looking good, is the plan to merge it and work on PRs for documentation and spec separately, or add those to this PR?

I do think that tests for `time.getISOFields()` should be added as part of this PR though.

As well, we could consider adding a calendar to `test/usercalendar.mjs` that exercises this functionality.

polyfill/lib/calendar.mjs
```diff
@@ -116,6 +157,18 @@ export class Calendar {
...
+    throw new Error('not implemented');
+  }
+  minute(time) {
+    void time;
+    throw new Error('not implemented');
+  }
+  second(time) {
+    void time;
+    throw new Error('not implemented');
+  }
```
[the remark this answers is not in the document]
ryzokuken: ouch, I have no idea why/how I missed those. Thanks!

polyfill/lib/time.mjs
ryzokuken: I removed these as part of one of my refactors, but then undid it to make the patch cleaner. This must've slipped through.

polyfill/lib/calendar.mjs
```diff
@@ -68,6 +89,26 @@ export class Calendar {
...
+  timeSubtract(time, duration, options, constructor) {
+    void time;
+    void duration;
+    void options;
+    void constructor;
+    throw new Error('not implemented');
+  }
+  timeDifference(one, two, options) {
```
ptomato, line 106: Sorry I didn't think of this before - I think we should name this `timeUntil` since we are renaming `dateDifference` to `dateUntil` in my PR as well.

polyfill/lib/calendar.mjs
```diff
@@ -158,6 +223,22 @@ class ISO8601Calendar extends Calendar {
     ({ year, month, day } = ES.RegulateDate(year, month, day, overflow));
     return new constructor(year, month, day, this);
   }
+  timeFromFields(fields, options, constructor) {
```
ptomato, line 226: I also didn't think of this before, but should the ISO calendar have an implementation of `timeAdd`, `timeSubtract`, and `timeUntil`?

polyfill/lib/datetime.mjs
```diff
@@ -449,18 +449,18 @@ export class DateTime {
     const roundingIncrement = ES.ToTemporalRoundingIncrement(options, maximumIncrements[smallestUnit], false);
 
     let { deltaDays, hours, minutes, seconds, milliseconds, microseconds, nanoseconds } = ES.DifferenceTime(
```
ptomato, line 451: I think this needs to be moved into `ISO8601Calendar.timeUntil()` and that method needs to be called here.

ptomato (changes requested):
Sorry, I missed the first time around, that Time.add, Time.subtract, Time.until, and Time.since need to use calendar.timeAdd, calendar.timeSubtract, calendar.timeUntil.

polyfill/lib/date.mjs
```diff
@@ -8,12 +8,12 @@ import {
...
   ISO_DAY,
-  HOUR,
-  MINUTE,
-  SECOND,
-  MILLISECOND,
-  MICROSECOND,
-  NANOSECOND,
+  ISO_HOUR,
+  ISO_MINUTE,
+  ISO_SECOND,
```
justingrant, line 13: For seconds and smaller units, do they need to be calendar-aware?  Are there any use cases we know about that require calendar-specific measurement of seconds or nanoseconds? Or are we just making them calendar-aware for consistency?

polyfill/lib/calendar.mjs
```diff
@@ -68,6 +89,26 @@ export class Calendar {
...
     throw new Error('not implemented');
   }
+  timeAdd(time, duration, options, constructor) {
+    void time;
+    void duration;
+    void options;
+    void constructor;
+    throw new Error('not implemented');
+  }
+  timeSubtract(time, duration, options, constructor) {
```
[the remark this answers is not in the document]
justingrant: Related: #1066 (for dates). But if we're adding time methods for the first time, why not just have one?

justingrant:
I didn't review this deeply because I'm not familiar enough with the implementation, but I left a few notes. 

One other related thing I noticed: the phrase "calendar units" shows up in a bunch of places in Temporal polyfill code and a few places in the spec. In the past these were synonyms for "may not be fixed-length units" but now that *all* units are calendar units, we probably need a different term to replace "calendar units" with. How about "fixed-length units" and "variable-length units" ?

polyfill/lib/date.mjs
ptomato: I think we should probably rename them to ISO in any case, but yes it does seem that there should not be use cases where 1000 milliseconds ≠ 1 second.

polyfill/lib/calendar.mjs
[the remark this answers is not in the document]
ptomato: The original calendar draft had only one add() and one subtract() method. I can't remember exactly, but it may not be possible to combine dateAdd() and timeAdd() since the allowed ranges for all the Plain types are different. But it may be possible to keep that range checking in the callers.

polyfill/lib/datetime.mjs
```diff
@@ -445,18 +445,18 @@ export class DateTime {
...
-      GetSlot(this, HOUR),
```
ptomato, line 448: `this` and `other` are swapped here. That is probably where the test failures are coming from.

polyfill/lib/datetime.mjs
```diff
@@ -233,16 +233,13 @@ export class DateTime {
...
-    ({ hour, minute, second, millisecond, microsecond, nanosecond } = ES.RegulateTime(
-      hour,
-      minute,
-      second,
-      millisecond,
-      microsecond,
-      nanosecond,
-      overflow
-    ));
+    const time = calendar.timeFromFields(fields, options, GetIntrinsic('%Temporal.Time%'));
```
ptomato, line 236:
```suggestion
    const time = calendar.timeFromFields(fields, { overflow }, GetIntrinsic('%Temporal.Time%'));
```
I think we are normally creating a new options object to pass to the calendar methods instead of passing through the user-supplied one.

polyfill/lib/datetime.mjs
```diff
@@ -444,19 +441,9 @@ export class DateTime {
...
-      GetSlot(this, MICROSECOND),
-      GetSlot(this, NANOSECOND),
-      GetSlot(other, HOUR),
-      GetSlot(other, MINUTE),
-      GetSlot(other, SECOND),
-      GetSlot(other, MILLISECOND),
-      GetSlot(other, MICROSECOND),
-      GetSlot(other, NANOSECOND)
+    let { days: deltaDays, hours, minutes, seconds, milliseconds, microseconds, nanoseconds } = calendar.timeUntil(
+      this.toTime(),
```
ptomato, line 445: These should be abstract operations so that the call to `toTime` isn't observable (though feel free to defer this to your next PR)

polyfill/lib/ecmascript.mjs
```diff
@@ -1140,6 +1161,17 @@ export const ES = ObjectAssign({}, ES2020, {
     const cal2 = ES.CalendarToString(two);
     return cal1 === cal2;
   },
+  ConsolidateCalendars: (one, two) => {
+    const sOne = ES.CalendarToString(one);
+    const sTwo = ES.CalendarToString(two);
+    if (sOne === sTwo || sOne === 'iso8601') {
```
ptomato, line 1167: I thought we had removed the special handling of the ISO calendar as a "null" calendar everywhere else ­— double check with @sffc who probably remembers better than I do.

polyfill/lib/time.mjs
```diff
@@ -146,137 +157,35 @@ export class Time {
...
   }
   until(other, options = undefined) {
     if (!ES.IsTemporalTime(this)) throw new TypeError('invalid receiver');
     other = ES.ToTemporalTime(other, Time);
+    const calendar = GetSlot(this, CALENDAR);
+    const otherCalendar = GetSlot(other, CALENDAR);
+    const calendarId = ES.CalendarToString(calendar);
+    const otherCalendarId = ES.CalendarToString(otherCalendar);
+    if (calendarId !== otherCalendarId) {
+      throw new RangeError(`cannot compute difference between dates of ${calendarId} and ${otherCalendarId} calendars`);
```
ptomato, line 186:
```suggestion
      throw new RangeError(`cannot compute difference between times of ${calendarId} and ${otherCalendarId} calendars`);
```

polyfill/lib/time.mjs
```diff
@@ -337,6 +233,14 @@ export class Time {
   since(other, options = undefined) {
     if (!ES.IsTemporalTime(this)) throw new TypeError('invalid receiver');
     other = ES.ToTemporalTime(other, Time);
+    const calendar = GetSlot(this, CALENDAR);
+    const otherCalendar = GetSlot(other, CALENDAR);
+    const calendarId = ES.CalendarToString(calendar);
+    const otherCalendarId = ES.CalendarToString(otherCalendar);
+    if (calendarId !== otherCalendarId) {
+      throw new RangeError(`cannot compute difference between dates of ${calendarId} and ${otherCalendarId} calendars`);
```
ptomato, line 241:
```suggestion
      throw new RangeError(`cannot compute difference between times of ${calendarId} and ${otherCalendarId} calendars`);
```

polyfill/lib/zoneddatetime.mjs
```diff
@@ -370,16 +370,16 @@ export class ZonedDateTime {
     const tz = GetSlot(this, TIME_ZONE);
     return {
       calendar: GetSlot(this, CALENDAR),
-      hour: GetSlot(dt, HOUR),
+      hour: GetSlot(dt, ISO_HOUR),
```
ptomato, line 373: Rename these fields to `iso` as well.

polyfill/lib/time.mjs
```diff
@@ -292,20 +201,7 @@ export class Time {
...
-      GetSlot(this, NANOSECOND),
-      GetSlot(other, HOUR),
-      GetSlot(other, MINUTE),
-      GetSlot(other, SECOND),
-      GetSlot(other, MILLISECOND),
-      GetSlot(other, MICROSECOND),
-      GetSlot(other, NANOSECOND)
-    );
+    let { hours, minutes, seconds, milliseconds, microseconds, nanoseconds } = calendar.timeUntil(this, other);
     ({ hours, minutes, seconds, milliseconds, microseconds, nanoseconds } = ES.RoundDuration(
```
ptomato, line 205: This is going to be a lot of work, but I believe RoundDuration will need to round according to the calendar's length of hours and minutes? It's fine to not block this PR on it.

polyfill/test/usercalendar.mjs
```diff
@@ -224,29 +224,61 @@ describe('Userland calendar', () => {
...
+    // years, 10-day months, 864-hour days, 10-minute hours and 10-second
+    // minutes, and the year zero is at the Unix epoch.
+    function decimalToISO(year, month, day, hour = 0, minute = 0, second = 0, overflow = 'constrain') {
       if (overflow === 'constrain') {
         if (month < 1) month = 1;
         if (month > 10) month = 10;
         if (day < 1) day = 1;
         if (day > 10) day = 10;
+        if (hour < 1) hour = 1;
+        if (hour > 864) hour = 864;
```
ptomato, line 236:
```suggestion
        if (hour < 0) hour = 0;
        if (hour > 863) hour = 863;
```

ptomato (approved):
I'm afraid that Duration rounding and balancing will be a big missing part of this, but let's make sure that what is in this PR can be merged soon.

polyfill/lib/datetime.mjs
ryzokuken: @ptomato check the `dateFromFields` call *right* above this. It passes in `options` directly. I spotted a lot of such issues early-on and even made a few refactor commits that I later dropped to keep this PR as small and on-topic as possible. Should I fix all sites in this PR, or would a single follow-up that takes care of all such cases suffice?

polyfill/lib/datetime.mjs
ryzokuken: Will do!

polyfill/lib/ecmascript.mjs
ryzokuken: I'll reach out to @sffc about this, thanks.

polyfill/lib/time.mjs
ryzokuken: Thanks! Copy+paste gone wrong. Fixed.

polyfill/lib/time.mjs
ryzokuken: Same as above. Thanks!

polyfill/lib/zoneddatetime.mjs
ryzokuken: Yes! Done.

polyfill/lib/time.mjs
ryzokuken: Well, wasn't the plan to keep Durations ISO-only? But I think I get your point more or less. Since `Time.p.add(duration)` interprets the duration in a calendar-specific manner, `Time.p.until` should produce a Duration based on the calendar. Let's make an issue to discuss this more in detail and land this? We can either generalize this as `calendar.durationRound` or just add the logic to the `until`/`since` methods on the calendar.

polyfill/test/usercalendar.mjs
ryzokuken: Fixed reject, but somehow missed this one! Thanks.

ryzokuken:
IIUC, all outstanding comments have been addressed and this should be ready to merge. However, a couple of things need to be worked out in follow-ups. In the meantime, I'd also start working on the spec text and docs for this change. Thanks everyone!

Ms2ger:
Travis is having issues, so I tested locally before merging.

PR #1063, opened 2020-10-27
polyfill: fix intl tests for japanese calendar

ryzokuken:
/cc @justingrant @ptomato

### DIFF
--- a/polyfill/test/intl.mjs
+++ b/polyfill/test/intl.mjs
@@ -66,11 +66,11 @@ describe('Intl', () => {
         second: 30,
         calendar: 'japanese'
       });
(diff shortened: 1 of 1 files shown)


ptomato:
Apparently this fixes broken tests on some platforms, but breaks working tests on others. I think we need to figure out the reason for this discrepancy. Maybe mark these tests as `it.skip(...)` for the time being?

justingrant:
Why not just keep working tests like #1061 until we figure out what's causing environments to be different?  Even if we can figure it out, other contributors who maybe have an older environment may end up with broken tests. Seems better to either skip the bad tests or to accept both formats like #1061. I like the latter better but either are OK.

ptomato:
Closing, #1061 is merged.

PR #1065, opened 2020-10-28
Fix bugs in rounding algorithm

ptomato:
Currently requires #1060, so review that first.

Closes: #1059

### DIFF
--- a/polyfill/test/duration.mjs
+++ b/polyfill/test/duration.mjs
@@ -941,10 +941,17 @@ describe('Duration', () => {
       equal(`${d.round({ smallestUnit: 'years', relativeTo })}`, 'P6Y');
       equal(`${d.negated().round({ smallestUnit: 'years', relativeTo })}`, '-P6Y');
     });
-    it('balances differently depending on relativeTo', () => {
+    it('balances up differently depending on relativeTo', () => {
       const fortyDays = Duration.from({ days: 40 });
       equal(`${fortyDays.round({ largestUnit: 'years', relativeTo: '2020-01-01' })}`, 'P1M9D');
       equal(`${fortyDays.round({ largestUnit: 'years', relativeTo: '2020-02-01' })}`, 'P1M11D');
+      equal(`${fortyDays.round({ largestUnit: 'years', relativeTo: '2020-03-01' })}`, 'P1M9D');
+      equal(`${fortyDays.round({ largestUnit: 'years', relativeTo: '2020-04-01' })}`, 'P1M10D');
+      const minusForty = Duration.from({ days: -40 });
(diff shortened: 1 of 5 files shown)


docs/calendar.md
```diff
@@ -269,36 +269,38 @@ date.day         // => 7
...
 - `options` (object): An object with properties representing options for the operation.
   The following options are recognized:
   - `largestUnit` (optional string): The largest unit of time to allow in the resulting `Temporal.Duration` object.
     Valid values are `'auto'`, `'years'`, `'months'`, and `'days'`.
     The default is `'auto'`.
 
-**Returns:** a `Temporal.Duration` representing the difference between `larger` and `smaller`.
+**Returns:** a `Temporal.Duration` representing the time elapsed after `one` and until `two`.
 
 If either of `smaller` or `larger` are not `Temporal.Date` objects, then they will be converted to one as if they were passed to `Temporal.Date.from()`.
```
justingrant, line 285: Don't you mean "`one` or `two`" not "`smaller` or `larger`"

polyfill/test/duration.mjs
```diff
@@ -941,10 +941,13 @@ describe('Duration', () => {
...
       equal(`${d.negated().round({ smallestUnit: 'years', relativeTo })}`, '-P6Y');
     });
-    it('balances differently depending on relativeTo', () => {
+    it('balances up differently depending on relativeTo', () => {
       const fortyDays = Duration.from({ days: 40 });
       equal(`${fortyDays.round({ largestUnit: 'years', relativeTo: '2020-01-01' })}`, 'P1M9D');
       equal(`${fortyDays.round({ largestUnit: 'years', relativeTo: '2020-02-01' })}`, 'P1M11D');
+      const minusForty = Duration.from({ days: -40 });
+      equal(`${minusForty.round({ largestUnit: 'years', relativeTo: '2020-01-01' })}`, '-P1M9D');
+      equal(`${minusForty.round({ largestUnit: 'years', relativeTo: '2020-02-01' })}`, '-P1M9D');
```
justingrant, line 950: It might be more useful to use 2020-03-01 and 2020-04-01 as the starting point where the durations would be different, as opposed to 2020-01-01 and 2020-02-01 where both cases only have to traverse 31-day months .

justingrant (approved):
Looks great, other than the handful of minor issues I noted.

polyfill/test/duration.mjs
ptomato: I wanted to check that the results were different for the same `relativeTo`, but this is also a good point. I've done both in both cases now.

Ms2ger (approved):

PR #1071, opened 2020-10-28
Remove Duration arithmetic overflow option

ptomato:
Closes: #857
See also: #856

### DIFF
--- a/docs/duration.md
+++ b/docs/duration.md
@@ -208,9 +208,7 @@ duration = duration.with({ years, months });
 - `other` (`Temporal.Duration` or value convertible to one): The duration to add.
 - `options` (optional object): An object with properties representing options for the addition.
   The following options are recognized:
-  - `overflow` (string): How to deal with additions that result in mixed-sign values or values that are larger than the next highest unit.
-    Allowed values are `constrain` and `balance`.
-    The default is `constrain`.
+  - (TODO)
 
 **Returns:** a new `Temporal.Duration` object which represents the sum of the durations of `duration` and `other`.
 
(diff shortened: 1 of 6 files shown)


polyfill/index.d.ts
```diff
@@ -351,8 +351,8 @@ export namespace Temporal {
     negated(): Temporal.Duration;
     abs(): Temporal.Duration;
     with(durationLike: DurationLike): Temporal.Duration;
-    add(other: Temporal.Duration | DurationLike | string, options?: DurationOptions): Temporal.Duration;
-    subtract(other: Temporal.Duration | DurationLike | string, options?: DurationOptions): Temporal.Duration;
+    add(other: Temporal.Duration | DurationLike | string, options?: DurationRoundOptions): Temporal.Duration;
```
justingrant, line 354: I assume that the options are defined here but not in the docs because you're waiting for #856 to document the new options?  OK if so.

ptomato: That's right.

docs/duration.md
```diff
@@ -289,15 +278,8 @@ If `duration` is not a `Temporal.Duration` object, then it will be converted to
...
-In order to be valid, the resulting duration must not have fields with mixed signs.
-However, before the result is balanced, it's possible that the intermediate result will have one or more negative fields while the overall duration is positive, or vice versa.
-For example, "4 hours and 15 minutes" minus "2 hours and 30 minutes" results in "2 hours and &minus;15 minutes".
-The `overflow` argument tells what to do in this case:
-
-- In `constrain` mode (the default), subtractions that result in mixed-sign fields will balance those fields with the next-highest field so that all the fields of the result are constrained to have the same sign.
-- In `balance` mode, all fields are balanced with the next highest field, no matter if they have mixed signs or not.
-
-For usage examples and a more complete explanation of how balancing works and why it is necessary, especially for subtracting `Temporal.Duration`, see [Duration balancing](./balancing.md#duration-arithmetic).
+In order to be valid, the resulting duration must not have fields with mixed signs, and so the result is balanced.
```
justingrant, line 281: When duration rounding in add/subtract lands, this text should probably be clarified (e.g. to define what "balanced" means, and to explain that balancing will only happen up to the largest unit), but no need to worry about it now.

justingrant (approved):
I didn't dig into the polyfill code or test changes, but the docs changes look good.

Ms2ger (approved):

PR #1078, opened 2020-10-30
Remove some references to 'era' in CalendarFields() calls.

Ms2ger:


ptomato (approved):
:+1:, I must have missed these

PR #1081, opened 2020-10-30
toString extra options

ptomato:
Closes: #703

### DIFF
--- a/docs/zoneddatetime.md
+++ b/docs/zoneddatetime.md
@@ -1078,6 +1078,9 @@ zdt1.equals(zdt1); // => true
 
 - `options` (optional object): An object with properties influencing the formatting.
   The following options are recognized:
+  - `calendar` (string): Whether to show the calendar annotation in the return value.
+    Valid values are `'auto'`, `'always'`, and `'never'`.
+    The default is `'auto'`.
   - `fractionalSecondDigits` (number or string): How many digits to print after the decimal point in the output string.
     Valid values are `'auto'`, 0, 1, 2, 3, 4, 5, 6, 7, 8, or 9.
     The default is `'auto'`.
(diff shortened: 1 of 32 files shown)


docs/zoneddatetime.md
```diff
@@ -1072,7 +1072,30 @@ zdt1.equals(zdt2); // => false (same offset but different time zones)
...
+    The default is `'auto'`.
+  - `timeZone` (string): Whether to show the time zone name annotation in the return value.
+    Valid values are `'auto'` and `'never'`.
+    The default is `'auto'`.
+  - `calendar` (string): Whether to show the calendar annotation in the return value.
+    Valid values are `'auto'`, `'always'`, and `'never'`.
+    The default is `'auto'`.
+  - `fractionalSecondDigits` (number or string): How many digits to print after the decimal point in the output string.
+    Valid values are `'auto'`, 0, 1, 2, 3, 4, 5, 6, 7, 8, or 9.
+    The default is `'auto'`.
```
justingrant, line 1092: Is the behavior of `auto` defined anywhere in the docs? I didn't see it here but might have missed it.

polyfill/lib/zoneddatetime.mjs
```diff
@@ -453,8 +467,9 @@ function zonedDateTimeToString(zdt, precision, options = undefined) {
   minute = ES.ISODateTimePartString(minute);
   const seconds = ES.FormatSecondsStringPart(second, millisecond, microsecond, nanosecond, precision);
   const tz = GetSlot(zdt, TIME_ZONE);
-  const offset = ES.GetOffsetStringFor(tz, GetSlot(zdt, INSTANT));
-  const zone = ES.TimeZoneToString(tz);
-  const calendar = ES.FormatCalendarAnnotation(GetSlot(zdt, CALENDAR));
-  return `${year}-${month}-${day}T${hour}:${minute}${seconds}${offset}[${zone}]${calendar}`;
+  let result = `${year}-${month}-${day}T${hour}:${minute}${seconds}`;
```
justingrant, line 470: I noticed that the `__repr__` property isn't using the same format as here. Is this intentional?

```js
ZonedDateTime {_repr_: "Temporal.ZonedDateTime <2020-11-01T01:30:00.0-07:00[America/Los_Angeles]>"}
```

Note the `.0` after the seconds, which looks odd.

justingrant (approved):
Quick review of the ZDT parts. Looks good.

polyfill/lib/zoneddatetime.mjs
ptomato: I added a commit fixing this.

docs/zoneddatetime.md
ptomato: Fixed this, I had neglected to copy a few paragraphs from DateTime.

Ms2ger (changes requested):

polyfill/lib/yearmonth.mjs
```diff
@@ -4,18 +4,29 @@ import { GetISO8601Calendar } from './calendar.mjs';
...
-function YearMonthToString(yearMonth) {
+function YearMonthToString(yearMonth, showCalendar = 'auto') {
   const year = ES.ISOYearString(GetSlot(yearMonth, ISO_YEAR));
   const month = ES.ISODateTimePartString(GetSlot(yearMonth, ISO_MONTH));
   let resultString = `${year}-${month}`;
-  const calendar = ES.FormatCalendarAnnotation(GetSlot(yearMonth, CALENDAR));
-  if (calendar) {
+  const calendar = GetSlot(yearMonth, CALENDAR);
+  const calendarString = ES.FormatCalendarAnnotation(calendar, showCalendar);
+  if (calendarString || !ES.IsTemporalCalendar(calendar) || GetSlot(calendar, CALENDAR_ID) !== 'iso8601') {
```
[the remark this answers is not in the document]
ptomato: The string has to include the day if the calendar is non-ISO, and omit the day if it is ISO. But in hindsight there is a better way to accomplish that than this confusing condition.

Ms2ger (approved):

PR #1083, opened 2020-10-31
Correct timezone offset allowed range 

akshgpt7:
Fixes #1036

### DIFF
--- a/polyfill/lib/regex.mjs
+++ b/polyfill/lib/regex.mjs
@@ -10,7 +10,7 @@ export const calendarID = new RegExp(`(?:${calComponent.source}(?:-${calComponen
 const yearpart = /(?:[+\u2212-]\d{6}|\d{4})/;
 export const datesplit = new RegExp(`(${yearpart.source})(?:-(\\d{2})-(\\d{2})|(\\d{2})(\\d{2}))`);
 const timesplit = /(\d{2})(?::(\d{2})(?::(\d{2})(?:[.,](\d{1,9}))?)?|(\d{2})(?:(\d{2})(?:[.,](\d{1,9}))?)?)?/;
-export const offset = /([+\u2212-])([0-2][0-9])(?::?([0-5][0-9])(?::?([0-5][0-9])(?:[.,](\d{1,9}))?)?)?/;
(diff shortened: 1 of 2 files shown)


polyfill/lib/regex.mjs
```diff
@@ -7,7 +7,7 @@ export const calendarID = new RegExp(`(?:${calComponent.source}(?:-${calComponen
 const yearpart = /(?:[+\u2212-]\d{6}|\d{4})/;
 export const datesplit = new RegExp(`(${yearpart.source})(?:-(\\d{2})-(\\d{2})|(\\d{2})(\\d{2}))`);
 const timesplit = /(\d{2})(?::(\d{2})(?::(\d{2})(?:[.,](\d{1,9}))?)?|(\d{2})(?:(\d{2})(?:[.,](\d{1,9}))?)?)?/;
-export const offset = /([+\u2212-])([0-2][0-9])(?::?([0-5][0-9])(?::?([0-5][0-9])(?:[.,](\d{1,9}))?)?)?/;
+export const offset = /([+\u2212-])([0-1][0-9]|[2][0-3])(?::?([0-5][0-9])(?::?([0-5][0-9])(?:[.,](\d{1,9}))?)?)?/;
```
ptomato, line 10: Small improvement: `[0-1]` can be simplified to `[01]` (since no range is necessary) and `[2]` can be simplified to `2` (since it is only one character, not a choice.)

ptomato (commented):
Very much appreciated, thanks! Would you be able to add a regression test for this as well? `polyfill/test/timezone.mjs` would be a good place to add that.

akshgpt7:
> Would you be able to add a regression test for this as well?
Sure! On it.

polyfill/lib/regex.mjs
akshgpt7: Thanks, makes sense, I'll make the change.

Ms2ger:
This will also need a rebase

akshgpt7:
I'll rebase and squash the commits once I'm done with writing the test.

ryzokuken (approved):

PR #1089, opened 2020-11-02
add github actions workflows

ryzokuken:
Only one action remains to be ported, which is WIP. It'll need me to check out how GitHub Secrets work.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,45 @@
+name: Run tests
+on: pull_request
+jobs:
+  test-polyfill:
+    strategy:
+      matrix:
+        node: [10.x, 12.x, 14.x, 15.x]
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,45 @@
+name: Run tests
+on: pull_request
+jobs:
+  test-polyfill:
+    strategy:
+      matrix:
+        node: [10.x, 12.x, 14.x, 15.x]
+        os: [ubuntu-latest, windows-latest, macos-latest]
```
Ms2ger, line 8: Is this running everything 12 times? That sounds like overkill.

ryzokuken: What about dropping 10.x?

Ms2ger (approved):
Let's try.

PR #1092, opened 2020-11-02
Install python in the test262 Action.

Ms2ger:


.github/workflows/test.yml
```diff
@@ -15,6 +15,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
+    - name: Setup python
```
ryzokuken, line 18:
```suggestion
    - name: use python v3.9
```

nit, for consistency.

ryzokuken (approved):
Thanks!

ryzokuken (approved):

PR #1103, opened 2020-11-03
ZonedDateTime hoursInDay, startOfDay, round, add, subtract

ptomato:
See: #569

### DIFF
--- a/polyfill/lib/ecmascript.mjs
+++ b/polyfill/lib/ecmascript.mjs
@@ -2418,6 +2418,19 @@ export const ES = ObjectAssign({}, ES2020, {
     ES.RejectDuration(years, months, weeks, days, hours, minutes, seconds, milliseconds, microseconds, nanoseconds);
     return { years, months, weeks, days, hours, minutes, seconds, milliseconds, microseconds, nanoseconds };
   },
+  AddInstant: (epochNanoseconds, h, min, s, ms, µs, ns) => {
+    let sum = bigInt(0);
+    sum = sum.plus(bigInt(ns));
+    sum = sum.plus(bigInt(µs).multiply(1e3));
+    sum = sum.plus(bigInt(ms).multiply(1e6));
+    sum = sum.plus(bigInt(s).multiply(1e9));
+    sum = sum.plus(bigInt(min).multiply(60 * 1e9));
+    sum = sum.plus(bigInt(h).multiply(60 * 60 * 1e9));
+
+    const result = bigInt(epochNanoseconds).plus(sum);
+    ES.RejectInstantRange(result);
+    return result;
+  },
   RoundNumberToIncrement: (quantity, increment, mode) => {
     const quotient = quantity / increment;
     let round;
--- a/polyfill/lib/ecmascript.mjs
+++ b/polyfill/lib/ecmascript.mjs
@@ -2431,6 +2431,72 @@ export const ES = ObjectAssign({}, ES2020, {
     ES.RejectInstantRange(result);
     return result;
   },
+  AddZonedDateTime: (instant, timeZone, calendar, years, months, weeks, days, h, min, s, ms, µs, ns, overflow) => {
+    // If only time is to be added, then use Instant math. It's not OK to fall
+    // through to the date/time code below because compatible disambiguation in
+    // the PlainDateTime=>Instant conversion will change the offset of any
+    // ZonedDateTtime in the repeated clock time after a backwards transition.
+    // When adding/subtracting time units and not dates, this disambiguation is
+    // not expected and so is avoided below via a fast path for time-only
+    // arithmetic.
+    // BTW, this behavior is similar in spirit to offset: 'prefer' in `with`.
+    if (ES.DurationSign(years, months, weeks, days, 0, 0, 0, 0, 0, 0) === 0) {
+      return ES.AddInstant(GetSlot(instant, EPOCHNANOSECONDS), h, min, s, ms, µs, ns);
+    }
+
+    // RFC 5545 requires the date portion to be added in calendar days and the
+    // time portion to be added in exact time.
+    // FIXME: "op" and the dateAdd/dateSubtract conditional will not be needed
+    // after #993 lands, changing the order of operations to be the same for
+    // both addition and subtraction.
+    let dt = ES.GetTemporalDateTimeFor(timeZone, instant, calendar);
+    const TemporalDate = GetIntrinsic('%Temporal.Date%');
+    const datePart = new TemporalDate(GetSlot(dt, ISO_YEAR), GetSlot(dt, ISO_MONTH), GetSlot(dt, ISO_DAY), calendar);
+    const addedDate = calendar.dateAdd(datePart, { years, months, weeks, days }, { overflow }, TemporalDate);
+    const TemporalDateTime = GetIntrinsic('%Temporal.DateTime%');
+    const dtIntermediate = new TemporalDateTime(
+      GetSlot(addedDate, ISO_YEAR),
+      GetSlot(addedDate, ISO_MONTH),
+      GetSlot(addedDate, ISO_DAY),
+      GetSlot(dt, ISO_HOUR),
+      GetSlot(dt, ISO_MINUTE),
+      GetSlot(dt, ISO_SECOND),
+      GetSlot(dt, ISO_MILLISECOND),
+      GetSlot(dt, ISO_MICROSECOND),
+      GetSlot(dt, ISO_NANOSECOND),
+      calendar
+    );
+
+    // Note that 'compatible' is used below because this disambiguation behavior
+    // is required by RFC 5545.
+    const instantIntermediate = ES.GetTemporalInstantFor(timeZone, dtIntermediate, 'compatible');
+    return ES.AddInstant(GetSlot(instantIntermediate, EPOCHNANOSECONDS), h, min, s, ms, µs, ns);
+  },
+  SubtractZonedDateTime: (instant, timeZone, calendar, years, months, weeks, days, h, min, s, ms, µs, ns, overflow) => {
+    // FIXME: to be consolidated into AddZonedDateTime by #993
+    if (ES.DurationSign(years, months, weeks, days, 0, 0, 0, 0, 0, 0) === 0) {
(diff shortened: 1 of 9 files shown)


polyfill/lib/zoneddatetime.mjs
```diff
@@ -298,7 +298,15 @@ export class ZonedDateTime {
   }
   startOfDay() {
     if (!ES.IsTemporalZonedDateTime(this)) throw new TypeError('invalid receiver');
-    throw new Error('startOfDay() not implemented yet');
+    const dt = dateTime(this);
+    const DateTime = GetIntrinsic('%Temporal.DateTime%');
+    const dtStart = new DateTime(GetSlot(dt, ISO_YEAR), GetSlot(dt, ISO_MONTH), GetSlot(dt, ISO_DAY), 0, 0, 0, 0, 0, 0);
```
justingrant, line 456: For non-ISO time calendars, is all-zero time ISO time units guaranteed to be the start of the day?  If so then this code is good. If not, then do you need to ask the calendar when the start of the day is?

polyfill/lib/zoneddatetime.mjs
```diff
@@ -128,7 +128,51 @@ export class ZonedDateTime {
...
+    const year = GetSlot(dt, ISO_YEAR);
+    const month = GetSlot(dt, ISO_MONTH);
+    const day = GetSlot(dt, ISO_DAY);
+    const today = new DateTime(year, month, day, 0, 0, 0, 0, 0, 0);
+    const tomorrowFields = ES.AddDate(year, month, day, 0, 0, 0, 1, 'reject');
+    const tomorrow = new DateTime(tomorrowFields.year, tomorrowFields.month, tomorrowFields.day, 0, 0, 0, 0, 0, 0);
+    const timeZone = GetSlot(this, TIME_ZONE);
+    const todayNs = GetSlot(ES.GetTemporalInstantFor(timeZone, today, 'compatible'), EPOCHNANOSECONDS);
+    const tomorrowNs = GetSlot(ES.GetTemporalInstantFor(timeZone, tomorrow, 'compatible'), EPOCHNANOSECONDS);
+    let { seconds, milliseconds, microseconds, nanoseconds } = ES.DifferenceInstant(
```
justingrant, line 142: Why is all the code below here needed? Could it be replaced with something like this?
```js
return Number(tomorrowNs - todayNs) / 3.6e12;
```

polyfill/test/zoneddatetime.mjs
```diff
@@ -1064,6 +1064,30 @@ describe('ZonedDateTime', () => {
...
+    });
+    it('hoursInDay works with non-hour DST change', () => {
+      const zdt1 = ZonedDateTime.from('2020-10-04T12:00[Australia/Lord_Howe]');
+      equal(zdt1.hoursInDay, 23.5);
+      const zdt2 = ZonedDateTime.from('2020-04-05T12:00[Australia/Lord_Howe]');
+      equal(zdt2.hoursInDay, 24.5);
+    });
+    it('hoursInDay works with non-half-hour DST change', () => {
+      const zdt = ZonedDateTime.from('1933-01-01T12:00[Asia/Singapore]');
+      assert(Math.abs(zdt.hoursInDay - 23.6666666666666666) < Number.EPSILON);
```
justingrant, line 1396: I had no idea that `Number.EPSILON` existed.  Very useful. Thanks for the education!

polyfill/lib/zoneddatetime.mjs
```diff
@@ -309,9 +309,78 @@ export class ZonedDateTime {
...
+      millisecond,
+      microsecond,
+      nanosecond,
+      calendar
+    );
+
+    const timeZone = GetSlot(this, TIME_ZONE);
+    const offsetNs = ES.GetOffsetNanosecondsFor(timeZone, GetSlot(this, INSTANT));
+
+    // See the 'prefer' algorithm in ToTemporalZonedDateTime
```
justingrant, line 408: I don't fully follow this logic here, but as long as tests pass I assume it's fine.

polyfill/test/zoneddatetime.mjs
```diff
@@ -730,6 +730,185 @@ describe('ZonedDateTime', () => {
...
+    });
+    it('accepts plural units', () => {
+      assert(zdt.round({ smallestUnit: 'hours' }).equals(zdt.round({ smallestUnit: 'hour' })));
+      assert(zdt.round({ smallestUnit: 'minutes' }).equals(zdt.round({ smallestUnit: 'minute' })));
+      assert(zdt.round({ smallestUnit: 'seconds' }).equals(zdt.round({ smallestUnit: 'second' })));
+      assert(zdt.round({ smallestUnit: 'milliseconds' }).equals(zdt.round({ smallestUnit: 'millisecond' })));
+      assert(zdt.round({ smallestUnit: 'microseconds' }).equals(zdt.round({ smallestUnit: 'microsecond' })));
+      assert(zdt.round({ smallestUnit: 'nanoseconds' }).equals(zdt.round({ smallestUnit: 'nanosecond' })));
+    });
+    it.skip('rounds correctly to a 25-hour day', () => {
```
justingrant, line 1028: Very smart to put skipped tests in for these cases.

polyfill/lib/ecmascript.mjs
```diff
@@ -2431,6 +2431,72 @@ export const ES = ObjectAssign({}, ES2020, {
...
+    // not expected and so is avoided below via a fast path for time-only
+    // arithmetic.
+    // BTW, this behavior is similar in spirit to offset: 'prefer' in `with`.
+    if (ES.DurationSign(years, months, weeks, days, 0, 0, 0, 0, 0, 0) === 0) {
+      return ES.AddInstant(GetSlot(instant, EPOCHNANOSECONDS), h, min, s, ms, µs, ns);
+    }
+
+    // RFC 5545 requires the date portion to be added in calendar days and the
+    // time portion to be added in exact time.
+    // FIXME: "op" and the dateAdd/dateSubtract conditional will not be needed
```
justingrant, line 2449: This comment doesn't match the code below. What conditional?

polyfill/test/zoneddatetime.mjs
```diff
@@ -1291,6 +1423,71 @@ describe('ZonedDateTime', () => {
...
+      equal(result.toString(), '2020-03-01T00:00:00-08:00[America/Los_Angeles]');
+      equal(breakoutUnits('add', zdt, d, options).toString(), result.toString());
+    });
+    it('order of operations: add / reject', () => {
+      const zdt = ZonedDateTime.from('2020-01-31T00:00-08:00[America/Los_Angeles]');
+      const d = Temporal.Duration.from({ months: 1, days: 1 });
+      const options = { overflow: 'reject' };
+      throws(() => zdt.add(d, options), RangeError);
+    });
+    it.skip('order of operations: subtract / none', () => {
```
justingrant, line 1467: Thanks for these tests. Will be helpful in validating #993 fix.

justingrant (approved):
Looks great. It's so cool to see ZDT coming alive!  I made a few minor suggestions and notes. Nothing major.

polyfill/lib/zoneddatetime.mjs
justingrant: Or is there some non-ISO time calendar reason to want to do it the way you did?

Ms2ger (approved):
Thanks for the careful review, @justingrant. It seems like none of your/my comments particularly justify holding this up, so I'll merge now, and let @ptomato deal with them in the next PR.

polyfill/lib/zoneddatetime.mjs
ptomato: I believe that this is OK because we stipulated that non-ISO calendars must still define days to be the same 24-hour periods as the ISO calendar.

polyfill/test/zoneddatetime.mjs
ptomato: I'd actually prefer if the test framework provided an `assert.almostEqual()` for floats, as Jasmine does, but this works well enough.

polyfill/test/zoneddatetime.mjs
ptomato: I don't know if you remember, but you wrote them... :smile:

polyfill/lib/zoneddatetime.mjs
ptomato: Good point. We still do need to calculate the quotient and remainder separately because BigInt division is integer division, but this can be simplified a lot.

polyfill/lib/ecmascript.mjs
ptomato: Sorry, left over from a previous version where I was trying to keep both AddZonedDateTime and SubtractZonedDateTime consolidated into one function. I'll remove this.

polyfill/lib/zoneddatetime.mjs
justingrant: Why not just convert the difference to a `Number` and then use floating-point `Number` math?  If a "day" is longer than 140 24-hour days, some precision loss seems OK. ;-)

polyfill/lib/ecmascript.mjs
justingrant: Cool. Please leave the note about #993 though!  Will be helpful for me to find places to fix along with #993. I'm assuming that these AOs should be consolidated the #993 fix-- LMK if not.

justingrant: BTW, I'm hoping to finish the #993 fix today. Tomorrow at latest.

ptomato:
Thanks for the reviews, I've got fixes for all of these on my next branch.

polyfill/test/zoneddatetime.mjs
justingrant: Haha, I totally forgot. You should have taken the credit. I would have never remembered! ;-)

polyfill/lib/ecmascript.mjs
ptomato: There is still a note about #993 in SubtractZonedDateTime, so that should be fine.

polyfill/lib/zoneddatetime.mjs
ptomato: In the spec text, the epoch nanoseconds are mathematical values (ℝ values) so you have to be explicit about when you convert them to Number values, my guess is that this is the cleanest way.

PR #1108, opened 2020-11-03
Drop the integer argument to YearMonth.toDateOnDay and rename.

Ms2ger:
See #889.


spec/yearmonth.html
```diff
@@ -408,20 +408,27 @@ <h1>Temporal.YearMonth.prototype.valueOf ( )</h1>
       </emu-alg>
     </emu-clause>
 
-    <emu-clause id="sec-temporal.yearmonth.prototype.todateonday">
-      <h1>Temporal.YearMonth.prototype.toDateOnDay ( _day_ )</h1>
+    <emu-clause id="sec-temporal.yearmonth.prototype.todate">
+      <h1>Temporal.YearMonth.prototype.toDate ( _item_ )</h1>
       <p>
-        The `toDateOnDay` method takes one argument _day_.
+        The `toDate` method takes one argument _day_.
```
ptomato, line 414:
```suggestion
        The `toDate` method takes one argument _item_.
```

ptomato (approved):
Looks good. One change left to do in the spec text, but I don't think that needs to block this PR, so merging it now.

Issue #1072, opened 2020-10-29
Add "Plain" prefix to Date, Time, DateTime, YearMonth, MonthDay types (e.g. `PlainDate`)


justingrant:
Per the decisions in #707, we will be adding a `Plain` prefix to Temporal's unzoned date and/or time types. (The date/time type with a time zone will be called `ZonedDateTime`.) 

Adding this prefix is a straightforward but labor-intensive change, so it's a great first issue for someone who wants to help out.

Specific changes needed: 

Date => PlainDate
Time => PlainTime
DateTime => PlainDateTime
YearMonth => PlainYearMonth
MonthDay => PlainMonthDay

Changes need to be made in the polyfill code, polyfill tests, Temporal docs, cookbook sample, and the Temporal spec. 

You can read #707 for the history behind these changes, but here's the TL;DR: 
* We don't want to encourage users (via a short, prefixless name) to use an unzoned date/time type. Instead, we want users to think about their use case and decide on an zoned or unzoned date/time type without the names putting a thumb on the scale. Therefore, we'll prefix the unzoned date/time type.
* For consistency, we'll apply the same prefix to all other unzoned date or time types: Date, Time, YearMonth, MonthDay
* We came up with a short list of prefixes for the community to consider, and `Plain` won by a 3:1 margin over other possible prefixes choices in a [166-vote Twitter poll](https://twitter.com/justingrantjg/status/1316190780373958656).

juandopazo:
Joda and Threeten call a `PlainDate` a [`LocalDate`](https://www.threeten.org/articles/local-date.html), should Temporal follow that?

PR #1118, opened 2020-11-04
Rename types with plain prefix

ryzokuken:
Fixes: #1072 

- [x] Rename conversion methods, e.g. toDate=>toPlainDate
- [ ] Rename now methods
- [ ] Rename the types themselves
- [ ] Mop-up changes that are not publicly visible (e.g. renaming abstract operations)


ptomato (approved):
:+1:

PR #1129, opened 2020-11-05
Date/Time.toZonedDateTime

Ms2ger:
See #889.


Ms2ger:
Docs will need an update too.

justingrant (approved):
Looks great!

ptomato (approved):
I rebased these and it was straightforward. :+1:

ptomato:
Let's keep the docs change for a follow-up.

justingrant:
I'm already planning to do the TS types updates corresponding to this and similar PRs, so I can do the docs updates too as part of that.

PR #1135, opened 2020-11-05
plain rename: finally rename the classes

ryzokuken:
Rename the classes from `Foo` => `PlainFoo`. Tests pass, but the docs might've not be perfect. Let's deal with those as we find them.


ptomato (approved):
tl;dr

ryzokuken:
> tl;dr

mood rn

PR #1137, opened 2020-11-05
polyfill: add Temporal.now.plainTime

ryzokuken:
Add polyfill code and relevant tests for `Temporal.now.plainTime`.


ptomato (approved):
short and sweet

PR #1141, opened 2020-11-06
Fix spec lint issues

justingrant:
These problems were preventing builds from completing, which in turn prevented the Temporal website from updating.

The long-term solution is to ensure that ecmarkup must pass for a PR to merge. In the meantime, here's a PR to get builds working again.


ptomato (approved):
Thanks for fixing.

PR #1151, opened 2020-11-06
Lint the spec in CI.

Ms2ger:


justingrant (approved):
Are the TS types checked too here? If not they should be.

Ms2ger:
Dunno re: TS

ryzokuken (approved):

ryzokuken:
> Are the TS types checked too here? If not they should be.

`npm run lint` takes care of that. It runs `eslint` as well as `tscheck` IIRC.

PR #1153, opened 2020-11-06
ZonedDateTime.toLocaleString() and last cookbook example

ptomato:
See: #569

### DIFF
--- a/docs/cookbook/meetingPlanner.js
+++ b/docs/cookbook/meetingPlanner.js
@@ -1,19 +1,15 @@
 // Display local time zone and three others
-const here = Temporal.now.timeZone();
-const now = Temporal.now.instant();
+const browserCalendar = new Intl.DateTimeFormat().resolvedOptions().calendar;
+const now = Temporal.now.zonedDateTime(browserCalendar);
 const timeZones = [
-  { name: 'Here', tz: here },
+  { name: 'Here', tz: now.timeZone },
   { name: 'New York', tz: Temporal.TimeZone.from('America/New_York') },
   { name: 'London', tz: Temporal.TimeZone.from('Europe/London') },
   { name: 'Tokyo', tz: Temporal.TimeZone.from('Asia/Tokyo') }
 ];
 
 // Start the table at midnight local time
-const browserCalendar = new Intl.DateTimeFormat().resolvedOptions().calendar;
-const calendarNow = now.toPlainDateTime(here, browserCalendar);
-const startTime = calendarNow
-  .with(Temporal.PlainTime.from('00:00')) // midnight
-  .toInstant(here);
+const startTime = now.startOfDay();
 
 // Build the table
 const table = document.getElementById('meeting-planner');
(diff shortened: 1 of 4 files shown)


polyfill/lib/intl.mjs
```diff
@@ -319,15 +358,39 @@ function extractOverrides(temporalObj, main) {
...
+      instant: main[TZ_RESOLVED].getInstantFor(datetime),
       formatter: main[DATETIME]
     };
   }
 
+  if (ES.IsTemporalZonedDateTime(temporalObj)) {
+    const calendar = GetSlot(temporalObj, CALENDAR);
+    if (calendar.id !== 'iso8601' && calendar.id !== main[CAL_ID]) {
+      throw new RangeError(
+        `cannot format ZonedDateTime with calendar ${calendar.id} in locale with calendar ${main[CAL_ID]}`
```
Ms2ger, line 370: Use `CalendarToString` (and make sure to call it just once.)

spec/intl.html
```diff
@@ -315,11 +315,16 @@ <h1>PartitionDateTimePattern ( _dateTimeFormat_, _x_ )</h1>
             1. <ins>Let _pattern_ be _dateTimeFormat_.[[TemporalDateTimePattern]].</ins>
             1. <ins>Let _tm_ be { [[weekday]]: ! ToDayOfWeek(_x_.[[ISOYear]], _x_.[[ISOMonth]], _x_.[[ISODay]]), [[year]]: _x_.[[ISOYear]], [[month]]: _x_.[[ISOMonth]], [[day]]: _x_.[[ISODay]], [[hour]]: _x_.[[ISOHour]], [[minute]]: _x_.[[ISOMinute]], [[second]]: _x_.[[ISOSecond]] }.</ins>
           1. <ins>If _date_ has an [[InitializedTemporalZonedDateTime]] internal slot, then</ins>
-            1. <ins>Let _calendar_ be ? CalendarToString(_x_).</ins>
+            1. <ins>Let _timeZone_ be ? TimeZoneToString(_x_.[[TimeZone]]).</ins>
+            1. <ins>If _dateTimeFormat_.[[TimeZone]] is not equal to DefaultTimeZone(), and _timeZone_ is not equal to _dateTimeFormat_.[[TimeZone]], then</ins>
+              1. <ins>Throw a *RangeError* exception.</ins>
+            1. <ins>Let _calendar_ be ? CalendarToString(_x_.[[Calendar]]).</ins>
             1. <ins>If _calendar_ is not *"iso8601"* and not equal to _dateTimeFormat_.[[Calendar]], then</ins>
               1. <ins>Throw a *RangeError* exception.</ins>
```
Ms2ger, line 323: Is the polyfill doing these checks in the opposite order?

Ms2ger (changes requested):

polyfill/lib/intl.mjs
ptomato: Good point, I've added another commit to fix the existing instances of this problem as well.

docs/cookbook/meetingPlanner.js
```diff
@@ -1,46 +1,43 @@
...
+    cell.className = `time-${columnTime.hour}`;
 
     // Highlight the current hour in each row
-    if (hours === calendarNow.hour) cell.className += ' time-current';
+    if (hours === now.hour) cell.className += ' time-current';
 
     // Show the date in midnight cells
     let formatOptions;
-    if (dt.hour === 0) {
+    if (columnTime.hour === 0) {
```
justingrant, line 35: I think you probably want to compare columnTime to `.startOfDay` here, not zero, if you want to handle cases where DST skips an hour starting at midnight.

justingrant (approved):
Looks good. I admittedly don't know much about the innards of toLocaleString so I focused my review on the cookbook sample and the high-level expectations like throwing if zones are mismatched.

Ms2ger (approved):

docs/cookbook/meetingPlanner.js
ptomato: That's a good point, it does handle other skipped and repeated times gracefully, so it should here as well.

PR #1156, opened 2020-11-07
Specification for ZonedDateTime.until/since

ptomato:
See: #569

### DIFF
--- a/spec/zoneddatetime.html
+++ b/spec/zoneddatetime.html
@@ -673,8 +673,18 @@
         1. Let _roundingMode_ be ? ToTemporalRoundingMode(_options_).
         1. Let _maximum_ be ! MaximumTemporalDurationRoundingIncrement(_smallestUnit_).
         1. Let _roundingIncrement_ be ? ToTemporalRoundingIncrement(_options_, _maximum_, *false*).
-        1. <mark>TODO</mark>.
-        1. Return ? CreateTemporalDuration(<mark>TODO</mark>).
+        1. If _largestUnit_ is not one of *"years"*, *"months"*, *"weeks"*, or *"days"*, then
+          1. Let _differenceNs_ be ? DifferenceInstant(_zonedDateTime_.[[Nanoseconds]], _other_.[[Nanoseconds]], _roundingIncrement_, _smallestUnit_, _roundingMode_).
+          1. Let _balanceResult_ be ! BalanceDuration(0, 0, 0, 0, 0, 0, _differenceNs_, _largestUnit_).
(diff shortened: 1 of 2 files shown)


spec/zoneddatetime.html
```diff
@@ -673,8 +673,18 @@ <h1>Temporal.ZonedDateTime.prototype.until ( _other_ [ , _options_ ] )</h1>
         1. Let _roundingMode_ be ? ToTemporalRoundingMode(_options_).
         1. Let _maximum_ be ! MaximumTemporalDurationRoundingIncrement(_smallestUnit_).
         1. Let _roundingIncrement_ be ? ToTemporalRoundingIncrement(_options_, _maximum_, *false*).
-        1. <mark>TODO</mark>.
-        1. Return ? CreateTemporalDuration(<mark>TODO</mark>).
+        1. If _largestUnit_ is not one of *"years"*, *"months"*, *"weeks"*, or *"days"*, then
+          1. Let _difference_ be ? DifferenceInstant(_zonedDateTime_.[[Nanoseconds]], _other_.[[Nanoseconds]], _roundingIncrement_, _smallestUnit_, _roundingMode_).
+          1. Let _balanceResult_ be ! BalanceDuration(0, 0, 0, _difference_.[[Seconds]], _difference_.[[Milliseconds]], _difference_.[[Microseconds]], _difference_.[[Nanoseconds]], _largestUnit_).
```
Ms2ger, line 678: Doesn't DifferenceInstant return a number?

spec/zoneddatetime.html
```diff
@@ -673,8 +673,18 @@ <h1>Temporal.ZonedDateTime.prototype.until ( _other_ [ , _options_ ] )</h1>
...
+          1. Let _result_ be the new Record { [[Years]]: 0, [[Months]]: 0, [[Weeks]]: 0, [[Days]]: _balanceResult_.[[Days]], [[Hours]]: _balanceResult_.[[Hours]], [[Minutes]]: _balanceResult_.[[Minutes]], [[Seconds]]: _balanceResult_.[[Seconds]], [[Milliseconds]]: _balanceResult_.[[Milliseconds]], [[Microseconds]]: _balanceResult_.[[Microseconds]], [[Nanoseconds]]: _balanceResult_.[[Nanoseconds]] }.
+        1. Else,
+          1. If ? TimeZoneEquals(_zonedDateTime_.[[TimeZone]], _other_.[[TimeZone]]) is *false*, then
+            1. Throw a *RangeError* exception.
```
Ms2ger, line 682: Hmm, it seems a little strange to allow different time zones in the other branch, though it does indeed work. Probably fine.

Ms2ger (changes requested):

spec/zoneddatetime.html
ptomato: I double-checked the docs and this is how it's supposed to work. (The rationale is that the time zone DST rules don't matter for the answer in the time-units-only case, only the time zone offset; so it shouldn't throw.)

justingrant: @ptomato is correct. ZDT math (following RFC 5545) always treats hours and smaller units as exact time (like Intsant) while treating days and larger units as calendar dates (like PlainDateTime).  The difference in behavior here stems from this. If units are hours or less, then there's no ambiguity (and hence no exception). It's also why `'hours'` is the default for these operations, so that the default options won't throw.

Ms2ger (approved):

PR #1158, opened 2020-11-07
Fix rounding in ZonedDateTime.toString()

ptomato:
See: #569


justingrant (approved):
This is exactly the right way to do it. Smart test cases too.

Ms2ger (approved):

PR #1159, opened 2020-11-07
Change default of until/since rounding to 'trunc'

### LINKED ISSUE #1122: Default for until/since roundingMode should be 'trunc' not 'nearest'
While implementing a fix to #993, I noticed that the default rounding mode for `until` and `since` is `'nearest'`.

My intent in #827 was that the default rounding mode for difference would be `'trunc'` but I never actually wrote it down in the proposal. Doh!

I'm happy to PR a fix for this tomorrow. 

Justification: 

* Now that we have a `round` method, that's the place where I'd expect rounding to happen.
* Developers will depend on this method for use cases like billing where rounding up will be unexpected. Over-billing is bad!
* It's a common use case to use difference for "flipping", e.g. has time since some event flipped over to a new hour/minute/second/etc in order to display dynamic relative UI. If the flip happens in the middle of a day/hour/minute that's unexpected.

justingrant:
Updates the default rounding mode for until/since to match user expectations that by default these methods measure "whole units" not partial remainders.

Docs are also updated, including fixing code example results for until/since, many of which were wrong due to this PR as well as other recent changes to `until`/`since` behavior.

BTW, I ran Test262 locally and all tests passed.

Fixes #1122.

### DIFF
--- a/docs/date.md
+++ b/docs/date.md
@@ -318,7 +318,7 @@ date = Temporal.PlainDate.from('2006-01-24');
 date.with({ day: 1 }); // => 2006-01-01
 // What's the last day of the next month?
 const nextMonthDate = date.add({ months: 1 });
-nextMonthDate.with({day: nextMonthDate.daysInMonth }); // => 2006-02-28
+nextMonthDate.with({ day: nextMonthDate.daysInMonth }); // => 2006-02-28
 ```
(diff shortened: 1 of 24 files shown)


ptomato (approved):
:+1:

PR #1160, opened 2020-11-07
Fix Instant's _repr_ to use Z suffix not +00:00

justingrant:
Fixes #1110 by switching from using the `'UTC'` timezone to an `undefined` timezone when calling `ES.TemporalInstantToString`.

With the fix: 
image


ptomato (approved):
:+1:

ryzokuken (approved):

PR #1162, opened 2020-11-08
Update docs type diagram with PlainXxx names

justingrant:
This PR updates the object diagram on the Temporal home page with new PlainXxx type names.


ptomato (approved):
Thanks!

PR #1174, opened 2020-11-11
Update *FromFields calls.

Ms2ger:


ptomato (approved):
Great, this makes a lot of calendar things more readable and fills many gaps in the spec.

PR #1179, opened 2020-11-12
polyfill: durations accept fractional time values

ryzokuken:
Fixes: https://github.com/tc39/proposal-temporal/issues/938

/cc @justingrant @ptomato


polyfill/lib/regex.mjs
```diff
@@ -33,4 +33,8 @@ export const time = new RegExp(`^${timesplit.source}(?:${zonesplit.source})?(?:$
 export const yearmonth = new RegExp(`^(${yearpart.source})-?(\\d{2})$`);
 export const monthday = /^(?:--)?(\d{2})-?(\d{2})$/;
 
-export const duration = /^([+\u2212-])?P(?:(\d+)Y)?(?:(\d+)M)?(?:(\d+)W)?(?:(\d+)D)?(?:T(?!$)(?:(\d+)H)?(?:(\d+)M)?(?:(\d+)(?:[.,](\d{1,9}))?S)?)?$/i;
+const fraction = /(\d+)(?:[.,](\d+))?/;
```
justingrant, line 36: We should decide if 10+ fractional digits are OK or not, per comment above.

polyfill/test/duration.mjs
```diff
@@ -139,21 +139,37 @@ describe('Duration', () => {
...
+    it('decimal places only allowed in time units', () => {
       [
         'P0.5Y',
         'P1Y0,5M',
         'P1Y1M0.5W',
         'P1Y1M1W0,5D',
-        'P1Y1M1W1DT0.5H',
-        'P1Y1M1W1DT1H0,5M',
-        'P1Y1M1W1DT1H0.5M0.5S'
+        { years: 0.5 },
```
justingrant, line 153: IMHO, all these rounding tests should be done with both positive and negative durations, because a frequent cause of rounding errors in libraries is differences in handling negative vs. positive numbers.

justingrant (changes requested):
Looks good, with one question and a few suggestions re: test cases.

polyfill/lib/ecmascript.mjs
```diff
@@ -793,15 +918,15 @@ export const ES = ObjectAssign({}, ES2020, {
     if (validUnits.indexOf(unit1) > validUnits.indexOf(unit2)) return unit2;
     return unit1;
   },
-  ToPartialRecord: (bag, fields) => {
+  ToPartialRecord: (bag, fields, cast = ES.ToInteger) => {
```
ptomato, line 921: I'm not sure the spec language has first-class abstract operations like this, but, maybe consider instead splitting `BUILTIN_FIELDS` into `BUILTIN_DATETIME_FIELDS` and `DURATION_FIELDS`, and using `ToNumber` on duration fields?

ptomato (approved):
Looking good, the main change is that I think we should throw on more than our supported maximum number of decimal places, as per ISO 8601.

polyfill/lib/ecmascript.mjs
```diff
@@ -293,12 +295,37 @@ export const ES = ObjectAssign({}, ES2020, {
...
+    let nanoseconds = ES.ToInteger(fSeconds.slice(6, 9)) * sign;
+
+    fHours = fHours ? (sign * ES.ToInteger(fHours)) / 10 ** fHours.length : 0;
+    fMinutes = fMinutes ? (sign * ES.ToInteger(fMinutes)) / 10 ** fMinutes.length : 0;
+
+    const time = ES.DurationHandleFractions(
+      fHours,
+      minutes,
+      fMinutes,
+      seconds,
```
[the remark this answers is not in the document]
ryzokuken: > (Although then you do need to ensure that the seconds are not given in the string, if minutes are fractional.)

Exactly why I did it this way.

polyfill/lib/ecmascript.mjs
```diff
@@ -395,6 +422,78 @@ export const ES = ObjectAssign({}, ES2020, {
...
+        fMilliseconds,
+        microseconds,
+        fMicroseconds,
+        nanoseconds,
+        fNanoseconds
+      ].forEach((val) => {
+        if (val !== 0) throw new RangeError('only the smallest unit can be fractional');
+      });
+      let mins = fHours * 60;
+      minutes = ES.ToInteger(mins);
```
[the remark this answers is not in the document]
ryzokuken: done.

ptomato:
I noticed from the CI output that the "order of operations" test262 tests supply non-integer input, those are breaking now instead of truncating it.

littledan:
As I explained in https://github.com/tc39/proposal-temporal/issues/938#issuecomment-726896052 , I think this PR should be dropped.

ptomato:
This PR still implements support for ISO strings such as `PT0.5H`, which is unrelated to the Number concern, I think?

spec/abstractops.html
```diff
@@ -92,6 +92,44 @@ <h1>GetStringOrNumberOption ( _options_, _property_, _stringValues_, _minimum_,
...
   </emu-clause>
 
+
+  <emu-clause id="sec-temporal-durationhandlefractions" aoid="DurationHandleFractions">
+    <h1>DurationHandleFractions ( _fHours_, _minutes_, _fMinutes_, _seconds_, _fSeconds_, _milliseconds_, _fMilliseconds_, _microseconds_, _fMicroseconds_, _nanoseconds_, _fNanoseconds_ )</h1>
+    <emu-alg>
+      1. If _fHours_ is not equal to 0, then
+        1. If either of _minutes_, _fMinutes_, _seconds_, _fSeconds_, _milliseconds_, _fMilliseconds_, _microseconds_, _fMicroseconds_, _nanoseconds_, _fNanoseconds_ is not 0, throw a *RangeError* exception.
+        1. Let _mins_ be _fHours_ × 60.
+        1. Set _minutes_ to the integral Number nearest _mins_ in the direction of +0𝔽.
```
ptomato, line 102:
```suggestion
        1. Set _minutes_ to the integral Number nearest to _mins_ in the direction of +0<sub>𝔽</sub>.
```
(also below)

(although I think you can also do more simply `Set _minutes_ to floor(_mins_).`)

spec/abstractops.html
```diff
@@ -949,10 +987,10 @@ <h1>ISO 8601 grammar</h1>
           Digits TimeFraction? SecondsDesignator
 
       DurationMinutes :
-          Digits MinutesDesignator DurationSeconds?
+          Digits TimeFraction? MinutesDesignator DurationSeconds?
```
ptomato, line 990: I think the grammar should show that the fraction is incompatible with the seconds part:
```
DurationMinutes :
    Digits MinutesDesignator DurationSeconds?
    Digits TimeFraction MinutesDesignator
```

spec/abstractops.html
```diff
@@ -1198,42 +1236,51 @@ <h1>ParseTemporalDurationString ( _isoString_ )</h1>
       1. Assert: Type(_isoString_) is String.
       1. If _isoString_ does not satisfy the syntax of a |TemporalDurationString| (see <emu-xref href="#sec-temporal-iso8601grammar"></emu-xref>), then
         1. Throw a *RangeError* exception.
-      1. Let _sign_, _years_, _months_, _weeks_, _days_, _hours_, _minutes_, _seconds_, and _fraction_ be the parts of _isoString_ produced respectively by the |Sign|, |DurationYears|, |DurationMonths|, |DurationWeeks|, |DurationDays|, |DurationHours|, |DurationMinutes|, |DurationSeconds|, and |TimeFractionalPart| productions, or *undefined* if not present.
+      1. Let _sign_, _years_, _months_, _weeks_, _days_, _hours_, _fHours_, _minutes_, _fMinutes_, _seconds_, and _fSeconds_ be the parts of _isoString_ produced respectively by the |Sign|, |DurationYears|, |DurationMonths|, |DurationWeeks|, |DurationDays|, |DurationHours|, |DurationHoursFraction|, |DurationMinutes|, |DurationMinutesFraction|, |DurationSeconds|, and |DurationSecondsFraction| productions, or *undefined* if not present.
```
ptomato, line 1239: DurationHoursFraction, DurationMinutesFraction, and DurationSecondsFraction need to be defined.

ptomato (approved):
I don't think we should block this PR on any of the remaining comments, since they are minor improvements, so I think punting them to a followup is fine.

spec/abstractops.html
ptomato: The linter was complaining about this line anyway so I'll push a new commit to fix it.

ptomato:
OK, merging this now, we should solve the remaining comments in a follow up soon, ideally when the discussion in #938 is resolved.

spec/abstractops.html
ryzokuken: What we're doing right now is accept a duration with all the fractions, set the units to zero appropriately and handle them that way. We could change the spec text and instead of throwing at a later time, removing all those checks etc.

spec/abstractops.html
ryzokuken: can we use `|TimeFractionalPart|` three times over?

spec/abstractops.html
ptomato: I am not 100% sure, but I think `PT0.5H30M` is actually illegal according to ISO 8601. In the polyfill it doesn't matter whether we throw because it's not accepted by the regex or we throw right after, but I think the spec grammar should match what's a legal string as much as possible.

spec/abstractops.html
ptomato: Yes, I think you can just define them like `DurationHoursFraction : TimeFractionalPart` etc.

spec/abstractops.html
justingrant: Correct. Only the smallest unit is permitted to be fractional in ISO 8601.

ryzokuken: The string still throws, it just throws after the parsing stage (in `DurationHandleFractions`). I thought it was okay as long as we threw on the wrong format.

PR #1191, opened 2020-11-14
Add object diagram and persistence diagram to spec

### LINKED ISSUE #1180: Add object model and persistence diagrams to the spec
Per https://github.com/tc39/proposal-temporal/issues/655#issuecomment-725889571, @ljharb strongly encouraged us to add the persistence model diagram (https://tc39.es/proposal-temporal/docs/persistence-model.svg) to the spec. 

While we're at it, it probably makes sense to add the object model diagram too. https://tc39.es/proposal-temporal/docs/object-model.svg.

Both files are present in the /docs/ folder. 

I'll volunteer to PR this before next week's TC39 meeting.

justingrant:
Fixes #1180.

### DIFF
--- a/docs/object-model.svg
+++ b/docs/object-model.svg
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<svg 
+<svg
   viewBox="116 540 4040 1484"
   width="100%"
   height="100%"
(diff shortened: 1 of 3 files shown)


ptomato (approved):
This seems as good a spot as any, for now! Thanks for filling in the introduction as well.

PR #1201, opened 2020-11-16
Reduce potential precision loss in ZDT `hoursInDay`

justingrant:
Do all calculations using integer math and only do a single integer-division operation at the end.

### DIFF
--- a/polyfill/lib/zoneddatetime.mjs
+++ b/polyfill/lib/zoneddatetime.mjs
@@ -145,8 +145,7 @@ export class ZonedDateTime {
     const timeZone = GetSlot(this, TIME_ZONE);
     const todayNs = GetSlot(ES.GetTemporalInstantFor(timeZone, today, 'compatible'), EPOCHNANOSECONDS);
     const tomorrowNs = GetSlot(ES.GetTemporalInstantFor(timeZone, tomorrow, 'compatible'), EPOCHNANOSECONDS);
-    const { quotient, remainder } = tomorrowNs.subtract(todayNs).divmod(3.6e12);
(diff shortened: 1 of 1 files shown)


polyfill/lib/zoneddatetime.mjs
```diff
@@ -146,7 +146,7 @@ export class ZonedDateTime {
     const todayNs = GetSlot(ES.GetTemporalInstantFor(timeZone, today, 'compatible'), EPOCHNANOSECONDS);
     const tomorrowNs = GetSlot(ES.GetTemporalInstantFor(timeZone, tomorrow, 'compatible'), EPOCHNANOSECONDS);
     const { quotient, remainder } = tomorrowNs.subtract(todayNs).divmod(3.6e12);
-    return quotient.toJSNumber() + remainder.toJSNumber() / 3.6e12;
+    return (quotient.toJSNumber() * 3.6e12 + remainder.toJSNumber()) / 3.6e12;
```
ptomato, line 149: Maybe `quotient.multiply(3.6e12).add(remainder).toJSNumber() / 3.6e12` could be better, to keep everything in BigInt domain until the last division?

justingrant: Or, even easier, AFACT there's no need for the `divmod`, right?
```javascript
return tomorrowNs.subtract(todayNs).toJSNumber() / 3.6e12;
```

Code is updated with this solution.

ptomato: Oh right, that works, tomorrowNs − todayNs is guaranteed to be a safe integer.

ptomato (approved):

PR #1202, opened 2020-11-16
Fill in the remaining gaps in the TimeZone spec

ptomato:


spec/timezone.html
```diff
@@ -213,9 +202,28 @@ <h1>Temporal.TimeZone.prototype.getInstantFor ( _dateTime_ [ , _options_ ] )</h1
...
+        1. If _disambiguation_ is *"earlier"*, then
+          1. Let _earlier_ be ? AddDateTime(_dateTime_.[[ISOYear]], _dateTime_.[[ISOMonth]], _dateTime_.[[ISODay]], _dateTime_.[[ISOHour]], _dateTime_.[[ISOMinute]], _dateTime_.[[ISOSecond]], _dateTime_.[[ISOMillisecond]], _dateTime_.[[ISOMicrosecond]], _dateTime_.[[ISONanosecond]], 0, 0, 0, 0, 0, 0, 0, 0, 0, −_nanoseconds_, *"constrain"*).
+          1. Set _possibleInstants_ to ? GetPossibleInstantsFor(_timeZone_, _earlier_).
+          1. Assert: _possibleInstants_'s length is not 0.
+          1. Return _possibleInstants_[0].
+        1. Assert: _disambiguation_ is *"compatible"* or *"later"*.
+        1. Let _later_ be ? AddDateTime(_dateTime_.[[ISOYear]], _dateTime_.[[ISOMonth]], _dateTime_.[[ISODay]], _dateTime_.[[ISOHour]], _dateTime_.[[ISOMinute]], _dateTime_.[[ISOSecond]], _dateTime_.[[ISOMillisecond]], _dateTime_.[[ISOMicrosecond]], _dateTime_.[[ISONanosecond]], 0, 0, 0, 0, 0, 0, 0, 0, 0, _nanoseconds_, *"constrain"*).
+        1. Set _possibleInstants_ to ? GetPossibleInstantsFor(_timeZone_, _later_).
+        1. Set _n_ to _possibleInstants_'s length.
+        1. Assert: _n_ is not 0.
```
Ms2ger, line 225: You can't know that - `GetPossibleInstantsFor` can return whatever it wants.

spec/timezone.html
```diff
@@ -232,7 +240,16 @@ <h1>Temporal.TimeZone.prototype.getPossibleInstantsFor ( _dateTime_ )</h1>
         1. Let _timeZone_ be the *this* value.
         1. Perform ? RequireInternalSlot(_timeZone_, [[InitializedTemporalTimezone]]).
         1. Set _dateTime_ to ? ToTemporalDateTime(_dateTime_).
-        1. <mark>TODO.</mark>
+        1. If _timeZone_.[[IsOffset]] is *true*, then
+          1. Let _epochNanoseconds_ be ! GetEpochFromParts(_dateTime_.[[ISOYear]], _dateTime_.[[ISOMonth]], _dateTime_.[[ISODay]], _dateTime_.[[ISOHour]], _dateTime_.[[ISOMinute]], _dateTime_.[[ISOSecond]], _dateTime_.[[ISOMillisecond]], _dateTime_.[[ISOMicrosecond]], _dateTime_.[[ISONanosecond]]).
+          1. Let _instant_ be ? CreateTemporalInstant(_epochNanoseconds_ − _timeZone_.[[OffsetNanoseconds]]).
+          1. Return ? CreateArrayFromList(« _instant_ »).
+        1. Let _possibleEpochNanoseconds_ be ? GetIANATimeZoneEpochValue(_timeZone_.[[Identifier]], _dateTime_.[[ISOYear]], _dateTime_.[[ISOMonth]], _dateTime_.[[ISODay]], _dateTime_.[[ISOHour]], _dateTime_.[[ISOMinute]], _dateTime_.[[ISOSecond]], _dateTime_.[[ISOMillisecond]], _dateTime_.[[ISOMicrosecond]], _dateTime_.[[ISONanosecond]]).
+        1. Let _possibleInstants_ be the empty List.
```
Ms2ger, line 248: Nit: "a new empty List" or "« »". (Unlike strings, lists are mutable, so "the" empty list doesn't make sense.)

spec/timezone.html
```diff
@@ -362,12 +401,83 @@ <h1>CreateTemporalTimeZone ( _identifier_ [ , _newTarget_ ] )</h1>
...
         1. If _newTarget_ is not given, set it to %Temporal.TimeZone%.
-        1. Let _object_ be ? OrdinaryCreateFromConstructor(_newTarget_, *"%Temporal.TimeZone.prototype%"*, « [[InitializedTemporalTimeZone]], [[Identifier]] »).
+        1. Let _object_ be ? OrdinaryCreateFromConstructor(_newTarget_, *"%Temporal.TimeZone.prototype%"*, « [[InitializedTemporalTimeZone]], [[Identifier]], [[IsOffset]], [[OffsetNanoseconds]] »).
         1. Set _object_.[[Identifier]] to _identifier_.
+        1. If _identifier_ satisfies the syntax of a |TimeZoneNumericUTCOffset| (see <emu-xref href="#sec-temporal-iso8601grammar"></emu-xref>), then
+          1. Set _object_.[[IsOffset]] to *true*.
+          1. Set _object_.[[OffsetNanoseconds]] to ! ParseTimeZoneOffsetString(_identifier_).
+        1. Else,
+          1. Set _object_.[[IsOffset]] to *false*.
+          1. Set _object_.[[OffsetNanoseconds]] to *undefined*.
```
Ms2ger, line 411: Do you need both internal slots, or does "[[OffsetNanoseconds]] is not **undefined**" work?

Ms2ger (changes requested):

spec/timezone.html
ptomato: Hmm, you're right. I wonder what is best to do in this case. If getPossibleInstantsFor is inconsistent, then I think probably the best thing to do is throw.

spec/timezone.html
ptomato: That works just as well; I was thinking that this might be more understandable, but I've changed it.

spec/timezone.html
ptomato: I got rid of the assertions, in any case.

Ms2ger (approved):

PR #1213, opened 2020-11-20
Make PartitionDateTimePattern calendar-aware.

Ms2ger:
Fixes #591.

---

Does this approach make sense?

### DIFF
--- a/spec/biblio.json
+++ b/spec/biblio.json
@@ -69,6 +69,16 @@
     },
     {
       "type": "op",
+      "aoid": "PartitionPattern",
+      "id": "sec-partitionpattern"
+    },
+    {
+      "type": "op",
+      "aoid": "ResolveLocale",
+      "id": "sec-resolvelocale"
+    },
+    {
+      "type": "op",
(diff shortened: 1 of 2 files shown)


ptomato:
This general approach makes sense to me.

However, I was originally thinking it wouldn't be necessary at all, since Intl.DateTimeFormat already knows how to format a legacy Date in another calendar, so we wouldn't need to Get the properties from the calendar. But, on closer inspection, it seems that that isn't covered in the current edition of 402 at all: https://tc39.es/ecma402/#sec-partitiondatetimepattern By my reading, the month and era formats may be calendar-dependent if they are `'narrow'`, `'short'`, or `'long'`, but anything (including year and day) that is `'numeric'` or `'2-digit'` will always print as the ISO calendar according to the spec text. So upon further investigation this is maybe something that needs to be fixed in 402 separately from Temporal, and then on the Temporal side we can just pass in the epoch nanoseconds value, time zone, and calendar? I'm not sure.

ptomato:
Can be closed because #1218 was merged?

PR #1216, opened 2020-11-23
Clarify ZDT docs for `from` behavior of date/time fields

justingrant:
The ZDT `from` docs incorrectly stated that date/time input was in context of the user-provided calendar, but that's not true for string inputs. 

This PR fixes this mistake. 

Thanks @ryzokuken for finding this!


ptomato (approved):
Thanks!

PR #1218, opened 2020-11-25
Specify Intl.DateTimeFormat based on nanoseconds since the epoch.

Ms2ger:


Ms2ger:
Alternative to #1213 .

spec/intl.html
```diff
@@ -292,46 +292,62 @@ <h1>PartitionDateTimePattern ( _dateTimeFormat_, _x_ )</h1>
...
             1. <ins>Let _pattern_ be _dateTimeFormat_.[[TemporalPlainDatePattern]].</ins>
-            1. <ins>Let _tm_ be { [[weekday]]: ! ToDayOfWeek(_x_.[[ISOYear]], _x_.[[ISOMonth]], _x_.[[ISODay]]), [[year]]: _x_.[[ISOYear]], [[month]]: _x_.[[ISOMonth]], [[day]]: _x_.[[ISODay]] }.</ins>
-          1. <ins>If _x_ has an [[InitializedTemporalYearMonth]] internal slot, then</ins>
             1. <ins>Let _calendar_ be ? CalendarToString(_x_.[[Calendar]]).</ins>
-            1. <ins>If _calendar_ is not equal to _dateTimeFormat_.[[Calendar]], throw a *RangeError* exception.</ins>
+            1. <ins>If _calendar_ is _dateTimeFormat_.[[Calendar]], then</ins>
+              1. <ins>Let _calendarOverride_ be _x_.[[Calendar]].</ins>
+            1. <ins>Else if _calendar_ is *"iso8601"*, then</ins>
+              1. <ins>Let _calendarOverride_ be ? CalendarFrom(_dateTimeFormat_.[[Calendar]]).</ins>
+              1. <mark>Note that this calls `Temporal.Calendar.from`, which may be author code. Desirable?.</mark>
```
ptomato, line 301: Not entirely sure what's correct, but I think you could use GetBuiltinCalendar here? _dateTimeFormat_.[[Calendar]] is always a built-in calendar, and I think we don't want user code to be able to change the behaviour of toLocaleString by monkeypatching Temporal.Calendar.from().

spec/intl.html
```diff
@@ -340,7 +356,7 @@ <h1>PartitionDateTimePattern ( _dateTimeFormat_, _x_ )</h1>
             1. <ins>Set _x_ to ? ToNumber(_x_).</ins>
           1. Set _x_ to TimeClip(_x_).
           1. If _x_ is *NaN*, throw a *RangeError* exception.
-          1. <ins>Let _tm_ be ToLocalTime(_x_, _dateTimeFormat_.[[Calendar]], _dateTimeFormat_.[[TimeZone]]).</ins>
+          1. <ins>Let _epochNanoseconds be ℤ(_x_) × 10<sup>6</sup>.</ins>
```
ptomato, line 359:
```suggestion
          1. <ins>Let _epochNanoseconds_ be ℤ(_x_) × 10<sup>6</sup>.</ins>
```

spec/intl.html
```diff
@@ -495,7 +511,9 @@ <h1>ToLocalTime ( _t_, _calendar_, _timeZone_ )</h1>
       </p>
 
       <emu-alg>
-        1. Assert: Type(_t_) is Number.
+        1. Assert: Type(_t_) is <del>Number</del><ins>BigInt</ins>.
+        1. <ins>Set _t_ to 𝔽(floor(ℝ(_t_) / 10<sup>6</sup>)).</ins>
+        1. <mark>TODO: This probably causes issues with fractional seconds. Review.</mark>
```
ptomato, line 516: Would we have to patch LocalTZA to accept epoch nanoseconds as well?

ptomato (commented):
This looks like a quite nice solution.

spec/intl.html
Ms2ger: Fixed by not using LocalTZA.

spec/intl.html
```diff
@@ -292,46 +292,61 @@ <h1>PartitionDateTimePattern ( _dateTimeFormat_, _x_ )</h1>
...
             1. <ins>Let _pattern_ be _dateTimeFormat_.[[TemporalPlainYearMonthPattern]].</ins>
-            1. <ins>Let _tm_ be { [[year]]: _x_.[[ISOYear]], [[month]]: _x_.[[ISOMonth]] }.</ins>
-          1. <ins>If _x_ has an [[InitializedTemporalMonthDay]] internal slot, then</ins>
             1. <ins>Let _calendar_ be ? CalendarToString(_x_.[[Calendar]]).</ins>
-            1. <ins>If _calendar_ is not equal to _dateTimeFormat_.[[Calendar]], throw a *RangeError* exception.</ins>
+            1. <ins>If _calendar_ is _dateTimeFormat_.[[Calendar]], then</ins>
+              1. <ins>Let _calendarOverride_ be _x_.[[Calendar]].</ins>
+            1. <ins>Else if _calendar_ is *"iso8601"*, then</ins>
+              1. <ins>Let _calendarOverride_ be ? GetBuiltinCalendar(_dateTimeFormat_.[[Calendar]]).</ins>
+              1. <mark>Note that this calls `Temporal.Calendar.from`, which may be author code. Desirable?.</mark>
```
ptomato, line 312: I think this comment (and the same ones below) are obsolete now.

ptomato (approved):
Seems OK to me, I was ambivalent on whether to use LocalTZA or not.

PR #1226, opened 2020-12-11
Remove incorrect parameter from toString sample in Instant docs

justingrant:
Some of the toString sample code had a wrong first parameter `undefined`.  This PR removes it.


ptomato:
:+1:, thanks!

ptomato (approved):

PR #1271, opened 2021-01-13
Omit fractional Numbers support in Duration.from for now.

Ms2ger:
Fixes #938.


littledan (approved):
LGTM; given the signedness of durations, it wouldn't make sense to reuse https://tc39.es/ecma262/#sec-canonicalnumericindexstring

ryzokuken (approved):
LGTM

PR #1272, opened 2021-01-13
Add some examples to the documentation readme.

### LINKED ISSUE #1219: Improve introductory documentation
The API surface is motivated by use cases, and it should be possible to tell a strong user story for each of those use cases in the docs. This could include one usage per class and one usage per method name/category, as [the recent presentation](https://docs.google.com/presentation/d/1xP3_UaXlS4-SilVpFu6UFOL8QQw0Dr_VsaR8mSSEATI/edit) does. This could be presented in two ways:
- In the main page of the documentation, the basics could be explained in a way that's parallel to the presentation
- The cookbook could include examples of of using each main part of the API

Ms2ger:
Fixes #1219.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -20,11 +20,23 @@ Temporal fixes these problems by:
 - Parsing a strictly specified string format
 - Supporting non-Gregorian calendars
 
+Temporal provides a number of separate types, in order to
+
+- Represent the information you have
+- Avoid buggy pattern of filling in 0 or UTC for missing info
+- Do appropriate calculations based on type
+
+
 ## Cookbook
 
 A cookbook to help you get started and learn the ins and outs of Temporal is available [here](./cookbook.md).
 
 ## API Documentation
+
+Types whose name starts with "Plain" (like `Temporal.PlainDate`, `Temporal.PlainTime`, and `Temporal.PlainDateTime`) do not, by convention, have an associated time zone.
+Converting between such types and exact time types (`Temporal.Instant` and `Temporal.ZonedDateTime`) can be ambiguous because of time zones and daylight saving time.
+
+Read more about [handling time zones, DST, and ambiguity in `Temporal`](./ambiguity.md).
 
 ### **Temporal.now**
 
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -115,8 +206,19 @@ See [Duration balancing](./balancing.md) for more on this topic.
...
 
 A `Temporal.TimeZone` represents an IANA time zone, a specific UTC offset, or UTC itself.
+Time zones translate from a date/time in UTC to a local date/time.
 Because of this `Temporal.TimeZone` can be used to convert between `Temporal.Instant` and `Temporal.PlainDateTime` as well as finding out the offset at a specific `Temporal.Instant`.
 
+It is also possible to implement your own time zones.
+
+```js
+const timeZone = Temporal.TimeZone.from('Africa/Cairo');
+timeZone.getPossibleInstantsFor(...)
```
littledan, line 216: It would be good if some of these had actual inputs and outputs...

docs/README.md
```diff
@@ -20,12 +20,24 @@ Temporal fixes these problems by:
...
+- Do appropriate calculations based on type
+
+
 ## Cookbook
 
 A cookbook to help you get started and learn the ins and outs of Temporal is available [here](./cookbook.md).
 
 ## API Documentation
 
+Types whose name starts with "Plain" (like `Temporal.PlainDate`, `Temporal.PlainTime`, and `Temporal.PlainDateTime`) do not, by convention, have an associated time zone.
```
littledan, line 36: "By convention" sticks out strangely here (as if it's a convention that Temporal users should follow, as opposed to one that the library follows). I would word this as,

> The Temporal API follows a convention of using types whose names start with "Plain" "Plain" (like `Temporal.PlainDate`, `Temporal.PlainTime`, and `Temporal.PlainDateTime`) for objects which do not associated time zone.

docs/README.md
```diff
@@ -20,12 +20,24 @@ Temporal fixes these problems by:
...
+Converting between such types and exact time types (`Temporal.Instant` and `Temporal.ZonedDateTime`) can be ambiguous because of time zones and daylight saving time.
```
littledan, line 37: Maybe we can give a hint of what to come, saying briefly something like, "and the Temporal API lets developers configure the resolution of this ambiguity".

littledan (approved):
This PR is a great improvement to the docs!!

justingrant:
Damn, you guys move fast!  One belated comment on the code samples: the output of `Temporal.now.instant().toString()` would never match what's shown with an even minute and no seconds, unless there's options used in the `toString` call. A more realistic output would be something like `"2021-01-13T20:57:01.500944804Z"`

ptomato:
Always happy to address comments after the fact. I'll include that in another PR today.

Issue #1276, opened 2021-01-14
Calendar: iso8601 `era()` calls invalid Temporal.Date intrinsic


cjtenny:
https://github.com/tc39/proposal-temporal/blob/157f6d7227c18798992c10368e2f838c79da2db8/polyfill/lib/calendar.mjs#L213

Should be Temporal.PlainDate. Quick fix, pulling out from #1245

PR #1277, opened 2021-01-14
Change Temporal.Date intrinsic to Temporal.PlainDate in iso8601 `era()` impl

cjtenny:
Quick fix for #1276


ptomato (approved):
Indeed, saw this in #1245, :+1:

justingrant:
Oops!  Good catch. Heads up: new checkin coming shortly on #1245 to fix some other bugs with `overflow`.

BTW, looks like there's the same mistakes in the specs too. #1278

PR #1289, opened 2021-01-15
[polyfill] Clarify constructor error messages for missing required arguments.

Ms2ger:
Fixes #1279.


ptomato (approved):
How did you write all those test262 tests so quickly?!

PR #1299, opened 2021-01-16
Allow calendars to validate fields and coerce types

cjtenny:
Do not coerce {day, month, year} to integers except in ISO calendar implementation.

Fixes #1229.

### DIFF
--- a/polyfill/lib/calendar.mjs
+++ b/polyfill/lib/calendar.mjs
@@ -310,16 +310,44 @@ impl['gregory'] = ObjectAssign({}, impl['iso8601'], {
   },
 
   dateFromFields(fields, overflow) {
+    const result = {};
     // Intentionally alphabetical
-    fields = ES.ToRecord(fields, [['day'], ['era', 'ad'], ['month'], ['year']]);
-    const isoYear = gre.isoYear(fields.year, fields.era);
-    return impl['iso8601'].dateFromFields({ ...fields, year: isoYear }, overflow);
+    for (const [name, defaultValue] of [['day'], ['era', 'ad'], ['month'], ['year']]) {
+      let value = fields[name];
+      if (value === undefined) {
+        if (defaultValue === undefined) {
+          throw new TypeError(`required property '${name}' missing or undefined`);
+        } else {
+          value = defaultValue;
+        }
+      }
+      if (name != 'era') {
+        value = ES.ToInteger(value);
+      }
+      result[name] = value;
+    }
+    const isoYear = gre.isoYear(result.year, result.era);
+    return impl['iso8601'].dateFromFields({ ...result, year: isoYear }, overflow);
   },
   yearMonthFromFields(fields, overflow) {
+    const result = {};
     // Intentionally alphabetical
-    fields = ES.ToRecord(fields, [['era', 'ad'], ['month'], ['year']]);
-    const isoYear = gre.isoYear(fields.year, fields.era);
-    return impl['iso8601'].yearMonthFromFields({ ...fields, year: isoYear }, overflow);
+    for (const [name, defaultValue] of [['era', 'ad'], ['month'], ['year']]) {
+      let value = fields[name];
+      if (value === undefined) {
+        if (defaultValue === undefined) {
+          throw new TypeError(`required property '${name}' missing or undefined`);
+        } else {
+          value = defaultValue;
+        }
+      }
+      if (name != 'era') {
+        value = ES.ToInteger(value);
+      }
+      result[name] = value;
+    }
+    const isoYear = gre.isoYear(result.year, result.era);
+    return impl['iso8601'].yearMonthFromFields({ ...result, year: isoYear }, overflow);
   }
 });
 
@@ -407,15 +435,35 @@ impl['japanese'] = ObjectAssign({}, impl['iso8601'], {
 
   dateFromFields(fields, overflow) {
     // Intentionally alphabetical
-    fields = ES.ToRecord(fields, [['day'], ['era'], ['month'], ['year']]);
-    const isoYear = jpn.isoYear(fields.year, fields.era);
-    return impl['iso8601'].dateFromFields({ ...fields, year: isoYear }, overflow);
+    const result = {};
+    for (const name of ['day', 'era', 'month', 'year']) {
+      let value = fields[name];
+      if (value === undefined) {
+        throw new TypeError(`required property '${name}' missing or undefined`);
+      }
+      if (name != 'era') {
+        value = ES.ToInteger(value);
+      }
+      result[name] = value;
+    }
+    const isoYear = jpn.isoYear(result.year, result.era);
+    return impl['iso8601'].dateFromFields({ ...result, year: isoYear }, overflow);
   },
   yearMonthFromFields(fields, overflow) {
+    const result = {};
     // Intentionally alphabetical
-    fields = ES.ToRecord(fields, [['era'], ['month'], ['year']]);
-    const isoYear = jpn.isoYear(fields.year, fields.era);
-    return impl['iso8601'].yearMonthFromFields({ ...fields, year: isoYear }, overflow);
+    for (const name of ['era', 'month', 'year']) {
+      let value = fields[name];
+      if (value === undefined) {
+        throw new TypeError(`required property '${name}' missing or undefined`);
+      }
+      if (name != 'era') {
+        value = ES.ToInteger(value);
+      }
+      result[name] = value;
+    }
+    const isoYear = jpn.isoYear(result.year, result.era);
+    return impl['iso8601'].yearMonthFromFields({ ...result, year: isoYear }, overflow);
   }
 });
 
--- a/polyfill/lib/calendar.mjs
+++ b/polyfill/lib/calendar.mjs
@@ -171,22 +171,39 @@ DefineIntrinsic('Temporal.Calendar.prototype.toString', Calendar.prototype.toStr
 
 impl['iso8601'] = {
   dateFromFields(fields, overflow) {
-    let { year, month, day } = ES.ToTemporalDateFields(fields, []);
-    year = ES.ToInteger(year);
-    month = ES.ToInteger(month);
-    day = ES.ToInteger(day);
+    const result = {};
+    for (const name of ['day', 'month', 'year']) {
+      let value = fields[name];
+      if (value === undefined) {
+        throw new TypeError(`required property '${name}' missing or undefined`);
+      }
+      result[name] = ES.ToInteger(value);
+    }
+    const { day, month, year } = result;
     return ES.RegulateDate(year, month, day, overflow);
   },
   yearMonthFromFields(fields, overflow) {
-    let { year, month } = ES.ToTemporalYearMonthFields(fields, []);
-    year = ES.ToInteger(year);
-    month = ES.ToInteger(month);
+    const result = {};
+    for (const name of ['month', 'year']) {
+      let value = fields[name];
+      if (value === undefined) {
+        throw new TypeError(`required property '${name}' missing or undefined`);
+      }
+      result[name] = ES.ToInteger(value);
+    }
+    const { month, year } = result;
     return ES.RegulateYearMonth(year, month, overflow);
   },
   monthDayFromFields(fields, overflow) {
-    let { month, day } = ES.ToTemporalMonthDayFields(fields, []);
-    month = ES.ToInteger(month);
-    day = ES.ToInteger(day);
+    const result = {};
+    for (const name of ['day', 'month']) {
+      let value = fields[name];
+      if (value === undefined) {
+        throw new TypeError(`required property '${name}' missing or undefined`);
+      }
+      result[name] = ES.ToInteger(value);
+    }
+    const { day, month } = result;
     return ES.RegulateMonthDay(month, day, overflow);
   },
   fields(fields) {
--- a/polyfill/lib/calendar.mjs
+++ b/polyfill/lib/calendar.mjs
@@ -171,15 +171,22 @@ DefineIntrinsic('Temporal.Calendar.prototype.toString', Calendar.prototype.toStr
 
 impl['iso8601'] = {
   dateFromFields(fields, overflow) {
-    const { year, month, day } = ES.ToRecord(fields, [['day'], ['month'], ['year']]);
+    let { year, month, day } = ES.ToTemporalDateFields(fields, []);
+    year = ES.ToInteger(year);
+    month = ES.ToInteger(month);
+    day = ES.ToInteger(day);
     return ES.RegulateDate(year, month, day, overflow);
   },
   yearMonthFromFields(fields, overflow) {
-    const { year, month } = ES.ToRecord(fields, [['month'], ['year']]);
+    let { year, month } = ES.ToTemporalYearMonthFields(fields, []);
+    year = ES.ToInteger(year);
+    month = ES.ToInteger(month);
     return ES.RegulateYearMonth(year, month, overflow);
   },
   monthDayFromFields(fields, overflow) {
-    const { month, day } = ES.ToRecord(fields, [['day'], ['month']]);
+    let { month, day } = ES.ToTemporalMonthDayFields(fields, []);
+    month = ES.ToInteger(month);
+    day = ES.ToInteger(day);
     return ES.RegulateMonthDay(month, day, overflow);
   },
   fields(fields) {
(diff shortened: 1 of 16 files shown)


justingrant:
Overall this looks good. A few notes:
* It'd be good to add some tests to this PR that validate that a userland calendar can: 
  * pass non-built-in fields to `from` and the calendar will receive them as-is.
  * change the type of built-in fields, e.g. `day` can be an object or a string
  * pass `undefined` for one or more of the built-in fields
* AFAICT, this PR won't work for `with` (more specifically: it won't let the calendar distinguish between original fields and new fields, which breaks interpretation of overlapping fields like `month` vs. `monthCode`), for the reasons noted in https://github.com/tc39/proposal-temporal/issues/1235#issuecomment-760716368. I assume that this omission is intentional and a solution for `with` coming in another PR. Correct?
* An assumption made in this PR is that *all* validation and coercion should be done by the calendar. One thing we haven't yet discussed is whether there are any field invariants that should be applied to all calendars.  For example, should `day` be required to be a positive integer or `undefined`, regardless of calendar?  Should `monthCode` be required to be a non-empty string or `undefined`? And so on. If yes, then we may want to do some validation in abstract ops for all calendars, and to delegate additional validation to the calendar. Or we might make these "documentation-only" restrictions that the calendar must enforce on its own. I don't think this concern should block merging of this PR, because if the answer is "no invariants" or "docs only" then no action is needed and this PR is fine as-is. If the answer is "abstract ops should enforce" then we can always add this validation later.

cjtenny:
Thanks for the comments, Justin!

* Re: tests, I agree. I'll get to work on it. Right now I'm fixing up the test262 failures and figuring out whether to write new tests for `getFields` after this - for better or for worse, this changes some of the alphabetical order accesses including changing some datetime accesses to be {date fields} then {time fields}. Do you have thoughts on that? Preserving the alphabetical access pattern for `get` would require some annoying spec changes, I'm trying to understand whether or not it's important. Preserving the alphabetical access pattern for `valueOf` wouldn't make sense since calendars can choose how to treat non-primitive {`day`, `month`, `year`}.
* Re: #1235 , yeah, the extent of @ptomato 's and my discussion was that this should be treated separately. I don't have another PR underway for it at the moment and have no strong feelings on what I do next - happy to work on that but if you wanted to go at it, feel free. I hope Tuesday's discussion about `monthCode` & associated fields will at least make the choice of how to address it clear.
* It is an interesting assertion. Would a custom calendar want to take something like "first monday" as a name for `day`? Would we want to support that? I don't have a strong opinion here. Isn't requiring those invariants of {`day`, `month`, `year`} the status quo without this change, though? (maybe sans 'positive', but the spec required them all to be integers).

cjtenny:
This change gives some room for more strict validation in the ISO calendar without imposing it on all calendars. Should the ISO calendar fail more loudly if e.g. conversions to number don't produce something equal to the input? A temporal object `.from({day: {day: "5"}, ...})` having `day: 0` may be frustrating, though I imagine things like this were already discussed (sorry! catching up).

cjtenny:
>       * pass non-built-in fields to `from` and the calendar will receive them as-is.

This is already tested with a ''seasons" calendar in the polyfill tests. Maybe it should be part of the test262 tests, but if so let's separate moving that.

>       * change the type of built-in fields, e.g. `day` can be an object or a string

Done.

>       * pass `undefined` for one or more of the built-in fields

Actually, I wrote a test for this then remembered this PR doesn't implement that - PlainYearMonth still requires 'year' and 'month' to have non-undefined values, for example. Do you think we should not even check that they have values? I wrote a calendar that assumes y2k if undefined; would it make sense for users of such a calendar to have to pass in _something_ for year?

I can see two ways to change the spec: require that the parameter be passed, but allow it to be explicitly set to undefined; or, stop checking for presence of calendar fields at all, in which case: are they still built in fields? Given that we have 'month', 'day', 'year' methods already specified throughout, I think it makes sense to continue to require they be passed, but I could see the case for allowing them to be explicitly undefined (but not implicitly) - it would be inconsistent and weird, though, so my preference is to not change anything.

ptomato:
> I can see two ways to change the spec: require that the parameter be passed, but allow it to be explicitly set to undefined; or, stop checking for presence of calendar fields at all, in which case: are they still built in fields? Given that we have 'month', 'day', 'year' methods already specified throughout, I think it makes sense to continue to require they be passed, but I could see the case for allowing them to be explicitly undefined (but not implicitly) - it would be inconsistent and weird, though, so my preference is to not change anything.

Treating `foo: undefined` differently from `foo` not being present, would be out of character for JS and would be difficult to get through committee, so there would have to be a very strong reason.

polyfill/lib/ecmascript.mjs
```diff
@@ -982,13 +982,13 @@ export const ES = ObjectAssign({}, ES2020, {
       } else {
         any = true;
       }
-      if (BUILTIN_FIELDS.has(property)) {
+      if (BUILTIN_FIELDS.has(property) && !CALENDAR_FIELDS.has(property)) {
```
ptomato, line 985: Nothing against your code in particular, but this line makes me think that too much has been bolted on to `ToRecord` already, making it too confusing, and I'd like to refactor it. However since this part is already different than the spec text, that can be done later.

polyfill/lib/calendar.mjs
```diff
@@ -171,15 +171,39 @@ DefineIntrinsic('Temporal.Calendar.prototype.toString', Calendar.prototype.toStr
 
 impl['iso8601'] = {
   dateFromFields(fields, overflow) {
-    const { year, month, day } = ES.ToRecord(fields, [['day'], ['month'], ['year']]);
```
ptomato, line 174: I think a similar change needs to be made below in the `gregory` and `japanese` calendars.

polyfill/test/PlainDate/prototype/getFields/order-of-operations.js
```diff
@@ -41,7 +38,7 @@ Object.defineProperties(ObservedDate.prototype, {
 const instance = new ObservedDate(2000, 5, 2);
 
 const result = instance.getFields();
-assert.sameValue(result.year, 2000, "year result");
-assert.sameValue(result.month, 5, "month result");
-assert.sameValue(result.day, 2, "day result");
 assert.compareArray(actual, expected, "order of operations");
+assert.sameValue(result.year.valueOf(), 2000, "year result");
```
ptomato, line 42: Should we remove the trap on valueOf now that we're not checking for it? Or is it important to check that we _don't_ call valueOf?

spec/plaintime.html
```diff
@@ -885,7 +885,8 @@ <h1>ToTemporalTimeRecord ( _temporalTimeLike_ )</h1>
           1. Let _property_ be the Property value of the current row.
           1. Let _value_ be ? Get(_temporalTimeLike_, _property_).
           1. If _value_ is *undefined*, then
-            1. Throw a *TypeError* exception.
+            1. If _someRequired_ is not *false*, then
```
ptomato, line 888: If I'm reading correctly, this will behave differently from the code. This would throw the exception if _any_ properties were missing, rather than if _all_ properties were missing.

ptomato (approved):
I also agree that the current state of this PR (no validation outside the calendar at all) is fine.

Minor comments, feel free to squash and merge when they are fixed.

polyfill/lib/ecmascript.mjs
cjtenny: agreed

justingrant:
> Actually, I wrote a test for this then remembered this PR doesn't implement that - PlainYearMonth still requires 'year' and 'month' to have non-undefined values, for example. Do you think we should not even check that they have values? I wrote a calendar that assumes y2k if undefined; would it make sense for users of such a calendar to have to pass in _something_ for year?

The reason to allow `undefined` is to allow calendars to support alternatives to `month` or `year`. For example, the following code (that works in #1245) throws in current Temporal because YMD fields can't be undefined:

```js
Temporal.PlainDate.from({ era: 'reiwa', eraYear: 1, month: 7, day: 1, calendar: 'japanese' });
Temporal.PlainDate.from({ year: 5782, monthCode: '5L', day: 1, calendar: 'hebrew' });
```

My preference would be to simply allow `undefined` for all fields and then to let the calendar determine which combination of fields are allowed vs. rejected. Without this capability, calendars would never be able to provide alternative ways to specify a year, month, or day. These alternatives seem to be common across calendars, e.g. chinese named years, hebrew has different types of years, some calendars may sometimes use Zodiac signs instead of traditional month names, etc.  Even if we decide that built-in calendars shouldn't support these alternatives, IMHO we should still allow userland calendars to provide alternative ways to describe a year, month, or day.

FWIW, the implementation in #1245 allows `undefined` but coerces non-`undefined` types to the expected type (integer for all fields except `era` and `monthCode` which must be strings).

> * Re: #1235 , yeah, the extent of @ptomato 's and my discussion was that this should be treated separately. I don't have another PR underway for it at the moment and have no strong feelings on what I do next - happy to work on that but if you wanted to go at it, feel free. I hope Tuesday's discussion about `monthCode` & associated fields will at least make the choice of how to address it clear.

Nope, go ahead. I don't think I'll have time for that PR this week. 

Ironically, the `with` issue in #1235 is essentially the same problem as this PR: dealing with fields that have multiple possible ways to describe them. Example: 

```js
Temporal.PlainDate.from({ year: 5782, month: 1, day: 1, calendar: 'hebrew' }).with({ monthCode: '5L' });
Temporal.PlainDate.from({ year: 5782, monthCode: '5L', day: 1, calendar: 'hebrew' }).with({ month: 1 });
```

The problem in current Temporal is that both lines above will call the calendar's `dateFromFields` method with this bag: `{ year: 5782, month: 1, day: 1, monthCode: '5L' }`.  There's no way for it to know if `month` or `monthCode` should "win". What #1245 does is to strip `month` from the bag before the calendar gets it by removing all fields returned by `fields(userProvidedFields, {role: 'merge'}`.  There's certainly other ways to solve this problem-- I don't have a strong opinion on the solution as long as it addresses the use-case of overlapping calendar fields.

spec/plaintime.html
cjtenny: oops, correct; will fix

polyfill/lib/calendar.mjs
cjtenny: The reason I made this change was to ensure that `ISO{Date,...}FromFields` preserve `[get, valueOf, get, valueOf, ...]` order in their accesses. Custom calendars don't need to preserve this order. However, I can't find whether all ICU calendar implementations are part of the standard - where should I look for that?

polyfill/test/PlainDate/prototype/getFields/order-of-operations.js
cjtenny: Good question, I'm not sure. As is, the current test does check that we don't call valueOf to do any coercion. I was having trouble understanding the logic around why the ISO calendar shouldn't be converting these to integers when they're custom getters for fields on a temporal object with the ISO calendar.

Looking now I actually think something might be broken (I might have broken something here). I'll dig into it.

cjtenny:
> ```js
> Temporal.PlainDate.from({ era: 'reiwa', eraYear: 1, month: 7, day: 1, calendar: 'japanese' });
> Temporal.PlainDate.from({ year: 5782, monthCode: '5L', day: 1, calendar: 'hebrew' });
> ```

This could actually be rewritten to use 'year' instead of 'eraYear'; the calendar would interpret year differently with 'era' present. I wonder if that solves the issue entirely?

> My preference would be to simply allow `undefined` for all fields and then to let the calendar determine which combination of fields are allowed vs. rejected. Without this capability, calendars would never be able to provide alternative ways to specify a year, month, or day. These alternatives seem to be common across calendars, e.g. chinese named years, hebrew has different types of years, some calendars may sometimes use Zodiac signs instead of traditional month names, etc. Even if we decide that built-in calendars shouldn't support these alternatives, IMHO we should still allow userland calendars to provide alternative ways to describe a year, month, or day.

By 'all fields' am I correct in understanding you mean 'year', 'month', and 'day' where applicable? DateTime objects still need to check for time fields, and you're not proposing delegating naming of and validation of time fields to calendars, right?

I can see the case for this but I'm not sure of the impact. Would like to hear what others think.

I wonder if this actually addresses the use case of overlapping fields entirely: letting the type and preserce and meaning of 'month', 'year' be _completely_ calendar-defined such that there's no need to specify further fields. That seems too simple, but maybe I'm missing something.

polyfill/lib/calendar.mjs
ptomato: The ICU calendar implementations will be built-in to Intl-aware JavaScript engines, but they won't be _specified_, at least not as part of this proposal. That's because, like time zones, calendars may change for political and administrative reasons, and new calendars may be added to ICU.

I think the general consensus has been that built-in calendars should adhere to the same standards as the ISO calendar, but userland calendars are only "recommended" to do these things. So I do think that the implementations in the polyfill ought to do the same things as the ISO calendar, and I believe that's what #1245 does as well.

justingrant:
> > ```js
> > Temporal.PlainDate.from({ era: 'reiwa', eraYear: 1, month: 7, day: 1, calendar: 'japanese' });
> > Temporal.PlainDate.from({ year: 5782, monthCode: '5L', day: 1, calendar: 'hebrew' });
> > ```
> 
> This could actually be rewritten to use 'year' instead of 'eraYear'; the calendar would interpret year differently with 'era' present. I wonder if that solves the issue entirely?

For `year`, requiring `era` for some calendars but not others would "solve" the issue at the cost of not being able to write calendar-neutral code (or making calendar-neutral code hard enough that most developers won't bother).  I'm not sure that's the right tradeoff if our goal is to enable lots of code to work across calendars.

Similarly for `monthCode`, we could offer a different API (e.g. a `month` field that accepts either a string or a number) but that seems much less intuitive for callers than to allow separate field names, especially because getters need to have different names if we want to expose a month index vs. month code. It seems like an odd API if the name of an input field doesn't match the name of the output (getter) fields.

> > My preference would be to simply allow `undefined` for all fields and then to let the calendar determine which combination of fields are allowed vs. rejected. Without this capability, calendars would never be able to provide alternative ways to specify a year, month, or day. These alternatives seem to be common across calendars, e.g. chinese named years, hebrew has different types of years, some calendars may sometimes use Zodiac signs instead of traditional month names, etc. Even if we decide that built-in calendars shouldn't support these alternatives, IMHO we should still allow userland calendars to provide alternative ways to describe a year, month, or day.
> 
> By 'all fields' am I correct in understanding you mean 'year', 'month', and 'day' where applicable? DateTime objects still need to check for time fields, and you're not proposing delegating naming of and validation of time fields to calendars, right?

Yep, I should have said "all date fields" or "YMD fields". Good catch.

> I wonder if this actually addresses the use case of overlapping fields entirely: letting the type and preserce and meaning of 'month', 'year' be _completely_ calendar-defined such that there's no need to specify further fields. That seems too simple, but maybe I'm missing something.

IMHO, there are two reasonable options: 
1. What you propose: calendars are entirely responsible for YMD field coercion/validation.
2. YMD fields can be undefined, but if not undefined then Temporal will coerce them (and perhaps do basic validation, e.g. ensuring that `day` is a positive integer) before calling the calendar which can do further validation.

Note that there's a similar decision we have to make in getters of PlainDate and other date-having types. If a calendar returns `'abc'` from its `day` method, should Temporal throw?

I'm in favor of (2) for both inputs and outputs (and I implemented it in #1245) because it makes it forces calendar authors to write code that's more likely to work with existing Temporal code.  But I don't have a strong opinion on this: (2) could be "enforced" by documentation even if (1) is the implementation we choose.

My only strong opinion on this topic is that we should allow `undefined` for all YMD fields because so many calendars seem to have alternate ways to specify YMD values.

polyfill/test/PlainDate/prototype/getFields/order-of-operations.js
cjtenny: (Result of discussion: nothing is broken, this is correct; it's important to check that we don't call valueOf, in this case)

spec/plaintime.html
cjtenny: Fixed in update

polyfill/lib/calendar.mjs
cjtenny: Fixed in update

cjtenny: Fixed in update

cjtenny:
Justin: good points. I'll keep it short here:

> My only strong opinion on this topic is that we should allow `undefined` for all YMD fields because so many calendars seem to have alternate ways to specify YMD values.

I agree. I'll address this in a follow up change and link the discussion / carry on there.

ptomato:
Per the outcome of the Jan 19 meeting on #1299, we'll unfortunately have to undo part of this...

PR #1303, opened 2021-01-18
Remove CalendarToString and TimeZoneToString

ptomato:
See: #1294 (but does not close the discussion)

### DIFF
--- a/spec/calendar.html
+++ b/spec/calendar.html
@@ -35,15 +35,6 @@
       </emu-alg>
     </emu-clause>
 
-    <emu-clause id="sec-temporal-calendartostring" aoid="CalendarToString">
-      <h1>CalendarToString ( _calendar_ )</h1>
-      <emu-alg>
-        1. Let _toString_ be ? Get(_calendar_, *"toString"*).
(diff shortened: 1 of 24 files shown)


Ms2ger (approved):
Fair enough.

spec/calendar.html
```diff
@@ -879,7 +867,7 @@ <h1>Temporal.Calendar.prototype.toJSON ( )</h1>
       </p>
       <emu-alg>
         1. Let _calendar_ be the *this* value.
-        1. Return ? CalendarToString(_calendar_).
+        1. Return ? ToString(_calendar_).
```
ljharb, line 870: a reasonable alternative would be to Call %Temporal.Calendar.toString%, but this way plays better with subclassing i suppose.

ljharb (approved):

spec/calendar.html
ptomato: This way does an additional observable Get(calendar, @@toPrimitive) but I think that's OK?

ljharb: Yes, either a Symbol.toPrimitive or a toString lookup would be required to invoke subclass code.

Issue #1307, opened 2021-01-19
Calendar should distinguish between existing and new fields in with()


ptomato:
Follow up from #1235. Paraphrasing Justin's description:

The with() method is problematic if input fields can overlap or be aliased, because the merging of user-inputted fields and existing fields happens inside `ecmascript.mjs`, not inside the calendar. This means that `date.with({monthCode: '5L'})` will fail because by the time the calendar gets the list of fields, there's no way for it to know if `month` or `monthCode` is the correct input field to use when constructing the new date.

Ideas from #1235:
- a `role` option on Calendar.fields()
- a Calendar.mergeFields() method that custom calendars implement
- a Calendar.dateWith(date, fields) and analogous methods for PlainYearMonth/PlainMonthDay
- allowing multiple fields bags in Calendar.dateFromFields(), yearMonthFromFields(), monthDayFromFields()

ptomato:
@cjtenny and I discussed this this afternoon and we thought a Calendar.mergeFields() method would be the best option here. If a custom calendar adds fields but doesn't add the capability to specify a calendar unit in more than one way (in other words, they don't add anything like `month` and `monthCode`), then they can override Calendar.fields() but inherit the default Calendar.mergeFields().

#1310 has a proof of concept of this, with a test calendar illustrating how it would be used (to specify years as a `century` plus a `centuryYear`). I'll put this on the agenda for Thursday but if you can, let's try to resolve it even earlier; if I get enough positive reactions then I'll go ahead and write the spec text and flesh out the tests.

Note that #1310 _doesn't_ have the correct implementation of mergeFields() for what we discussed today, with `year`/`era`/`eraYear`/`month`/`monthCode`. This is because we don't have `eraYear` and `monthCode` in the polyfill yet. The actual implementation for the ISO calendar would look something like this (untested and not caring about observable side effects):
```js
mergeFields(fields, extraFields) {
  if ('month' in extraFields) delete fields.monthCode;
  if ('monthCode' in extraFields) delete fields.month;
  return Object.assign(fields, extraFields);
}
```
and then the implementation for the Gregorian/Japanese/other calendars with eras would add similar code for `year` vs `era`+`eraYear`.

PR #1310, opened 2021-01-20
Calendar.mergeFields()

ptomato:
Closes: #1307

### DIFF
--- a/docs/calendar.md
+++ b/docs/calendar.md
@@ -321,7 +321,7 @@ Temporal.Calendar.from('chinese').dateUntil(
 ); // => P1M2D
 ```
 
-### calendar.**fields**(fields: array<string>) : array<string>
+### calendar.**fields**(_fields_: array<string>) : array<string>
 
 **Parameters:**
 
@@ -330,7 +330,7 @@ Temporal.Calendar.from('chinese').dateUntil(
 **Returns:** a new list of field names.
 
 This method does not need to be called directly except in specialized code.
-It is called indirectly when using the `from()` static methods and `with()` methods of `Temporal.PlainDateTime`, `Temporal.PlainDate`, and `Temporal.PlainYearMonth`.
+It is called indirectly when using the `from()` static methods and `with()` methods of `Temporal.PlainDateTime`, `Temporal.PlainDate`, `Temporal.PlainMonthDay`, `Temporal.PlainYearMonth`, and `Temporal.ZonedDateTime`.
 
 Custom calendars should override this method if they require more fields with which to denote the date than the standard `year`, `month`, and `day` (for example, `era`).
 The input array contains the field names that are necessary for a particular operation (for example, `'month'` and `'day'` for `Temporal.PlainMonthDay.prototype.with()`), and the method should make a copy of the array and add whichever extra fields are necessary.
--- a/docs/calendar.md
+++ b/docs/calendar.md
@@ -52,7 +52,7 @@ The other, more difficult, way to create a custom calendar is to create a plain 
 The object must implement all of the `Temporal.Calendar` properties and methods except for `id`, `fields()`, and `toJSON()`.
 Any object with the required methods will return the correct output from any Temporal property or method.
 However, most other code will assume that custom calendars act like built-in `Temporal.Calendar` objects.
-To interoperate with libraries or other code that you didn't write, then you should implement the `id` property and the `fields()` method as well.
+To interoperate with libraries or other code that you didn't write, then you should implement the `id` property and the `fields()` and `toJSON()` methods as well.
 Your object must not have a `calendar` property, so that it can be distinguished in `Temporal.Calendar.from()` from other Temporal objects that have a calendar.
 
 The identifier of a custom calendar must consist of one or more components of between 3 and 8 ASCII alphanumeric characters each, separated by dashes, as described in [Unicode Technical Standard 35](https://unicode.org/reports/tr35/tr35.html#Unicode_locale_identifier).
--- a/docs/calendar.md
+++ b/docs/calendar.md
@@ -347,6 +347,39 @@ Temporal.Calendar.from('iso8601').fields(['month', 'day']);
 ```
 <!-- prettier-ignore-end -->
 
+### calendar.**mergeFields**(_fields_: object, _additionalFields_: object) : object
+
+**Parameters:**
+
+- `fields` (object): A plain object with properties representing calendar units.
+- `additionalFields` (object): Another plain object with properties representing calendar units.
+
+**Returns:** a new object with properties from both `fields` and `additionalFields`.
+
+This method does not need to be called directly except in specialized code.
+It is called indirectly when using the `with()` methods of `Temporal.PlainDateTime`, `Temporal.PlainDate`, `Temporal.PlainMonthDay`, `Temporal.PlainYearMonth`, and `Temporal.ZonedDateTime`.
+
+Custom calendars should override this method if they allow a calendar unit to be specified in more than one way.
+(For example, the Gregorian calendar allows years to be specified either by a `year` property or a combination of `era` and `eraYear`.)
+The overridden implementation should return an object with some or all of the properties from the original `fields` object and `additionalFields` copied onto it.
+
+When subclassing `Temporal.Calendar`, this method doesn't need to be overridden, unless your calendar adds more ways to specify a unit other than the built-in properties `monthCode`, `era`, and `eraYear`.
+The default implementation copies all properties from `additionalFields` onto `fields`, taking into account that months may be specified either by `month` or `monthCode` properties, and any other special cases required by built-in calendars.
+
+Usage example:
+
+<!-- prettier-ignore-start -->
+```js
+// In built-in calendars, this method copies properties, taking `month`
+// and `monthCode` into account
+Temporal.Calendar.from('iso8601').mergeFields(
+  { year: 2006, month: 7, day: 31 },
+  { monthCode: '8' }
+);
+// => { year: 2006, monthCode: '8', day: 31 }
+```
+<!-- prettier-ignore-end -->
+
 ### calendar.**toString**() : string
 
 **Returns:** The string given by `calendar.id`.
--- a/polyfill/lib/calendar.mjs
+++ b/polyfill/lib/calendar.mjs
@@ -166,7 +166,6 @@ export class Calendar {
 
 MakeIntrinsicClass(Calendar, 'Temporal.Calendar');
 DefineIntrinsic('Temporal.Calendar.from', Calendar.from);
-DefineIntrinsic('Temporal.Calendar.prototype.fields', Calendar.prototype.fields);
 
 impl['iso8601'] = {
   dateFromFields(fields, overflow) {
--- a/polyfill/lib/calendar.mjs
+++ b/polyfill/lib/calendar.mjs
@@ -64,6 +64,10 @@ export class Calendar {
     fields = ES.CreateListFromArrayLike(fields, ['String']);
     return impl[GetSlot(this, CALENDAR_ID)].fields(fields);
   }
+  mergeFields(fields, additionalFields) {
+    if (!ES.IsTemporalCalendar(this)) throw new TypeError('invalid receiver');
+    return impl[GetSlot(this, CALENDAR_ID)].mergeFields(fields, additionalFields);
+  }
   dateAdd(date, duration, options, constructor) {
     if (!ES.IsTemporalCalendar(this)) throw new TypeError('invalid receiver');
     date = ES.ToTemporalDate(date, GetIntrinsic('%Temporal.PlainDate%'));
@@ -182,6 +186,9 @@ impl['iso8601'] = {
   },
   fields(fields) {
     return fields;
+  },
+  mergeFields(fields, additionalFields) {
+    return { ...fields, ...additionalFields };
   },
   dateAdd(date, duration, overflow) {
     const { years, months, weeks, days } = duration;
--- a/polyfill/lib/ecmascript.mjs
+++ b/polyfill/lib/ecmascript.mjs
@@ -1445,9 +1445,8 @@ export const ES = ObjectAssign({}, ES2020, {
   },
   CalendarFields: (calendar, fieldNames) => {
     let fields = calendar.fields;
-    if (fields === undefined) fields = GetIntrinsic('%Temporal.Calendar.prototype.fields%');
-    const array = ES.Call(fields, calendar, [fieldNames]);
-    return ES.CreateListFromArrayLike(array, ['String']);
+    if (fields !== undefined) fieldNames = ES.Call(fields, calendar, [fieldNames]);
+    return ES.CreateListFromArrayLike(fieldNames, ['String']);
   },
 
   ToTemporalCalendar: (calendarLike) => {
--- a/polyfill/lib/ecmascript.mjs
+++ b/polyfill/lib/ecmascript.mjs
@@ -1448,6 +1448,11 @@ export const ES = ObjectAssign({}, ES2020, {
     if (fields !== undefined) fieldNames = ES.Call(fields, calendar, [fieldNames]);
     return ES.CreateListFromArrayLike(fieldNames, ['String']);
   },
+  CalendarMergeFields: (calendar, fields, additionalFields) => {
+    let mergeFields = calendar.mergeFields;
+    if (mergeFields === undefined) return { ...fields, ...additionalFields };
+    return ES.Call(mergeFields, calendar, [fields, additionalFields]);
+  },
 
   ToTemporalCalendar: (calendarLike) => {
     if (ES.Type(calendarLike) === 'Object') {
--- a/polyfill/lib/plaindate.mjs
+++ b/polyfill/lib/plaindate.mjs
@@ -21,8 +21,6 @@ import {
   SetSlot
 } from './slots.mjs';
 
-const ObjectAssign = Object.assign;
-
 function TemporalDateToString(date, showCalendar = 'auto') {
   const year = ES.ISOYearString(GetSlot(date, ISO_YEAR));
   const month = ES.ISODateTimePartString(GetSlot(date, ISO_MONTH));
@@ -133,8 +131,8 @@ export class PlainDate {
     if (!props) {
       throw new TypeError('invalid date-like');
     }
-    const fields = ES.ToTemporalDateFields(this, fieldNames);
-    ObjectAssign(fields, props);
+    let fields = ES.ToTemporalDateFields(this, fieldNames);
+    fields = ES.CalendarMergeFields(calendar, fields, props);
 
     options = ES.NormalizeOptionsObject(options);
     const overflow = ES.ToTemporalOverflow(options);
--- a/polyfill/test/usercalendar.mjs
+++ b/polyfill/test/usercalendar.mjs
@@ -232,9 +232,6 @@ describe('Userland calendar', () => {
       return { year, days };
     }
     const obj = {
-      get id() {
-        return this.toString();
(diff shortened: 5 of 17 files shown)


polyfill/lib/plaindate.mjs
```diff
@@ -133,8 +133,8 @@ export class PlainDate {
     if (!props) {
```
justingrant, line 133: After we add monthCode/era/eraYear, will those fields be added to `['day', 'month', 'year']` in the call to `ES.CalendarFields` 2 lines above here?  I'm asking because calendars will need to know whether their `fields` implementations must add those fields or not.

polyfill/test/usercalendar.mjs
```diff
@@ -293,6 +293,9 @@ describe('Userland calendar', () => {
       },
       fields(fields) {
         return fields.slice();
```
justingrant, line 295: Unrelated to this PR, but why does this implementation copy the object if it's not needed to be copied?

justingrant (commented):
Overall looks good, but in addition to minor comments I have two potentially significant questions: 
* Should `fields` object be a Temporal object or a plain object bag?
* Is it OK that this implementation calls a bunch of getters that are never used?

Details are inline.

polyfill/lib/calendar.mjs
```diff
@@ -185,6 +190,9 @@ impl['iso8601'] = {
   fields(fields) {
     return fields;
   },
+  mergeFields(fields, extraFields) {
+    return ObjectAssign(fields, extraFields);
```
cjtenny, line 194: Will it be spec to always modify the first bag? To only call mergeFields on a potentially user-derived but not user-created bag? (i.e. how it's called on ToTemporal____Fields throughout the rest of this PR)

cjtenny (commented):
I like it overall, but had one naming question and one semantic question

docs/calendar.md
```diff
@@ -347,6 +347,38 @@ Temporal.Calendar.from('iso8601').fields(['month', 'day']);
...
+**Returns:** an object with properties from both `fields` and `extraFields`.
+
+This method does not need to be called directly except in specialized code.
+It is called indirectly when using the `with()` methods of `Temporal.PlainDateTime`, `Temporal.PlainDate`, `Temporal.PlainMonthDay`, `Temporal.PlainYearMonth`, and `Temporal.ZonedDateTime`.
+
+Custom calendars should override this method if they allow a calendar unit to be specified in more than one way.
+(For example, the Gregorian calendar allows years to be specified either by a `year` property or a combination of `era` and `eraYear`.)
+
+When subclassing `Temporal.Calendar`, this method doesn't need to be overridden, unless your calendar allows this with other properties than the built-in calendars do.
+The default implementation copies all properties from `extraFields` onto `fields`, taking into account that months may be specified either by `month` or `monthCode` properties, and any other special cases required by built-in calendars.
```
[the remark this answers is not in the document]
cjtenny: As far as I can tell, for all potential usages, fields will have already been created by `ToTemporal____Fields` as the validated new-fields-bag; since those will always return new objects, I think this is explicitly intended but agree it should be explicitly stated and mutation (allowed, expected).

docs/calendar.md
```diff
@@ -347,6 +347,38 @@ Temporal.Calendar.from('iso8601').fields(['month', 'day']);
 ```
 <!-- prettier-ignore-end -->
 
+### calendar.**mergeFields**(_fields_: object, _extraFields_: object) : object
+
+**Parameters:**
+
+- `fields` (object): An object with properties representing calendar units.
+- `extraFields` (object): Another object with properties representing calendar units.
```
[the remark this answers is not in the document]
cjtenny: All implementation paths based on how this code works have it also be a plain object via ToPartialRecords as well, and it should be explicitly stated

docs/calendar.md
```diff
@@ -347,6 +347,38 @@ Temporal.Calendar.from('iso8601').fields(['month', 'day']);
 ```
 <!-- prettier-ignore-end -->
 
+### calendar.**mergeFields**(_fields_: object, _extraFields_: object) : object
```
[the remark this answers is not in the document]
cjtenny: I don't think any cases for any of the ICU calendars would require merging fields to know what the output context is; the logic depends only on the fields present on the input. In the `role: 'merge'` logic in #1245, do you think there's any use case that requires knowing the output context for merging?

docs/calendar.md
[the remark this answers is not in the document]
cjtenny: Technically neither is an 'existing object' in the sense that it's visible to the users of temporal; both are new objects minimize getter calls. But yeah, specifying the intent of 'properties representing an existing Temporal object's calendar units' and 'properties representing new user-specified calendar units' in some way would be :+1:

docs/calendar.md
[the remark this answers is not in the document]
justingrant: Nope, I'm not aware of any. But we previously named a bunch of methods "dateXxx" for future compatibility reasons.  Wanted to make sure that those reasons don't apply here. If not, then name is fine as-is.

cjtenny:
Oops, somehow despite refreshing; Justin, I didn't see your comments when I added mine last night. Sorry for the redundancy, Philip.

docs/calendar.md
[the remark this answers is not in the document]
ptomato: My reasoning is that this can be `mergeFields` for the same reason `fields` isn't `dateFields`; it will work for both date and time fields if time calendars are added in the future.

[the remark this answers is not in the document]
ptomato: An options argument could be added on later in a backwards-compatible way if that became necessary. (The contrived "century" calendar in the tests is an example of a calendar with a multi-valued field like that, but I don't think it is diminished by not having such an option.)

polyfill/lib/plaindate.mjs
```diff
@@ -133,8 +133,8 @@ export class PlainDate {
...
       throw new TypeError('invalid date-like');
     }
-    const fields = ES.ToTemporalDateFields(this, fieldNames);
-    ObjectAssign(fields, props);
+    let fields = ES.ToTemporalDateFields(this, fieldNames);
```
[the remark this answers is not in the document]
ptomato: This already existed before this change: `plainDate.with({ day: 15 })` calls the `day` getter on the receiver, even though the value is overwritten by 15.

I think having the Calendar method return a list of fields that should be deleted from the fields of the Temporal object is a valid way of achieving the same thing, but it's my opinion that mergeFields() is an easier operation to reason about.

I don't think we should pass the Temporal object through to the calendar, because then mergeFields() would have to Get different fields depending on the type of the Temporal object.

polyfill/lib/plaindate.mjs
ptomato: I'm assuming that `monthCode` should be added to that line since it's part of all calendars and is present in implementations without Intl. `era` and `eraYear` should not, since they aren't present on the ISO calendar and aren't present in implementations without Intl.

polyfill/test/usercalendar.mjs
ptomato: It's specified that Calendar.fields() should make a copy. There's probably no need to actually do it, but this is just a throwaway calendar anyway.

docs/calendar.md
```diff
@@ -347,6 +347,38 @@ Temporal.Calendar.from('iso8601').fields(['month', 'day']);
...
+
+**Returns:** an object with properties from both `fields` and `extraFields`.
+
+This method does not need to be called directly except in specialized code.
+It is called indirectly when using the `with()` methods of `Temporal.PlainDateTime`, `Temporal.PlainDate`, `Temporal.PlainMonthDay`, `Temporal.PlainYearMonth`, and `Temporal.ZonedDateTime`.
+
+Custom calendars should override this method if they allow a calendar unit to be specified in more than one way.
+(For example, the Gregorian calendar allows years to be specified either by a `year` property or a combination of `era` and `eraYear`.)
+
+When subclassing `Temporal.Calendar`, this method doesn't need to be overridden, unless your calendar allows this with other properties than the built-in calendars do.
```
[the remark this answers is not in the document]
ptomato: I'm a bit hesitant to spend a lot of space on an explanation, because I think very few people are ever seriously going to write userland code that does this, but I have rephrased it a bit. :smile:

polyfill/lib/plaindate.mjs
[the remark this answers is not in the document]
justingrant: Seems reasonable. I don't have a strong opinion either way.

polyfill/test/usercalendar.mjs
```diff
@@ -293,6 +293,9 @@ describe('Userland calendar', () => {
...
+      },
+      mergeFields(fields, extraFields) {
+        return Object.assign(fields, extraFields);
```
sffc, line 298: Isn't this where you want to have the default logic that handles monthCode, era, and eraYear?

ptomato: Correct, but none of those are implemented in the polyfill yet, so there's nothing else for this method to do at this time. That will be added later.

docs/calendar.md
```diff
@@ -347,6 +347,39 @@ Temporal.Calendar.from('iso8601').fields(['month', 'day']);
...
 <!-- prettier-ignore-end -->
 
+### calendar.**mergeFields**(_fields_: object, _additionalFields_: object) : object
+
+**Parameters:**
+
+- `fields` (object): A plain object with properties representing calendar units.
+- `additionalFields` (object): Another plain object with properties representing calendar units.
+
+**Returns:** the same object that was passed in as `fields`, with its properties modified.
```
ljharb, line 357: why mutate, rather than returning a new object? The naive implementation would be `return { ...fields, ...additionalFields };`

ptomato: Good question; before Temporal calls this method, we have already copied all the properties onto a new object with ToTemporalDateFields etc. Returning a new object here would create a second new object unnecessarily.

ljharb: Hmm. Mutation is widely considered a bad practice, and potentially deoptimizes object access as well. I don't think the builtins should be worried about creating a new object "unnecessarily", especially since in the common case, engines will be able to silently optimize away the new object.

polyfill/lib/plaindate.mjs
cjtenny: (commented above, @ `ES.CalendarMergeFields` implementation; commenting here for reference)

polyfill/lib/ecmascript.mjs
```diff
@@ -1445,9 +1445,13 @@ export const ES = ObjectAssign({}, ES2020, {
...
     let fields = calendar.fields;
-    if (fields === undefined) fields = GetIntrinsic('%Temporal.Calendar.prototype.fields%');
-    const array = ES.Call(fields, calendar, [fieldNames]);
-    return ES.CreateListFromArrayLike(array, ['String']);
+    if (fields !== undefined) fieldNames = ES.Call(fields, calendar, [fieldNames]);
+    return ES.CreateListFromArrayLike(fieldNames, ['String']);
+  },
+  CalendarMergeFields: (calendar, fields, additionalFields) => {
+    let mergeFields = calendar.mergeFields;
+    if (mergeFields === undefined) return ObjectAssign(fields, additionalFields);
```
cjtenny, line 1453: Re: Justin's later comment about `monthCode`:

If we add `monthCode` where `month` is requested, then the base implementation of `MergeFields` needs to process `month` <--> `monthCode` conflicts for the ISO calendar; `Object.assign` wouldn't be sufficient in that case, and undefined `mergeFields` as `Object.assign` seems a potential source of bugs. I'm not sure if this means we don't default in `ES.CalendarMergeFields` or `Calendar.prototype.mergeFields`, or something else.

polyfill/lib/calendar.mjs
```diff
@@ -184,6 +187,9 @@ impl['iso8601'] = {
   fields(fields) {
     return fields;
   },
+  mergeFields(fields, additionalFields) {
+    return ObjectAssign(fields, additionalFields);
+  },
```
cjtenny, line 192: Will `Object.Assign` being the `ES.CalendarMergeFields` default if `mergeFields` is undefined make this code unnecessary? I think it can be omitted. Or is `mergeFields` only undefined if the prototype is overridden?

spec/calendar.html
```diff
@@ -392,6 +405,26 @@ <h1>ISODay ( _dateOrDateTime_ )</h1>
...
+        1. If _keys_ contains *"year"*, then
+          1. Perform ? DeletePropertyOrThrow(_fields_, *"era"*).
+          1. Perform ? DeletePropertyOrThrow(_fields_, *"eraYear"*).
+        1. If _keys_ contains *"era"* and *"eraYear"*, then
+          1. Perform ? DeletePropertyOrThrow(_fields_, *"year"*).
+        1. For each element _nextKey_ of _keys_, do
+          1. Let _propValue_ be ? Get(_additionalFields_, _nextKey_).
+          1. If _propValue_ is not *undefined*, then
+            1. Perform ? Set(_fields_, _nextKey_, _propValue_, *true*).
+        1. Return _fields_.
```
cjtenny, line 425: `Object.Assign` doesn't seem to match this :thinking: and shouldn't ISOMergeFields reject era/eraYear being present at all? (FWIW, validation should already do that, but I haven't specified it yet; however, if that text is removed, Object.Assign would be sufficient assuming validation has already rejected era parameters and ISOMergeFields is called after ToTemporal____Fields has been called on `_fields_` and `_additionalFields_`)

spec/calendar.html
```diff
@@ -843,9 +876,26 @@ <h1>Temporal.Calendar.prototype.fields ( _fields_ )</h1>
...
-        This function is the <dfn>%Temporal.Calendar.prototype.fields%</dfn> intrinsic object.
+        The `mergeFields` method takes two arguments, _fields_ and _additionalFields_.
+        The following steps are taken:
       </p>
+      <emu-alg>
+        1. Let _calendar_ be the *this* value.
+        1. Perform ? RequireInternalSlot(_calendar_, [[InitializedTemporalCalendar]]).
+        1. Assert: _calendar_.[[Identifier]] is *"iso8601"*.
+        1. Set _fields_ to ? ToObject(_fields_).
+        1. Set _additionalFields_ to ? ToObject(_additionalFields_).
```
cjtenny, line 896: if we call ToTemporal____Fields on the bags, we can change these to `Assert`s

spec/intl.html
```diff
@@ -1184,9 +1184,24 @@ <h1>Temporal.Calendar.prototype.fields ( _fields_ )</h1>
...
           <p>
-            This function is the <dfn>%Temporal.Calendar.prototype.fields%</dfn> intrinsic object.
+            The `mergeFields` method takes two arguments, _fields_ and _additionalFields_.
+            The following steps are taken:
           </p>
+          <emu-alg>
+            1. Let _calendar_ be the *this* value.
+            1. Perform ? RequireInternalSlot(_calendar_, [[InitializedTemporalCalendar]]).
+            1. Set _fields_ to ? ToObject(_fields_).
+            1. Set _additionalFields_ to ? ToObject(_additionalFields_).
```
cjtenny, line 1200: (again)

cjtenny (commented):
Looks good. Main question is around defaults before shipping it and how future handling of monthCode will impact when it's acceptable to default to Object.Assign

docs/calendar.md
ptomato: Fair enough, Justin brought up the same question but I hadn't taken into account that you can optimize out the call completely if it's a builtin calendar and Temporal.Calendar.prototype is unchanged. I'll change this.

polyfill/lib/ecmascript.mjs
ptomato: I wasn't taking `monthCode` into account in the polyfill since it's not implemented there yet. You're right that once `monthCode` is implemented, the default mergeFields can no longer be Object.assign. Once that happens, I'll split out a separate `ES.ISOMergeFields()` like it is done in the spec text, and call that here.

polyfill/lib/calendar.mjs
ptomato: This should be here nonetheless so that other calendars can chain up to it.

spec/calendar.html
ptomato: > `Object.Assign` doesn't seem to match this

You're right about that, but see a few comments back, this will change.

> shouldn't ISOMergeFields reject era/eraYear being present at all?

Oops, you're right about that also.

spec/calendar.html
ptomato: I'm not sure I agree with that, because user code can call `calendar.mergeFields(Math.PI, false)` regardless of what Temporal does.

docs/calendar.md
```diff
@@ -347,6 +347,39 @@ Temporal.Calendar.from('iso8601').fields(['month', 'day']);
...
+- `additionalFields` (object): Another plain object with properties representing calendar units.
+
+**Returns:** a new object with properties from both `fields` and `additionalFields`.
+
+This method does not need to be called directly except in specialized code.
+It is called indirectly when using the `with()` methods of `Temporal.PlainDateTime`, `Temporal.PlainDate`, `Temporal.PlainMonthDay`, `Temporal.PlainYearMonth`, and `Temporal.ZonedDateTime`.
+
+Custom calendars should override this method if they allow a calendar unit to be specified in more than one way.
+(For example, the Gregorian calendar allows years to be specified either by a `year` property or a combination of `era` and `eraYear`.)
+The overridden implementation should return an object with the properties from the original `fields` object and `additionalFields` copied onto it, and possibly properties removed.
```
cjtenny, line 364: Nit: properties are not removed from new object; something like "with some or all of the properties from..." might work better

spec/calendar.html
```diff
@@ -392,6 +405,27 @@ <h1>ISODay ( _dateOrDateTime_ )</h1>
...
+      </p>
+      <emu-alg>
+        1. Let _merged_ be ? OrdinaryObjectCreate(%Object.prototype%).
+        1. Let _originalKeys_ be ? _fields_.[[OwnPropertyKeys]]().
+        1. For each element _nextKey_ in _originalKeys_, do
+          1. Let _propValue_ be ? Get(_fields_, _nextKey_).
+          1. If _propValue_ is not *undefined*, then
+            1. Perform ? Set(_merged_, _nextKey_, _propValue_, *true*).
+        1. Let _newKeys_ be ? _additionalFields_.[[OwnPropertyKeys]]().
+        1. For each element _nextKey_ of _keys_, do
```
cjtenny, line 422: nit: `_keys_` undefined; DYM `_newKeys_`?

cjtenny (approved):
One language change, one typo, then lgtm!

spec/calendar.html
```diff
@@ -38,14 +38,27 @@ <h1>GetISO8601Calendar ( )</h1>
     <emu-clause id="sec-temporal-calendarfields" aoid="CalendarFields">
       <h1>CalendarFields ( _calendar_, _fieldNames_ )</h1>
       <p>
-        The CalendarFields abstract operation transforms a List of String values _fieldNames_ into another List of String values by calling the `fields` method of the given _calendar_ Object, falling back to the intrinsic operation if not present.
+        The CalendarFields abstract operation transforms a List of String values _fieldNames_ into another List of String values by calling the `fields` method of the given _calendar_ Object, falling back to a default operation if not present.
       </p>
       <emu-alg>
         1. Let _fields_ be ? Get(_calendar_, *"fields"*).
```
ljharb, line 44: should this use GetMethod, to ensure it's either nullish or a function?

ptomato: Agreed, this is another abstract op that I wasn't familiar with. I think it should be used consistently throughout the spec text, so I've opened #1323 for that.

PR #1311, opened 2021-01-20
Specify era/eraYear.

### LINKED ISSUE #1308: Add extra calendar field getters to the 402 portion of the spec text
I overlooked this before, the `era` getters and the Calendar.era() method have to be present in the 402 portion of the spec. According to the consensus at the 2021-01-19 meeting, so should `eraYear`. All of these return `undefined` for ISO calendar dates.

Ms2ger:
Fixes #1308.

### DIFF
--- a/spec/intl.html
+++ b/spec/intl.html
@@ -927,6 +927,38 @@
           </emu-alg>
         </emu-clause>
 
+        <emu-clause id="sec-temporal.calendar.prototype.era">
+          <h1>Temporal.Calendar.prototype.era ( _dateOrDateTime_ )</h1>
+          <p>
+            The `era` method takes one argument _dateOrDateTime_.
+            The following steps are taken:
+          </p>
+          <emu-alg>
+            1. Let _calendar_ be the *this* value.
+            1. Perform ? RequireInternalSlot(_calendar_, [[InitializedTemporalCalendar]]).
+            1. If _calendar_.[[Identifier]] is *"iso8601"*, then
+              1. Return *undefined*.
+            1. Let _era_ be the result of implementation-defined processing of _dateOrDateTime_ and the value of _calendar_.[[Identifier]].
+            1. Return 𝔽(_era_).
+          </emu-alg>
+        </emu-clause>
+
+        <emu-clause id="sec-temporal.calendar.prototype.erayear">
+          <h1>Temporal.Calendar.prototype.eraYear ( _dateOrDateTime_ )</h1>
+          <p>
+            The `eraYear` method takes one argument _dateOrDateTime_.
+            The following steps are taken:
+          </p>
+          <emu-alg>
+            1. Let _calendar_ be the *this* value.
+            1. Perform ? RequireInternalSlot(_calendar_, [[InitializedTemporalCalendar]]).
+            1. If _calendar_.[[Identifier]] is *"iso8601"*, then
+              1. Return *undefined*.
+            1. Let _eraYear_ be the result of implementation-defined processing of _dateOrDateTime_ and the value of _calendar_.[[Identifier]].
+            1. Return 𝔽(_eraYear_).
+          </emu-alg>
+        </emu-clause>
+
         <emu-clause id="sup-temporal.calendar.prototype.year">
           <h1>Temporal.Calendar.prototype.year ( _dateOrDateTime_ )</h1>
           <p>This definition supersedes the definition provided in <emu-xref href="#sec-temporal.calendar.prototype.year"></emu-xref>.</p>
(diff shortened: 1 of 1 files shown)


spec/intl.html
```diff
@@ -927,6 +927,38 @@ <h1>Temporal.Calendar.prototype.dateUntil ( _one_, _two_, _options_ )</h1>
...
+          <p>
+            The `era` method takes one argument _dateOrDateTime_.
+            The following steps are taken:
+          </p>
+          <emu-alg>
+            1. Let _calendar_ be the *this* value.
+            1. Perform ? RequireInternalSlot(_calendar_, [[InitializedTemporalCalendar]]).
+            1. If _calendar_.[[Identifier]] is *"iso8601"*, then
+              1. Return *undefined*.
+            1. Let _era_ be the result of implementation-defined processing of _dateOrDateTime_ and the value of _calendar_.[[Identifier]].
```
ljharb, line 941: this is very hand-wavey. can this be elaborated in, or made less imprecise?

spec/intl.html
```diff
@@ -927,6 +927,38 @@ <h1>Temporal.Calendar.prototype.dateUntil ( _one_, _two_, _options_ )</h1>
...
+          <p>
+            The `eraYear` method takes one argument _dateOrDateTime_.
+            The following steps are taken:
+          </p>
+          <emu-alg>
+            1. Let _calendar_ be the *this* value.
+            1. Perform ? RequireInternalSlot(_calendar_, [[InitializedTemporalCalendar]]).
+            1. If _calendar_.[[Identifier]] is *"iso8601"*, then
+              1. Return *undefined*.
+            1. Let _era_ be the result of implementation-defined processing of _dateOrDateTime_ and the value of _calendar_.[[Identifier]].
```
ljharb, line 957: also here

spec/intl.html
```diff
@@ -927,6 +927,38 @@ <h1>Temporal.Calendar.prototype.dateUntil ( _one_, _two_, _options_ )</h1>
...
+            1. Return 𝔽(_era_).
+          </emu-alg>
+        </emu-clause>
+
+        <emu-clause id="sup-temporal.calendar.prototype.erayear">
+          <h1>Temporal.Calendar.prototype.eraYear ( _dateOrDateTime_ )</h1>
+          <p>
+            The `eraYear` method takes one argument _dateOrDateTime_.
```
ljharb, line 949:
```suggestion
            The `eraYear` method takes one argument, _dateOrDateTime_.
```
What kind of argument?

spec/intl.html
```diff
@@ -927,6 +927,38 @@ <h1>Temporal.Calendar.prototype.dateUntil ( _one_, _two_, _options_ )</h1>
           </emu-alg>
         </emu-clause>
 
+        <emu-clause id="sup-temporal.calendar.prototype.era">
+          <h1>Temporal.Calendar.prototype.era ( _dateOrDateTime_ )</h1>
+          <p>
+            The `era` method takes one argument _dateOrDateTime_.
```
ljharb, line 933:
```suggestion
            The `era` method takes one argument, _dateOrDateTime_.
```
What kind of argument?

spec/intl.html
```diff
@@ -1187,6 +1219,36 @@ <h1>Temporal.PlainDate.prototype.toLocaleString ( [ _locales_ [ , _options_ ] ]
...
+          <p>
+            `Temporal.PlainDate.prototype.era` is an accessor property whose set accessor function is *undefined*.
+            Its get accessor function performs the following steps:
+          </p>
+          <emu-alg>
+            1. Let _plainDate_ be the *this* value.
+            1. Perform ? RequireInternalSlot(_plainDate_, [[InitializedTemporalDate]]).
+            1. Let _calendar_ be _plainDate_.[[Calendar]].
+            1. Let _method_ be ? Get(_calendar_, *"era"*).
+            1. Return ? Call(_method_, _calendar_, « _plainDate_ »).
```
ljharb, line 1234: Any reason not to use Invoke here? (same with all the Get + Calls)

ptomato: I hadn't encountered Invoke before! ... Good suggestion, I'll open a new issue for that.

spec/intl.html
```diff
@@ -927,6 +927,38 @@ <h1>Temporal.Calendar.prototype.dateUntil ( _one_, _two_, _options_ )</h1>
           </emu-alg>
         </emu-clause>
 
+        <emu-clause id="sup-temporal.calendar.prototype.era">
```
ptomato, line 930: These should be `sec-` (here and below) because they don't supersede anything in the 262 part of the document.

ptomato (approved):
Not sure what to do about the vagueness of the "implementation-defined processing" bits. We can't really specify out all the ICU calendars, but on the other hand I do take the point that it is weird to have methods that return either `undefined` or an implementation-defined result with no explanation.

Would that situation be sufficiently improved when we add the corresponding explanation in the documentation, or does there need to be some explanation in the spec text as well?

ljharb:
I think it needs to be clear in the spec steps.

Ms2ger:
Editorial improvements are still welcome, of course, but I've merged this in the interest of stabilizing the normative parts of the spec.

ljharb:
Given there’s still a number of open review comments, I’m not sure why merging prior to addressing them is stabilizing.

PR #1313, opened 2021-01-20
Fix paths for codecov.

Ms2ger:


Ms2ger:
Per https://docs.codecov.io/docs/fixing-paths

ptomato:
I'm assuming this is to get the compare view working in Codecov?

ptomato (approved):

PR #1318, opened 2021-01-21
Remove getFields methods from Duration object.

cjtenny:
Ref: #1306, #1312


cjtenny:
I'm not sure if we also intended to remove these as well, but if so, here's a change doing so. My validation WIP broke the test262 Duration.getFields tests and I realized it might be a holdout.

Ms2ger (approved):
Please also update `spec/duration.html`

cjtenny:
Oops, thought I did but didn't cherry-pick it out. Wilco!

PR #1328, opened 2021-01-26
Options handling

ptomato:
Closes: #1253

### DIFF
--- a/spec/abstractops.html
+++ b/spec/abstractops.html
@@ -571,19 +571,9 @@
         1. Let _calendar_ be ? GetOptionalTemporalCalendar(_value_).
         1. Let _fieldNames_ be ? CalendarFields(_calendar_, « *"day"*, *"month"*, *"year"* »).
         1. Let _fields_ be ? ToTemporalDateTimeFields(_value_, _fieldNames_).
-        1. Let _temporalDate_ be ? DateFromFields(_calendar_, _fields_, %Temporal.PlainDate%).
-        1. Let _timeResult_ be ? ToTemporalTimeRecord(_fields_).
-        1. Let _result_ be the new Record {
-            [[Year]]: _temporalDate_.[[ISOYear]],
-            [[Month]]: _temporalDate_.[[ISOMonth]],
-            [[Day]]: _temporalDate_.[[ISODay]],
-            [[Hour]]: _timeResult_.[[Hour]],
-            [[Minute]]: _timeResult_.[[Minute]],
-            [[Second]]: _timeResult_.[[Second]],
-            [[Millisecond]]: _timeResult_.[[Millisecond]],
-            [[Microsecond]]: _timeResult_.[[Microsecond]],
-            [[Nanosecond]]: _timeResult_.[[Nanosecond]]
-          }.
+        1. Let _options_ be ! OrdinaryObjectCreate(%Object.prototype%).
+        1. Perform ! CreateDataPropertyOrThrow(_options_, *"overflow"*, *"constrain"*).
+        1. Let _result_ be ? InterpretTemporalDateTimeFields(_calendar_, _fields_, _options_).
         1. Let _offset_ be ? Get(_value_, *"offset"*).
         1. Let _timeZone_ be ? Get(_value_, *"timeZone"*).
       1. Else,
@@ -645,6 +635,22 @@
       1. If either _u1_ or _u2_ is *"milliseconds"*, return *"milliseconds"*.
       1. If either _u1_ or _u2_ is *"microseconds"*, return *"microseconds"*.
       1. Return *"nanoseconds"*.
+    </emu-alg>
+  </emu-clause>
+
+  <emu-clause id="sec-temporal-mergelargestunitoption" aoid="MergeLargestUnitOption">
+    <h1>MergeLargestUnitOption ( _options_, _largestUnit_ )</h1>
+    <p>
+      The abstract operation MergeLargestUnitOption returns a new plain Object with enumerable own String properties copied from the _options_ Object, and the `largestUnit` property set to _largestUnit_.
+    </p>
+    <emu-alg>
+      1. Let _merged_ be ! OrdinaryObjectCreate(%Object.prototype%).
+      1. Let _keys_ be ? EnumerableOwnPropertyNames(_options_, ~key~).
+      1. For each element _nextKey_ of _keys_, do
+        1. Let _propValue_ be ? Get(_options_, _nextKey_).
+        1. Perform ! CreateDataPropertyOrThrow(_merged_, _nextKey_, _propValue_).
+      1. Perform ! CreateDataPropertyOrThrow(_merged_, *"largestUnit"*, _largestUnit_).
+      1. Return _merged_.
     </emu-alg>
   </emu-clause>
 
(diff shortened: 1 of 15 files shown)


spec/abstractops.html
```diff
@@ -571,19 +571,9 @@ <h1>ToRelativeTemporalObject ( _options_ )</h1>
...
-            [[Month]]: _temporalDate_.[[ISOMonth]],
-            [[Day]]: _temporalDate_.[[ISODay]],
-            [[Hour]]: _timeResult_.[[Hour]],
-            [[Minute]]: _timeResult_.[[Minute]],
-            [[Second]]: _timeResult_.[[Second]],
-            [[Millisecond]]: _timeResult_.[[Millisecond]],
-            [[Microsecond]]: _timeResult_.[[Microsecond]],
-            [[Nanosecond]]: _timeResult_.[[Nanosecond]]
-          }.
+        1. Let _options_ be ? OrdinaryObjectCreate(%Object.prototype%).
```
ljharb, line 574: can this throw? I wouldn’t expect so.
```suggestion
        1. Let _options_ be ! OrdinaryObjectCreate(%Object.prototype%).
```

spec/abstractops.html
```diff
@@ -648,6 +638,24 @@ <h1>LargerOfTwoTemporalDurationUnits ( _u1_, _u2_ )</h1>
...
   </emu-clause>
 
+  <emu-clause id="sec-temporal-mergelargestunitoption" aoid="MergeLargestUnitOption">
+    <h1>MergeLargestUnitOption ( _options_, _largestUnit_ )</h1>
+    <p>
+      The abstract operation MergeLargestUnitOption returns a new plain Object with the properties copied from the _options_ Object, and the `largestUnit` property set to _largestUnit_.
+    </p>
+    <emu-alg>
+      1. Let _merged_ be ! OrdinaryObjectCreate(%Object.prototype%).
+      1. Let _keys_ be ? _options_.[[OwnPropertyKeys]]().
```
ljharb, line 648: do you mean to allow symbols here too?

spec/abstractops.html
```diff
@@ -648,6 +638,24 @@ <h1>LargerOfTwoTemporalDurationUnits ( _u1_, _u2_ )</h1>
...
+      1. For each element _nextKey_ of _keys_, do
+        1. Let _desc_ be ? _options_.[[GetOwnProperty]](_nextKey_).
+        1. If _desc_ is not *undefined* and _desc_.[[Enumerable]] is *true*, then
```
ljharb, line 651: if you’re already filtering to own enumerable properties, and if symbol keys aren’t needed, could this whole algorithm use https://tc39.es/ecma262/#sec-enumerableownpropertynames ?

another option is to call `%Object.assign%` directly, since that’s basically what this is doing?

spec/plaindate.html
```diff
@@ -715,21 +708,21 @@ <h1>CreateTemporalDateFromStatic ( _constructor_, _isoYear_, _isoMonth_, _isoDay
...
-      <h1>ToTemporalDate ( _item_ [ , _constructor_ [ , _overflow_ ] ] )</h1>
+      <h1>ToTemporalDate ( _item_ [ , _constructor_ [ , _options_ ] ] )</h1>
       <p>
         The abstract operation ToTemporalDate returns its argument _item_ if it is already a Temporal.PlainDate instance, converts _item_ to a new Temporal.PlainDate instance if possible, and throws otherwise.
       </p>
       <emu-alg>
         1. If _constructor_ is not given, set it to %Temporal.PlainDate%.
-        1. If _overflow_ is not given, set it to *"constrain"*.
-        1. Assert: _overflow_ is either *"constrain"* or *"reject"*.
+        1. If _options_ is not given, set it to ! OrdinaryObjectCreate(%Object.prototype%).
```
ljharb, line 717:
```suggestion
        1. If _options_ is not present, set it to ! OrdinaryObjectCreate(%Object.prototype%).
```
altho the editorial direction is to remove optional arguments from AOs, so it’d probably be better to provide an explicit `~empty~` in the “no options” case.

justingrant (commented):
Overall LGTM. One thing though: this PR expands the scope of calendars so that any calendar method that accepts options needs to be in charge of interpreting and validating options. Doesn't there need to be changes in the Temporal.Calendar docs that notes that calendar authors are responsible for interpreting and validating options?  I assume that we should also include guidance for how this interpretation and validation should work. Here's some text as a starting point:

> Custom calendars are responsible for interpreting and validating all inputs, including `options`. Calendars should (and built-in calendars will) throw a TypeError if a required option is missing or has the wrong type, but throw a RangeError if it's present but has an invalid value.

> Calendars are also responsible for assigning default values. For example, if the `overflow` option is `undefined`, it will be interpreted by built-in calendars as `'constrain'`. Custom calendars should maintain this behavior unless there's a good reason not to.  Calendars can also accept additional non-default values for existing options or can accept new options that built-in calendars don't. When adding new options, calendar authors should use a unique prefix, e.g. the name of the calendar, to avoid potential conflicts with future options which may be used by Temporal.

spec/abstractops.html
ptomato: Probably not!

spec/plaindate.html
ptomato: I've made this change, but I think it would be a large change to remove all optional arguments from the Temporal spec text, so I'll leave that out of scope of this PR for now. That sounds like an editorial thing that could be done before Stage 4. Would you mind opening a new issue with more info about what is currently considered the best practice?

ptomato:
That text seems great to me.

spec/abstractops.html
```diff
@@ -648,6 +638,22 @@ <h1>LargerOfTwoTemporalDurationUnits ( _u1_, _u2_ )</h1>
     </emu-alg>
   </emu-clause>
 
+  <emu-clause id="sec-temporal-mergelargestunitoption" aoid="MergeLargestUnitOption">
+    <h1>MergeLargestUnitOption ( _options_, _largestUnit_ )</h1>
+    <p>
+      The abstract operation MergeLargestUnitOption returns a new plain Object with the properties copied from the _options_ Object, and the `largestUnit` property set to _largestUnit_.
```
ljharb, line 644:
```suggestion
      The abstract operation MergeLargestUnitOption returns a new plain Object with enumerable own String properties copied from the _options_ Object, and the `largestUnit` property set to _largestUnit_.
```

spec/abstractops.html
```diff
@@ -648,6 +638,22 @@ <h1>LargerOfTwoTemporalDurationUnits ( _u1_, _u2_ )</h1>
...
+    </p>
+    <emu-alg>
+      1. Let _merged_ be ! OrdinaryObjectCreate(%Object.prototype%).
+      1. Let _keys_ be ? EnumerableOwnPropertyNames(_options_, ~key~).
+      1. For each element _nextKey_ of _keys_, do
+        1. Let _propValue_ be ? Get(_options_, _nextKey_).
+        1. Perform ? Set(_merged_, _nextKey_, _propValue_, *true*).
+      1. Perform ? Set(_merged_, *"largestUnit"*, _largestUnit_, *true*).
```
ljharb, line 652: Why Set here? I'd expect to not invoke random Object.prototype setters that users have installed:
```suggestion
        1. Perform ! CreateDataPropertyOrThrow(_merged_, _nextKey_, _propValue_, *true*).
      1. Perform ! CreateDataPropertyOrThrow(_merged_, *"largestUnit"*, _largestUnit_, *true*).
```

cjtenny (approved):
Agree with both Justin's & Jordan's comments; once those are addressed, good to go.

PR #1333, opened 2021-01-27
Update samples on documentation

tegandbiscuits:
A comma was missing from a code sample so I fixed that so it's copy and pastable.

The snippet

```js
const dateTIme1 = dateTime.with({
  minutes: 17,
  seconds: 19,
});
```

raised the error `Uncaught TypeError: invalid date-time-like` using the plurals. I assume that the docs should show what currently works, but if it should support `minutes` and `seconds` then I can change it back.


ptomato (approved):
Thanks very much, this is indeed correct!

PR #1342, opened 2021-02-01
fix: remove balance word

HERBOD7:
**Description of the Change**
remove balance word from calendar doc

**Release Notes**

- Fix issue [#1320](https://github.com/tc39/proposal-temporal/issues/1320)


ryzokuken (approved):
Thanks for the PR!

PR #1343, opened 2021-02-02
Clarify invalid duration field error message

### LINKED ISSUE #1233: error messages resulting because instant is not calendar aware
java.time has the same distinction between calendar-aware-or-not entities - and in my experience as a maintainer of libraries around that [1], beginners and even intermediates trip up on this again and again. 

My proposal would be to improve the error messages when this mistake is made. For example

```
Temporal.now.instant().add(Temporal.Duration.from({years: 1}))  
```

The error message is correct, but it would save many people a lot of time if the message pointed people to a link about calendar-awareness or somehow explained the problem sufficiently.

[1]  https://github.com/henryw374/cljc.java-time/issues/18

justingrant:
Fixes #1233

### DIFF
--- a/polyfill/lib/ecmascript.mjs
+++ b/polyfill/lib/ecmascript.mjs
@@ -571,7 +571,9 @@ export const ES = ObjectAssign({}, ES2020, {
     ES.RejectDuration(years, months, weeks, days, hours, minutes, seconds, milliseconds, microseconds, nanoseconds);
     for (const property of disallowedProperties) {
       if (record[property] !== 0) {
-        throw new RangeError(`invalid duration field ${property}`);
+        throw new RangeError(
+          `Duration field ${property} not supported by Temporal.Instant. Try Temporal.ZonedDateTime instead.`
+        );
       }
     }
     return record;


polyfill/lib/ecmascript.mjs
```diff
@@ -571,7 +571,9 @@ export const ES = ObjectAssign({}, ES2020, {
     ES.RejectDuration(years, months, weeks, days, hours, minutes, seconds, milliseconds, microseconds, nanoseconds);
     for (const property of disallowedProperties) {
       if (record[property] !== 0) {
-        throw new RangeError(`invalid duration field ${property}`);
+        throw new RangeError(
+          `Duration field ${property} not supported by Temporal.Instant. Try Temporal.ZonedDateTime instead.`
```
justingrant, line 575: Even though this Abstract Operation is generic in theory, in practice it's only used by the Instant type (and we're not changing the API anymore!) so this type-specific error message seems safe.

ryzokuken: @justingrant also since the error messages aren't specified, I think it's not a problem :)

ryzokuken (approved):
Looks great, thanks!

PR #1347, opened 2021-02-04
Convert "bc"/"ad" eras to "bce"/"ce" in Gregorian calendar

justingrant:
This PR changes the `bc` and `ad` eras in the `gregory` calendar to `ce` and `bce`, respectively, per   
https://github.com/unicode-org/icu4x/issues/470#issuecomment-772735333.


cjtenny (approved):
Ship it!

PR #1354, opened 2021-02-08
Docs: show clipped content at right edge of persistence-model.svg

justingrant:
This PR fixes the size of persistence-model.svg so the right edge of the diagram won't be clipped in the docs home page.

Current: 
image

Fixed: 
image


ptomato (approved):
:+1:

PR #1359, opened 2021-02-11
Use GetMethod for looking up calendar and time zone methods

### LINKED ISSUE #1323: Use GetMethod for looking up calendar and time zone methods
Follow up from #1310, using [GetMethod](https://tc39.es/ecma262/#sec-getmethod) instead of Get will ensure that the property is either nullish or callable.

cjtenny:
Fixes #1323.


ptomato (approved):
:+1:

ljharb (approved):

PR #1362, opened 2021-02-11
Bugfix: Make ISO monthCode abstract operations match polyfill.

cjtenny:
This is subject to further change based on IETF standardization of monthCodes; it is not going to be defined by temporal.

Fixes #1348.


ptomato (approved):
:+1:

PR #1366, opened 2021-02-12
docs: Clarify that the from() overflow option doesn't apply to strings

ptomato:
Closes: #797


ptomato:
(So cool that my documentation change increased code coverage)

cjtenny (approved):
codecov bot knows something we don't, and it's holding it over us. Someday...

ryzokuken (approved):
Thanks!

PR #1372, opened 2021-02-19
Fix Duration docs code samples

justingrant:
There were a few incomplete or incorrect code samples in the Duration docs. This PR fixes them. I made a few minor tweaks in the Duration docs while I was in the neighborhood.


ptomato (approved):
:+1:

PR #1373, opened 2021-02-19
polyfill: Fix iso week number

redneb:
The current algorithm for calculating the ISO week number is wrong for dates that belong to an ISO week from the previous year. For example `Temporal.Calendar.from("iso8601").weekOfYear("2021-01-01")` returns 52 but the correct answer is 53.

The problem is in the function `ES.WeekOfYear`: This function computes first the expression `const week = MathFloor((doy - dow + 10) / 7);` and then, if the answer is 0, this means that the day belongs to the last ISO week of the previous year. So the result of the function must be the number of ISO weeks of the *previous* year, which is 53 if the previous year ends on a Thursday (regardless if its a leap or not) or if ends on a Friday and is a leap. Otherwise the result must be 52 ([source](https://en.wikipedia.org/wiki/ISO_week_date#Weeks_per_year)).

### DIFF
--- a/polyfill/lib/ecmascript.mjs
+++ b/polyfill/lib/ecmascript.mjs
@@ -2028,7 +2028,7 @@ export const ES = ObjectAssign({}, ES2020, {
     const week = MathFloor((doy - dow + 10) / 7);
 
     if (week < 1) {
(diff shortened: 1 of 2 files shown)


ptomato (approved):
:+1: Thank you!

PR #1375, opened 2021-02-19
Make Duration _repr_ match toString() output

### LINKED ISSUE #1371: Unexpected zero seconds shown in Duration.prototype._repr_ output
The debugging-only `_repr_` getter (which is used to show the contents of a Duration in the polyfill while debugging with Chrome devtools console) has an extra `0.0S` appended to its output that should not be there. The output should match `toString()`.

Repro: run the following code in Chrome devtools console:

```js
Temporal.Duration.from('P73DT4H')
```

Expected: `Duration {_repr_: "Temporal.Duration <P73DT4H>"}` (`_repr_` should have same output as `toString()`)
Actual: `Duration {_repr_: "Temporal.Duration <P73DT4H0.0S>"}`

Note that the `_repr_` is only for debugging the pre-production polyfill. It's not present in the 262 spec so fixing this won't affect the spec.

justingrant:
Fixes #1371

### DIFF
--- a/polyfill/lib/duration.mjs
+++ b/polyfill/lib/duration.mjs
@@ -73,8 +73,10 @@ export class Duration {
     SetSlot(this, NANOSECONDS, nanoseconds);
 
     if (typeof __debug__ !== 'undefined' && __debug__) {
+      const { precision, unit, increment } = ES.ToDurationSecondsStringPrecision({});
+      const s = ES.TemporalDurationToString(this, precision, { unit, increment, roundingMode: 'trunc' });
       Object.defineProperty(this, '_repr_', {
-        value: `${this[Symbol.toStringTag]} <${ES.TemporalDurationToString(this)}>`,
+        value: `${this[Symbol.toStringTag]} <${s}>`,
         writable: false,
         enumerable: false,
         configurable: false


ptomato (commented):
Good catch, I had found the other day that this was broken in `toJSON()` and included a [fix](https://github.com/tc39/proposal-temporal/pull/1370/commits/3b9e891dacf74410e530051df1ab85b173acd660) in #1370 but had missed this one.

It should be OK to just add `ES.TemporalDurationToString(this, 'auto')` though.

justingrant:
> Good catch, I had found the other day that this was broken in `toJSON()` and included a [fix](https://github.com/tc39/proposal-temporal/pull/1370/commits/3b9e891dacf74410e530051df1ab85b173acd660) in #1370 but had missed this one.
> 
> It should be OK to just add `ES.TemporalDurationToString(this, 'auto')` though.

Given that #1370 seems to fix most of these cases, do you want to just fix _repr_ in #1370 too instead of spreading fixes across multiple PRs?

BTW, instead of updating all callsites to use an `'auto'` parameter, would a better fix be to treat `undefined` as `'auto'` in `ES.TemporalDurationToString`?  It seems odd that the default behavior of ES.TemporalDurationToString isn't the same as `'auto'`.

ptomato:
Yeah, makes sense. I'll close this and pull out that fix from the other PR into a new PR.

PR #1378, opened 2021-02-20
Update polyfill to use zero-padded month codes

### LINKED ISSUE #1369: 'monthCode' for built-in calendars does not behave as documented.
Document, Chapter `Temporal.PlainDate`, about `monthCode`:

"For common (non-leap) months, `monthCode` should be `M${month}`, where _month is zero padded up to two digits_."

Extracting `monthCode` from a date is `iso8601`:

`Temporal.PlainDate.from('2021-02-18').monthCode`
`//> "M2"`  
should be "M02" as per definition.`

Same behaviour with hebrew calendar ('M5', 'M5L'...), chinese...

cjtenny:
Fixes #1369.

### DIFF
--- a/polyfill/lib/calendar.mjs
+++ b/polyfill/lib/calendar.mjs
@@ -1038,7 +1038,7 @@ const helperHebrew = ObjectAssign({}, nonIsoHelperBase, {
       this.validateCalendarDate(calendarDate);
       if (month === undefined) {
         if (monthCode.endsWith('L')) {
-          if (monthCode !== 'M05L') throw new RangeError(`Hebrew leap month must have monthCode M5L, not ${monthCode}`);
+          if (monthCode !== 'M05L') throw new RangeError(`Hebrew leap month must have monthCode M05L, not ${monthCode}`);
           month = 6;
           if (!this.inLeapYear({ year })) {
             if (overflow === 'reject') {
--- a/polyfill/lib/calendar.mjs
+++ b/polyfill/lib/calendar.mjs
@@ -254,7 +254,7 @@ impl['iso8601'] = {
   },
   monthCode(date) {
     if (!HasSlot(date, ISO_MONTH)) date = ES.ToTemporalDate(date, GetIntrinsic('%Temporal.PlainDate%'));
-    return `M${ES.ToString(GetSlot(date, ISO_MONTH))}`;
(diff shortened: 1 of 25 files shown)


ptomato (approved):
Looks correct. (I only skimmed the giant search-and-replace part.) It looks like there is some more search-and-replace necessary in the test262 suite though.

polyfill/lib/calendar.mjs
```diff
@@ -538,7 +538,7 @@ const nonIsoHelperBase = {
       if (typeof monthCode !== 'string') {
         throw new RangeError(`monthCode must be a string, not ${ES.Type(monthCode).toLowerCase()}`);
       }
-      if (!/^M(1?\d)(L?)$/.test(monthCode)) throw new RangeError(`Invalid monthCode: ${monthCode}`);
+      if (!/^M(\d?\d)(L?)$/.test(monthCode)) throw new RangeError(`Invalid monthCode: ${monthCode}`);
```
justingrant, line 541:
```suggestion
      if (!/^M([01]?\d)(L?)$/.test(monthCode)) throw new RangeError(`Invalid monthCode: ${monthCode}`);
```

polyfill/lib/calendar.mjs
```diff
@@ -323,13 +323,13 @@ function resolveNonLunisolarMonth(calendarDate, overflow = undefined, monthsPerY
     // rely on this function to constrain/reject out-of-range `month` values.
     if (overflow === 'reject') ES.RejectToRange(month, 1, monthsPerYear);
     if (overflow === 'constrain') month = ES.ConstrainToRange(month, 1, monthsPerYear);
-    monthCode = `M${month}`;
+    monthCode = `M${month.toString().padStart(2, '0')}`;
```
justingrant, line 326: Given how often this pattern is repeated, seems worthy of pulling out into a `buildMonthCode` function either locally in this file or as an `ES` method.  Would shave some bytes off the polyfill.

justingrant (commented):
Looks good with a few tweaks

polyfill/lib/calendar.mjs
cjtenny: done!

ptomato (approved):
Just one straggler that I noticed in an exception message, please go ahead and merge after fixing that.

PR #1380, opened 2021-02-21
Duration.prototype.round: Throw on no largest or smallest unit provided.

cjtenny:
Fix bug where largestUnit: 'auto' is considered to be no unit provided.

Fixes #1379.

### DIFF
--- a/polyfill/lib/ecmascript.mjs
+++ b/polyfill/lib/ecmascript.mjs
@@ -776,19 +776,7 @@ export const ES = ObjectAssign({}, ES2020, {
       ['microsecond', 'microseconds'],
       ['nanosecond', 'nanoseconds']
     ]);
-    const allowed = new Set([
-      'years',
-      'months',
-      'weeks',
-      'days',
-      'hours',
-      'minutes',
-      'seconds',
-      'milliseconds',
-      'microseconds',
-      'nanoseconds'
-    ]);
-    const retval = ES.GetOption(options, 'largestUnit', ['auto', ...allowed, ...plural.keys()]);
+    const retval = ES.GetOption(options, 'largestUnit', ['auto', ...plural.keys(), ...plural.values(), 'weeks']);
     if (plural.has(retval)) return plural.get(retval);
     return retval;
   },
--- a/polyfill/lib/ecmascript.mjs
+++ b/polyfill/lib/ecmascript.mjs
@@ -764,6 +764,34 @@ export const ES = ObjectAssign({}, ES2020, {
     if (plural.has(retval)) return plural.get(retval);
     return retval;
   },
+  ToLargestTemporalDurationUnit: (options) => {
+    const plural = new Map([
+      ['year', 'years'],
+      ['month', 'months'],
+      ['day', 'days'],
+      ['hour', 'hours'],
+      ['minute', 'minutes'],
+      ['second', 'seconds'],
+      ['millisecond', 'milliseconds'],
+      ['microsecond', 'microseconds'],
+      ['nanosecond', 'nanoseconds']
+    ]);
+    const allowed = new Set([
+      'years',
+      'months',
+      'weeks',
+      'days',
+      'hours',
+      'minutes',
+      'seconds',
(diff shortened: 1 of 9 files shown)


cjtenny:
(should probably add a test for these specific cases before merging, but will wait for agreement on #1379 first)

spec/duration.html
```diff
@@ -431,16 +431,20 @@ <h1>Temporal.Duration.prototype.round ( _options_ )</h1>
...
         1. Let _defaultLargestUnit_ be ! DefaultTemporalLargestUnit(_duration_.[[Years]], _duration_.[[Months]], _duration_.[[Weeks]], _duration_.[[Days]], _duration_.[[Hours]], _duration_.[[Minutes]], _duration_.[[Seconds]], _duration_.[[Milliseconds]], _duration_.[[Microseconds]], _duration_.[[Nanoseconds]]).
         1. Set _defaultLargestUnit_ to ! LargerOfTwoTemporalDurationUnits(_defaultLargestUnit_, _smallestUnit_).
-        1. Let _largestUnit_ be ? ToLargestTemporalUnit(_options_, *undefined*).
+        1. Let _largestUnit_ be ? ToLargestTemporalDurationUnit(_options_).
         1. If _largestUnit_ is *undefined*, then
           1. Set _largestUnitPresent_ to *false*.
           1. Set _largestUnit_ to _defaultLargestUnit_.
+        1. Else if _largestUnit_ is *"auto"*, then
+          1. Set _largestUnit_ to _defaultLargestUnit_.
+        1. If _smallestUnitPresent_ is *false* and -largestUnitPresent_ is *false*, then
```
ptomato, line 446:
```suggestion
        1. If _smallestUnitPresent_ is *false* and _largestUnitPresent_ is *false*, then
```

spec/duration.html
```diff
@@ -1452,7 +1456,7 @@ <h1>AdjustRoundedDurationDays ( _years_, _months_, _weeks_, _days_, _hours_, _mi
       </emu-alg>
     </emu-clause>
 
-    <emu-clause id="sec-temporal-tolimitedtemporalduration" aoid="ToLimitedTemporalDuration">
+    <emu-clause id="sec-temporal-tolimitedtemporalduration" aoid="ToiLmitedTemporalDuration">
```
ptomato, line 1459: mistake?

polyfill/lib/ecmascript.mjs
```diff
@@ -764,6 +764,34 @@ export const ES = ObjectAssign({}, ES2020, {
...
+      'weeks',
+      'days',
+      'hours',
+      'minutes',
+      'seconds',
+      'milliseconds',
+      'microseconds',
+      'nanoseconds'
+    ]);
+    const retval = ES.GetOption(options, 'largestUnit', ['auto', ...allowed, ...plural.keys()]);
```
ptomato, line 791: Could be written as `['auto', ...plural.keys(), ...plural.values(), 'weeks']` as in ToTemporalDurationTotalUnit, but it's not really consequential; up to you.

ptomato (commented):
Agreed, this should have a test before merging. Otherwise looks fine if this is the behaviour we indeed want.

spec/duration.html
cjtenny: :flushed:

polyfill/lib/ecmascript.mjs
cjtenny: Sounds good, I just did the quick and dirty copy of ToLargestTemporalUnit to start.

cjtenny:
I added a commit fixing `To(Largest|Smallest)?Temporal(Duration)?Unit` calls because many of them had incorrect argument order. The smallest way to do it was change `ToSmallestTemproalDurationUnit` to take the same argument order as `ToLargestTemporalUnit` and fix the inconsistent callsites. The inconsistency between those two abstract ops is probably why this was wrong in a number of places to begin with...

ptomato (approved):
Looks good, please do the relevant squashes before merging so that only the two commits remain.

PR #1393, opened 2021-02-26
Various bugfixes and making things consistent

ptomato:
Encountered while investigating #1294


cjtenny (approved):
All looks good, thanks for fixing these! I'm surprised it didn't require any tests changes. I don't think it's worth adding test/test262 changes right now for all of these (e.g. verifying that exceptions don't call timeZone.toString), but I think we should add tests (doesn't have to be now; test262 seems fine) for:
- Properly checking callable on retrieved methods
- That timeZone.getInstantFor properly throws on non-Temporal.Instant values
(eventually). Do we have a list of 'things missing from the test262 tests' somewhere?

ptomato:
Cam and I discussed that we do need to start keeping a checklist of things that need to be test262'd.

PR #1395, opened 2021-02-26
ISOMonthDayFromFields: Reject leap day in appropriate cases when calendar is specified.

cjtenny:
ToTemporalMonthDay accepts 'month' without 'monthCode' or 'year' when no calendar has been specified, for users writing calendar-insensitive code. However, when a calendar has been specified, ISOMonthDayFromFields will still accept 2/29/2001 specified via month and year. This change makes ISOMonthDayFromFields reject or constrain 02-29 when month + year are specified, and keeps acceptance of 02-29 when only monthCode + day are specified. See the added test case for some examples.


ptomato (approved):
This looks like a correct implementation of what we had discussed! Unfortunate that I didn't get it right the first time, but good that you noticed we can remove all those unused abstract operations.

PR #1397, opened 2021-02-27
Remove some internal TimeZone method calls

ptomato:
Found this bug while investigating #1294.

### DIFF
--- a/polyfill/test/ZonedDateTime/prototype/toPlainDateTime/plain-custom-timezone.js
+++ b/polyfill/test/ZonedDateTime/prototype/toPlainDateTime/plain-custom-timezone.js
@@ -8,15 +8,14 @@ includes: [compareArray.js]
 
 const actual = [];
 const expected = [
-  "get timeZone.getPlainDateTimeFor",
-  "call timeZone.getPlainDateTimeFor",
+  "get timeZone.getOffsetNanosecondsFor",
+  "call timeZone.getOffsetNanosecondsFor",
 ];
 
-const dateTime = Temporal.PlainDateTime.from("1963-07-02T12:00:00.987654321");
 const timeZone = new Proxy({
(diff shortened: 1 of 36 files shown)


cjtenny (approved):
Looks good. I'll collect observability & ordering of this into the test262 work items issue as well!

ptomato:
I went to fix the test262 suite which I had forgotten to do on Friday, and realized that I'd overlooked changing GetOffsetStringFor into BuiltinTimeZoneGetOffsetStringFor. @cjtenny would you mind taking another quick look?

polyfill/test/ZonedDateTime/prototype/toPlainDateTime/plain-custom-timezone.js
```diff
@@ -8,15 +8,14 @@ includes: [compareArray.js]
...
 ];
 
-const dateTime = Temporal.PlainDateTime.from("1963-07-02T12:00:00.987654321");
 const timeZone = new Proxy({
-  getPlainDateTimeFor() {
-    actual.push("call timeZone.getPlainDateTimeFor");
-    return dateTime;
+  getOffsetNanosecondsFor() {
+    actual.push("call timeZone.getOffsetNanosecondsFor");
+    return -8735135802468;
```
cjtenny, line 18: are these numbers just different keyboard mashings?

cjtenny (commented):
lgtm!

polyfill/test/ZonedDateTime/prototype/toPlainDateTime/plain-custom-timezone.js
ptomato: I don't know where `160583136123456789n` came from below, but this is the number that yields a PlainDateTime whose time is 12:00:00.987654321.

PR #1402, opened 2021-02-27
Clarify host time zone information updates to refer to time zone database

### LINKED ISSUE #1386: Delegate review: Clarify "updating the time zone information"
From #1383. E.g. in the note at the end of https://tc39.es/proposal-temporal/#sec-temporal-tolocaltime, it's unclear whether a host environment updating "the time zone information" means the current time zone or the time zone database. Clarify that it's the latter.

cc @bmeck

cjtenny:
Fixes #1386.

### DIFF
--- a/spec/intl.html
+++ b/spec/intl.html
@@ -830,9 +830,9 @@
       </emu-note>
       <emu-note>
         <ins>
-          Time zone information is subject to change, and host environments may update their time zone information at any time.
+          Time zone information is subject to change, and host environments may update their time zone database at any time.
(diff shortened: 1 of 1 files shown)


cjtenny:
Aside from this change, "time zone information" is used to consistently refer to the information for a single time zone and "time zone database" is used to refer to the data fro all time zones. Any references to the current time zone use that language ('current time zone'). Let me know if you think explicitly clarifying that _time zone information/database does not refer to the host environment's current timezone_ is worth adding as well.

PR #1405, opened 2021-02-28
polyfill: fix Duration precision bug

cjtenny:
Same bug as #1351 in another place, oops.

Fixes #1404.

### DIFF
--- a/polyfill/test/Duration/prototype/toString/precision.js
+++ b/polyfill/test/Duration/prototype/toString/precision.js
@@ -1,0 +1,13 @@
+// Copyright (C) 2021 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+const { Duration } = Temporal;
+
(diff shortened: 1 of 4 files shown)


cjtenny:
thought to check, just noticed: this bug exists in spec as well. will update, hold the presses

cjtenny:
Spec text is going to have rebase conflicts with #1403 so I'll merge that first.

polyfill/test/Duration/prototype/toString/precision.js
```diff
@@ -0,0 +1,13 @@
+// Copyright (C) 2021 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD licecnse found in the LICENSE file.
```
ptomato, line 2:
```suggestion
// This code is governed by the BSD license found in the LICENSE file.
```

ptomato (approved):

PR #1417, opened 2021-03-05
spec: add missing `_` to fix rendering

ljharb:


ptomato (approved):
Thanks! I'm guessing the CI failure is a hiccup in codecov's service, so I'll just re-run it.

PR #1419, opened 2021-03-05
remove observable calls to "from"

ryzokuken:
Fixes: #1293 

/cc @ptomato @ljharb


spec/calendar.html
```diff
@@ -260,12 +260,14 @@ <h1>GetOptionalTemporalCalendar ( _item_ )</h1>
...
-        1. Let _from_ be ? GetMethod(%Temporal.Calendar%, *"from"*).
-        1. If _from_ is *undefined*, set _from_ to %Temporal.Calendar.from%.
-        1. Let _calendar_ be ? Call(_from_, %Temporal.Calendar%, « _identifier_ »).
-        1. If Type(_calendar_) is not Object, then
-          1. Throw a *TypeError* exception.
-        1. Return _calendar_.
+        1. If Type(_item_) is Object, then
+          1. If ? HasProperty(_item_, *"calendar"*) is *false*, return _item_.
+          1. Set _item_ to ? Get(_item_, *"calendar"*).
+          1. If Type(_item_) is Object and ? HasProperty(_item_, *"calendar"*) is *false*, return _item_.
```
ljharb, line 266: you already know HasProperty is true here, since you returned early 2 lines up

spec/timezone.html
```diff
@@ -578,9 +568,14 @@ <h1>ToTemporalTimeZone ( _temporalTimeZoneLike_ )</h1>
     <emu-clause id="sec-temporal-timezonefrom" aoid="TimeZoneFrom">
       <h1>TimeZoneFrom ( _identifier_ )</h1>
       <emu-alg>
-        1. Let _from_ be ? GetMethod(%Temporal.TimeZone%, *"from"*).
-        1. If _from_ is *undefined*, set _from_ to %Temporal.TimeZone.from%.
-        1. Return ? Call(_from_, %Temporal.TimeZone%, « _identifier_ »).
+        1. If Type(_item_) is Object, then
+          1. If ? HasProperty(_item_, *"timeZone"*) is *false*, return _item_.
+          1. Set _item_ to ? Get(_item_, *"timeZone"*).
+          1. If Type(_item_) is Object and ? HasProperty(_item_, *"timeZone"*) is *false*, return _item_.
```
ljharb, line 574: same here

polyfill/lib/ecmascript.mjs
```diff
@@ -1513,17 +1514,23 @@ export const ES = ObjectAssign({}, ES2020, {
     const TemporalCalendar = GetIntrinsic('%Temporal.Calendar%');
     return new TemporalCalendar('iso8601');
   },
-  CalendarFrom: (calendarLike) => {
-    const TemporalCalendar = GetIntrinsic('%Temporal.Calendar%');
-    let from = TemporalCalendar.from;
-    if (from === undefined) {
-      from = GetIntrinsic('%Temporal.Calendar.from%');
+  CalendarFrom: (item) => {
```
ptomato, line 1517: I think these AOs will need to have a `constructor` parameter passed to them, and construct instances of that class, rather than the intrinsic %Temporal.Calendar% and %Temporal.TimeZone%. This is because we need to construct instances of a subclass if the static method is called on a subclass. (The previous code did this correctly for TimeZone but not for Calendar; my bad.)

polyfill/test/PlainDateTime/constructor/constructor/order-of-operations.js
```diff
@@ -54,17 +52,6 @@ const calendar = {
...
-    actual.push("get Temporal.Calendar.from");
-    return function(argument) {
-      actual.push("call Temporal.Calendar.from");
-      assert.sameValue(argument, "iso8601");
-      return calendar;
-    };
-  },
-});
-
 const dateTime = new Temporal.PlainDateTime(...dateTimeArgs, "iso8601");
```
ptomato, line 55: I guess you'd want to pass in the `calendar` object here instead of `iso8601`?

polyfill/test/regex.mjs
```diff
@@ -588,7 +588,7 @@ describe('fromString regex', () => {
     }
   });
 
-  describe('time zone ID', () => {
+  describe.skip('time zone ID', () => {
```
ptomato, line 591: Maybe worth a comment explaining that we can unskip these tests when we add the resolver parameter to from()?

spec/calendar.html
ptomato: _item_ is a different value from what it was 2 lines earlier, here. (This is for the case of `Temporal.Calendar.from(someTemporalObject)` where `someTemporalObject` has a `.calendar` property.)

ptomato (approved):
Thanks! A few changes still to be made. I think it's fine to un-"draft" this PR by the way.

spec/calendar.html
ljharb: ah, i see. that's confusing; is there a reason we can't create a new spec variable?

ljharb: if it's a temporal object, why would it use `.calendar` instead of an internal slot?

ptomato: That's discussed in #925, to handle the case of a property bag standing in for a Temporal object. I think it'd be reasonable to additionally check if there was a [[Calendar]] internal slot and use that, but I don't feel comfortable reversing a decision that the champions group already had a lengthy discussion on, just as an aside in this PR. So I'd prefer that this PR preserve the existing algorithm. If you feel strongly about it, could you open another issue to discuss it?

ljharb: You're saying that this is to support a *third* type of thing (Temporal objects, strings, and Temporal-like objects)?

Yes, I'll file an issue about that.

ljharb: https://github.com/tc39/proposal-temporal/issues/1428

ptomato: I meant filing an issue about changing this particular site to check [[Calendar]], which is — as I said above — just fine with me. I just didn't want to change the algorithm _in this PR_ where the other champions might not have a chance to comment. But I guess we can handle it as part of #1428. I'll respond there separately.

ptomato (approved):
Thanks, CreateTemporalCalendarFromStatic is the way to go.

Issue #1416, opened 2021-03-05
editor review: missing abstract operations


ljharb:
referenced in https://tc39.es/proposal-temporal/#sec-temporal-builtintimezonegetplaindatetimefor:
 - `GetPartsFromEpoch` has no definition (possibly meant to be https://tc39.es/proposal-temporal/#sec-temporal-getisopartsfromepoch ?)
 - `BalanceDateTime` has no definition (possibly meant to be https://tc39.es/proposal-temporal/#sec-temporal-balanceisodatetime ?)

referenced in https://tc39.es/proposal-temporal/#sec-temporal-builtintimezonegetinstantfor:
 - `GetEpochFromParts` has no definition (possibly meant to be https://tc39.es/proposal-temporal/#sec-temporal-getisopartsfromepoch ?)

ptomato:
Thanks! I think this probably came from a rebase error. I sure wish ecmarkup would catch these things :smile:

PR #1421, opened 2021-03-06
Fix abstract operation names

ptomato:
These operations were renamed to have "ISO" in the name while I was
working on the patch that defined the operations in which they were
called.

Closes: #1416


ljharb (approved):

cjtenny (approved):
:+1:

PR #1430, opened 2021-03-08
Ensure that implementation-defined Intl operations access properties …

### LINKED ISSUE #1388: Delegate review: Specify in which order properties are accessed when passing object to implementation-defined operation
From #1383. In implementation-defined operations such as non-ISO calendars (e.g. https://tc39.es/proposal-temporal/#sup-temporal.calendar.prototype.datefromfields) it must be specified in which order properties from these objects are accessed. The objects should be copied and objects containing data properties should be passed to the operations, so that getters are not called more than once.

cc @bmeck

cjtenny:
…in consistent order and only call getters once.

Fixes #1388.

### DIFF
--- a/spec/intl.html
+++ b/spec/intl.html
@@ -1144,7 +1144,9 @@
             1. If _calendar_.[[Identifier]] is *"iso8601"*, then
               1. Let _result_ be ? ISODateFromFields(_fields_, _options_).
             1. Else,
-              1. Let _result_ be a Record with [[Year]], [[Month]], and [[Day]] fields, that is the result of implementation-defined processing of _fields_, _options_, and the value of _calendar_.[[Identifier]].
+              1. Let _overflow_ be ? ToTemporalOverflow(_options_).
+              1. Set _fields_ to ? PrepareTemporalFields(_fields_, « *"day"*, *"era"*, *"eraYear"*, *"month"*, *"monthCode"*, *"year"* », « *"day"* »).
+              1. Let _result_ be a Record with [[Year]], [[Month]], and [[Day]] fields, that is the result of implementation-defined processing of _fields_, _overflow_, and the value of _calendar_.[[Identifier]].
             1. Return ? CreateTemporalDateFromStatic(_constructor_, _result_.[[Year]], _result_.[[Month]], _result_.[[Day]], _calendar_).
           </emu-alg>
         </emu-clause>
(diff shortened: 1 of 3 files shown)


spec/intl.html
```diff
@@ -1258,6 +1266,8 @@ <h1>Temporal.Calendar.prototype.eraYear ( _dateOrDateTime_ )</h1>
           <emu-alg>
             1. Let _calendar_ be the *this* value.
             1. Perform ? RequireInternalSlot(_calendar_, [[InitializedTemporalCalendar]]).
+            1. If _dateOrDateTime_ does not have an [[ISOYear]] internal slot, then
```
ljharb, line 1269: I’m not sure the spec asks this about something it doesn’t already know is an Object. It seems intuitive that a primitive would neatly fail this test, and i don’t have a specific change in mind, but i wanted to call it out.

cjtenny: That's a good point. Do you think it'd be better as "`If _dateOrDateTime_ is not an Object or does not have an [[ISOYear]] internal slot, then`"? Or would you prefer the more concise version here?

ljharb: I think I'll defer to @bakkot, @syg, and @michaelficarra on this one; there's a few other cases in this PR (including "if it has the slot") where it'd apply as well.

bakkot: I'd probably say `If Type(_dateOrDateTime_) is not Object or _dateOrDateTime_ does not have an [[ISOYear]] internal slot, then`, which is close to the phrasing in [`RequireInternalSlot`](https://tc39.es/ecma262/#sec-requireinternalslot).

cjtenny: Wilco, thanks

polyfill/lib/ecmascript.mjs
```diff
@@ -1010,7 +1010,7 @@ export const ES = ObjectAssign({}, ES2020, {
     }
     return any ? any : false;
   },
-  PrepareTemporalFields: (bag, fields) => {
+  PrepareTemporalFields: (bag, fields, convert = true) => {
```
ptomato, line 1013: This parameter seems to be no longer used?

ptomato (approved):
Thanks, all looks correct.

polyfill/lib/ecmascript.mjs
```diff
@@ -1024,7 +1024,7 @@ export const ES = ObjectAssign({}, ES2020, {
         value = defaultValue;
       } else {
         any = true;
-        if (BUILTIN_CASTS.has(property)) {
+        if (convert && BUILTIN_CASTS.has(property)) {
```
ptomato, line 1027: This one has to go too, I'll fix it up

cjtenny: oof, where is my mind. thank you.

PR #1438, opened 2021-03-10
Create calendar-review.md

sffc:
I imported the Google Doc that @justingrant and I have collaborated on to explain the design decisions behind calendars in Temporal.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ For a detailed breakdown of motivations, see:
 ### Principles:
 
 - All Temporal objects are immutable.
-- Date values can be represented in local calendar systems, but they should be convertable to and from the [Proleptic Gregorian Calendar](https://en.wikipedia.org/wiki/Proleptic_Gregorian_calendar).
+- Date values can be represented in local calendar systems, but they should be convertable to and from the [Proleptic Gregorian Calendar](https://en.wikipedia.org/wiki/Proleptic_Gregorian_calendar). Read more on [non-Gregorian calendar motivation](./docs/calendar-review.md).
 - All time-of-day values are based on a standard 24-hour clock.
 - [Leap seconds](https://en.wikipedia.org/wiki/Leap_second) are not represented.
 
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ For a detailed breakdown of motivations, see:
 ### Principles:
 
 - All Temporal objects are immutable.
-- Date values can be represented in local calendar systems, but they should be convertable to and from the [Proleptic Gregorian Calendar](https://en.wikipedia.org/wiki/Proleptic_Gregorian_calendar). Read more on [non-Gregorian calendar motivation](./docs/calendar-review.md).
+- Date values can be represented in local calendar systems ([why?](./docs/calendar-review.md)), but they should be convertable to and from the [Proleptic Gregorian Calendar](https://en.wikipedia.org/wiki/Proleptic_Gregorian_calendar).
 - All time-of-day values are based on a standard 24-hour clock.
 - [Leap seconds](https://en.wikipedia.org/wiki/Leap_second) are not represented.
 
--- a/docs/calendar-review.md
+++ b/docs/calendar-review.md
@@ -80,9 +80,9 @@ date.month; // => 6
 date.day; // => 2
 date.inLeapYear; // => true
 date.calendar.id; // => hebrew
-inTwoMonths = date.add({ months: 4 }); // => 23 Sivan 5779
-inTwoMonths.withCalendar('iso8601'); // => 2019-06-26
-date.until(inTwoMonths, { largestUnit: 'months' }); // => P4M
+inFourMonths = date.add({ months: 4 }); // => 23 Sivan 5779
+inFourMonths.withCalendar('iso8601'); // => 2019-06-26
+date.until(inFourMonths, { largestUnit: 'months' }); // => P4M
 ```
 
 By definition, creating a bag of fields with year, month, and day requires choosing a calendar system. We elide the ISO calendar if no other calendar system is specified in this situation, and it is the only place Temporal performs calendar elision:
@@ -170,7 +170,7 @@ An alternative approach to the current Temporal design would have been to store 
 *   Chaining multiple operations (e.g. `.add({months: 2}).with({day: 1})`) gets less ergonomic and more bug-prone because the same calendar value must be used in all chained calls. 
 *   The `Temporal.PlainMonthDay` type would become challenging to reason about without a calendar attached because month/day values like birthdays and holidays are inherently calendar-specific and have no ISO counterpart. This is especially true in lunisolar calendars like Hebrew and Chinese where ordinal month numbers vary from year to year for the same month. (Current Temporal uses a string <code>[monthCode](https://tc39.es/proposal-temporal/docs/plaindate.html#monthCode)</code> field to describe year-independent months.) For example, a birthday on 12 Adar I in the Hebrew calendar is currently modelled as \
 <code>  Temporal.PlainMonthDay.from( \
-    { monthCode: '05L', day: 12, calendar: 'hebrew' } \
+    { monthCode: 'M05L', day: 12, calendar: 'hebrew' } \
   ); \
 </code>
 
@@ -191,7 +191,7 @@ It would be possible to retain the calendar in Temporal objects' data model with
 *   `Temporal.Duration` would gain a `calendar` field.  
     *   Instances using the ISO calendar would only support ISO fields: `days`, `weeks`, `months`, and `years`. 
     *   Instances using a non-ISO calendar would only support calendar fields: `calendarDays`, `calendarWeeks`, `calendarMonths`, and `calendarYears`. 
-    *   String parsing and serialization would use calendar annotations like other Temporal types, e.g. `'P2D[u-ca-chinese]'`. Strings without an annotation would assume the ISO calendar.
+    *   String parsing and serialization would use calendar annotations like other Temporal types, e.g. `'P2D[u-ca=chinese]'`. Strings without an annotation would assume the ISO calendar.
     *   Question: should `date.add({calendarMonths: 1})` use `date.calendar`? Or should `date.add({calendarMonths: 1, calendar: 'chinese'})` be required even if `date.calendar==='chinese'`?
     *   `until` and `since` are now ambiguous, so they’d need a way to indicate whether the output should be an ISO duration or a non-ISO one. Options include:
         *   Split each method into "iso" and "calendar" variants, e.g. `isoUntil` vs. `calendarUntil` or `untilISO` vs. `untilCalendar`
--- a/docs/calendar-review.md
+++ b/docs/calendar-review.md
@@ -1,0 +1,212 @@
+# Supporting Calendar Systems in Temporal
+
+Authors: Temporal champions [justingrant](https://twitter.com/justingrantjg) and [sffc](https://twitter.com/_sffc)
+
+[ECMAScript Temporal](https://github.com/tc39/proposal-temporal/) is the new date and time API for ECMAScript.  Because ECMAScript is designed for global audiences, high-quality internationalization is a design goal.
+
+This document discusses why and how calendar systems are included in the Temporal API, explains problems we expect developers to encounter with calendars, and outlines how to avoid those problems. A final section explains alternative designs for using calendar systems in Temporal, along with an explanation of why the current approach was chosen instead.
+
+## Background and Motivation
+
+Calendar systems provide a way to assign human-readable numbers, such as year, month, and day, to a particular period of time.
+
+Much of the world uses the Gregorian calendar system, which uses solar years, months, and days: over time, it is aligned with the position of the Earth relative to the sun.  However, there are other calendar systems used throughout much of the world.  Some popular calendar systems include Hebrew, Islamic, Buddhist, Hindu, Chinese, Japanese, and Ethiopic.
+
+Today's date is March 4, 2021 in the Gregorian calendar.  Here are several other representations of that date:
+
+*   20 Adar 5781 (Hebrew)
+*   Rajab 20, 1442 AH (Islamic, Umm al-Qura)
+*   March 4, 2564 BE (Thai Buddhist)
+*   令和3年3月4日 == March 4, 3 Reiwa (Japanese)
+
+Non-Gregorian calendar systems are used in day-to-day life in several countries, particularly those with weaker cultural and economic ties to Gregorian-dominant Western countries, as well as in religious, cultural, and academic applications.  For example, several countries including Japan, China, Iran, Afghanistan, Saudi Arabia, and Taiwan use non-Gregorian calendars for at least some official government purposes. Beyond those official uses, a majority of the world's population lives in countries that use non-Gregorian calendars to determine the dates of religious or cultural holidays. 
+
+### Calendars in Business Logic (Not Only String Localization)
+
+Calendar systems play an important role in localization (l10n).  They enable a human-friendly display representation of timestamps, dates, and times.
+
+However, calendar systems also play an important role in the business logic of an application.  Applications involving dates intended for human consumption must consider the calendar system when performing arithmetic or any other business-logic operation that uses months and years.  Examples of calendar-sensitive operations include:
+
+*   What is the current day plus one month?
+*   What is the first day of the month?
+*   On what month and day do I celebrate my birthday or anniversary?
+*   On what day do I celebrate a religious or cultural holiday, like Ramadan, Passover, Easter, or the Chinese Lunar New Year?
+
+In other words, units of time greater than a solar day are inherently dependent on the calendar system.  The only universal solution to date arithmetic would be to omit the concept of months and years, but a date API without months and years would be insufficient for many use cases.
+
+### Intl-First Design
+
+ECMAScript supports internationalization (i18n) as a primary feature. The i18n subcommittee, TC39-TG2, evangelizes "Intl-first design." Two principles of Intl-first design are (1) language- and region-dependent operations should be data-driven, and (2) user preferences should be an explicit input to APIs. These principles help ensure that APIs work well for developers targeting a wide range of end-users around the world.
+
+Temporal applies these design principles by abstracting calendar-specific logic and by making the calendar an explicit choice when constructing objects. The result is an API that supports use cases like the ones listed above for regions that use only Gregorian and for those that have other calendar systems in common use.
+
+### Using Calendars in Temporal Code
+
+The following section shows how calendar systems are exposed in Temporal code. For a more thorough discussion, see the [documentation](https://tc39.es/proposal-temporal/docs/calendar.html). 
+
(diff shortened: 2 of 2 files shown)


sffc:
Ping @ryzokuken @justingrant 

Also, @jasonwilliams, once this is checked in, we should send it to the email thread.

docs/calendar-review.md
```diff
@@ -0,0 +1,212 @@
...
+isLastDayOfYear(hebrewNewYearsEve); // => expected: true, actual: false
+```
+
+### Mitigating the Unexpected Calendar Problem
+
+There are two ways to avoid the Unexpected Calendar Problem: the preferred solution of writing calendar-safe code, or a fallback option to validate or coerce inputs to a known calendar.
+
+#### Writing Calendar-Safe Code
+
+The best way to mitigate the Unexpected Calendar Problem is to write “calendar-safe” code that works for all built-in calendars. For example, the function above can easily be written to be calendar-safe by using the `monthsInYear` and `daysInMonth` properties instead of hard-coding ISO constants. Developers who follow documented [best practices for writing calendar-safe Temporal code](https://tc39.es/proposal-temporal/docs/calendar.html#writing-cross-calendar-code) will be able to use the same code for all built-in calendars.
```
justingrant, line 128:
```suggestion
The best way to mitigate the Unexpected Calendar Problem is to write "calendar-safe" code that works for all built-in calendars. For example, the function above can easily be written to be calendar-safe by using the `monthsInYear` and `daysInMonth` properties instead of hard-coding ISO constants. Developers who follow documented [best practices for writing calendar-safe Temporal code](https://tc39.es/proposal-temporal/docs/calendar.html#writing-cross-calendar-code) will be able to use the same code for all built-in calendars.
```

docs/calendar-review.md
```diff
@@ -0,0 +1,212 @@
...
+
+Beyond documentation, the Temporal API itself has been designed to make it easier to write calendar-safe code. For example:
+
+*   All built-in calendars (including ISO) use the same set of date fields: `year`, `month`, `day`, and `monthCode`. These fields act the same way in all built-in calendars.
+*   The `month` property is a continuous (no gaps), 1-based index of the month in the current year. This aligns with ISO behavior and avoids unexpected behavior from lunisolar calendars where each year may have a different set of months.  A separate `monthCode` string field is used for a year-independent month identifier.
+*   The `year` property is a signed value that’s relative to a specific calendar era. This matches ISO calendar behavior and prevents bugs from, for example, eras like B.C. where years count backwards.  Separate `eraYear` and `era` fields are used for measuring years within the context of a specific calendar era.
```
justingrant, line 134: Avoiding the repetition of "specific era" may make this section clearer.

```suggestion
*   The `year` property is a signed value that's relative to a specific "default era" for each calendar. This matches ISO calendar behavior and prevents bugs from, for example, eras like B.C. where years count backwards.  Separate `eraYear` and `era` fields are used for measuring years within the context of other eras.
```

docs/calendar-review.md
```diff
@@ -0,0 +1,212 @@
...
+*   Differences between calendars are exposed via convenience properties available in every calendar, e.g. `monthsInYear`, `daysInMonth`, `inLeapYear`, etc. Userland code can completely avoid the use of calendar-specific constants (e.g. 12 months per year) by using these properties. 
+*   There is never an implicit default calendar system. Developers must explicitly decide whether they’re using the ISO calendar or another calendar. For example, shortcut methods on the `Temporal.now` object are named with an ISO prefix to avoid ambiguity, such as `Temporal.now.zonedDateTimeISO()`.
```
justingrant, line 136:
```suggestion
*   There is never an implicit default calendar system. Developers must explicitly decide whether they're using the ISO calendar or another calendar. For example, shortcut methods on the `Temporal.now` object are named with an ISO prefix to avoid ambiguity, such as `Temporal.now.zonedDateTimeISO()`.
```

docs/calendar-review.md
```diff
@@ -0,0 +1,212 @@
...
+*   Temporal never infers a calendar from the user’s environment.  Introducing a calendar system into a program requires a conscious, opt-in decision on the part of the developer of the app, library, or input source.
```
justingrant, line 137:
```suggestion
*   Temporal never infers a calendar from the user's environment.  Introducing a calendar system into a program requires a conscious, opt-in decision on the part of the developer of the app, library, or input source.
```

docs/calendar-review.md
```diff
@@ -0,0 +1,212 @@
...
+
+#### Validating or Coercing Inputs
+
+Even though Temporal makes it straightforward to write calendar-safe code, doing so still requires work from developers. Some developers won’t do this work. Sometimes this will be because they don’t know about or don’t understand non-ISO calendars. In other cases, developers won’t bother to follow best practices because they expect to only be dealing with ISO-calendar data. Even if a developer conscientiously follows best practices in their own code, most ECMAScript apps have many library dependencies that may not do so. Or they may have to interoperate with existing code that cannot be changed due to time constraints.
```
justingrant, line 141:
```suggestion
Even though Temporal makes it straightforward to write calendar-safe code, doing so still requires work from developers. Some developers won't do this work. Sometimes this will be because they don't know about or don't understand non-ISO calendars. In other cases, developers won't bother to follow best practices because they expect to only be dealing with ISO-calendar data. Even if a developer conscientiously follows best practices in their own code, most ECMAScript apps have many library dependencies that may not do so. Or they may have to interoperate with existing code that cannot be changed due to time constraints.
```

docs/calendar-review.md
```diff
@@ -0,0 +1,212 @@
...
+`date.add({months: 2}, {calendar: 'chinese'})`
+*   Chaining multiple operations (e.g. `.add({months: 2}).with({day: 1})`) gets less ergonomic and more bug-prone because the same calendar value must be used in all chained calls. 
+*   The `Temporal.PlainMonthDay` type would become challenging to reason about without a calendar attached because month/day values like birthdays and holidays are inherently calendar-specific and have no ISO counterpart. This is especially true in lunisolar calendars like Hebrew and Chinese where ordinal month numbers vary from year to year for the same month. (Current Temporal uses a string <code>[monthCode](https://tc39.es/proposal-temporal/docs/plaindate.html#monthCode)</code> field to describe year-independent months.) For example, a birthday on 12 Adar I in the Hebrew calendar is currently modelled as \
+<code>  Temporal.PlainMonthDay.from( \
+    { monthCode: '05L', day: 12, calendar: 'hebrew' } \
+  ); \
+</code>It’s not clear how we’d model holidays if the calendar isn’t stored in the instance.
```
justingrant, line 175:
```suggestion
</code>

It's not clear how we'd model holidays if the calendar isn't stored in the instance.
```

docs/calendar-review.md
```diff
@@ -0,0 +1,212 @@
...
+*   Code that wants to return a date along with its calendar can no longer return a single Temporal instance. Instead, it must return a compound object or serialized string. For example, a date picker that returned a date in the user’s preferred calendar could no longer return a <code>Temporal.PlainDate</code>. Instead, it’d have to return a tuple or compound object like <code>{date, calendar}</code>.  Passing around compound objects (or extra calendar parameters) would complicate all calendar-using Temporal code.
```
justingrant, line 176:
```suggestion
*   Code that wants to return a date along with its calendar can no longer return a single Temporal instance. Instead, it must return a compound object or serialized string. For example, a date picker that returned a date in the user's preferred calendar could no longer return a <code>Temporal.PlainDate</code>. Instead, it'd have to return a tuple or compound object like <code>{date, calendar}</code>.  Passing around compound objects (or extra calendar parameters) would complicate all calendar-using Temporal code.
```

docs/calendar-review.md
```diff
@@ -0,0 +1,212 @@
...
+
+The Temporal champions believe that the ergonomic challenges of this alternative are worse than the work required to mitigate the Unexpected Calendar Problem. 
+
+### Alternative 2: Separate ISO and Calendar Properties
+
+It would be possible to retain the calendar in Temporal objects’ data model without being vulnerable to the Unexpected Calendar Problem, but doing so would involve a significant increase in Temporal surface area. Below is one way it could be done.
```
justingrant, line 182:
```suggestion
It would be possible to retain the calendar in Temporal objects' data model without being vulnerable to the Unexpected Calendar Problem, but doing so would involve a significant increase in Temporal surface area. Below is one way it could be done. 
```

docs/calendar-review.md
```diff
@@ -0,0 +1,212 @@
...
+    *   Question: should `date.add({calendarMonths: 1})` use `date.calendar`? Or should `date.add({calendarMonths: 1, calendar: 'chinese'})` be required even if `date.calendar==='chinese'`?
+    *   `until` and `since` are now ambiguous, so they’d need a way to indicate whether the output should be an ISO duration or a non-ISO one. Options include:
+        *   Split each method into "iso" and "calendar" variants, e.g. `isoUntil` vs. `calendarUntil` or `untilISO` vs. `untilCalendar`
+        *   Always return both flavors, e.g. `foo.until(bar).isoDuration` and `foo.until(bar).calendarDuration`.
+        *   Use an option, e.g. `{ resultFields: 'iso' | 'calendar' }`. FWIW, I'm not sure that this option would work well with TS in cases where options are not specified as a literal in the API call.
+    *   There would NOT be a `withCalendar` method on durations, because durations don’t have the ability to convert between calendar and ISO data. Instead, they would either have ISO fields or calendar fields, but not both.
```
justingrant, line 198:
```suggestion
    *   There would NOT be a `withCalendar` method on durations, because durations don't have the ability to convert between calendar and ISO data. Instead, they would either have ISO fields or calendar fields, but not both.
```

docs/calendar-review.md
```diff
@@ -0,0 +1,212 @@
...
+
+After considering this and similar options, the champions decided that the additional complexity wasn’t worth it. In particular, we were concerned that the dual-property approach would be more confusing and would result in more bugs than teaching users to validate or coerce inputs.
```
justingrant, line 200:
```suggestion
After considering this and similar options, the champions decided that the additional complexity wasn't worth it. In particular, we were concerned that the dual-property approach would be more confusing and would result in more bugs than teaching users to validate or coerce inputs. 
```

docs/calendar-review.md
```diff
@@ -0,0 +1,212 @@
...
+date = date.withCalendar('iso8601');
+```
+
+To validate:
+
+```javascript
+if (date.calendar.id !== 'iso8601') throw new Error ('invalid calendar');
+```
+
+Note that “inputs” are not only external data. Inputs could include values returned from library dependencies, e.g. date pickers. Unless the library or input source guarantees a certain calendar system, defensive code should assume Temporal objects (or strings or objects that will be converted into a Temporal object) to be “external” and subject to validation or coercion to the ISO calendar if the code that will use that data is not calendar-safe.
```
justingrant, line 157:
```suggestion
Note that "inputs" are not only external data. Inputs could include values returned from library dependencies, e.g. date pickers. Unless the library or input source guarantees a certain calendar system, defensive code should assume Temporal objects (or strings or objects that will be converted into a Temporal object) to be "external" and subject to validation or coercion to the ISO calendar if the code that will use that data is not calendar-safe.
```

docs/calendar-review.md
```diff
@@ -0,0 +1,212 @@
...
+Temporal.PlainDate.from({ year: 2019, month: 2, day: 28 });
+```
+
+The following Temporal types all carry a calendar: `Temporal.ZonedDateTime`, `Temporal.PlainDateTime`, `Temporal.PlainDate`, `Temporal.PlainYearMonth`, `Temporal.PlainMonthDay`
+
+## The Unexpected Calendar Problem
+
+Integrating calendar awareness into many Temporal operations is essential for a global audience, but it introduces a new problem: if code is written to assume one calendar system, bugs may result if a different calendar system is introduced instead. Introducing unexpected calendars could also be a vector for malicious activity.
+
+This section describes this problem, which we call the “Unexpected Calendar Problem,” and how to mitigate it in userland Temporal code.
```
justingrant, line 100:
```suggestion
This section describes this problem, which we call the "Unexpected Calendar Problem," and how to mitigate it in userland Temporal code.
```

justingrant (approved):
Looks good. I made one minor wording suggestion and a global suggestion to change fancy smart quotes to ASCII quotes.

sffc:
Thanks!

docs/calendar-review.md
```diff
@@ -0,0 +1,214 @@
...
+Non-Gregorian calendar systems are used in day-to-day life in several countries, particularly those with weaker cultural and economic ties to Gregorian-dominant Western countries, as well as in religious, cultural, and academic applications.  For example, several countries including Japan, China, Iran, Afghanistan, Saudi Arabia, and Taiwan use non-Gregorian calendars for at least some official government purposes. Beyond those official uses, a majority of the world's population lives in countries that use non-Gregorian calendars to determine the dates of religious or cultural holidays. 
+
+### Calendars in Business Logic (Not Only String Localization)
+
+Calendar systems play an important role in localization (l10n).  They enable a human-friendly display representation of timestamps, dates, and times.
+
+However, calendar systems also play an important role in the business logic of an application.  Applications involving dates intended for human consumption must consider the calendar system when performing arithmetic or any other business-logic operation that uses months and years.  Examples of calendar-sensitive operations include:
+
+*   What is the current day plus one month?
```
ptomato, line 30: Maybe to sound slightly more natural, "What is the date one month from today?"

docs/calendar-review.md
```diff
@@ -0,0 +1,214 @@
...
+date.with({ day: 1 }); // => 1 Adar I 5779
+date.year; // => 5779
+date.monthCode; // => M05L
+date.month; // => 6
+date.day; // => 2
+date.inLeapYear; // => true
+date.calendar.id; // => hebrew
+inTwoMonths = date.add({ months: 4 }); // => 23 Sivan 5779
+inTwoMonths.withCalendar('iso8601'); // => 2019-06-26
+date.until(inTwoMonths, { largestUnit: 'months' }); // => P4M
```
ptomato, line 85:
```suggestion
inFourMonths = date.add({ months: 4 }); // => 23 Sivan 5779
inFourMonths.withCalendar('iso8601'); // => 2019-06-26
date.until(inFourMonths, { largestUnit: 'months' }); // => P4M
```

docs/calendar-review.md
```diff
@@ -0,0 +1,214 @@
...
+*   Many methods (`add`, `subtract`, `round`, `until`, `since`, `with`, maybe `toPlainMonthDay`) would require a calendar option to be able to use calendar-specific months, days, or years. Example: \
+`date.add({months: 2}, {calendar: 'chinese'})`
+*   Chaining multiple operations (e.g. `.add({months: 2}).with({day: 1})`) gets less ergonomic and more bug-prone because the same calendar value must be used in all chained calls. 
+*   The `Temporal.PlainMonthDay` type would become challenging to reason about without a calendar attached because month/day values like birthdays and holidays are inherently calendar-specific and have no ISO counterpart. This is especially true in lunisolar calendars like Hebrew and Chinese where ordinal month numbers vary from year to year for the same month. (Current Temporal uses a string <code>[monthCode](https://tc39.es/proposal-temporal/docs/plaindate.html#monthCode)</code> field to describe year-independent months.) For example, a birthday on 12 Adar I in the Hebrew calendar is currently modelled as \
+<code>  Temporal.PlainMonthDay.from( \
+    { monthCode: '05L', day: 12, calendar: 'hebrew' } \
```
ptomato, line 173:
```suggestion
    { monthCode: 'M05L', day: 12, calendar: 'hebrew' } \
```

docs/calendar-review.md
```diff
@@ -0,0 +1,214 @@
...
+    *   Temporal APIs that accept property bag parameters will accept either calendar fields or ISO fields, BUT NOT BOTH. `date.with({calendarYear: 5780, isoMonth: 12}) `should throw. BTW, this may be enforceable via TS and ESLint for a nice design-time DX.
+    *   Trying to read `calendar*` fields would throw if the calendar is ISO because there is no human calendar defined.
+    *   Non-field convenience properties (e.g. `dayOfYear`, `inLeapYear`, etc.) would also be duplicated. The non-`calendar` variants of these fields would report the ISO values.
+*   `Temporal.Duration` would gain a `calendar` field.  
+    *   Instances using the ISO calendar would only support ISO fields: `days`, `weeks`, `months`, and `years`. 
+    *   Instances using a non-ISO calendar would only support calendar fields: `calendarDays`, `calendarWeeks`, `calendarMonths`, and `calendarYears`. 
+    *   String parsing and serialization would use calendar annotations like other Temporal types, e.g. `'P2D[u-ca-chinese]'`. Strings without an annotation would assume the ISO calendar.
```
ptomato, line 194:
```suggestion
    *   String parsing and serialization would use calendar annotations like other Temporal types, e.g. `'P2D[u-ca=chinese]'`. Strings without an annotation would assume the ISO calendar.
```

docs/calendar-review.md
```diff
@@ -0,0 +1,214 @@
...
+    *   Question: should `date.add({calendarMonths: 1})` use `date.calendar`? Or should `date.add({calendarMonths: 1, calendar: 'chinese'})` be required even if `date.calendar==='chinese'`?
```
ptomato, line 195: This section kind of mixes up the modifications that would need to be made in order to take this approach, with the consequences of making those modifications. FWIW, I think the document is fine this way, so no need to block merging on this, but it does read slightly confusingly.

ptomato (approved):
Nice job, this document is really good.

docs/calendar-review.md
```diff
@@ -0,0 +1,214 @@
+# Supporting Calendar Systems in Temporal
+
+Authors: Temporal champions [justingrant](https://twitter.com/justingrantjg) and [sffc](https://twitter.com/_sffc)
```
justingrant, line 3: One more note: do we need to call out authors here?  We don't do that anywhere else in Temporal docs, so not sure there's a need for it here.

README.md
```diff
@@ -40,7 +40,7 @@ For a detailed breakdown of motivations, see:
 ### Principles:
 
 - All Temporal objects are immutable.
-- Date values can be represented in local calendar systems, but they should be convertable to and from the [Proleptic Gregorian Calendar](https://en.wikipedia.org/wiki/Proleptic_Gregorian_calendar).
+- Date values can be represented in local calendar systems, but they should be convertable to and from the [Proleptic Gregorian Calendar](https://en.wikipedia.org/wiki/Proleptic_Gregorian_calendar). Read more on [non-Gregorian calendar motivation](./docs/calendar-review.md).
```
justingrant, line 43: This is the TL;DR of the entire proposal, so we should keep words to a minimum here.

```suggestion
- Date values can be represented in local calendar systems ([why?](./docs/calendar-review.md)), but they should be convertable to and from the [Proleptic Gregorian Calendar](https://en.wikipedia.org/wiki/Proleptic_Gregorian_calendar).
```

docs/calendar-review.md
sffc: Removed

docs/calendar-review.md
sffc: I removed it.  I think there are enough other bullets in this section.  (OK @justingrant?)

sffc:
Updated with the feedback from @ptomato, @justingrant, and @Louis-Aime.

PR #1445, opened 2021-03-11
polyfill: fix the Instant methods types

artalar:
Based on https://github.com/tc39/proposal-temporal/blob/732751277ed13c65f82fb55496354c428ebc58b3/polyfill/lib/instant.mjs#L63

### DIFF
--- a/polyfill/index.d.ts
+++ b/polyfill/index.d.ts
@@ -547,8 +547,12 @@ export namespace Temporal {
     readonly epochMicroseconds: bigint;
     readonly epochNanoseconds: bigint;
     equals(other: Temporal.Instant | string): boolean;
-    add(durationLike: Temporal.Duration | DurationLike | string): Temporal.Instant;
-    subtract(durationLike: Temporal.Duration | DurationLike | string): Temporal.Instant;
+    add(
+      durationLike: Omit<Temporal.Duration | DurationLike, 'years' | 'months' | 'weeks' | 'days'> | string
+    ): Temporal.Instant;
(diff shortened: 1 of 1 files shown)


artalar:
Sorry, I need to test it first

artalar:
I rechecked a few cases, looks ok.

ryzokuken (approved):
This is great stuff, @artalar! Thanks a ton!

For TS noobs like me, https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys should help,

justingrant:
Sorry, I'm very late in reviewing this! @artalar, I have a few questions: 

1) Does this PR cause new compiler errors compared to the previous .d.ts content?  For example, will either of the lines below cause a TS error with this PR?  (They'll both cause runtime errors.)
```ts
Temporal.now.instant().add(Temporal.Duration.from('P10DT2H'));
Temporal.now.instant().add({ days: 10, hours: 2 });
```
2) If the answer to (1) is "no" then is the main benefit of this PR that it will limit autocomplete to valid properties?  That's a good justification for this PR, but I'm just wondering if there are other benefits besides that one. 
3) What's the benefit of wrapping the `Omit` around the Duration type (the class type), as opposed to only wrapping `DurationLike` (the object literal type)?  I don't think `Omit`-ing the class type hurts anything, but I was wondering if it helps anything.

ryzokuken:
I think it will infact throw a compile error in the second example, not sure about the first but I guess @artalar could confirm.

PR #1470, opened 2021-04-06
Copy objects before passing them to user code twice

ptomato:
Closes: #1426

### DIFF
--- a/polyfill/test/Duration/constructor/compare/read-time-fields-before-datefromfields.js
+++ b/polyfill/test/Duration/constructor/compare/read-time-fields-before-datefromfields.js
@@ -1,0 +1,20 @@
+// Copyright (C) 2021 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.duration.compare
+info: |
+    sec-temporal.duration.compare step 4:
+      4. Let _relativeTo_ be ? ToRelativeTemporalObject(_options_).
+    sec-temporal-torelativetemporalobject step 4.g:
+      g. Let _result_ be ? InterpretTemporalDateTimeFields(_calendar_, _fields_, _options_).
+    sec-temporal-interprettemporaldatetimefields steps 1–2:
+      1. Let _timeResult_ be ? ToTemporalTimeRecord(_fields_).
+      2. Let _temporalDate_ be ? DateFromFields(_calendar_, _fields_, _options_).
+includes: [temporalHelpers.js]
+---*/
+
+const calendar = TemporalHelpers.calendarMakeInvalidGettersTime();
+const duration1 = new Temporal.Duration(1);
+const duration2 = new Temporal.Duration(0, 12);
+Temporal.Duration.compare(duration1, duration2, { relativeTo: { year: 2000, month: 1, day: 1, calendar } });
--- a/polyfill/test/helpers/temporalHelpers.js
+++ b/polyfill/test/helpers/temporalHelpers.js
@@ -326,6 +326,61 @@ var TemporalHelpers = {
   },
 
   /*
+   * A custom calendar that modifies the fields object passed in to
+   * dateFromFields, sabotaging its time properties.
+   */
+  calendarMakeInfinityTime() {
+    class CalendarMakeInfinityTime extends Temporal.Calendar {
+      constructor() {
+        super("iso8601");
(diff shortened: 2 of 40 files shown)


spec/abstractops.html
```diff
@@ -1510,7 +1510,9 @@ <h1>ToPositiveIntegerOrInfinity ( _argument_ )</h1>
   <emu-clause id="sec-temporal-preparetemporalfields" aoid="PrepareTemporalFields">
     <h1>PrepareTemporalFields ( _fields_, _fieldNames_, _requiredFields_ )</h1>
     <p>
-      The abstract operation PrepareTemporalFields reads the relevant properties of an Object _fields_; ensures all the required properties are present; and creates a new Object with all the properties converted to the correct type. It performs the following steps:
+      The abstract operation PrepareTemporalFields reads the relevant properties of an Object _fields_; ensures all the required properties are present; and creates a new Object with all the properties converted to the correct type.
+      The relevant property names are given by the List of Strings _fieldNames_, and the required property names are given by the List of Strings _requiredFields_.
+      It performs the following steps:
```
ljharb, line 1515: so for a required field, "present and undefined" and "absent" will throw, but "present and null" is property-dependent?

ptomato: That seems in line with other places that use ToIntegerOrInfinity or ToString in JS — `era: null` is coerced to `era: "null"`, and `year: null` is coerced to `year: 0`

ljharb: `undefined` in those cases is coerced to `"undefined"` or `NaN` (probably `0` for ToIntegerOrInfinity) - i'm asking about the throw for undefined, not the null :-)

ptomato: I see, I thought you were confirming the throw for undefined and complaining about the lack of throwing for null. Yes, present but undefined does throw - steps 3a-b of PrepareTemporalFields.

polyfill/test/helpers/temporalHelpers.js
```diff
@@ -324,4 +324,122 @@ var TemporalHelpers = {
...
+
+      monthDayFromFields(fields, options) {
+        assert.sameValue(fields.shouldNotBeCopied, undefined, "extra fields should not be copied");
+        return super.monthDayFromFields(fields, options);
+      }
+
+      mergeFields(fields, additionalFields) {
+        const retval = super.mergeFields(fields, additionalFields);
+        retval._calendar = this;
+        retval.shouldNotBeCopied = true;
```
cjtenny, line 417: Though never directly referenced, the assertions around order of operations would reflect if this field were accessed, correct?

cjtenny (approved):
Looks great. Thanks for deciding we can ditch & then cleaning up the ToTemporal<>Fields operations in the spec, I like it! Also, very clever usage of infinity calendar / invalid values. Two clarifying questions for my understanding, no changes needed, lgtm.

polyfill/test/Duration/constructor/compare/read-time-fields-before-datefromfields.js
```diff
@@ -0,0 +1,20 @@
...
+    sec-temporal-interprettemporaldatetimefields steps 1–2:
+      1. Let _timeResult_ be ? ToTemporalTimeRecord(_fields_).
+      2. Let _temporalDate_ be ? DateFromFields(_calendar_, _fields_, _options_).
+includes: [temporalHelpers.js]
+---*/
+
+const calendar = TemporalHelpers.calendarMakeInvalidGettersTime();
+const duration1 = new Temporal.Duration(1);
+const duration2 = new Temporal.Duration(0, 12);
+Temporal.Duration.compare(duration1, duration2, { relativeTo: { year: 2000, month: 1, day: 1, calendar } });
```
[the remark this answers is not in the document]
ptomato: Yes. (It may be that there is a more intentional way to express this in test262 but I haven't found it)

polyfill/test/helpers/temporalHelpers.js
ptomato: That's correct, and also it's checked directly in e.g. https://github.com/tc39/proposal-temporal/pull/1470/files/f607e9253bac10eb0c13c3a543e2b5fbd7eaacf0#diff-e430a79407f82088e99ad1ef05fd1bf0e6d043135b7e42a1483f631c7fa66c9aR400 that the field wasn't copied.

PR #1482, opened 2021-04-09
Remove dangling IsValidCalendarID

### LINKED ISSUE #1477: Dangling IsValidCalendarID
13.9 IsValidCalendarID ( id ) defined in https://tc39.es/proposal-temporal/#sec-temporal-isvalidcalendarid 

is not refereed by any spec text in Temporal nor in ECMA262 or ECMA402

What is it for? No other part of spec is referring to this. Is it designed for some other proposal to refer to it?

@littledan @sffc @ptomato

FrankYFTang:
Remove IsValidCalendarID which no spec text referring to
close #1477

@pipobscure @maggiepint @mattjohnsonpint @bterlson @sffc @ryzokuken @ptomato @jasonwilliams @justingrant

### DIFF
--- a/spec/abstractops.html
+++ b/spec/abstractops.html
@@ -143,22 +143,6 @@
     </emu-alg>
   </emu-clause>
 
-  <emu-clause id="sec-temporal-isvalidcalendarid" aoid="IsValidCalendarID">
-    <h1>IsValidCalendarID ( _id_ )</h1>
-    <p>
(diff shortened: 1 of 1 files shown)


ptomato (approved):
Thanks. This is indeed no longer needed, as per #300 callers can no longer pass in their own calendar ID to the constructor. I was going to ask if you could move the editor's note to https://tc39.es/proposal-temporal/#sec-temporal-iso8601grammar to apply to the CalendarName production, pending #1450, but I don't _think_ that's necessary since there is already text there explaining the IETF draft.

PR #1485, opened 2021-04-14
Handle opposite date/time signs in DifferenceISODateTime.

Ms2ger:
This ensures cases such as months=1, hours=-1 are balanced correctly.

Fixes #1415.

### DIFF
--- a/polyfill/lib/ecmascript.mjs
+++ b/polyfill/lib/ecmascript.mjs
@@ -3290,7 +3290,35 @@ export const ES = ObjectAssign({}, ES2020, {
       µs2,
       ns2
     );
+
+    const timeSign = ES.DurationSign(
+      0,
+      0,
+      0,
+      deltaDays,
+      hours,
+      minutes,
+      seconds,
+      milliseconds,
+      microseconds,
+      nanoseconds
+    );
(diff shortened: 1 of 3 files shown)


Ms2ger:
Spec to be done if this seems like a good approach.

polyfill/lib/ecmascript.mjs
```diff
@@ -2960,7 +2960,24 @@ export const ES = ObjectAssign({}, ES2020, {
       µs2,
       ns2
     );
+
+    const timeSign = ES.DurationSign(0, 0, 0, deltaDays, hours, minutes, seconds, milliseconds, microseconds, nanoseconds);
     ({ year: y1, month: mon1, day: d1 } = ES.BalanceISODate(y1, mon1, d1 + deltaDays));
+    const dateSign = ES.CompareISODate(y1, mon1, d1, y2, mon2, d2);
+    if (dateSign === timeSign) {
```
ptomato, line 2967: Using `===` here confused me, but I guess it's because the ComparisonResult (as in, C strcmp order) has the opposite sign from DurationSign? Maybe it would be clearer to `CompareISODate(y2, mon2, d2, y1, mon1, d1)` or something like that.

ptomato (commented):
The approach of rebalancing the time part if the signs are opposite seems like a good one to me.

ryzokuken (approved):

Issue #1398, opened 2021-02-27
test262 work items


cjtenny:
In #1393 @ptomato discussed wanting a place to track tests that will need to be added to test262. We decided a single meta-issue should suffice. Feel free to edit, add, and sign up for any items! This is a living list and should be expected to grow.

- [ ] From #1393: Does Temporal check that retrieved methods from user-supplied objects are callable?
- [ ] From #1393: `timeZone.getInstantFor` throw for non-Temporal.Instant values
- [ ] #1047: calendar.fields return includes 'calendar'
- [ ] #1047: calendar.fields return includes non-calendar fields, e.g. 'nanoseconds'
- [ ] #1134: Time zones for Temporal.now()
- [ ] From #1397: Observable lookups & calls for timeZone objects from other temporal methods
- [ ] #1388: Calendar order of operations (with and)? without Intl
- [ ] From #1396: `PlainMonthDay.prototype.toPlainDate` correctly handle `[['year', undefined]]` for input fields
- [ ] From #1489: Test [CalendarFields calls](https://github.com/tc39/proposal-temporal/pull/1489#discussion_r615068081), including with [fields added](https://github.com/tc39/proposal-temporal/pull/1489#discussion_r615082466), and [getPossibleInstantsFor type checks](https://github.com/tc39/proposal-temporal/pull/1489#discussion_r615141800)
- [ ] Check fast-paths from #1428:
  - [ ] PlainDateTime -> PlainTime
  - [ ] ZonedDateTime -> TimeZone
  - [ ] PlainTime -> Calendar
  - [ ] PlainDate -> relativeTo object
  - [ ] Callers of GetOptionalTemporalCalendar and ToTemporalMonthDay, with a not-previously-covered Temporal object (e.g. ToTemporalDate with some object like `Object.assign(new Temporal.MonthDay(1, 1), { year: 2000 }))`
  - [ ] ZonedDateTime -> PlainDate, PlainDateTime, PlainTime
- [ ] Check that with() methods don't call getters on Temporal objects
- [ ] Check all remaining callers of GetPossibleInstantsFor with checkTimeZonePossibleInstantsIterable
- [ ] Check that things of the wrong type in an iterable (calendar.fields return, calendar.fields input, timeZone.getPossibleInstantsFor return) throw before exhausting the iterator
- [ ] Check that `calendar.fields` is always called with the correct list of fields
- [ ] Check that Gets occur for every extra field returned by `calendar.fields`
[...]

PR #1489, opened 2021-04-16
Consume iterables where possible

ptomato:
Closes: #1427

### DIFF
--- a/polyfill/lib/ecmascript.mjs
+++ b/polyfill/lib/ecmascript.mjs
@@ -1,7 +1,5 @@
 /* global __debug__ */
 
-const ArrayIsArray = Array.isArray;
-const ArrayPrototypeIndexOf = Array.prototype.indexOf;
 const ArrayPrototypePush = Array.prototype.push;
 const IntlDateTimeFormat = globalThis.Intl.DateTimeFormat;
 const MathMin = Math.min;
@@ -4342,36 +4340,6 @@ export const ES = ObjectAssign({}, ES2020, {
       throw new RangeError(`${property} must be between ${minimum} and ${maximum}, not ${value}`);
     }
     return MathFloor(value);
-  },
-  // Following two operations are overridden because the es-abstract version of
-  // ES.Get() unconditionally uses util.inspect
-  LengthOfArrayLike: (obj) => {
-    if (ES.Type(obj) !== 'Object') {
-      throw new TypeError('Assertion failed: `obj` must be an Object');
-    }
-    return ES.ToLength(obj.length);
-  },
-  CreateListFromArrayLike: (obj, elementTypes) => {
-    if (ES.Type(obj) !== 'Object') {
-      throw new TypeError('Assertion failed: `obj` must be an Object');
-    }
-    if (!ArrayIsArray(elementTypes)) {
-      throw new TypeError('Assertion failed: `elementTypes`, if provided, must be an array');
-    }
-    var len = ES.LengthOfArrayLike(obj);
-    var list = [];
-    var index = 0;
-    while (index < len) {
-      var indexName = ES.ToString(index);
-      var next = obj[indexName];
-      var nextType = ES.Type(next);
-      if (ArrayPrototypeIndexOf.call(elementTypes, nextType) < 0) {
-        throw new TypeError(`item type ${nextType} is not a valid elementType`);
-      }
-      ArrayPrototypePush.call(list, next);
-      index += 1;
-    }
-    return list;
   }
 });
 
--- a/polyfill/lib/ecmascript.mjs
+++ b/polyfill/lib/ecmascript.mjs
@@ -1236,7 +1236,18 @@ export const ES = ObjectAssign({}, ES2020, {
       }
 
       calendar = ES.GetOptionalTemporalCalendar(item);
-      const fieldNames = ES.CalendarFields(calendar, ['day', 'month', 'monthCode', 'year']);
+      const fieldNames = ES.CalendarFields(calendar, [
+        'day',
+        'hour',
+        'microsecond',
+        'millisecond',
+        'minute',
+        'month',
+        'monthCode',
+        'nanosecond',
+        'second',
+        'year'
+      ]);
       const fields = ES.ToTemporalDateTimeFields(item, fieldNames);
       ({
         year,
@@ -1508,7 +1519,18 @@ export const ES = ObjectAssign({}, ES2020, {
     if (ES.Type(item) === 'Object') {
       if (ES.IsTemporalZonedDateTime(item)) return item;
       calendar = ES.GetOptionalTemporalCalendar(item);
-      const fieldNames = ES.CalendarFields(calendar, ['day', 'month', 'year']);
+      const fieldNames = ES.CalendarFields(calendar, [
+        'day',
+        'hour',
+        'microsecond',
+        'millisecond',
+        'minute',
+        'month',
+        'monthCode',
+        'nanosecond',
+        'second',
+        'year'
+      ]);
       const fields = ES.ToTemporalZonedDateTimeFields(item, fieldNames);
       ({
         year,
--- a/polyfill/lib/ecmascript.mjs
+++ b/polyfill/lib/ecmascript.mjs
@@ -2849,9 +2849,7 @@ export const ES = ObjectAssign({}, ES2020, {
 
     let calendar;
     if (relativeTo) {
-      if (!(ES.IsTemporalDateTime(relativeTo) || ES.IsTemporalZonedDateTime(relativeTo))) {
-        throw new TypeError('starting point must be PlainDateTime or ZonedDateTime');
-      }
+      relativeTo = ES.ToTemporalDateTime(relativeTo);
       calendar = GetSlot(relativeTo, CALENDAR);
     }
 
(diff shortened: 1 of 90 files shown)


spec/calendar.html
```diff
@@ -56,7 +56,10 @@ <h1>CalendarFields ( _calendar_, _fieldNames_ )</h1>
         1. Let _fieldsArray_ be ? CreateArrayFromList(_fieldNames_).
         1. If _fields_ is not *undefined*, then
           1. Set _fieldsArray_ to ? Call(_fields_, _calendar_, « _fieldsArray_ »).
-        1. Return ? CreateListFromArrayLike(_fieldsArray_, « String »).
+        1. Let _result_ be ? IterableToList(_fieldsArray_).
+        1. For each element _name_ of _result_, do
+          1. If Type(_name_) is not String, throw a *TypeError*.
+        1. Return _result_.
```
ljharb, line 62: it seems like this should be throwing the error as soon as the item is encountered - in other words, if i pass an infinite iterable where the first item is not a string, i'd expect an exception rather than freezing the thread.

Promise.all (and the other combinators) work this way; i'm not sure if there's a strong reason not to do the validation per-item.

spec/timezone.html
```diff
@@ -626,7 +626,7 @@ <h1>BuiltinTimeZoneGetInstantFor ( _timeZone_, _dateTime_, _disambiguation_ )</h
       <h1>GetPossibleInstantsFor ( _timeZone_, _dateTime_ )</h1>
       <emu-alg>
         1. Let _possibleInstants_ be ? Invoke(_timeZone_, *"getPossibleInstantsFor"*, « _dateTime_ »).
-        1. Let _list_ be ? CreateListFromArrayLike(_possibleInstants_, « Object »).
+        1. Let _list_ be ? IterableToList(_possibleInstants_).
```
ljharb, line 629: this is actually normative, because https://tc39.es/ecma262/#sec-createlistfromarraylike indeed does the validation on each item.

ptomato: I agree, this is normative, it's one of the normative changes that I included in my slides last month as part of the conditions for Stage 3. (Even without the validation, CreateListFromArrayLike does Get(_obj_, *"length"*).) I was looking through the commit messages to see if I had mentioned anywhere that it was editorial, but I don't think that's the case?

spec/calendar.html
ptomato: Makes sense. I'm not sure how to cleanly do an AO with two independently optional parameters (_elementTypes_ and _method_) so I'm going with a separate operation IterableToListOfType( _items_, _elementTypes_ [ , _method_ ] ), does that sound good?

ljharb: Sure, that works

spec/timezone.html
ljharb: Nah it’s not a contradiction or anything - but i didn’t realize it was intentional. Per above, i think iterable items should be validated per-item, not post-iteration.

spec/calendar.html
bakkot: Works for me, though if you aren't going to use the `method` argument you should just leave it out of the AO - no reason to add complexity which isn't used.

ptomato: Updated.

spec/abstractops.html
```diff
@@ -4,6 +4,29 @@
...
+    </p>
+    <emu-alg>
+      1. Let _iteratorRecord_ be ? GetIterator(_items_, ~sync~).
+      1. Let _values_ be a new empty List.
+      1. Let _next_ be *true*.
+      1. Repeat, while _next_ is not *false*,
+        1. Set _next_ to ? IteratorStep(_iteratorRecord_).
+        1. If _next_ is not *false*, then
+          1. Let _nextValue_ be ? IteratorValue(_next_).
+          1. If Type(_nextValue_) is not an element of _elementTypes_, throw a *TypeError* exception.
```
bakkot, line 24: I believe you should close the iterator before throwing, as in

```suggestion
          1. If Type(_nextValue_) is not an element of _elementTypes_, then
            1. Let _completion_ be ThrowCompletion(a newly created TypeError object).
            1. Return ? IteratorClose(_iteratorRecord_, _completion_).
```

spec/abstractops.html
```diff
@@ -4,6 +4,29 @@
 <emu-clause id="sec-temporal-abstract-ops">
   <h1>Abstract operations</h1>
 
+  <!-- Based on ECMA-262 IterableToList -->
+  <emu-clause id="sec-iterabletolistoftype" aoid="IterableToListOfType">
+    <h1>IterableToList ( _items_, _elementTypes_ )</h1>
```
ljharb, line 9:
```suggestion
    <h1>IterableToListOfType ( _items_, _elementTypes_ )</h1>
```

spec/timezone.html
```diff
@@ -626,9 +626,15 @@ <h1>BuiltinTimeZoneGetInstantFor ( _timeZone_, _dateTime_, _disambiguation_ )</h
...
-        1. For each element _instant_ in _list_ in List order, do
-          1. Perform ? RequireInternalSlot(_instant_, [[InitializedTemporalInstant]]).
+        1. Let _iteratorRecord_ be ? GetIterator(_possibleInstants_, ~sync~).
+        1. Let _list_ be a new empty List.
+        1. Let _next_ be *true*.
+        1. Repeat, while _next_ is not *false*,
+          1. Set _next_ to ? IteratorStep(_iteratorRecord_).
+          1. If _next_ is not *false*, then
+            1. Let _nextValue_ be ? IteratorValue(_next_).
+            1. Perform ? RequireInternalSlot(_nextValue_, [[InitializedTemporalInstant]]).
```
ljharb, line 636: This also likely needs to close the iterator when throwing

polyfill/lib/ecmascript.mjs
```diff
@@ -1236,7 +1234,18 @@ export const ES = ObjectAssign({}, ES2020, {
       }
 
       calendar = ES.GetOptionalTemporalCalendar(item);
-      const fieldNames = ES.CalendarFields(calendar, ['day', 'month', 'monthCode', 'year']);
+      const fieldNames = ES.CalendarFields(calendar, [
```
cjtenny, line 1237: Oh dear, this bug must've been here a while. Will add a test262 work item for this and the one below in ToTemporalZonedDateTime, good catch.

polyfill/test/helpers/temporalHelpers.js
```diff
@@ -480,6 +553,40 @@ var TemporalHelpers = {
     assert.compareArray(actual, expected, "property getters not called");
   },
 
+  /*
+   * A custom calendar that returns an iterable instead of an array from its
+   * fields() method, otherwise identical to the ISO calendar.
+   */
+  calendarFieldsIterable() {
```
cjtenny, line 560: Should we have a follow up test262 work item (not needed for this PR) to test that returning an iterator with additional fields beyond the input works as well, or is that too redundant?

spec/timezone.html
cjtenny, line 636: Follow up test262 work item: test that iterator values are instants here (current test262 only checks that toString matches)

cjtenny (approved):
Looks great to me. Pending addressing Jordan's & Kevin's suggestions, ship it. Only comments are for some #1398 additions.

polyfill/test/helpers/temporalHelpers.js
ptomato: Yes, it should probably test that there are Gets done for the additional fields.

spec/timezone.html
```diff
@@ -626,9 +626,17 @@ <h1>BuiltinTimeZoneGetInstantFor ( _timeZone_, _dateTime_, _disambiguation_ )</h
...
+        1. Let _iteratorRecord_ be ? GetIterator(_possibleInstants_, ~sync~).
+        1. Let _list_ be a new empty List.
+        1. Let _next_ be *true*.
+        1. Repeat, while _next_ is not *false*,
+          1. Set _next_ to ? IteratorStep(_iteratorRecord_).
+          1. If _next_ is not *false*, then
+            1. Let _nextValue_ be ? IteratorValue(_next_).
+            1. If Type(_nextValue_) is not Object or _nextValue_ does not have an [[InitializedTemporalInstant]] internal slot, then
+              1. Let _completion_ be ThrowCompletion(a newly created *TypeError* object).
+              1. Return ? IteratorClose(_iteratorRecord_, _completion_).
```
ljharb, line 638: Do IteratorStep and IteratorValue's abrupt return need to also call IteratorClose?

spec/abstractops.html
```diff
@@ -4,6 +4,31 @@
...
+      It performs the following steps when called:
+    </p>
+    <emu-alg>
+      1. Let _iteratorRecord_ be ? GetIterator(_items_, ~sync~).
+      1. Let _values_ be a new empty List.
+      1. Let _next_ be *true*.
+      1. Repeat, while _next_ is not *false*,
+        1. Set _next_ to ? IteratorStep(_iteratorRecord_).
+        1. If _next_ is not *false*, then
+          1. Let _nextValue_ be ? IteratorValue(_next_).
```
ljharb, line 23: same here re IteratorClose

spec/timezone.html
bakkot: No; if those throw it represents error an in the _iterator itself_, rather than the iterator working correctly but the consumer having an issue.

ljharb: thanks, i wasn't sure about that (and didn't check the main spec first)

ptomato: They don't do so in https://tc39.es/ecma262/#sec-iterabletolist, so is this something that should be resolved at the ecma262 level?

spec/timezone.html
ptomato: Do you mean, test that in all built-in time zones the elements of the array returned from getPossibleInstantsFor are Instants?

spec/timezone.html
ptomato: Oh, I guess my comment crossed with Kevin's; resolving.

spec/timezone.html
cjtenny: and that returning a non-Instant from getPossibleInstantsFor properly throws at this point, that is: when consuming the iterable

PR #1494, opened 2021-04-17
Various normative issues that are part of Stage 3 conditions

ptomato:
Fixes the normative issues identified in #1424 (and a few various editorial things)

See: #1424

### DIFF
--- a/polyfill/lib/ecmascript.mjs
+++ b/polyfill/lib/ecmascript.mjs
@@ -4352,7 +4352,7 @@ export const ES = ObjectAssign({}, ES2020, {
       const ms = Date.now();
       const result = bigInt(ms).multiply(1e6).plus(ns);
       ns = ms % 1e6;
-      return result;
+      return bigInt.min(NS_MAX, bigInt.max(NS_MIN, result));
     };
   })(),
   SystemTimeZone: () => {
--- a/polyfill/lib/ecmascript.mjs
+++ b/polyfill/lib/ecmascript.mjs
@@ -1745,14 +1745,18 @@ export const ES = ObjectAssign({}, ES2020, {
     if (mergeFields === undefined) return { ...fields, ...additionalFields };
     return ES.Call(mergeFields, calendar, [fields, additionalFields]);
   },
-  CalendarDateAdd: (calendar, date, duration, options) => {
-    const dateAdd = ES.GetMethod(calendar, 'dateAdd');
+  CalendarDateAdd: (calendar, date, duration, options, dateAdd) => {
+    if (dateAdd === undefined) {
+      dateAdd = ES.GetMethod(calendar, 'dateAdd');
+    }
     const result = ES.Call(dateAdd, calendar, [date, duration, options]);
     if (!ES.IsTemporalDate(result)) throw new TypeError('invalid result');
     return result;
   },
-  CalendarDateUntil: (calendar, date, otherDate, options) => {
-    const dateUntil = ES.GetMethod(calendar, 'dateUntil');
+  CalendarDateUntil: (calendar, date, otherDate, options, dateUntil) => {
+    if (dateUntil === undefined) {
+      dateUntil = ES.GetMethod(calendar, 'dateUntil');
+    }
     const result = ES.Call(dateUntil, calendar, [date, otherDate, options]);
     if (!ES.IsTemporalDuration(result)) throw new TypeError('invalid result');
     return result;
@@ -2858,15 +2862,24 @@ export const ES = ObjectAssign({}, ES2020, {
         // no-op
         break;
       case 'months':
-        if (!calendar) throw new RangeError('a starting point is required for months balancing');
-        // balance years down to months
-        while (MathAbs(years) > 0) {
-          const newRelativeTo = ES.CalendarDateAdd(calendar, relativeTo, oneYear, {});
-          const oneYearMonths = ES.CalendarDateUntil(calendar, relativeTo, newRelativeTo, { largestUnit: 'months' })
-            .months;
-          relativeTo = newRelativeTo;
-          months += oneYearMonths;
-          years -= sign;
+        {
+          if (!calendar) throw new RangeError('a starting point is required for months balancing');
+          // balance years down to months
+          const dateAdd = ES.GetMethod(calendar, 'dateAdd');
+          const dateUntil = ES.GetMethod(calendar, 'dateUntil');
+          while (MathAbs(years) > 0) {
+            const newRelativeTo = ES.CalendarDateAdd(calendar, relativeTo, oneYear, {}, dateAdd);
+            const oneYearMonths = ES.CalendarDateUntil(
+              calendar,
+              relativeTo,
+              newRelativeTo,
+              { largestUnit: 'months' },
+              dateUntil
+            ).months;
+            relativeTo = newRelativeTo;
+            months += oneYearMonths;
+            years -= sign;
+          }
         }
         break;
       case 'weeks':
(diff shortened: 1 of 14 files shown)


polyfill/lib/ecmascript.mjs
```diff
@@ -2837,15 +2841,21 @@ export const ES = ObjectAssign({}, ES2020, {
...
-          const newRelativeTo = ES.CalendarDateAdd(calendar, relativeTo, oneYear, {});
-          const oneYearMonths = ES.CalendarDateUntil(calendar, relativeTo, newRelativeTo, { largestUnit: 'months' })
-            .months;
-          relativeTo = newRelativeTo;
-          months += oneYearMonths;
-          years -= sign;
+        {
+          if (!calendar) throw new RangeError('a starting point is required for months balancing');
+          // balance years down to months
+          const dateAdd = ES.GetMethod(calendar, 'dateAdd');
```
cjtenny, line 2847: This and all following similar modifications change two things semantically:
- No more `if (!ES.IsTemporalDate(result)) throw new TypeError('invalid result');` aka `Perform ? RequireInternalSlot(addedDate, [[InitializedTemporalDate]]).` (not good?)
- Reduced lookups (good)

I think the spec doesn't match either (same with second instance in later comment, from previous optimization)

spec/abstractops.html
```diff
@@ -1508,7 +1508,7 @@ <h1>PrepareTemporalFields ( _fields_, _fieldNames_, _requiredFields_ )</h1>
     <emu-note>The value of ? ToIntegerOrInfinity(*undefined*) is 0.</emu-note>
     <emu-alg>
       1. Assert: Type(_fields_) is Object.
-      1. Let _result_ be ? OrdinaryObjectCreate(%Object.prototype%).
+      1. Let _result_ be ! OrdinaryObjectCreate(%Object.prototype%).
```
cjtenny, line 1511: Though this isn't an options object, do we still want it doing lookups on Object.prototype? Does it get passed back to user code and that's why?

spec/abstractops.html
```diff
@@ -1671,7 +1671,7 @@ <h1>PreparePartialTemporalFields ( _fields_, _fieldNames_ )</h1>
     <emu-alg>
       1. If Type(_fields_) is not Object, then
         1. Throw a *TypeError* exception.
-      1. Let _result_ be ? OrdinaryObjectCreate(%Object.prototype%).
+      1. Let _result_ be ! OrdinaryObjectCreate(%Object.prototype%).
```
cjtenny, line 1674: (repeat)

spec/calendar.html
```diff
@@ -436,7 +447,7 @@ <h1>ResolveISOMonth ( _fields_ )</h1>
         1. If _numberPart_ is *NaN*, throw a *RangeError* exception.
```
cjtenny, line 447: this line needs to go? I guess not normative but as long as you're in here

spec/duration.html
```diff
@@ -1107,11 +1110,12 @@ <h1>AddDuration ( _y1_, _mon1_, _w1_, _d1_, _h1_, _min1_, _s1_, _ms1_, _mus1_, _
           1. Let _dateDuration1_ be ? CreateTemporalDuration(_y1_, _mon1_, _w1_, _d1_, 0, 0, 0, 0, 0, 0).
           1. Let _dateDuration2_ be ? CreateTemporalDuration(_y2_, _mon2_, _w2_, _d2_, 0, 0, 0, 0, 0, 0).
           1. Let _dateAdd_ be ? GetMethod(_calendar_, *"dateAdd"*).
-          1. Let _options_ be ! OrdinaryObjectCreate(%Object.prototype%).
-          1. Let _intermediate_ be ? Call(_dateAdd_, _calendar_, « _datePart_, _dateDuration1_, _options_ »).
-          1. Let _end_ be ? Call(_dateAdd_, _calendar_, « _intermediate_, _dateDuration2_, _options_ »).
+          1. Let _firstAddOptions_ be ! OrdinaryObjectCreate(*null*).
+          1. Let _intermediate_ be ? Call(_dateAdd_, _calendar_, « _datePart_, _dateDuration1_, _firstAddOptions_ »).
+          1. Let _secondAddOptions_ be ! OrdinaryObjectCreate(*null*).
+          1. Let _end_ be ? Call(_dateAdd_, _calendar_, « _intermediate_, _dateDuration2_, _secondAddOptions_ »).
```
cjtenny, line 1116: When this was previously changed to cache the GetMethod, I believe I reviewed it but I didn't realize then that this comment applies here too:
> No more `if (!ES.IsTemporalDate(result)) throw new TypeError('invalid result')`; aka `Perform ? RequireInternalSlot(addedDate, [[InitializedTemporalDate]])`. (not good?)

spec/intl.html
```diff
@@ -1858,7 +1858,7 @@ <h1>PrepareTemporalFields ( _fields_, _fieldNames_, _requiredFields_ )</h1>
         <p>This definition supersedes the definition provided in <emu-xref href="#sec-temporal-preparetemporalfields"></emu-xref>.</p>
         <emu-alg>
           1. Assert: Type(_fields_) is Object.
-          1. Let _result_ be ? OrdinaryObjectCreate(%Object.prototype%).
+          1. Let _result_ be ! OrdinaryObjectCreate(%Object.prototype%).
```
cjtenny, line 1861: (same question for abstractops.html version)

cjtenny (commented):
Looks great; however, I think the change of CalendarDateAdd to caching the GetMethod both in this PR and from a previous PR has introduced some bugs. Also, a quick question about object prototype in PrepareTemporalFields

spec/calendar.html
```diff
@@ -416,7 +416,18 @@ <h1>ToISOWeekOfYear ( _year_, _month_, _day_ )</h1>
       <emu-note>Beware that dates at the beginning of a year may be part of a week from the preceding year, and dates at the end of a year may be part of a week at the beginning of the next year, as the first week of any year is defined as the week that contains the first Thursday of the year.</emu-note>
     </emu-clause>
 
-    <emu-clause id="sec-temporal-resolvenonlunisolarmonth" aoid="ResolveISOMonth">
+    <emu-clause id="sec-buildisomonthcode" aoid="BuildISOMonthCode">
+      <h1>BuildISOMonthCode ( _month_ )</h1>
+      <p>
+        The abstract operation BuildISOMonthCode returns the corresponding string month code, given the numeric ISO calendar month _month_.
```
ljharb, line 422: is “numeric iso calendar month” defined somewhere explicitly so that this description functions as a type assertion?

spec/duration.html
```diff
@@ -1244,23 +1246,28 @@ <h1>RoundDuration ( _years_, _months_, _weeks_, _days_, _hours_, _minutes_, _sec
           1. Let _fractionalSeconds_ be _nanoseconds_ × 10<sup>−9</sup> + _microseconds_ × 10<sup>−6</sup> + _milliseconds_ × 10<sup>−3</sup> + _seconds_.
         1. Let _remainder_ be *undefined*.
         1. If _unit_ is *"years"*, then
+          1. Let _dateAdd_ be ? GetMethod(_calendar_, *"dateAdd"*).
```
ljharb, line 1249: is it intentional to throw this error here, before the CreateTemporalDuration ones, and not at the time it’s first invoked?

spec/abstractops.html
ptomato: `Object.create(null)` objects behave weirdly, and I think are to be avoided in general. I agreed with @bakkot's suggestion here because ECMA-402 has [GetOptionsObject](https://tc39.es/ecma402/#sec-getoptionsobject) which does `Object.create(null)` if you pass it `undefined`, and options objects should be consistent throughout JS.

However, I don't think we need that consistency with every object that's created internally within Temporal, and there are instances elsewhere in ECMA-262 where normal objects are created.

(For example, changing this to `null` would mean that something like `plainDate.getFields().hasOwnProperty('year')` would throw, because `hasOwnProperty` isn't looked up on Object.prototype)

spec/calendar.html
ptomato: I can add that it's a number between 1 and 12, or write an explicit assertion algorithm step.

spec/duration.html
ptomato: It has to be in some defined order, but it doesn't have to be this particular order. I can move it up to just before it's first invoked.

spec/calendar.html
ljharb: As long as the assertion is explicit in either prose or steps, I’m happy :-) preference beyond that is up to the editors.

cjtenny (approved):
lgtm once the minor open comments are addressed!

Issue #1249, opened 2021-01-08
Sync GetOption operations with ECMA-402 2021 edition


ptomato:
If/when https://github.com/tc39/ecma402/pull/493 lands in ECMA-402, as is expected for the 2021 edition, update our NormalizeOptionsObject operation to the new name (GetOptionsObject?) and move it from the abstract operations section to the amendments section, and there delete it from ECMA-402 and insert it in ECMA-262.

PR #1495, opened 2021-04-17
Update to the version of NormalizeOptionsObject in ECMA-402

ptomato:
This eventually landed as GetOptionsObject in the 2021 edition of ECMA-402
so update the name, and note in the amendments section that the operations
are to be moved from ECMA-402 into ECMA-262.

Closes: #1249


cjtenny (approved):
lgtm!

Issue #1464, opened 2021-03-30
[polyfil] Two suggestions


sla100:
1. Temporal polyfil has a hard dependency to other polyfil: 
```json
"dependencies": {
  "big-integer": "^1.6.48",
}
```
and them import `big-integer` package expicitly (ex. `lib/ecmascript.mjs`):
```js 
import bigInt from 'big-integer';
```
It means that polyfil testing other polyfil even if [`bigint` is available on any modern browsers and Node > 10 implementations ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt#browser_compatibility).

IMHO your polyfil should using `big-integer` only as `devDepnedency` for tests and simply call global factory: 
```js
const NS_MIN = BigInt(-86400).multiply(1e17);
```

2. You can use the new functionality `exports` to to indicate the location of `esm` modules:
```json
"main": "dist/index.js",
"exports": {
  "import": "./lib/index.mjs",
  "require": "./dist/index.js"
}
```
[See dual_commonjs_es_module_packages](https://nodejs.org/dist/latest-v14.x/docs/api/packages.html#packages_dual_commonjs_es_module_packages)

This will allow to reduce import path from 
```js
import { Temporal } from 'proposal-temporal/lib/index.mjs';
```
to
```js
import { Temporal } from 'proposal-temporal';
```

Best regards

sla100:
> 
> `big-integer` isn't a polyfill, it's a bigint library. 

Forgive my mistake. You have absolutely right.

PR #1499, opened 2021-04-19
Add exports to package.json

ptomato:
Closes: #1464

### DIFF
--- a/polyfill/package.json
+++ b/polyfill/package.json
@@ -5,6 +5,10 @@
   "type": "commonjs",
   "main": "dist/index.js",
   "browser": "dist/index.umd.js",
+  "exports": {
+    "import": "./lib/index.mjs",
+    "require": "./dist/index.js"
+  },
   "types": "index.d.ts",
   "scripts": {
     "coverage": "c8 report --reporter html",
--- a/polyfill/README.md
+++ b/polyfill/README.md
@@ -31,7 +31,7 @@ const { Temporal } = require('proposal-temporal');
 Or, import the polyfill as an ES6 module:
 
 ```javascript
-import { Temporal } from 'proposal-temporal/lib/index.mjs';
+import { Temporal } from 'proposal-temporal';
 ```
 
 ## Node REPL with Temporal


cjtenny (approved):
:goal_net:

ljharb:
Note that adding this is a breaking change; the format you've used won't work in every node version, and I'm not sure why this improvement is needed given that the polyfill should theoretically be deprecated _very_ soon?

polyfill/package.json
```diff
@@ -5,6 +5,10 @@
   "type": "commonjs",
   "main": "dist/index.js",
   "browser": "dist/index.umd.js",
+  "exports": {
+    "import": "./lib/index.mjs",
+    "require": "./dist/index.js"
+  },
```
ljharb, line 11: this is what would be needed if you wanted to actually support every node version:
```suggestion
  "exports": {
     ".": [
       {
        "import": "./lib/index.mjs",
        "default": "./dist/index.js"
       },
       "./dist/index.js"
     ]
  },
```

but note that this is still a breaking change unless these are the only two files included in the package (which i can see they aren't from the "browser" field above)

ljharb (changes requested):

ptomato:
> Note that adding this is a breaking change; the format you've used won't work in every node version, and I'm not sure why this improvement is needed given that the polyfill should theoretically be deprecated _very_ soon? (arguably, should have been deprecated a month ago, per prior agreements)

I don't agree that that precludes changes to leave it in a good place, but I didn't realize this was a breaking change.

ljharb:
@ptomato "leaving it in a good place" is fine within a reasonable timeframe, ofc - adding new features tho, breaking or otherwise, seems a bit out of scope of that.

littledan:
What's wrong with a breaking change on something at version 0.8?

ptomato:
Certainly breaking changes are fine at version 0.8, but once I realized this was breaking, it seemed like just churn without a clear benefit, unlike the other breaking changes we've had.

PR #1503, opened 2021-04-28
Update object-model.svg

bkardell:
Keep this useful simplification, but clarify that PlainTime doesn't have a calendar


ryzokuken (approved):
Awesome! @ptomato wdyt?

PR #1504, opened 2021-04-28
Add Alternative 4: Separate Types for Calendared Dates

sffc:
Adding another alternative that was missing from calendar-review.md


ptomato (approved):
:+1:

PR #1505, opened 2021-04-29
Misc editorial improvements

ptomato:
~~Marked as draft, because I would like to write tests where appropriate before merging.~~

See: #1411

### DIFF
--- a/spec/abstractops.html
+++ b/spec/abstractops.html
@@ -644,18 +644,6 @@
     </emu-alg>
   </emu-clause>
 
-  <emu-clause id="sec-temporal-nonnegativemodulo" aoid="NonNegativeModulo">
-    <h1>NonNegativeModulo ( _x_, _y_ )</h1>
-    <emu-alg>
-      1. Let _result_ be _x_ modulo _y_.
-      1. If _result_ is *−0*, then
-        1. Return 0.
-      1. If _result_ &lt; 0, then
-        1. Set _result_ to _result_ + _y_.
-      1. Return _result_.
-    </emu-alg>
-  </emu-clause>
-
   <emu-clause id="sec-temporal-sign" aoid="Sign">
     <h1>Sign ( _n_ )</h1>
     <p>
--- a/spec/abstractops.html
+++ b/spec/abstractops.html
@@ -1140,10 +1140,8 @@
         1. Let _millisecond_ be 0.
         1. Let _microsecond_ be 0.
         1. Let _nanosecond_ be 0.
-      1. If ! ValidateISODate(_year_, _month_, _day_) is *false*, then
-        1. Throw a *RangeError* exception.
-      1. If ! ValidateTime(_hour_, _minute_, _second_, _millisecond_, _microsecond_, _nanosecond_) is *false*, then
-        1. Throw a *RangeError* exception.
+      1. If ! IsValidISODate(_year_, _month_, _day_) is *false*, throw a *RangeError* exception.
+      1. If ! IsValidTime(_hour_, _minute_, _second_, _millisecond_, _microsecond_, _nanosecond_) is *false*, throw a *RangeError* exception.
       1. Return the new Record {
         [[Year]]: _year_,
         [[Month]]: _month_,
--- a/spec/abstractops.html
+++ b/spec/abstractops.html
@@ -514,7 +514,7 @@
           1. Return _value_.
         1. If _value_ has an [[InitializedTemporalDate]] internal slot, then
           1. Return ? CreateTemporalDateTime(_value_.[[ISOYear]], _value_.[[ISOMonth]], _value_.[[ISODay]], 0, 0, 0, 0, 0, 0, _value_.[[Calendar]]).
-        1. Let _calendar_ be ? GetOptionalTemporalCalendar(_value_).
+        1. Let _calendar_ be ? GetTemporalCalendarWithISODefault(_value_).
         1. Let _fieldNames_ be ? CalendarFields(_calendar_, « *"day"*, *"month"*, *"monthCode"*, *"year"* »).
         1. Let _fields_ be ? PrepareTemporalFields(_value_, _fieldNames_, «»).
         1. Let _dateOptions_ be ! OrdinaryObjectCreate(*null*).
@@ -525,7 +525,7 @@
       1. Else,
         1. Let _string_ be ? ToString(_value_).
         1. Let _result_ be ? ParseISODateTime(_string_).
-        1. Let _calendar_ be ? ToOptionalTemporalCalendar(_result_.[[Calendar]]).
+        1. Let _calendar_ be ? ToTemporalCalendarWithISODefault(_result_.[[Calendar]]).
         1. Let _offset_ be _result_.[[TimeZoneOffset]].
         1. Let _timeZone_ be _result_.[[TimeZoneIANAName]].
       1. If _timeZone_ is not *undefined*, then
--- a/spec/instant.html
+++ b/spec/instant.html
@@ -220,7 +220,6 @@
         1. Let _instant_ be the *this* value.
         1. Perform ? RequireInternalSlot(_instant_, [[InitializedTemporalInstant]]).
         1. Let _duration_ be ? ToLimitedTemporalDuration(_temporalDurationLike_, « *"years"*, *"months"*, *"weeks"*, *"days"* »).
-        1. Perform ? RejectDurationSign(_duration_.[[Years]], _duration_.[[Months]], _duration_.[[Weeks]], _duration_.[[Days]], _duration_.[[Hours]], _duration_.[[Minutes]], _duration_.[[Seconds]], _duration_.[[Milliseconds]], _duration_.[[Microseconds]], _duration_.[[Nanoseconds]]).
         1. Let _ns_ be ! AddInstant(_instant_.[[EpochNanoseconds]], _duration_.[[Hours]], _duration_.[[Minutes]], _duration_.[[Seconds]], _duration_.[[Milliseconds]], _duration_.[[Microseconds]], _duration_.[[Nanoseconds]]).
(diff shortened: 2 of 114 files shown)


spec/abstractops.html
```diff
@@ -57,47 +57,19 @@ <h1>GetOption ( _options_, _property_, _type_, _values_, _fallback_ )</h1>
       1. Assert: Type(_options_) is Object.
       1. Let _value_ be ? Get(_options_, _property_).
       1. If _value_ is *undefined*, return _fallback_.
-      1. Assert: _type_ is *"boolean"* or *"string"*.
+      1. Assert: _type_ is *"boolean"*, *"string"*, *"number"*, or *"stringOrNumber"*.
```
ljharb, line 60: rather than `stringOrNumber`, why not accept a List of valid types? I think we do this in a couple places already.

spec/instant.html
```diff
@@ -65,7 +65,7 @@ <h1>Temporal.Instant.fromEpochSeconds ( _epochSeconds_ )</h1>
         1. Set _epochSeconds_ to ? ToNumber(_epochSeconds_).
         1. Set _epochSeconds_ to ? NumberToBigInt(_epochSeconds_).
         1. Let _epochNanoseconds_ be _epochSeconds_ × 1,000,000,000.
```
ljharb, line 67: i'm not sure this math works, unless 1e9 is written with the bigint notation?
```suggestion
        1. Let _epochNanoseconds_ be _epochSeconds_ × 1000000000<sub>ℤ</sub>.
```

spec/instant.html
```diff
@@ -80,7 +80,7 @@ <h1>Temporal.Instant.fromEpochMilliseconds ( _epochMilliseconds_ )</h1>
         1. Set _epochMilliseconds_ to ? ToNumber(_epochMilliseconds_).
         1. Set _epochMilliseconds_ to ? NumberToBigInt(_epochMilliseconds_).
         1. Let _epochNanoseconds_ be _epochMilliseconds_ × 1,000,000.
```
ljharb, line 82: also here
```suggestion
        1. Let _epochNanoseconds_ be _epochMilliseconds_ × 1000000<sub>ℤ</sub>.
```

spec/instant.html
```diff
@@ -94,7 +94,7 @@ <h1>Temporal.Instant.fromEpochMicroseconds ( _epochMicroseconds_ )</h1>
       <emu-alg>
         1. Set _epochMicroseconds_ to ? ToBigInt(_epochMicroseconds_).
         1. Let _epochNanoseconds_ be _epochMicroseconds_ × 1,000.
```
ljharb, line 96: and here
```suggestion
        1. Let _epochNanoseconds_ be _epochMicroseconds_ × 1000<sub>ℤ</sub>.
```

spec/instant.html
```diff
@@ -482,30 +480,24 @@ <h1>Properties of Temporal.Instant Instances</h1>
...
+    <emu-clause id="sec-temporal-isvalidepochnanoseconds" aoid="IsValidEpochNanoseconds">
+      <h1>IsValidEpochNanoseconds ( _epochNanoseconds_ )</h1>
+      <p>
+        The abstract operation IsValidEpochNanoseconds returns *true* if its argument _epochNanoseconds_ is within the allowed range of nanoseconds since the Unix epoch for a Temporal.Instant and Temporal.ZonedDateTime, and *false* otherwise.
+      </p>
       <emu-alg>
-        1. Assert: Type(_ns_) is BigInt.
-        1. If _ns_ &lt; −8.64 × 10<sup>21</sup><sub>ℝ</sub> or _ns_ &gt; 8.64 × 10<sup>21</sup><sub>ℝ</sub>, then
+        1. Assert: Type(_epochNanoseconds_) is BigInt.
+        1. If _epochNanoseconds_ &lt; −8.64 × 10<sup>21</sup><sub>ℝ</sub> or _epochNanoseconds_ &gt; 8.64 × 10<sup>21</sup><sub>ℝ</sub>, then
```
ljharb, line 490: the R notation
```suggestion
        1. If _epochNanoseconds_ &lt; −8.64 × 10<sup>21</sup><sub>ℤ</sub> or _epochNanoseconds_ &gt; 8.64 × 10<sup>21</sup><sub>ℤ</sub>, then
```

spec/instant.html
```diff
@@ -567,7 +559,7 @@ <h1>AddInstant ( _epochNanoseconds_, _hours_, _minutes_, _seconds_, _millisecond
             _seconds_ × 1,000,000,000 +
             _minutes_ × 60,000,000,000 +
             _hours_ × 3,600,000,000,000.
```
ljharb, line 561: `AddInstant` needs to add type assertions to its arguments, and also be explicit about which kind of numbers are being used in this equation.

spec/abstractops.html
ptomato: Indeed, as in CreateListFromArrayLike? I thought about that, but here `stringOrNumber` is not just allowing either a string or a number, but has to be specific about calling ToString if the value is neither a number or a string.

Maybe having the List of valid types be in order of priority would work? I'll check if that makes things clearer / more general or not.

ljharb: hm, it's a bit tricky if the valid type translates to a coercion algorithm - having `string` or `number` implicitly map to `ToString` or `ToNumber` is awkward, and i definitely see why you'd have gone with `stringOrNumber`.

ptomato: I've done this in https://github.com/tc39/proposal-temporal/pull/1505/commits/ff399a30e119f40c2c978f0952641061bcd5bdc5 but kept it separate for now, I don't have a strong opinion on one being better than the other.

ljharb: Nice, that seems like an improvement to me.

spec/instant.html
```diff
@@ -64,8 +64,8 @@ <h1>Temporal.Instant.fromEpochSeconds ( _epochSeconds_ )</h1>
       <emu-alg>
         1. Set _epochSeconds_ to ? ToNumber(_epochSeconds_).
         1. Set _epochSeconds_ to ? NumberToBigInt(_epochSeconds_).
-        1. Let _epochNanoseconds_ be _epochSeconds_ × 1,000,000,000.
-        1. Perform ? RejectInstant(_epochNanoseconds_).
+        1. Let _epochNanoseconds_ be _epochSeconds_ × *1,000,000,000*<sub>ℤ</sub>.
```
ljharb, line 67: the spec doesn't currently use any `,` separators in long numbers (like 1000).
```suggestion
        1. Let _epochNanoseconds_ be _epochSeconds_ × *1000000000*<sub>ℤ</sub>.
```

ptomato: OK, in that case I've gone with scientific notation — I'd assume we don't want people to make mistakes counting the zeroes in 3600000000000.

spec/instant.html
```diff
@@ -561,13 +553,14 @@ <h1>AddInstant ( _epochNanoseconds_, _hours_, _minutes_, _seconds_, _millisecond
...
-            _minutes_ × 60,000,000,000 +
-            _hours_ × 3,600,000,000,000.
-        1. Perform ? RejectInstant(_result_).
+        1. Assert: _hours_, _minutes_, _seconds_, _milliseconds_, _microseconds_, and _nanoseconds_ are integer Number values.
+        1. Let _result_ be _epochNanoseconds_ + ℤ(_nanoseconds_) +
+            ℤ(_microseconds_) × *1000*<sub>ℤ</sub> +
+            ℤ(_milliseconds_) × *10<sup>6</sup>*<sub>ℤ</sub> +
+            ℤ(_seconds_) × *10<sup>9</sup>*<sub>ℤ</sub> +
+            ℤ(_minutes_) × *6 × 10<sup>10</sup>*<sub>ℤ</sub> +
+            ℤ(_hours_) × *3.6 × 10<sup>12</sup>*<sub>ℤ</sub>.
```
ljharb, line 562:
```suggestion
            ℤ(_minutes_) × *6<sub>ℤ</sub> × 10<sup>10</sup>*<sub>ℤ</sub> +
            ℤ(_hours_) × *3.6<sub>ℤ</sub> × 10<sup>12</sup>*<sub>ℤ</sub>.
```

ljharb (approved):
Spec seems good to me!

spec/abstractops.html
```diff
@@ -45,59 +45,35 @@ <h1>GetOptionsObject ( _options_ )</h1>
...
     <emu-alg>
       1. Assert: Type(_options_) is Object.
       1. Let _value_ be ? Get(_options_, _property_).
       1. If _value_ is *undefined*, return _fallback_.
-      1. Assert: _type_ is *"boolean"* or *"string"*.
-      1. If _type_ is *"boolean"*, then
-        1. Let _value_ be ! ToBoolean(_value_).
-      1. If _type_ is *"string"*, then
-        1. Let _value_ be ? ToString(_value_).
+      1. If _types_ contains an element equal to Type(_value_), then
```
Ms2ger, line 60:
```suggestion
      1. If _types_ contains Type(_value_), then
```

spec/instant.html
ljharb: ah right, duh. Either of your suggestions is fine; as long as every number is in the same space.

ptomato: I'll go with 3600<sub>ℤ</sub> × 10<sup>9</sup><sub>ℤ</sub> since if we are not doing proper scientific notation, this is the most illustrative way to write the multiplication in my opinion.

ptomato:
What's changed since the last time: I removed the GetOption commit since I'm still working on ensuring that all the entry points that use GetOption are covered by test262 tests. I added test262 tests that cover "Remove NonNegativeModulo abstract operation". This is ready for merging if the test262 tests pass review. I'll open a separate PR for the GetOption changes so that this one isn't delayed on that any longer.

Ms2ger (approved):

PR #1506, opened 2021-05-01
spec: refactor Duration rounding

ryzokuken:
Refactor the logic in `Duration.prototype.round` into an abstract
operation "RoundDurationComplete" so that it could be called in
`Intl.DurationFormat`.

### DIFF
--- a/spec/duration.html
+++ b/spec/duration.html
@@ -434,14 +434,7 @@
         1. Let _maximum_ be ! MaximumTemporalDurationRoundingIncrement(_smallestUnit_).
         1. Let _roundingIncrement_ be ? ToTemporalRoundingIncrement(_options_, _maximum_, *false*).
         1. Let _relativeTo_ be ? ToRelativeTemporalObject(_options_).
-        1. Let _unbalanceResult_ be ? UnbalanceDurationRelative(_duration_.[[Years]], _duration_.[[Months]], _duration_.[[Weeks]], _duration_.[[Days]], _largestUnit_, _relativeTo_).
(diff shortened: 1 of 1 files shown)


spec/duration.html
```diff
@@ -1370,6 +1363,20 @@ <h1>RoundDuration ( _years_, _months_, _weeks_, _days_, _hours_, _minutes_, _sec
       </emu-alg>
     </emu-clause>
 
+    <emu-clause id="sec-temporal-rounddurationcomplete" aoid="RoundDurationComplete">
+      <h1>RoundDurationComplete ( _duration_, _smallestUnit_, _largestUnit_, _relativeTo_, _roundingIncrement_, _roundingMode_ )</h1>
```
ptomato, line 1367: I'm trying to make sure we add a `<p>` with a blurb about each abstract operation going forward, and possibly information about the types of the arguments. Would you mind adding one here?

ptomato (approved):
I wonder if it's necessary to note somewhere that this abstract op is for the benefit of ECMA-402 being able to call it?

And it occurred to me, are all the relativeTo/roundingIncrement/etc args relevant for 402?

ryzokuken:
This one is intentionally stalled since it is unclear again whether we would actually need it in this next iteration of `Intl.DurationFormat` (yes, we need to reiterate again). If it's not a problem. let's leave this open for a week or so while we work out more details regarding what the semantics of the proposal should be.

ryzokuken:
This is no longer needed for `Intl.DurationFormat`. Thanks @ptomato.

PR #1511, opened 2021-05-14
Editorial fixes

Ms2ger:


spec/plaindate.html
```diff
@@ -687,13 +687,11 @@ <h1>ToTemporalDate ( _item_ [ , _options_ ] )</h1>
           1. Let _fieldNames_ be ? CalendarFields(_calendar_, « *"day"*, *"month"*, *"monthCode"*, *"year"* »).
           1. Let _fields_ be ? PrepareTemporalFields(_item_, _fieldNames_, «»).
           1. Return ? DateFromFields(_calendar_, _fields_, _options_).
-        1. Let _overflow_ be ? ToTemporalOverflow(_options_).
+        1. Let _overflow_ be ? ToTemporalOverflow(_options_). <mark>TODO: _overflow_ is unused</mark>.
```
ptomato, line 690: It's used in the "_item_ is an Object" branch above. I think what happens in similar cases elsewhere is "Perform ? ToTemporalOverflow(_options_)."

ptomato (approved):
Thanks, looks good.

At this point I am trying to make sure every change is covered with test262 tests; even if the change is editorial then I double-check to make sure that there is a test262 test covering it. In this case the only test I can think of is making sure that invalid values of `{overflow: "..."}` will cause an exception in Temporal.PlainDate.from() regardless of whether the first argument is an object or string.

ptomato (approved):
:+1:

PR #1513, opened 2021-05-18
Editorial fixes

Ms2ger:


ptomato (approved):
All :+1: (and I don't really see a useful opportunity to increase test262 coverage for these, so it's OK that they have no tests)

PR #1524, opened 2021-06-02
Add tests for the shape of prototype methods.

Ms2ger:


ptomato (approved):
Obviously I skimmed it...

PR #1526, opened 2021-06-02
Add tests for the shape of static methods.

Ms2ger:


ptomato (approved):
These are correct, so feel free to merge it. But there seem to be still missing a few, like `test/Calendar/constructor/from/name.js`, I don't know if you want to add those in this same batch.

Ms2ger:
Will follow up for those

PR #1530, opened 2021-06-03
Polyfill: Add missing brand checks in TimeZone.

Ms2ger:


ptomato (commented):
Looks good, but it seems there was a polyfill test somehow relying on not having the brand check; that should be fixed or removed as well.

ryzokuken (approved):

PR #1533, opened 2021-06-11
Add more tests for the shape of prototype methods

ptomato:
(PR submitted on behalf of @Ms2ger)


ptomato (commented):
Skimmed it, and added the eslint rules so that we match the conventions in upstream test262.

Ms2ger (approved):

PR #1541, opened 2021-06-17
Normative: Add Temporal[@@toStringTag].

Ms2ger:
Fixes #1539.

### DIFF
--- a/spec/temporal.html
+++ b/spec/temporal.html
@@ -14,6 +14,15 @@
     <li>does not have a [[Call]] internal method; it cannot be invoked as a function.</li>
   </ul>
 
+  <emu-clause id="sec-value-properties-of-the-temporal-object">
+    <h1>Value Properties of the Temporal Object</h1>
+    <emu-clause id="sec-temporal-@@tostringtag">
+      <h1>Temporal [ @@toStringTag ]</h1>
+      <p>The initial value of the @@toStringTag property is the String value *"Temporal"*.</p>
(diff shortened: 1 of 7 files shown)


spec/temporal.html
```diff
@@ -14,6 +14,15 @@ <h1>The Temporal Object</h1>
     <li>does not have a [[Call]] internal method; it cannot be invoked as a function.</li>
   </ul>
 
+  <emu-clause id="sec-value-properties-of-the-temporal-object">
+    <h1>Value Properties of the Temporal Object</h1>
+    <emu-clause id="sec-temporal-@@tostringtag">
+      <h1>Intl[ @@toStringTag ]</h1>
```
ljharb, line 20:
```suggestion
      <h1>Temporal [ @@toStringTag ]</h1>
```

ljharb (changes requested):

spec/temporal.html
Ms2ger: Doh.

ljharb (approved):

ptomato (approved):
I will mark this as "draft" so it's not accidentally merged before being presented to plenary.

ptomato:
If we go with Temporal.now[@@toStringTag] as discussed on #1539 then this PR will need to be updated.

ptomato:
I believe this is covered by the consensus achieved at the July 2021 TC39 meeting as part of the "Guidance on nested namespaces" item. I will add the updates to `Temporal.Now` as per that item and un-draft this.

PR #1545, opened 2021-06-17
Editorial: Clarify PlainYearMonth uses integer mathematical values internally

Ms2ger:

PR #1546, opened 2021-06-17
Editorial: Clarify PlainMonthDay uses integer mathematical values internally.

Ms2ger:


Ms2ger:
Depends on #1545.

ptomato (approved):
This all looks good, after considering the same comments as on #1545.

PR #1550, opened 2021-06-18
Change `PlainMonthDay.p.equals` docs code sample to use correct type

justingrant:
Replaced `PlainDateTime` (wrong) with `PlainMonthDay` in the docs for `PlainMonthDay.p.equals`. Also added examples using string and object parameters.


ptomato (approved):
Those pesky copy-paste errors...

PR #1554, opened 2021-06-22
Editorial: Fix some mathematical type issues around PlainDateTime.

Ms2ger:


polyfill/lib/plaindatetime.mjs
```diff
@@ -32,15 +32,15 @@ export class PlainDateTime {
...
-    isoYear = ES.ToInteger(isoYear);
-    isoMonth = ES.ToInteger(isoMonth);
-    isoDay = ES.ToInteger(isoDay);
-    hour = ES.ToInteger(hour);
-    minute = ES.ToInteger(minute);
-    second = ES.ToInteger(second);
-    millisecond = ES.ToInteger(millisecond);
-    microsecond = ES.ToInteger(microsecond);
-    nanosecond = ES.ToInteger(nanosecond);
+    isoYear = ES.ToFiniteInteger(isoYear);
```
ljharb, line 35: Is “ToFiniteInteger” a thing in the spec text?

ptomato: It is currently not, my guess is that @Ms2ger is planning to add it as an editorial change as part of #1413 if it makes things more readable once the other spec text changes are complete?

ptomato (approved):
:+1:

polyfill/lib/plaindatetime.mjs
ljharb: No concern about adding the AO, but it seems a bit strange to use it in the polyfill before there's spec text for it.

ptomato: Since it was pre-existing, I guess this would be appropriate for a follow-up.

PR #1558, opened 2021-06-23
Editorial: Mark AddDuration argument relativeTo as required.

Ms2ger:
It is always passed.


ptomato (approved):
:+1: but I think we also need a follow-up making the argument required in AdjustRoundedDurationDays

PR #1561, opened 2021-06-23
Update ambiguity.md to remove outdated "DateTime" type name

justingrant:
Fix references to outdated "DateTime" type name


ptomato (approved):
:+1:

PR #1567, opened 2021-06-24
Editorial: Fix various issues

Ms2ger:


spec/plaindatetime.html
```diff
@@ -1032,6 +1032,7 @@ <h1>RoundISODateTime ( _year_, _month_, _day_, _hour_, _minute_, _second_, _mill
       </p>
       <emu-alg>
         1. Assert: _year_, _month_, _day_, _hour_, _minute_, _second_, _millisecond_, _microsecond_, and _nanosecond_ are integers.
+        1. If _dayLength_ is not present, set _dayLength_ to 8.64 × 10<sup>13</sup>.
```
ljharb, line 1035: This magic number seems like it should be a constant/AO defined like https://tc39.es/ecma262/#sec-hours-minutes-second-and-milliseconds

ptomato: I think this is a good suggestion, this constant shows up several times elsewhere in the spec text (as well as 8.64×10^21, the maximum range in nanoseconds before and after the epoch)

ptomato (approved):
The constant could also be addressed separately, since it applies in several other places than just here.

spec/plaindatetime.html
ptomato: → #1589

PR #1571, opened 2021-06-25
Move PlainDate test coverage to test262.

Ms2ger:


polyfill/test/PlainDate/prototype/valueOf/basic.js
```diff
@@ -0,0 +1,17 @@
...
+---*/
+
+const plainDate = Temporal.PlainDate.from("1963-02-13");
+
+assert.throws(TypeError, () => plainDate.valueOf(), "valueOf");
+assert.throws(TypeError, () => plainDate < plainDate, "<");
+assert.throws(TypeError, () => plainDate <= plainDate, "<=");
+assert.throws(TypeError, () => plainDate > plainDate, ">");
+assert.throws(TypeError, () => plainDate >= plainDate, ">=");
+assert.sameValue(plainDate === plainDate, true, "===");
```
ptomato, line 16: Should we add a separate test to check that `===` is object equality? That was present in the old test suite.

polyfill/test/PlainDate/prototype/since/calendar-mismatch.js
```diff
@@ -0,0 +1,11 @@
+// Copyright (C) 2021 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.plaindate.prototype.since
+features: [Temporal, Intl]
+---*/
+
+const plainDate1 = new Temporal.PlainDate(2000, 1, 1);
+const plainDate2 = new Temporal.PlainDate(2000, 1, 1, Temporal.Calendar.from('japanese'));
```
ptomato, line 10: Would be nice to use a custom calendar here so that the test could execute just as well in an Intl-less environment, but this is something we can go through and do later.

ptomato (approved):
A few minor comments, looks good!

PR #1580, opened 2021-06-28
Try to clarify Calendar#fields documentation.

Ms2ger:
Fixes #1425.


ptomato (approved):
You're right, that was out of date.

PR #1581, opened 2021-06-28
Editorial: Reverse loop in DifferenceISODate for clarity.

Ms2ger:


ptomato (approved):
I agree, this makes the spec text much easier to read. The _years_ variable must have been a holdover from before the different largest unit paths were separated...

PR #1582, opened 2021-06-28
Editorial: Avoid dividing by zero in ZonedDateTime#round.

Ms2ger:
Ref. #1502.


ptomato (approved):
:+1:

ptomato:
I considered that, but I thought it was an editorial error because the division operation on two mathematical values in [RoundTime](https://tc39.es/proposal-temporal/#sec-temporal-roundtime) step 3.b would not be defined in ℝ if the divisor is 0.

However, I'm glad you prompted me to take another look, because it seems there is still an editorial error. The result of [AddInstant](https://tc39.es/proposal-temporal/#sec-temporal-addinstant) and therefore [AddZonedDateTime](https://tc39.es/proposal-temporal/#sec-temporal-addzoneddatetime) is a BigInt, meaning the division in RoundTime 3.b is on operands of mixed type, which is an editorial error according to [Mathematical operations](https://tc39.es/ecma262/#sec-mathematical-operations). @Ms2ger care to fix it up?

Ms2ger:
Done in #1585.

PR #1584, opened 2021-06-29
fix type of offset to string from number

FrankYFTang:
@ptomato @justingrant @Ms2ger


ptomato (approved):
Thanks!

justingrant:
Good catch @FrankYFTang. Copypasta strikes again!

PR #1592, opened 2021-06-30
Normative: Remove extra "secondsPart"

FrankYFTang:
_secondsPart_ is already concatated into _timePart_ 4 lines above and should not be concancated again.

@ptomato @justingrant @sffc @Ms2ger

### DIFF
--- a/spec/duration.html
+++ b/spec/duration.html
@@ -1487,7 +1487,7 @@
         1. Let _signPart_ be the code unit 0x002D (HYPHEN-MINUS) if _sign_ &lt; 0, and otherwise the empty String.
         1. Let _result_ be the string concatenation of _signPart_, the code unit 0x0050 (LATIN CAPITAL LETTER P) and _datePart_.
         1. If _timePart_ is not *""*, then
-          1. Set _result_ to the string concatenation of _result_, the code unit 0x0054 (LATIN CAPITAL LETTER T), and _timePart__.
+          1. Set _result_ to the string concatenation of _result_, the code unit 0x0054 (LATIN CAPITAL LETTER T), and _timePart_.
         1. Return _result_.
       </emu-alg>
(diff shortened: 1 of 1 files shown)


spec/duration.html
```diff
@@ -1487,7 +1487,7 @@ <h1>TemporalDurationToString ( _years_, _months_, _weeks_, _days_, _hours_, _min
         1. Let _signPart_ be the code unit 0x002D (HYPHEN-MINUS) if _sign_ &lt; 0, and otherwise the empty String.
         1. Let _result_ be the string concatenation of _signPart_, the code unit 0x0050 (LATIN CAPITAL LETTER P) and _datePart_.
         1. If _timePart_ is not *""*, then
-          1. Set _result_ to the string concatenation of _result_, the code unit 0x0054 (LATIN CAPITAL LETTER T), _timePart_, and _secondsPart_.
+          1. Set _result_ to the string concatenation of _result_, the code unit 0x0054 (LATIN CAPITAL LETTER T), and _timePart__.
```
ptomato, line 1490:
```suggestion
          1. Set _result_ to the string concatenation of _result_, the code unit 0x0054 (LATIN CAPITAL LETTER T), and _timePart_.
```

ptomato (approved):
I believe this change is correct.

It's not clear to me whether this is a normative change that we would have to take to plenary or not. The current spec text is a mistake, for sure, but it's not invalid, and this would be an observable change if someone had implemented the steps as written.

(Unless there is some scoping rule that makes _timePart_ local to step 21? But I didn't think the spec worked like that)

Ms2ger:
I'm guessing this is already covered by tests?

ptomato:
Converted to draft until we determine whether it is a normative change that we have to take to plenary or not.

ptomato:

PR #1597, opened 2021-07-01
fix typo in TemporalTimeToString

FrankYFTang:
Step 3 should be 
3. Let minute be minute formatted as a two-digit decimal number, padded to the left with a zero if necessary.
not
3. Let hour be minute formatted as a two-digit decimal number, padded to the left with a zero if necessary.

@sffc @ptomato @justingrant @Ms2ger @ryzokuken


ptomato (approved):
Typo indeed, thanks!

PR #1599, opened 2021-07-01
Editorial: Various fixes

Ms2ger:


spec/duration.html
```diff
@@ -1228,7 +1226,7 @@ <h1>RoundDuration ( _years_, _months_, _weeks_, _days_, _hours_, _minutes_, _sec
             1. Set _zonedRelativeTo_ to _relativeTo_.
             1. Set _relativeTo_ to ? BuiltinTimeZoneGetPlainDateTimeFor(_relativeTo_.[[TimeZone]], _instant_, _relativeTo_.[[Calendar]]).
           1. Else,
-            1. Perform ? RequireInternalSlot(_relativeTo_, [[InitializedTemporalDateTime]]).
+            1. Assert: _relativeTo_ has an [[InitializedTemporalDateTime]] internal slot.
```
ptomato, line 1229: A possible follow-up improvement could be to move the assert to the beginning of the AO, and assert that it is either undefined or an Object with either an [[InitializedTemporalZonedDateTime]] or [[InitializedTemporalDateTime]] internal slot.

ptomato (approved):
:+1:

PR #1612, opened 2021-07-04
Change unsupported {hebrew} to [u-ca=hebrew]

FrankYFTang:
This is a typo of {hebrew} right? I cannot find {} syntax in the ISO8601 grammar


FrankYFTang:
@sffc @justingrant @Ms2ger @ryzokuken @ptomato

ptomato (approved):
Thanks, I think this was left over from another proposed syntax.

PR #1613, opened 2021-07-04
Editorial: Add ", [[InitializedTemporalDateTime]]" to list

FrankYFTang:
Somehow  [[InitializedTemporalDateTime]] is missing from these places.

### DIFF
--- a/spec/calendar.html
+++ b/spec/calendar.html
@@ -239,7 +239,7 @@
       <h1>ToTemporalCalendar ( _temporalCalendarLike_ )</h1>
       <emu-alg>
         1. If Type(_temporalCalendarLike_) is Object, then
-          1. If _temporalCalendarLike_ has an [[InitializedTemporalDate]], [[InitializedTemporalDateTime]], [[InitializedTemporalMonthDay]], [[InitializedTemporalTime]], [[InitializedTemporalYearMonth]], or [[InitializedTemporalZonedDateTime]] internal slot, then
+          1. If _temporalCalendarLike_ has an [[InitializedTemporalDate]], [[InitializedTemporalDateTime]], [[InitializedTemporalDateTime]], [[InitializedTemporalMonthDay]], [[InitializedTemporalTime]], [[InitializedTemporalYearMonth]], or [[InitializedTemporalZonedDateTime]] internal slot, then
             1. Return _temporalCalendarLike_.[[Calendar]].
           1. If ? HasProperty(_temporalCalendarLike_, *"calendar"*) is *false*, return _temporalCalendarLike_.
           1. Set _temporalCalendarLike_ to ? Get(_temporalCalendarLike_, *"calendar"*).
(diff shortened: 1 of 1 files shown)


FrankYFTang:
@justingrant @ptomato @sffc @ryzokuken @Ms2ger

ptomato:
This was done in order to limit the different types that Calendar methods in userland calendars should expect to be called with; see https://github.com/tc39/proposal-temporal/commit/1b47def71d9439efc57ec0f1727dc1fe93ca3e8d.  Note that there is already a fast path for PlainDateTime in step 3.c of [ToTemporalDate](https://tc39.es/proposal-temporal/#sec-temporal-totemporaldate), which is called in these Calendar methods.

I would not be opposed to adding PlainDateTime in that set of types that user calendars can expect, but this would be a normative change.

FrankYFTang:
> Note that there is already a fast path for PlainDateTime in step 3.c of [ToTemporalDate](https://tc39.es/proposal-temporal/#sec-temporal-totemporaldate), which is called in these Calendar methods.

 "fast path" in  PlainDateTime in step 3.c of [ToTemporalDate]??? It is calling  CreateTemporalDate(item.[[ISOYear]], item.[[ISOMonth]], item.[[ISODay]], item.[[Calendar]]) which is not very fast.

If your argument of "because there is a fast path in [ToTemporalDate], therefore..." hold (which is NOT I think, because it now create object), we should then remove the checking of PlainDate here too, right?

ptomato:
The idea was that creating PlainDate from PlainDateTime is a fast path, because it does not have to go through the calendar and potentially call into user code. It's a good point that it still creates an object, so I'll put this on the agenda for the following Temporal champions meeting. Thanks for the feedback.

FrankYFTang:
ok, after I take a closer look and try to implement it, it seems this issue is not in the path of the normal execution. (correct me if I am wrong) - while the Plain* call the Calendar via the CalendarXXX SOs, the fileds got pass to call these method are a ordinary object (prepared by the PrepareTemporalFields ) and are not a Temporal.* object as fields, therefore adding additional types checkin will not help the speed up of the regular execution came from the Temporal.Plain* calls. It looks like the speed up by add more types will only impact if the developer directly calling these calendear method with a Temporal objects, which probably happen much less frequent. Is my assessment correct?

FrankYFTang:
Feel free to close this PR if you think it is not worthy to take it post Stage 3- After carefully exam the normal usage of these method, I believe it won't impact the normal performance too mcuh since most of the time Temporal.PlainDateTime won't be the one pass to Calendar anyway.

ptomato:
Meeting, July 22: We decided to add this change to the next batch of normative changes to ask for consensus on, since it may still impact other implementations' performance.

ptomato:
Based on trying to write a test for these, I have actually become convinced that the change is not observable (and therefore not normative). My reasoning goes like this, let's take `get PlainDateTime.prototype.monthCode` as an example, but the reasoning is the same for the other methods:
1. [get PlainDateTime.prototype.monthCode](https://tc39.es/proposal-temporal/#sec-get-temporal.plaindatetime.prototype.monthcode) calls CalendarMonthCode with a PlainDateTime object.
2. [CalendarMonthCode](https://tc39.es/proposal-temporal/#sec-temporal-calendarmonthcode) invokes the `monthCode` method on the calendar with a PlainDateTime object.
3. If the implementation doesn't have Intl, ISO is the only possible calendar, so,
   1. [Calendar.prototype.monthCode](https://tc39.es/proposal-temporal/#sec-temporal.calendar.prototype.monthcode) converts the PlainDateTime to a PlainDate and calls ISOMonthCode with it. (The conversion of PlainDateTime to PlainDate isn't observable.)
   2. [ISOMonthCode](https://tc39.es/proposal-temporal/#sec-temporal-isomonthcode) expects an object with an [[ISOMonth]] internal slot. So the implementation would be free to optimize away the conversion of PlainDateTime to PlainDate, and pass the PlainDateTime to ISOMonthCode.
4. If the implementation does have Intl, [Calendar.prototype.monthCode](https://tc39.es/proposal-temporal/#sup-temporal.calendar.prototype.monthcode) converts the PlainDateTime to a PlainDate (again, not observable) and either calls ISOMonthCode with it, for the ISO calendar, or performs implementation-defined processing, for other calendars. So in this case the implementation is also free to optimize away this conversion.

So, I think we can just merge this, after all. Making it explicit that PlainDateTime is OK here, does make the spec text clearer, in my opinion. @Ms2ger could you take a look and confirm?

Ms2ger (approved):

PR #1616, opened 2021-07-06
Normative: Use internal slot in UnbalanceDurationRelative.

Ms2ger:


ptomato (approved):
Oops, this looks like my mistake, and I agree it's a normative change

ptomato:
I've added Frank's additional fixes from #1673 here, to consolidate the two pull requests, and added test coverage for each entry point where this happens.

FrankYFTang:
thanks

ptomato:

PR #1617, opened 2021-07-06
Editorial: Mark BalanceTime as infallible.

Ms2ger:
And do the same for operations that are now infallible as well.


ptomato (approved):
:+1:

PR #1627, opened 2021-07-09
Editoral: rm test _calendarString_ is empty before concat

FrankYFTang:
No need to check the if, the concat of empty string is the same.


FrankYFTang:
Simplify these two AOs, there are no need to test the string is empty before append it.

@ptomato @justingrant @sffc @ryzokuken @Ms2ger

ptomato (approved):
Thanks.

PR #1628, opened 2021-07-09
Missing the first "-" in TemporalMonthDayToString

FrankYFTang:
When the calendar is "iso8601", PlainMonthDay of 
July 4, 2021 should be format as "-07-04" not  "07-04".


FrankYFTang:
@justingrant @sffc @ptomato @Ms2ger @ryzokuken

FrankYFTang:
sorry, I think this is due to my misunderstanding

PR #1629, opened 2021-07-10
Speed up TimeZone and ZonedDateTime code and tests

justingrant:
Creating `Intl.DateTimeFormat` instances in V8 is slow and memory heavy. `ES.GetFormatterParts` and `ES.GetCanonicalTimeZoneIdentifier` are functions that are called many times when using Temporal, and they used to create new instances of Intl.DateTimeFormat for each call. In this commit, we cache them using the time zone identifier as the key.

It should be noted that doing the same to SystemTimeZone was avoided. This is due to the fact that user's time zone may change during the execution of a program. An example: `Temporal.now.zonedDateTimeISO()` should always output the correct time zone. This shouldn't be a problem for server-side code that usually doesn't (or rather, shouldn't) use the time zone from the environment for calculations.

This PR was ported from https://github.com/js-temporal/temporal-polyfill/pull/10.

### DIFF
--- a/polyfill/lib/ecmascript.mjs
+++ b/polyfill/lib/ecmascript.mjs
@@ -148,6 +148,27 @@ const ES2020 = {
   Type
 };
 
+const IntlDateTimeFormatEnUsCache = new Map();
+
+function getIntlDateTimeFormatEnUsForTimeZone(timeZoneIdentifier) {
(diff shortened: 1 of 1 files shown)


ptomato:
This speeds up the tests which are painfully slow!

ptomato (approved):

PR #1633, opened 2021-07-12
Editorial: communicate infallable time zone ops

jugglinmike:


spec/temporal.html
```diff
@@ -79,7 +79,7 @@ <h1>Temporal.now.timeZone ( )</h1>
         The following steps are taken:
       </p>
       <emu-alg>
-        1. Return ? SystemTimeZone().
+        1. Return SystemTimeZone().
```
ljharb, line 82:
```suggestion
        1. Return ! SystemTimeZone().
```

spec/temporal.html
```diff
@@ -221,7 +221,7 @@ <h1>SystemInstant ( )</h1>
       <h1>SystemDateTime ( _temporalTimeZoneLike_, _calendarLike_ )</h1>
       <emu-alg>
         1. If _temporalTimeZoneLike_ is *undefined*, then
-          1. Let _timeZone_ be ? SystemTimeZone().
+          1. Let _timeZone_ be SystemTimeZone().
```
ljharb, line 224:
```suggestion
          1. Let _timeZone_ be ! SystemTimeZone().
```

spec/temporal.html
```diff
@@ -234,7 +234,7 @@ <h1>SystemDateTime ( _temporalTimeZoneLike_, _calendarLike_ )</h1>
       <h1>SystemZonedDateTime ( _temporalTimeZoneLike_, _calendarLike_ )</h1>
       <emu-alg>
         1. If _temporalTimeZoneLike_ is *undefined*, then
-          1. Let _timeZone_ be ? SystemTimeZone().
+          1. Let _timeZone_ be SystemTimeZone().
```
ljharb, line 237:
```suggestion
          1. Let _timeZone_ be ! SystemTimeZone().
```

ljharb (changes requested):
unless the completion record is introspected, a call to a non-predicate AO should always have a `?` or `!`.

ptomato:
Thanks! My understanding of the notation matches Jordan's. Otherwise this looks correct.

jugglinmike:
@ptomato you got it. Bear in mind that [applying `!` in this way may not be supported by the editors](https://github.com/tc39/proposal-resizablearraybuffer/pull/62#issuecomment-877338412).

ljharb:
@jugglinmike before the planned "completion reform", applying it in this way is the rule. The editors aren't a fan, but so far have preferred consistency pending an overhaul of all AOs.

ptomato:
Like most things I've encountered in TC39, I wasn't aware there was a controversy about it. Nonetheless, being consistent until the editors provide guidance seems fine to me.

ptomato (approved):

PR #1638, opened 2021-07-14
Normative: Throw Range in PrepareTemporalFields for +∞ or -∞

FrankYFTang:
Fix one of the issue in https://github.com/tc39/proposal-temporal/issues/1467

@ptomato @justingrant @Ms2ger


spec/abstractops.html
```diff
@@ -1446,6 +1446,7 @@ <h1>PrepareTemporalFields ( _fields_, _fieldNames_, _requiredFields_ )</h1>
           1. If _property_ is in the Property column of <emu-xref href="#table-temporal-field-requirements"></emu-xref> and there is a Conversion value in the same row, then
             1. Let _Conversion_ represent the abstract operation named by the Conversion value of the same row.
             1. Set _value_ to ? _Conversion_(_value_).
+            1. If _value_ is +∞ or -∞, throw a *RangeError* exception.
```
ljharb, line 1449: would an alternative approach be to make an AO in the Conversion table that throws when it results in an infinity, and then this line wouldn't be needed?

FrankYFTang: Good Idea! I add a new AO ToInteger and rename ToPositiveIntegerOrInfinite to ToPositiveInteger and change all the places which call them and throw on +/- Infinity. I didn't touch anywhere else which does not throw after calling them.

FrankYFTang: PTAL

FrankYFTang:
This PR has normative change to PrepareTemporalFields and PreparePartialTemporalFields but also a lot of editorial changes on other part of the spec which simplfy the spec by using the new AO (ToInteger)

FrankYFTang:
or should we rename the new AO ToIntegerThrowOnInfinity instead of ToInteger so it won't got confused with the old ToInteger in ECMA262 before?

ljharb:
Renaming sounds like a great idea.

FrankYFTang:
> Renaming sounds like a great idea.

ok, let me take some time to change it later today

Ms2ger:
Fwiw, I used "ToFiniteInteger" in the polyfill.

Ms2ger:
I would also recommend adding the new abstract operation and updating the existing call sites that already throw, in a separate PR. That way we can land those changes before the normative change here.

FrankYFTang:
>Fwiw, I used "ToFiniteInteger" in the polyfill.

I don't like that name becaue it imply Integer could be Infinity while it is not.

> I would also recommend adding the new abstract operation and updating the existing call sites that already throw, in a separate PR. That way we can land those changes before the normative change here.

good point. I like that. I will change that.

ptomato:
@FrankYFTang Have you had time to split out the non-normative part from the normative part? If not, let me know and I can do it.

FrankYFTang:
> @FrankYFTang Have you had time to split out the non-normative part from the normative part? If not, let me know and I can do it.

Sorry, I took the whole week off last week for a personal issue and didn't read my email till now.  I can work on it coming week if it is not urgent.

ptomato:
Hi Frank, that's no problem. I hope everything is OK.

I'd like to get this done in plenty of time to put the normative part on the agenda for next TC39 meeting. If you can do it sometime this week, great. If not, it's no problem - just let me know so I can do it.

ptomato:
I split out the non-normative part into #1711.

ptomato:
I've added tests to this and based it on top of the editorial change in #1711. It should be ready to present for consensus now.

ptomato:

ptomato (approved):
:+1:

PR #1640, opened 2021-07-14
Normative: Wrap optional _options_ with [ ]

FrankYFTang:


ljharb (approved):

FrankYFTang:
try to fix https://github.com/tc39/proposal-temporal/issues/1639

spec/duration.html
```diff
@@ -448,7 +448,7 @@ <h1>Temporal.Duration.prototype.round ( _options_ )</h1>
     </emu-clause>
 
     <emu-clause id="sec-temporal.duration.prototype.total">
-      <h1>Temporal.Duration.prototype.total ( _options_ )</h1>
+      <h1>Temporal.Duration.prototype.total ( [ _options_ ] )</h1>
```
ptomato, line 451: This one is not optional because the `unit` is required. (Consider adding an `undefined` -> TypeError check in #1646?)

ptomato (changes requested):
This changes the value of the `length` property of the affected functions, could you modify the appropriate tests as well to reflect the new `length` value?

ptomato:
(Converting normative PR to draft so it is not accidentally merged without consensus)

ptomato:
I added tests to this, it should be ready to present for consensus now.

ljharb (approved):

ptomato (approved):

ptomato:

PR #1642, opened 2021-07-14
Editorial: document infallability of AO invocation

jugglinmike:
CreateTemporalCalendar cannot return an abrupt completion record when
invoked with a single parameter.

### DIFF
--- a/spec/calendar.html
+++ b/spec/calendar.html
@@ -247,7 +247,7 @@
         1. Let _identifier_ be ? ToString(_temporalCalendarLike_).
         1. If ! IsBuiltinCalendar(_identifier_) is *false*, then
           1. Let _identifier_ be ? ParseTemporalCalendarString(_identifier_).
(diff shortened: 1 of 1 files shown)


ljharb (approved):

Ms2ger (changes requested):
I'm generally not convinced that OrdinaryCreateFromConstructor is infallible. If you are, could you share your reasoning?

jugglinmike:
@Ms2ger I'm generally not convinced of that, either. For the specific invocation in question, though, OrdinaryCreateFromConstructor can't return an abrupt completion because it's being called with a well-known intrinsic object that is a constructor.

jugglinmike:
@Ms2ger It's been a little while, so I'm commenting here to make sure this patch didn't slip off your radar.

Ms2ger (approved):
ok

PR #1645, opened 2021-07-15
Rename Temporal.now => Temporal.Now

justingrant:
Fixes #1583 per today's consensus decision in TC39 plenary meeting.

Note that this PR does *not* include a `toStringTag` implementation.  See #1539.


ljharb (approved):
i only reviewed the one spec file; while i despise the capitalized form, this PR seems correct to me.

polyfill/test/exports.mjs
```diff
@@ -52,7 +52,7 @@ describe('Exports', () => {
     assert(named.includes('Calendar'));
   });
   it('should contain `now`', () => {
```
Ms2ger, line 54:
```suggestion
  it('should contain `Now`', () => {
```

polyfill/test/now/instant/length.js
```diff
@@ -16,7 +16,7 @@ includes: [propertyHelper.js]
 features: [Temporal]
```
Ms2ger, line 16: Should move this entire folder to `polyfill/test/Now/`, but do it separately, please.

spec/temporal.html
```diff
@@ -1,5 +1,5 @@
-<!doctype html>
-<meta charset="utf8">
+<!DOCTYPE html>
+<meta charset="utf8" />
```
Ms2ger, line 2:
```suggestion
<meta charset="utf8">
```

spec/temporal.html
```diff
@@ -71,7 +71,7 @@ <h1>The Temporal.now Object</h1>
   </ul>
 
   <emu-clause id="sec-function-properties-of-the-temporal-now-object">
-    <h1>Function Properties of the Temporal.now Object</h1>
+    <h1>Function Properties of the Temporal.Now Object</h1>
 
     <emu-clause id="sec-temporal.now.timezone">
       <h1>Temporal.now.timeZone ( )</h1>
```
Ms2ger, line 77:
```suggestion
      <h1>Temporal.Now.timeZone ( )</h1>
```

And you're missing a couple more below.

Ms2ger (changes requested):

spec/temporal.html
justingrant: Suggested revert is fine with me because the missing slash didn't break anything before, but this change and the line above it was automatically made by prettier. Is prettier wrong here?

ljharb: the trailing slash isn't required in html5, but personally i find it a much better style to always include it on a self-closing element.

spec/temporal.html
justingrant: Haha, this was me trying to manually undo the carnage caused by prettier. Clearly I was a bit too eager in undoing and missed a few valid changes!

For some reason I can't get the same overeager prettier behavior to reproduce anymore. Might be that my prettier plugin was auto-updated and I didn't read the release notes last time. ;-)

image

polyfill/test/now/instant/length.js
justingrant: I'll assume that a separate commit in this PR is OK. LMK if not. Also was fun to to learn how to rename things case-sensitively in git, because of course it didn't work the easy way.

spec/temporal.html
Ms2ger: It's meaningless cruft, and inconsistent with the other files.

Ms2ger (approved):

PR #1650, opened 2021-07-16
Fix division issues caught by tsc in RoundDuration

justingrant:
Port of https://github.com/js-temporal/temporal-polyfill/pull/14

### DIFF
--- a/polyfill/lib/ecmascript.mjs
+++ b/polyfill/lib/ecmascript.mjs
@@ -3898,8 +3898,8 @@ export const ES = ObjectAssign({}, ES2020, {
         oneYearDays = MathAbs(oneYearDays);
         const divisor = bigInt(oneYearDays).multiply(dayLengthNs);
         nanoseconds = divisor.multiply(years).plus(bigInt(days).multiply(dayLengthNs)).plus(nanoseconds);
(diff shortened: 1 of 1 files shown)


Ms2ger:
Are those testable?

justingrant:
> Are those testable?

Don't know. It was @12wrigja's PR over in the temporal-polyfill repo. James, is this testable?

12wrigja:
Apologies - I don't know. I'm not yet familiar enough with the Temporal codebase to know why seemingly this codepath was not exercised before? Maybe there should be a 262 test for this?

 Looking over the big-integer docs it seems that using native math operators causes the bigInt valueOf to be called before the math operation, so I'd assume that a relevant testcase here would be one that has sufficiently large numbers involved in these operations that some loss of precision was happening before. I think if `RoundDuration` was called with a sufficiently large value for `increment` then there would be some loss of precision at least in the multiplication step, but whether it ends up being meaningful I'm unsure.

I'm about to be OOO till the 26th, but at that point I'll return back to this if nobody else has and see if there's a way to write up a testcase for this change.

Ms2ger (approved):
Thanks for considering, I guess we can just merge already.

PR #1651, opened 2021-07-16
Speed up toLocaleString ~2.5x by optimizing creation of Intl.DateTimeFormat instances

justingrant:
Port of https://github.com/js-temporal/temporal-polyfill/pull/12

### DIFF
--- a/polyfill/test/intl.mjs
+++ b/polyfill/test/intl.mjs
@@ -1050,10 +1050,43 @@ describe('Intl', () => {
       it('should return an Array', () => assert(Array.isArray(Intl.DateTimeFormat.supportedLocalesOf())));
     });
 
-    const us = new Intl.DateTimeFormat('en-US', { timeZone: 'America/New_York' });
-    const at = new Intl.DateTimeFormat('de-AT', { timeZone: 'Europe/Vienna' });
+    // Verify that inputs to DateTimeFormat constructor are immune to mutation.
+    // Also verify that options properties are only read once.
+    const onlyOnce = (value) => {
+      const obj = {
+        calls: 0,
+        toString() {
+          if (++this.calls > 1) throw new RangeError('prop read twice');
+          return value;
+        }
+      };
+      return obj;
+    };
+    const optionsAT = {
+      timeZone: onlyOnce('Europe/Vienna')
+    };
+    const optionsUS = {
+      calls: 0,
+      value: 'America/New_York',
+      get timeZone() {
+        if (++this.calls > 1) throw new RangeError('prop read twice');
+        return this.value;
+      },
+      set timeZone(val) {
+        this.value = val;
+      }
+    };
+    const localesAT = ['de-AT'];
+    const us = new Intl.DateTimeFormat('en-US', optionsUS);
+    const at = new Intl.DateTimeFormat(localesAT, optionsAT);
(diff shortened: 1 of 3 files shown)


justingrant:
This PR wraps up a group of three DateTimeFormat optimization PRs. (The others were #1624 and #1629.)

Two weeks ago, demitasse tests took 46s-49s in CI. This PR only took 32s... 30% less time!

polyfill/package.json
```diff
@@ -8,7 +8,7 @@
   "types": "index.d.ts",
   "scripts": {
     "coverage": "c8 report --reporter html",
-    "test": "node --no-warnings --experimental-modules --icu-data-dir node_modules/full-icu --loader ./test/resolve.source.mjs ./test/all.mjs",
+    "test": "time node --no-warnings --experimental-modules --icu-data-dir node_modules/full-icu --loader ./test/resolve.source.mjs ./test/all.mjs",
```
Ms2ger, line 11: Revert

polyfill/test/intl.mjs
```diff
@@ -1050,10 +1050,47 @@ describe('Intl', () => {
...
+    const onlyOnce = (value) => {
+      const obj = {
+        calls: 0,
+        toString() {
+          if (++this.calls > 1) throw new RangeError('prop read twice');
+          return value;
+        },
+        toJSON() {
+          if (++this.calls > 1) throw new RangeError('prop read twice');
+          return value;
```
Ms2ger, line 1064: Should this ever be called?

Ms2ger (approved):

polyfill/test/intl.mjs
justingrant: Good catch. An early version of this PR used a JSON round-trip as a cheap way to do a deep copy of the input parameters, so this was previously useful. Now it's implemented differently so this `toJSON` test code will never run. I'll remove it, unless you think there's value in leaving it around in case some future implementation decides to use `JSON.stringify` on the inputs.

Ms2ger: I'm happy to keep it and add `valueOf` as well, or just drop it.

polyfill/package.json
justingrant: Will do. If OK with you, I'd like to leave it in the other repo. Especially as we start attracting new contributors over there, I'd like to instill awareness of perf regressions from everyone contributing code over there. Unless it hurts something?

polyfill/test/intl.mjs
justingrant: I just dropped it for simplicity. We already have to revisit this per https://github.com/js-temporal/temporal-polyfill/pull/12#discussion_r670048689 so any improvement in multi-call/proxy-calll tests could happen then.

I think we're good to merge. It's pretty late for me so I'm going to bed soon. If I can't stay up until the Test262 run finishes, I'll merge in the morning if you don't do it first.  Then I can go on vacation! :-)

Ms2ger (approved):

polyfill/package.json
Ms2ger: I don't suppose it hurts, but I'm slowly removing these tests anyway, so their runtime doesn't matter much here. If you want to think about ways to speed up the 262 tests, that'd help me more :)

PR #1656, opened 2021-07-20
Meta: bump ecmarkup to 8.0.1

bakkot:
This includes a fix for https://github.com/tc39/ecmarkup/issues/333, which should make the table of contents work correctly for sections such as [this one](https://tc39.es/proposal-temporal/#sup-time-zone-names), where the header of the section contains a link.


ptomato (approved):
Thanks for fixing that, those links were quite confusing when clicking through the table of contents.

PR #1665, opened 2021-07-21
Normative: Fix typo in DifferenceISODate

FrankYFTang:
Avoid Tautology in step 2.p in DifferenceISODate #1649

@ptomato @Ms2ger @justingrant @ryzokuken

### DIFF
--- a/spec/plaindate.html
+++ b/spec/plaindate.html
@@ -732,7 +732,7 @@
             1. Set _mid_ be ! AddISODate(_y1_, _m1_, _d1_, _years_, _months_, 0, 0, *"constrain"*).
             1. Let _midSign_ be -(! CompareISODate(_mid_.[[Year]], _mid_.[[Month]], _mid_.[[Day]], _y2_, _m2_, _d2_)).
           1. Let _days_ be 0.
-          1. If _mid_.[[Month]] is equal to _end_.[[Month]] and _mid_.[[Year]] is equal to _mid_.[[Year]], set _days_ to _end_.[[Day]] - _mid_.[[Day]].
+          1. If _mid_.[[Month]] is equal to _end_.[[Month]] and _mid_.[[Year]] is equal to _end_.[[Year]], set _days_ to _end_.[[Day]] - _mid_.[[Day]].
           1. Else if _sign_ &lt; 0, set _days_ to -_mid_.[[Day]] - (! ISODaysInMonth(_end_.[[Year]], _end_.[[Month]]) - _end_.[[Day]]).
           1. Else, set _days_ to _end_.[[Day]] + (! ISODaysInMonth(_mid_.[[Year]], _mid_.[[Month]]) - _mid_.[[Day]]).
           1. If _largestUnit_ is *"month"*, then


ptomato (approved):
If it's possible to have a test that would catch the difference between these, that would be great.

ptomato:
(Converting to draft so that normative change is not accidentally merged without consensus)

ptomato:
I have come to think this change is not observable, because the condition "_mid_.[[Year]] is equal to _end_.[[Year]]" is never false if "_mid_.[[Month]] is equal to _end_.[[Month]]" is true. By the time we get to step 2.p of [DifferenceISODate](https://tc39.es/proposal-temporal/#sec-temporal-differenceisodate), _mid_ is a date in either in the same month as _end_, or in the previous month, or the next month. It's never a nonzero multiple of 12 months apart (which would be the only situation in which _mid_.[[Month]] equals _end_.[[Month]] but _mid_.[[Year]] does not equal _end_.[[Year]]). So I think we can just change this line to

p. If _mid_.[[Month]] is equal to _end_.[[Month]], set _days_ to _end_.[[Day]] - _mid_.[[Day]].

as an editorial change. What do you think?

FrankYFTang:
> I have come to think this change is not observable, because the condition "_mid_.[[Year]] is equal to _end_.[[Year]]" is never false if "_mid_.[[Month]] is equal to _end_.[[Month]]" is true. By the time we get to step 2.p of [DifferenceISODate](https://tc39.es/proposal-temporal/#sec-temporal-differenceisodate), _mid_ is a date in either in the same month as _end_, or in the previous month, or the next month. It's never a nonzero multiple of 12 months apart (which would be the only situation in which _mid_.[[Month]] equals _end_.[[Month]] but _mid_.[[Year]] does not equal _end_.[[Year]]). So I think we can just change this line to
> 
> p. If _mid_.[[Month]] is equal to _end_.[[Month]], set _days_ to _end_.[[Day]] - _mid_.[[Day]].
> 
> as an editorial change. What do you think?

I was just trying to fix a typo and I have hard time to follow the math here. I am not disagree w/ you about that, but that is more than a simple typo change than I can follow now. Maybe get someone else to take another look? 

are you saying _mid_.[[Year]] is always the same as _end_.[[Year]] everything the code pass here?

ptomato:
What I think, after examining the algorithm, is that if _mid_.[[Year]] ≠ _end_.[[Year]], then _mid_.[[Month]] ≠ _end_.[[Month]] in every case, and if _mid_.[[Month]] = _end_.[[Month]] then _mid_.[[Year]] = _end_.[[Year]] in every case.

I came to suspect this because I was trying to write a test that would produce an incorrect result if the implementation followed the spec text as written, and I didn't manage to. Of course, that's not a guarantee that one doesn't exist, so someone can trivially prove me wrong by writing such a test :smile:

FrankYFTang:
> What I think, after examining the algorithm, is that if _mid_.[[Year]] ≠ _end_.[[Year]], then _mid_.[[Month]] ≠ _end_.[[Month]] in every case, and if _mid_.[[Month]] = _end_.[[Month]] then _mid_.[[Year]] = _end_.[[Year]] in every case.
> 
> I came to suspect this because I was trying to write a test that would produce an incorrect result if the implementation followed the spec text as written, and I didn't manage to. Of course, that's not a guarantee that one doesn't exist, so someone can trivially prove me wrong by writing such a test 😄

then, we might need at least add an Assert

ptomato:
I agree.

ptomato:
I've done this and added an Assert as you suggested in https://github.com/tc39/proposal-temporal/pull/1717

PR #1680, opened 2021-07-30
Drop some old-style structure tests.

Ms2ger:
Everything is covered in test262.


ptomato (approved):
es-abstract upgrade is unrelated, but was required to make CI work.

PR #1688, opened 2021-08-05
Normative: Optimize calls into user code in InterpretISODateTimeOffset

ptomato:
In the current spec text algorithm, there was an unnecessary call to
timeZone.getPossibleInstantsFor(), which potentially calls observably into
user code. Fix the algorithm so it doesn't make any more calls than
necessary.


ptomato:
Marking as draft so that the normative change isn't merged accidentally.

Ms2ger (approved):

PR #1691, opened 2021-08-05
Various editorial and tests changes

ptomato:


spec/abstractops.html
```diff
@@ -108,17 +108,14 @@ <h1>DurationHandleFractions ( _fHours_, _minutes_, _fMinutes_, _seconds_, _milli
         1. Set _seconds_ to floor(_secs_).
         1. Let _fSeconds_ be _secs_ modulo 1.
         1. If _fSeconds_ is not equal to 0, then
-          1. If any of _milliseconds_, _microseconds_, _nanoseconds_ is not 0, throw a *RangeError* exception.
```
ljharb, line 111: is this not normative?

ptomato: I don't believe so — it's dead code, these checks are already performed in step 2a, and the variables are not changed in the meantime.

polyfill/lib/calendar.mjs
```diff
@@ -221,8 +221,8 @@ impl['iso8601'] = {
     const { month, monthCode, ...original } = fields;
     const { month: newMonth, monthCode: newMonthCode } = additionalFields;
     if (newMonth === undefined && newMonthCode === undefined) {
-      original.month = month;
-      original.monthCode = monthCode;
+      if (month !== undefined) original.month = month;
+      if (monthCode !== undefined) original.monthCode = monthCode;
     }
     return { ...original, ...additionalFields };
```
Ms2ger, line 227: This still doesn't seem to quite match the spec, but this is better, at least.

Ms2ger (approved):

spec/timezone.html
```diff
@@ -528,7 +528,7 @@ <h1>FormatTimeZoneOffsetString ( _offsetNanoseconds_ )</h1>
       <emu-alg>
         1. Assert: _offsetNanoseconds_ is an integer.
         1. If _offsetNanoseconds_ ≥ 0, let _sign_ be *"+"*; otherwise, let _sign_ be *"-"*.
-        1. Let _nanoseconds_ be abs(_offsetNanoseconds_) modulo 10<sup>9</sup>.
+        1. Let _nanoseconds_ be _offsetNanoseconds_ modulo 10<sup>9</sup>.
```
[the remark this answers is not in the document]
ptomato: Agreed, I was thinking the wrong way about the definition of modulo. These two commits should just be removed.

polyfill/lib/calendar.mjs
ptomato: I've made a better corresponding version, let me know what you think.

Ms2ger (approved):
Thanks!

PR #1720, opened 2021-08-12
Normative: Throw TypeError on missing options from Duration.total

ptomato:
See also #1646, #1640

### DIFF
--- a/spec/duration.html
+++ b/spec/duration.html
@@ -461,6 +461,7 @@
       <emu-alg>
         1. Let _duration_ be the *this* value.
         1. Perform ? RequireInternalSlot(_duration_, [[InitializedTemporalDuration]]).
+        1. If _options_ is *undefined*, throw a *TypeError* exception.
         1. Set _options_ to ? GetOptionsObject(_options_).
         1. Let _relativeTo_ be ? ToRelativeTemporalObject(_options_).
         1. Let _unit_ be ? ToTemporalDurationTotalUnit(_options_).
--- a/polyfill/lib/duration.mjs
+++ b/polyfill/lib/duration.mjs
@@ -398,6 +398,7 @@ export class Duration {
     let microseconds = GetSlot(this, MICROSECONDS);
     let nanoseconds = GetSlot(this, NANOSECONDS);
 
+    if (options === undefined) throw new TypeError('options argument is required');
     options = ES.GetOptionsObject(options);
     const unit = ES.ToTemporalDurationTotalUnit(options, undefined);
     if (unit === undefined) throw new RangeError('unit option is required');
--- a/polyfill/test/Duration/prototype/total/options-wrong-type.js
+++ b/polyfill/test/Duration/prototype/total/options-wrong-type.js
@@ -1,0 +1,24 @@
+// Copyright (C) 2021 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.duration.prototype.total
+description: TypeError thrown when options argument is missing or a primitive
+features: [BigInt, Symbol, Temporal]
+---*/
+
+const values = [
+  undefined,
+  null,
+  true,
+  "string",
+  Symbol(),
+  1,
+  2n,
+];
+
+const instance = new Temporal.Duration(0, 0, 0, 0, 1);
+assert.throws(TypeError, () => instance.total(), "TypeError on missing argument");
+values.forEach((value) => {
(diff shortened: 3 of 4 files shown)


spec/duration.html
```diff
@@ -461,6 +461,7 @@ <h1>Temporal.Duration.prototype.total ( _options_ )</h1>
       <emu-alg>
         1. Let _duration_ be the *this* value.
         1. Perform ? RequireInternalSlot(_duration_, [[InitializedTemporalDuration]]).
+        1. If _options_ is *undefined*, throw a *TypeError* exception.
```
ljharb, line 464: why is an options object required here? Generally speaking I'd expect an options object to always be optional, and required things to be named positional arguments.

ptomato: That's how it was originally (see #789) but later was changed (https://github.com/tc39/proposal-temporal/issues/584#issuecomment-706387651). Unfortunately it looks like I missed the reason in https://github.com/tc39/proposal-temporal/blob/main/meetings/agenda-minutes-2020-10-09.md, but IIRC discussing that it was so that `.total({ unit, relativeTo })` was consistent with `.round({ smallestUnit, largestUnit, relativeTo })`. @justingrant Do you remember more clearly?

justingrant: IIRC, the reason is the same as for `round`: there's no sensible default. `seconds` might be assumed by some users, `milliseconds` by others, etc. Instead of users getting confused about what unit `total()` would return, we opted to make the `unit` option required so that users must explicitly opt in to a unit.

We discussed making the `unit` property a separate parameter from `options` but this behavior would be different from the rest of Temporal, so IIRC we thought that requiring `options` was more consistent and less weird than splitting into a new parameter.

We could choose to do a normative change to set a default (both here and in `round`) to a particular unit, but short of doing that I don't see how we can make `options` optional in this method.

ljharb: I think that having a required thing in an options object is different enough from most of the rest of the ecosystem that it would be far better as a positional argument, since it's necessary for the method.

justingrant: If it were just this one method, then I'd be inclined to agree, but the `smallestUnit` option is also required for `round` methods for the same reason as `unit` is required in `total`.  In `round`, moving `smallestUnit` to a positional argument would mean that the shape of the same input data would be different between `round` and other methods where that option is used, e.g. `until` and `since`. It seems confusing that the same option performing the same operation would be passed positionally in one method but in `options` in another.

If we must change the currently spec-ed behavior, assigning a default unit for rounding (`second` seems most reasonable) would be something I would support much sooner than changing `round` and `total` to use required positional arguments.

ljharb: The shape should be different, because the methods have different requirements.

ptomato:
Normative pull request -> draft

spec/duration.html
ptomato: We discussed this in the Temporal champions meeting yesterday. The preference for keeping round() and total() as is was unanimous but we thought it might be good to rename the argument to 'properties' instead of 'options' so it didn't carry a connotation of optional, which I've done in #1755.

Either way, I don't plan to make this change in this particular pull request because this is about making the spec text consistent as designed. @ljharb if you want to propose this as a normative change, please open a new issue where we can discuss alternatives.

ljharb: The connotation exists because it's an object of properties, the specified name for the argument is irrelevant.

If the spec currently has this argument as optional, then it's fine. Making it required would mean it should be a positional argument. My understanding is that without this PR, it remains optional - is that accurate?

ptomato: No, this just changes the error that is thrown when it's omitted, from RangeError to TypeError.

ljharb: Then I'd consider it something that should have been a stage 3 blocker but was obviously missed in review.

Changing the error type is fine, go nuts; but let's discuss this in the upcoming plenary.

ptomato: Can you open an issue for it so we're not discussing it in the comments on a vaguely-related pull request? I see where you're coming from but I think the status quo makes the right tradeoff here, and so did the other meeting attendees. Ideally we could have a discussion in its own thread about what the alternatives are and their pros and cons, maybe distilling it into a more general principle along the way, and not have to spend plenary time on it. And if we did have to spend plenary time on it, I think a separate issue thread would probably be helpful for the participants.

ljharb: Sure! Filed https://github.com/tc39/proposal-temporal/issues/1756

ptomato: Thanks.

ptomato:

@Ms2ger Would you mind taking a final look?

Ms2ger (approved):

PR #1727, opened 2021-08-14
Add w CalculateOffsetShift  pass to AddZonedDateTime

FrankYFTang:
somehow CalculateOffsetShift miss the w arg and didn't pass it to CalculateOffsetShift

@ptomato @justingrant @jugglinmike @Ms2ger @ryzokuken @ljharb


ptomato (approved):
Thanks! The spec text before this change was not implementable as written, so I believe this is an editorial change.

PR #1739, opened 2021-08-17
Editorial: Use consistent wording for Records.

Ms2ger:


spec/duration.html
```diff
@@ -900,7 +900,7 @@ <h1>BalanceDuration ( _days_, _hours_, _minutes_, _seconds_, _milliseconds_, _mi
           1. Set _nanoseconds_ to _nanoseconds_ modulo 1000.
         1. Else,
           1. Assert: _largestUnit_ is *"nanosecond"*.
-        1. Return the new Record {
+        1. Return the Record {
```
ljharb, line 903: the spec tends to omit the article for returns, a la https://tc39.es/ecma262/#sec-createlistiteratorRecord

Ms2ger: - "Return Record...": 1 hit
- "Return the Record...": 9 hits
- "Let x be Record...": no hits
- "Let x be the Record...": 13 hits

ljharb: obviously "be Record" would be wildly incorrect english; i guess i just was used to seeing that one hit :-) "Return the Record" seems like the better choice then.

ptomato (approved):
Seems like "Return the Record" is the most consistent.

PR #1742, opened 2021-08-18
Editorial: Consistently use dateTime instead

FrankYFTang:
Most places use dateTime but some places use temporalDateTime referring to "this". Make it consistent.

@ptomato @Ms2ger


ptomato (approved):
OK, thanks. We originally decided to use _temporalDate_ instead of _date_ to avoid confusion with the `Date` object, but _dateTime_ doesn't have that problem.

PR #1743, opened 2021-08-18
Editorial: Consistenly use yearMonth

FrankYFTang:
only one plcae use plainYearMonth to refer to this, change it to alos use yearMonth

@ptomato @Ms2ger


ptomato (approved):
:+1:

PR #1744, opened 2021-08-18
Editoral: Consistenly use monthDay for this

FrankYFTang:
One place use plainMonthDay, change it to name monthDay to be consisent.

@ptomato @Ms2ger


ptomato (approved):
:+1:

PR #1741, opened 2021-08-18
Move some Instant test coverage to test262.

Ms2ger:

PR #1746, opened 2021-08-19
Fix code coverage on main.

Ms2ger:


Ms2ger:
Btw, I noticed that before #1741, all the demitasse test runs (on node 12, 14, and 16) were racing to submit their test coverage. I believe the reduced coverage in `isoToCalendarDate` on the code paths that exist for old chrome bugs is because one of the older nodes won the last race on main.

ptomato (approved):
Thanks for looking into this!

PR #1747, opened 2021-08-19
Editorial: Add missing check to ParseTemporalCalendarString.

Ms2ger:


ptomato (approved):
I think I have figured out how this is editorial and not normative. I had my doubts because the exception is currently thrown in step 3 of [`new Temporal.Calendar()`](https://tc39.es/proposal-temporal/#sec-temporal-createtemporalcalendar), and there's a chance for step 1 to throw a different exception there if NewTarget is undefined. But when called from ToTemporalCalendar → CreateTemporalCalendar, it's not possible for NewTarget to be undefined. So I think I agree.

PR #1748, opened 2021-08-19
Copy options object for PlainYearMonth.{add,subtract} and InterpretTemporalDateTimeFields

### LINKED ISSUE #1686: Ensure the same object is not passed twice into user code
While investigating #1685, I discovered that there are a few instances of the problem described in #1426 that I didn't catch.

- PlainYearMonth.{[add](https://tc39.es/proposal-temporal/#sec-temporal.plainyearmonth.prototype.add),[subtract](https://tc39.es/proposal-temporal/#sec-temporal.plainyearmonth.prototype.subtract)} passes an options object into calendar.dateAdd(), which may modify it, and then passes the same object to calendar.yearMonthFromFields().
- [InterpretTemporalDateTimeFields](https://tc39.es/proposal-temporal/#sec-temporal-interprettemporaldatetimefields) (called indirectly from PlainDateTime.with(), ZonedDateTime.with(), and any entry point that takes a PlainDateTime-like or ZonedDateTime-like property bag) passes an options object into calendar.dateFromFields(), which may modify it, and then reads the `overflow` property from that object.

As per the reasoning in #1426, these instances should be fixed. These are normative changes.

cjtenny:
… to prevent user-modified objects from interfering with later operations.

Fixes #1686

### DIFF
--- a/polyfill/test/PlainYearMonth/prototype/add/calendar-arguments.js
+++ b/polyfill/test/PlainYearMonth/prototype/add/calendar-arguments.js
@@ -3,26 +3,57 @@
 
 /*---
 esid: sec-temporal.plainyearmonth.prototype.add
+description: PlainYearMonth.prototype.add should respect calendar arguments and pass copied options objects.
 info: |
     YearMonthFromFields ( calendar, fields [ , options ] )
 
     5. Let yearMonth be ? Invoke(calendar, "yearMonthFromFields", « fields, options »).
-includes: [temporalHelpers.js]
+includes: [compareArray.js, temporalHelpers.js]
 features: [Temporal]
 ---*/
 
-const options = {};
+const actual = [];
+const expected = [
+  "get overflow",
+  "get overflow",
+  "get overflow.toString",
+  "call overflow.toString",
+  "get overflow.toString",
+  "call overflow.toString",
+];
+const options = new Proxy({ overflow: "constrain" }, {
+  get(target, key) {
+    actual.push(`get ${key}`);
+    const result = target[key];
+    if (result === undefined) {
+      return undefined;
+    }
+    return TemporalHelpers.toPrimitiveObserver(actual, result, key);
+  },
+  has(target, key) {
+    actual.push(`has ${key}`);
+    return key in target;
+  },
+});
+
 class CustomCalendar extends Temporal.Calendar {
   constructor() {
     super("iso8601");
   }
+  dateAdd(date, duration, options) {
+    const result = super.dateAdd(date, duration, options);
+    options.overflow = 'meatloaf';
+    return result;
+  }
   yearMonthFromFields(...args) {
     assert.sameValue(args.length, 2, "args.length");
     assert.sameValue(typeof args[0], "object", "args[0]");
-    assert.sameValue(args[1], options, "args[1]");
+    assert.notSameValue(args[1], options, "args[1]");
     return super.yearMonthFromFields(...args);
   }
 }
+
 const plainYearMonth = new Temporal.PlainYearMonth(2000, 3, new CustomCalendar());
-const result = plainYearMonth.add({ months: 2 }, options);
-TemporalHelpers.assertPlainYearMonth(result, 2000, 5, "M05");
+const result = plainYearMonth.add({ months: 10 }, options);
+TemporalHelpers.assertPlainYearMonth(result, 2001, 1, "M01");
+assert.compareArray(actual, expected, "copied options object order of operations");
--- a/spec/plaindatetime.html
+++ b/spec/plaindatetime.html
@@ -879,8 +879,8 @@
       </p>
       <emu-alg>
         1. Let _timeResult_ be ? ToTemporalTimeRecord(_fields_).
-        1. Let _temporalDate_ be ? DateFromFields(_calendar_, _fields_, _options_).
         1. Let _overflow_ be ? ToTemporalOverflow(_options_).
+        1. Let _temporalDate_ be ? DateFromFields(_calendar_, _fields_, _options_).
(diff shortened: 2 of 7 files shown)


spec/plaindatetime.html
```diff
@@ -879,8 +879,10 @@ <h1>InterpretTemporalDateTimeFields ( _calendar_, _fields_, _options_ )</h1>
       </p>
       <emu-alg>
         1. Let _timeResult_ be ? ToTemporalTimeRecord(_fields_).
-        1. Let _temporalDate_ be ? DateFromFields(_calendar_, _fields_, _options_).
         1. Let _overflow_ be ? ToTemporalOverflow(_options_).
+        1. Let _optionsCopy_ be ! OrdinaryObjectCreate(*null*).
+        1. Perform ! CreateDataPropertyOrThrow(_optionsCopy_, *"overflow"*, _overflow_).
+        1. Let _temporalDate_ be ? DateFromFields(_calendar_, _fields_, _optionsCopy_).
```
ljharb, line 885: This seems like it drops all the options - was this meant to create with a proto of options rather than null?

cjtenny: Good point, that makes good sense for this call site since `_overflow_` is the only point we don't want to observe modifications. Do you agree for the other PlainYearMonth call sites it makes sense to have a null proto and only pass overflow?

ljharb: I’m not sure - i assume that if it’s only passing specific options across, then those are the only options that are listened to - but it seems like user code would want maximum information available, meaning they’d want all options to be copied over.

cjtenny: Ah, I've been away a little while, I'd forgotten about ToPrimitive. I'll create a proper copy :)

ptomato: I think for this case it'd be better to just switch the order of steps 2 and 3, reading the overflow property before handing the object over to user code.

spec/plainyearmonth.html
```diff
@@ -269,9 +269,14 @@ <h1>Temporal.PlainYearMonth.prototype.add ( _temporalDurationLike_ [ , _options_
           1. Let _day_ be 1.
         1. Let _date_ be ? CreateTemporalDate(_yearMonth_.[[ISOYear]], _yearMonth_.[[ISOMonth]], _day_, _calendar_).
         1. Let _durationToAdd_ be ? CreateTemporalDuration(_duration_.[[Years]], _duration_.[[Months]], _duration_.[[Weeks]], _balanceResult_.[[Days]], 0, 0, 0, 0, 0, 0).
-        1. Let _addedDate_ be ? CalendarDateAdd(_calendar_, _date_, _durationToAdd_, _options_).
+        1. Let _optionsCopy_ be ! OrdinaryObjectCreate(*null*).
+        1. Let _overflow_ be ? ToTemporalOverflow(_options_).
+        1. Perform ! CreateDataPropertyOrThrow(_optionsCopy_, *"overflow"*, _overflow_).
+        1. Let _addedDate_ be ? CalendarDateAdd(_calendar_, _date_, _durationToAdd_, _optionsCopy_).
```
ptomato, line 275: For PYM.add and subtract, I think it'd be better to copy the whole options object (example in [MergeLargestUnitOption](https://tc39.es/proposal-temporal/#sec-temporal-mergelargestunitoption)) and pass the original to CalendarDateAdd and the copy to YearMonthFromFields.

spec/plaindatetime.html
ljharb: True, but it’d still be better if a mutable object wasn’t persisted through every call :-)

ptomato:
Marking the normative PR as draft to avoid accidental merging

spec/plaindatetime.html
cjtenny: only one call into user code in this case, though, so nbd! I'll go with the order swap.

ptomato:
This change achieved consensus in TC39. Can be merged when the wording issues are addressed and the tests are added.

ptomato (approved):
A minor comment on the tests.

polyfill/test/PlainYearMonth/prototype/add/calendar-arguments.js
```diff
@@ -7,19 +7,20 @@ info: |
     YearMonthFromFields ( calendar, fields [ , options ] )
 
     5. Let yearMonth be ? Invoke(calendar, "yearMonthFromFields", « fields, options »).
-includes: [temporalHelpers.js]
+includes: [deepEqual.js, temporalHelpers.js]
 features: [Temporal]
 ---*/
 
-const options = {};
+const options = { overflow: "constrain" };
```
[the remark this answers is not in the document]
cjtenny: will do both, and test that it still uses overflow properly in both paths

spec/plainyearmonth.html
```diff
@@ -269,9 +269,14 @@ <h1>Temporal.PlainYearMonth.prototype.add ( _temporalDurationLike_ [ , _options_
           1. Let _day_ be 1.
         1. Let _date_ be ? CreateTemporalDate(_yearMonth_.[[ISOYear]], _yearMonth_.[[ISOMonth]], _day_, _calendar_).
         1. Let _durationToAdd_ be ? CreateTemporalDuration(_duration_.[[Years]], _duration_.[[Months]], _duration_.[[Weeks]], _balanceResult_.[[Days]], 0, 0, 0, 0, 0, 0).
+        1. Let _optionsCopy_ be ! OrdinaryObjectCreate(%Object.prototype%).
+        1. Let _keys_ be ? EnumerableOwnPropertyNames(_options_, ~key~).
```
ljharb, line 273: Why not get both the key and entry here? Then the defining part won’t have to use ? Get.

cjtenny: Sounds fine by me, just didn't know that was an option. What's the ecmarkup syntax for iterating over the list returned by `EnumerableOwnPropertyNames(_options_, ~key+value~).`?

ljharb: The same, but you have to manually destructure the entry array

PR #1755, opened 2021-08-20
Editorial: Change the name of 'options' to 'properties' when not optional

ptomato:
See discussion in #1720.

### DIFF
--- a/docs/duration.md
+++ b/docs/duration.md
@@ -386,12 +386,12 @@ d = Temporal.Duration.from('-PT8H30M');
 d.abs(); // PT8H30M
 ```
 
-### duration.**round**(_options_: object) : Temporal.Duration
+### duration.**round**(_properties_: object) : Temporal.Duration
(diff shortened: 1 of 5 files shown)


ljharb:
I don't think the name of the argument is relevant - an object full of stuff is an options bag regardless of its name, and required things belong as positional arguments.

ljharb:
("properties" to me kinds of sounds like `['a', 'b']` - ie, an array of property names)

ptomato:
I agree, but I don't have an idea for a better name, do you have a suggestion?

(I do think that `options` is not the best name, so I think this should be changed regardless of the outcome of #1756. But I'm also happy to close this and roll it into whatever becomes the eventual solution of that issue.)

ptomato:
Superseded by #1875. I suggested there to change the terminology to "configuration object" which was suggested by Yulia during the plenary.

PR #1757, opened 2021-08-21
Champions meeting minutes

ptomato:


sffc (approved):

meetings/agenda-minutes-2021-08-05.md
```diff
@@ -0,0 +1,61 @@
...
+- USA: I talked to SFC a few days ago and his opinion was similar.
+
+### IETF discussion about conflicting TZDB versions
+- JGT: Would we be OK if IETF decides that the instant should "win" if there's a conflict with the TZID (e.g. new TZDB since the value was stored)? Background:
+there's some concerns on the IETF list about our current plan to throw exceptions if there's a conflict between the TZID and the offset. I suspect there may be enthusiasm on the list for having the timestamp "win". If that's the consensus of the IETF folks, is this something we could live with? Specifically, it'd mean a normative change to `ZonedDateTime.from()` from `{offset: 'reject'}` to `{offset: 'use'}`. Or we could choose to keep our reject default, but that would mean throwing for values that are accepted by other platforms that follow the new IETF spec. My impression is that the people on the SEDATE list are concerned about ambiguity. I described how our offset option works. One reply was that another option was to not allow any ambiguity at all. I don't agree with that, but I can see where they're coming from.
```
justingrant, line 51:
```suggestion
there's some concerns on the IETF list about our current plan to throw exceptions if there's a conflict between the TZID and the offset. I suspect there may be enthusiasm on the list for having the timestamp "win", which is unsurprising given that the current RFC 3339 spec is focused on timestamps alone and because prior art in Java's ZonedDateTime API has timestamp winning. Anyway, if "timestamp wins" is the consensus of the IETF folks, is this something we could live with? Specifically, it'd mean a normative change to `ZonedDateTime.from()` from a default of `{offset: 'reject'}` to a default of `{offset: 'use'}`. Or we could choose to keep our `reject` default, but that would mean throwing for values that are accepted by other platforms that follow the new IETF spec. My impression is that the people on the SEDATE list are concerned about ambiguity. I described how our offset option works. One reply was that another option was to not allow any ambiguity at all. I don't agree with that, but I can see where they're coming from.
```

justingrant (approved):
Minor wording changes in a suggestion.

PR #1771, opened 2021-08-24
Editoral: add "or *null* if no such transition exists"

### LINKED ISSUE #1764: specify return value of null in GetIANATimeZone*Transition
The caller of GetIANATimeZoneNextTransition and GetIANATimeZonePreviousTransition
https://tc39.es/proposal-temporal/#sec-temporal.timezone.prototype.getnexttransition
```
5. Let transition be ? GetIANATimeZoneNextTransition(startingPoint.[[Nanoseconds]], timeZone.[[Identifier]]).
6. If transition is null, return null.
```
https://tc39.es/proposal-temporal/#sec-temporal.timezone.prototype.getprevioustransition
```
5. Let transition be ? GetIANATimeZonePreviousTransition(startingPoint.[[Nanoseconds]], timeZone.[[Identifier]]).
6. If transition is null, return null.
```

However, the descriptions of these two AOs didn't mention they will return null but only mention it will return "an integer representing the number of nanoseconds....". I think we need to add more text to the description of these two AOs to specify what condition will return null .

https://tc39.es/proposal-temporal/#sec-temporal-getianatimezonenexttransition
11.6.6 GetIANATimeZoneNextTransition ( epochNanoseconds, timeZoneIdentifier )
The abstract operation GetIANATimeZoneNextTransition is an implementation-defined algorithm that returns an integer representing the number of nanoseconds since the Unix epoch in UTC that corresponds to the first time zone transition after epochNanoseconds in the IANA time zone identified by timeZoneIdentifier.

11.6.7 GetIANATimeZonePreviousTransition ( epochNanoseconds, timeZoneIdentifier )
The abstract operation GetIANATimeZonePreviousTransition is an implementation-defined algorithm that returns an integer representing the number of nanoseconds since the Unix epoch in UTC that corresponds to the last time zone transition before epochNanoseconds in the IANA time zone identified by timeZoneIdentifier.

@ptomato @sffc @ljharb @Ms2ger @anba @ryzokuken @justingrant

FrankYFTang:
Add " or *null* if no such transition exists" to GetIANATimeZoneNextTransition and GetIANATimeZonePreviousTransition

https://tc39.es/proposal-temporal/#sec-temporal-getianatimezonenexttransition

Fixes #1764

### DIFF
--- a/spec/timezone.html
+++ b/spec/timezone.html
@@ -484,14 +484,14 @@
     <emu-clause id="sec-temporal-getianatimezonenexttransition" aoid="GetIANATimeZoneNextTransition">
       <h1>GetIANATimeZoneNextTransition ( _epochNanoseconds_, _timeZoneIdentifier_ )</h1>
       <p>
-        The abstract operation GetIANATimeZoneNextTransition is an implementation-defined algorithm that returns an integer representing the number of nanoseconds since the Unix epoch in UTC that corresponds to the first time zone transition after _epochNanoseconds_ in the IANA time zone identified by _timeZoneIdentifier_.
+        The abstract operation GetIANATimeZoneNextTransition is an implementation-defined algorithm that returns an integer representing the number of nanoseconds since the Unix epoch in UTC that corresponds to the first time zone transition after _epochNanoseconds_ in the IANA time zone identified by _timeZoneIdentifier_ or *null* if no such transition exists.
       </p>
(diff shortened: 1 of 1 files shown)


FrankYFTang:
address https://github.com/tc39/proposal-temporal/issues/1764

Ms2ger (approved):

PR #1774, opened 2021-08-25
initialize whole if unit is "nanosecond" in Duration.total

FrankYFTang:
In Temporal.Duration.prototype.total
If unit is "nanosecond" currently the variable whole is uninitialized. Need to initialize it to the correct value.

https://tc39.es/proposal-temporal/#sec-temporal.duration.prototype.total

@ptomato @ljharb @Ms2ger @ryzokuken @justingrant


FrankYFTang:
I am not sure setting uninitialized value should be consider Editorial or Normative. The spec AS IS has undetermine behavior if whole is assign randomly in memory....

PR #1775, opened 2021-08-25
Editorial: use "modulo" operator

jugglinmike:
@ptomato as I understand it, this usage is not affected by the problem you are correcting with gh-1709. Does that sound right to you?


Ms2ger (approved):

ptomato:
Thanks! I would've missed this one in my search for "modulo" in the spec, but I believe it's not affected by the other problem I was fixing, because both operands are always positive.

PR #1777, opened 2021-08-25
Normative: Remove extra  NegateTemporalRoundingMode which revert the effect of the first one.

FrankYFTang:
Notice currently the NegateTemporalRoundingMode got called twice in 
https://tc39.es/proposal-temporal/#sec-temporal.plainyearmonth.prototype.since

```
12. Set roundingMode to ! NegateTemporalRoundingMode(roundingMode).
...
25. Set roundingMode to ! NegateTemporalRoundingMode(roundingMode).
```
and there are no code to use roundingMode between step 12 and step 25. I believe the one in step 25 is extra, which  "revert" the negation of the rounding mode in step 12.

@ptomato @justingrant @ljharb @ryzokuken @Ms2ger


ptomato:

PR #1788, opened 2021-08-29
Rename Temporal.now to Temporal.Now in Japanese docs

antonysiuming:
`Temporal.now` was changed to `Temporal.Now` in #1645.
The Japanese documentation added in #1668 is still using `Temporal.now`.

### DIFF
--- a/docs/ja/README.md
+++ b/docs/ja/README.md
@@ -31,18 +31,18 @@ Temporal の API では、タイムゾーンが関連付けられていない日
 
 ### **Temporal.now**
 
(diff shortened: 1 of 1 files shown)


ptomato (approved):
Thanks!
BTW, it looks like you might need to change it in `now.md` as well?

antonysiuming:
> Thanks!
> BTW, it looks like you might need to change it in `now.md` as well?

`now.md` is not yet translated to Japanese, but I found I had missed changing 2 sentences in this pull request. I may create another pull request later.

PR #1792, opened 2021-08-31
docs: add chinese translations

sxyazi:
I have translated a document into Chinese. I will continue this work in my free time :(


ptomato (approved):
Thanks!! :smile: 

Note that these docs will migrate to MDN in the future. We don't know when that will happen, possibly only when browsers ship Temporal.

PR #1806, opened 2021-09-07
Move some PlainMonthDay test coverage to test262.

Ms2ger:


Ms2ger:
Getting some
```
UNEXPECTED PASS Instant/prototype/toLocaleString/options-conflict.js (strict mode)
```

that I don't think are mine.

polyfill/test/PlainMonthDay/prototype/toString/calendarname-always.js
```diff
@@ -0,0 +1,22 @@
...
+description: If calendarName is "always", the calendar ID should be included.
+features: [Temporal]
+---*/
+
+const tests = [
+  [[], "05-02[u-ca=iso8601]"],
+  [[{ toString() { return "custom"; } }], "1972-05-02[u-ca=custom]"],
+  [[{ toString() { return "iso8601"; } }], "05-02[u-ca=iso8601]"],
+  [[{ toString() { return "ISO8601"; } }], "1972-05-02[u-ca=ISO8601]"],
+  [[{ toString() { return "\u0131so8601"; } }], "1972-05-02[u-ca=\u0131so8601]"], // dotless i
```
ptomato, line 15: This isn't a [valid calendar name](https://tc39.es/proposal-temporal/#prod-CalChar), so this ISO string won't deserialize back in — I can't remember off the top of my head if that was considered earlier as part of #665 and is intentional in a "well, don't do that, then" kind of way, or if it was just overlooked. Probably best not to have a test for this case. Maybe we should open an issue to check this decision as well?

polyfill/test/PlainMonthDay/prototype/with/options-invalid.js
```diff
@@ -0,0 +1,12 @@
+// Copyright (C) 2021 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.plainmonthday.prototype.with
```
ptomato, line 5: This will need a `description` when merging into test262 upstream

ptomato (approved):
Each test will need a `description` as that's a required field for upstream test262, otherwise looks good.

polyfill/test/PlainMonthDay/prototype/toString/calendarname-always.js
Ms2ger: Hrm, on the contrary, I think this makes it more important to have a test. I don't care what the outcome is, though - update the spec to check if the output is sensible or keep the current behavior and test. If you prefer, I can move it out of this PR, though.

ptomato: OK, let's keep this then.

ptomato:
> (Please let me hit the merge button.)

I think you have permission to? Wasn't it working?

PR #1810, opened 2021-09-08
Editorial: Mark TemporalTimeToString as infallible

linusg:
Currently defined as:

> 1. Assert: hour, minute, second, millisecond, microsecond and nanosecond are integers.
> 2. Let hour be hour formatted as a two-digit decimal number, padded to the left with a zero if necessary.
> 3. Let minute be minute formatted as a two-digit decimal number, padded to the left with a zero if necessary.
> 4. Let seconds be ! FormatSecondsStringPart(second, millisecond, microsecond, nanosecond, precision).
> 5. Return the string-concatenation of hour, the code unit 0x003A (COLON), minute, and seconds.

And clearly none of this can throw.


Ms2ger (approved):

Ms2ger:
Thanks!

PR #1823, opened 2021-09-16
Editorial: Return [[Calendar]] in ParseTemporalZonedDateTimeString

FrankYFTang:
Return result.[[Calendar]] in ParseTemporalZonedDateTimeString from 
```
3. Let result be ! ParseISODateTime(isoString).
```
so in ToTemporalZonedDateTime
step 4-i will have 
```
i. Let calendar be ? ToTemporalCalendarWithISODefault(result.[[Calendar]]).
```

Without this fix the spec won't work becuase result won't have [[Calendar]]

Address https://github.com/tc39/proposal-temporal/issues/1676

@ptomato @justingrant @Ms2ger @ryzokuken


ptomato (approved):
Agreed.

PR #1848, opened 2021-09-25
Treat System Epoch in nanoseconds as bigint.BigInteger.

12wrigja:
This is the same as https://github.com/js-temporal/temporal-polyfill/pull/58.


justingrant (changes requested):
See https://github.com/js-temporal/temporal-polyfill/pull/58/files#r716075108

12wrigja:
Updated with the outcome of the PR over in the js-temporal repo.

justingrant (approved):

ptomato (approved):
:+1:

PR #1854, opened 2021-09-30
Move some PlainDate tests.

Ms2ger:


ptomato (approved):
Looks good, but I would like to hold off on merging it because we have the PR open at test262 to move these tests upstream.

ptomato:
These will have to go to tc39/test262 now, we can merge the removal of the corresponding Demitasse tests afterwards.

Ms2ger:
https://github.com/tc39/test262/pull/3302

PR #1860, opened 2021-10-07
Normative: Fix precision 0 issue in TemporalDurationToString

FrankYFTang:
See https://github.com/tc39/proposal-temporal/issues/1859

@ptomato @ljharb @justingrant @ryzokuken @Ms2ger


ptomato (approved):
Thanks!

It sounds like there are already test262 tests in the tree that cover this.

I'll convert this to draft and present it at the upcoming TC39 meeting.

ptomato:

PR #1861, opened 2021-10-07
[polyfill] Match spec for PlainDate#since().

Ms2ger:
- [x] https://github.com/tc39/test262/issues/3249
- [ ] I wasn't sure how to fix `polyfill/test/datemath.mjs`; thoughts welcome.

### DIFF
--- a/polyfill/lib/plaindate.mjs
+++ b/polyfill/lib/plaindate.mjs
@@ -223,10 +223,10 @@ export class PlainDate {
     const roundingIncrement = ES.ToTemporalRoundingIncrement(options, undefined, false);
 
     const untilOptions = { ...options, largestUnit };
-    let { years, months, weeks, days } = ES.CalendarDateUntil(calendar, this, other, untilOptions);
+    let { years, months, weeks, days } = ES.CalendarDateUntil(calendar, other, this, untilOptions);
     const Duration = GetIntrinsic('%Temporal.Duration%');
     if (smallestUnit === 'day' && roundingIncrement === 1) {
-      return new Duration(-years, -months, -weeks, -days, 0, 0, 0, 0, 0, 0);
+      return new Duration(years, months, weeks, days, 0, 0, 0, 0, 0, 0);
     }
     const relativeTo = ES.CreateTemporalDateTime(
       GetSlot(this, ISO_YEAR),
@@ -241,10 +241,10 @@ export class PlainDate {
       GetSlot(this, CALENDAR)
     );
     ({ years, months, weeks, days } = ES.RoundDuration(
-      years,
-      months,
-      weeks,
-      days,
+      -years,
+      -months,
+      -weeks,
+      -days,
       0,
       0,
       0,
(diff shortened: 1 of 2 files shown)


ptomato:
> I wasn't sure how to fix `polyfill/test/datemath.mjs`; thoughts welcome.

Hmm, what exactly are the dates for which `two.subtract(dif).equals(one)` and `two.add(dif.negated()).equals(one)` no longer hold?

As for test262, probably need to update the submodule to the latest commit.

justingrant:
> Here is my understanding, when you calculate '2020-03-30').since('2020-02-28'), it first figure out there is 1 M from 2020-02-28 to 2020-03-28, and then figure out there are two day from 2020-03-28 to 2020-03-30 and therefore it is P1M2D

This is incorrect, at least according to my understanding of the algorithm (and hopefully incorrect according to the spec text too, but we need to verify that!). 

The problem is that you're starting from `other`, not from `this`. Both `until` and `since` should always start from `this`.  `until` should start from `this` and use `add` to get to `other`. `since` should start from `this` and use `subtract` to get to `other`. But in both cases, you need to start from `this`.  Your example above is starting from `other` which is why it gets the wrong result.

> and when you calculate '2020-03-30').subtract('P1M1D'), it first subtract 1M from '2020-03-30' and become 2020-02-30, but there are no such day so it change it to 2020-02-29 , and then it subtract a day and become 2020-02-28.

This is correct, and matches how `since` should behave also.

@FrankYFTang, when you're calling [`DifferenceISODate`](https://tc39.es/proposal-temporal/#sec-temporal-differenceisodate) from `since`, `y1, m1, d1` should come from `this`. Is that how your implementation currently works? If not, try reversing and let us know whether you get the expected results.

FrankYFTang:
> Is that how your implementation currently works? If not, try reversing and let us know whether you get the expected results.

What important is how the spec text currently said, not how my implementation currently works. If you follow the spec text, does it agree with your understanding.

justingrant:
I think the spec may have a problem. Specifically, there's only one `dateUntil` method on the Calendar/CalendarProtocol type. I think this means that there's no way for the Calendar implementation to know whether it's supposed to start from `this` or `other` when executing `dateUntil`. So I'm not sure how the current spec can correctly implement `since`.

@ptomato, could I get your opinion on this?  Am I missing something?

If I'm not missing something, then we'll need to figure out how to address this in the spec. Options include: 
* a) Adding a `dateSince` method to CalendarProtocol, which would be implemented identically to `dateUntil` except it'd reverse the arguments sent to DifferenceISODate.
* b) Dropping the `since` method from all Temporal types
* c) Changing the invariants associated with `since` and `until`, as described in https://github.com/tc39/proposal-temporal/pull/1861#issuecomment-946168234 and https://github.com/tc39/proposal-temporal/pull/1861#pullrequestreview-781221376

if we do have to make a change, my preference would be (a) because it would seem to be the least disruptive, because only Calendar implementations would have to change (Calendar authoring is an advanced use case so changes there affect relatively few developers) but no other Temporal-using code would be affected.

ptomato:
The Calendar.dateUntil doesn't have rounding, so it is reversible. The other types' until() and since() methods implement the rounding on top of that.

justingrant:
> I'll file an issue for the spec bug.

#1878

ptomato:
The spec was changed, so I think this PR no longer applies and can be closed.

PR #1865, opened 2021-10-08
Normative: Throw earlier in ZonedDateTime.with if offset field is missing

ptomato:
As discussed in https://github.com/tc39/proposal-temporal/pull/1762#discussion_r695948180

This changes ZonedDateTime.prototype.with() to throw slightly earlier
(before calling calendar methods in user code) in the unlikely case that
the offset field is messed up or missing.


ptomato:
This behaviour was previously not covered by tests, so new tests are in https://github.com/tc39/test262/pull/3293

PR #1866, opened 2021-10-11
[polyfill] PlainMonthDay#toPlainDate() rejects out-of-range dates.

Ms2ger:
Ref: tc39/test262#3252.


ptomato (approved):
Indeed, this matches the spec text.

ptomato:
Can't merge this until https://github.com/tc39/test262/pull/3274 is merged and we can update the git submodule to point to that.

PR #1881, opened 2021-10-22
Normative: Fix arg order bug in PD/PDT `since`

### LINKED ISSUE #1878: Spec for `since` of PlainDate & PlainDateTime incorrectly reverses arguments to `CalendarDateUntil`
The spec for `since` in for several Temporal types does not match the decision in https://github.com/tc39/proposal-temporal/issues/993#issuecomment-721889153. That decision says that `since` calculations should start with `this`, not `other`. Here's how all the Temporal types are spec-ed:
* `PlainYearMonth.p.since` seems to be spec-ed correctly. This spec text is a good model for fixing the others. 
* `PlainDate.p.since` has the parameter-reversal bug in Step 14 where `CalendarDateUntil` is called with `other` before `this`.
* `PlainDateTime.p.since` has the parameter-reversal bug in Step 14 where `DifferenceISODateTime` is called with `other` parameters before `this` parameters.
* `Instant.p.since` seems to be spec-ed correctly (but should validate this!). 
* `ZonedDateTime.p.since` seems to be spec-ed correctly (but should validate this!). 
* `PlainTime.p.since` seems to be spec-ed correctly (but should validate this!). 

Here's more info from https://github.com/tc39/proposal-temporal/pull/1861#issuecomment-946256137: (see discussion in that PR for more context)

> The spec for all the `since` methods is wrong. Instead, the spec should follow what the polyfill is doing, which is:
> 
> 1. Call CalendarDateUntil with `this` followed by `other` (not the other way around like the current spec)
> 2. Negate the result from CalendarDateUntil before returning it to the caller.
> 
> Here's an example from the polyfill showing this logic in practice:
> 
> ```js
> let { years, months, weeks, days } = ES.CalendarDateUntil(calendar, this, other, untilOptions);
> const Duration = GetIntrinsic('%Temporal.Duration%');
> if (smallestUnit === 'day' && roundingIncrement === 1) {
>   return new Duration(-years, -months, -weeks, -days, 0, 0, 0, 0, 0, 0);
> }
> const relativeTo = ES.CreateTemporalDateTime(
>   GetSlot(this, ISO_YEAR),
>   GetSlot(this, ISO_MONTH),
>   GetSlot(this, ISO_DAY),
>   0,
>   0,
>   0,
>   0,
>   0,
>   0,
>   GetSlot(this, CALENDAR)
> );
> ({ years, months, weeks, days } = ES.RoundDuration(
>   years,
>   months,
>   weeks,
>   days,
>   0,
>   0,
>   0,
>   0,
>   0,
>   0,
>   roundingIncrement,
>   smallestUnit,
>   ES.NegateTemporalRoundingMode(roundingMode),
>   relativeTo
> ));
> 
> return new Duration(-years, -months, -weeks, -days, 0, 0, 0, 0, 0, 0);
> ```

justingrant:
Fixes #1878 by ensuring that calculations in `since` start from `this` and proceed to `other`. The current spec has this reversed, which yields incorrect results in cases where intermediate values are constrained due to variable-length months.

The polyfill, docs, and tests were already working as expected, and this PR aligns the spec with that correct behavior by matching the corresponding logic in the polyfill.

Marking as draft until it can be discussed in a TC39 plenary meeting.


ptomato:

Tests are pending in https://github.com/tc39/test262/pull/3274 and I still plan to add additional ones to cover this behaviour for all the Plain types.

PR #1889, opened 2021-10-25
Normative: Fix TemporalDurationToString for negative duration

### LINKED ISSUE #1886: Is `TemporalDurationToString` using modulo correctly?
From https://tc39.es/proposal-temporal/#sec-temporal-temporaldurationtostring: 

```
1. Assert: precision is not "minute".
2. Set seconds to the mathematical value of seconds.
3. Set milliseconds to the mathematical value of milliseconds.
4. Set microseconds to the mathematical value of microseconds.
5. Set nanoseconds to the mathematical value of nanoseconds.
6. Let sign be ! DurationSign(years, months, weeks, days, hours, minutes, seconds, milliseconds, microseconds, nanoseconds).
7. Set microseconds to microseconds + the integral part of nanoseconds / 1000.
8. Set nanoseconds to nanoseconds modulo 1000.
9. Set milliseconds to milliseconds + the integral part of microseconds / 1000.
10. Set microseconds to microseconds modulo 1000.
11. Set seconds to seconds + the integral part of milliseconds / 1000.
12. Set milliseconds to milliseconds modulo 1000.
```

My understanding of [`x modulo y`](https://tc39.es/proposal-temporal/#eqn-modulo) is that its sign is solely dependent on the sign of `y`, which is always positive. If a duration is negative and has non-zero values in sub-second units, then I think we'd end up with off-by-one results because the (positive) `modulo` result would be the reverse of the sign of the negative slot(s) value. 

I'm not 100% sure I'm interpreting the spec correctly, but wanted to flag the issue.

justingrant:
Fixes #1886

### DIFF
--- a/spec/duration.html
+++ b/spec/duration.html
@@ -1474,11 +1474,11 @@
         1. Set _nanoseconds_ to the mathematical value of _nanoseconds_.
         1. Let _sign_ be ! DurationSign(_years_, _months_, _weeks_, _days_, _hours_, _minutes_, _seconds_, _milliseconds_, _microseconds_, _nanoseconds_).
         1. Set _microseconds_ to _microseconds_ + the integral part of _nanoseconds_ / 1000.
-        1. Set _nanoseconds_ to _nanoseconds_ modulo 1000.
+        1. Set _nanoseconds_ to remainder(_nanoseconds_, 1000).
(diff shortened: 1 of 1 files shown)


ptomato:

FrankYFTang:
Is there any test show the diff of this PR? I somehow see no change when I switch

ptomato:
It's on my to-do list to write tests that cover this.

Issue #1892, opened 2021-10-27
Unreachable code in ToTemporalZonedDateTime


FrankYFTang:
While I debug my v8 implementation against the following two tests I believe I found a unreachable code in ToTemporalZonedDateTime

https://github.com/tc39/test262/blob/main/test/built-ins/Temporal/ZonedDateTime/from/offset-undefined.js
https://github.com/tc39/test262/blob/main/test/built-ins/Temporal/ZonedDateTime/from/options-undefined.js

The problem is in step 2 of ToTemporalZonedDateTime
6.5.2 ToTemporalZonedDateTime ( item [ , options ] )
https://tc39.es/proposal-temporal/#sec-temporal-totemporalzoneddatetime
```
2. If Type(item) is Object, then
a. If item has an [[InitializedTemporalZonedDateTime]] internal slot, then
i. Return item.
b. Let calendar be ? GetTemporalCalendarWithISODefault(item).
c. Let fieldNames be ? CalendarFields(calendar, « "day", "hour", "microsecond", "millisecond", "minute", "month", "monthCode", "nanosecond", "second", "year" »).
d. Append "timeZone" to fieldNames.
e. Let fields be ? PrepareTemporalFields(item, fieldNames, « "timeZone" »).
f. Let timeZone be ? Get(fields, "timeZone").
g. Set timeZone to ? ToTemporalTimeZone(timeZone).
h. Let offsetString be ? Get(fields, "offset").
i. If offsetString is undefined, then
i. Set offsetBehaviour to wall.
j. Else,
i. Set offsetString to ? ToString(offsetString).

```
Notice after step 2-d fieldNames does not include 'offset' therefore
step 2-e Let fields be ? PrepareTemporalFields(item, fieldNames, « "timeZone" »).
 for ISO8601 calendar is 
2-e Let fields be ? PrepareTemporalFields(item, « "day", "hour", "microsecond", "millisecond", "minute", "month", "monthCode", "nanosecond", "second", "year", "timeZone"  », « "timeZone" »).

and therefore, after ste 2-3 fields does not contain "offset" even if item has  "offset" property. and therefore the IF test on step 2-i.
``` 
2-i If offsetString is undefined, then
```
will always be TRUE and step 2-i-i will always be execute and step 2-j and 2-j-i are dead code and will never be run.
```
2-i. If offsetString is undefined, then
2-i-i. Set offsetBehaviour to wall.
2-j. Else,
[...]

PR #1893, opened 2021-10-27
Normative: add "offset" to fieldNames in ToTemporalZonedDateTime

FrankYFTang:
Fix https://github.com/tc39/proposal-temporal/issues/1892 to avoid deadcode
Needed to fulfil 
https://github.com/tc39/test262/blob/main/test/built-ins/Temporal/ZonedDateTime/from/offset-undefined.js
https://github.com/tc39/test262/blob/main/test/built-ins/Temporal/ZonedDateTime/from/options-undefined.js

@pdunkel @ptomato @jugglinmike @Ms2ger @ljharb @justingrant


ptomato:

PR #1902, opened 2021-11-01
Handle other Node.js versions in localized calendar tests

ptomato:
This test would otherwise fail with Node.js 17.x because there wasn't a
snapshot key for `node17`. Add an `other` fallback key which is equal to
the latest `node16` snapshot. (If that later becomes incorrect, the
snapshots can be updated.)


justingrant (approved):
Look at you, Mr Living on the Bleeding Edge of Node! :-)

PR #1903, opened 2021-11-04
Editorial: Use "set" instead "let" when updating instant in ZonedDateTime#withPlainTime

Lubrsi:
On step 6 of this algorithm, instant is created with "Let instant be".
On step 10, it uses "Let instant be" again, when instant already exists.
This changes it to "Set instant to" to match withPlainDate.


ptomato (approved):
Thanks!

PR #1914, opened 2021-11-09
Editorial: Mark RoundTemporalInstant as infallible

Lubrsi:
Current `RoundTemporalInstant` definition:
```
1. Assert: Type(ns) is BigInt.
2. If unit is "hour", then
    a. Let incrementNs be increment × 3.6 × 1012.
3. Else if unit is "minute", then
    a. Let incrementNs be increment × 6 × 1010.
4. Else if unit is "second", then
    a. Let incrementNs be increment × 109.
5. Else if unit is "millisecond", then
    a. Let incrementNs be increment × 106.
6. Else if unit is "microsecond", then
    a. Let incrementNs be increment × 103.
7. Else,
    a. Assert: unit is "nanosecond".
    b. Let incrementNs be increment.
8. Return ! RoundNumberToIncrement(ℝ(ns), incrementNs, roundingMode).
```

Current `RoundNumberToIncrement` definition:
```
1. Assert: x and increment are mathematical values.
2. Assert: roundingMode is "ceil", "floor", "trunc", or "halfExpand".
3. Let quotient be x / increment.
4. If roundingMode is "ceil", then
    a. Let rounded be −floor(−quotient).
5. Else if roundingMode is "floor", then
    a. Let rounded be floor(quotient).
6. Else if roundingMode is "trunc", then
    a. Let rounded be the integral part of quotient, removing any fractional digits.
7. Else,
    a. Let rounded be ! RoundHalfAwayFromZero(quotient).
8. Return rounded × increment.
```

Current `RoundHalfAwayFromZero` definition:
```
1. Assert: x is a mathematical value.
2. Return the mathematical value that is closest to x and is an integer. If two integers are equally close to x, then the result is the integer that is farther away from 0. If x is already an integer, then the result is x.
```

### DIFF
--- a/spec/duration.html
+++ b/spec/duration.html
@@ -1427,7 +1427,7 @@
             [[Microseconds]]: _microseconds_,
             [[Nanoseconds]]: _nanoseconds_
             }.
-        1. Set _timeRemainderNs_ to ? RoundTemporalInstant(ℤ(_timeRemainderNs_ − _dayLengthNs_), _increment_, _unit_, _roundingMode_).
+        1. Set _timeRemainderNs_ to ! RoundTemporalInstant(ℤ(_timeRemainderNs_ − _dayLengthNs_), _increment_, _unit_, _roundingMode_).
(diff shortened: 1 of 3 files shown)


ryzokuken (approved):

ptomato (approved):
Thanks!

PR #1918, opened 2021-11-13
Editorial: Turn Temporal.Duration.prototype.total unit if branches into else-if's

linusg:
This is both more performant when implemented literally, and much more common across the spec in general when testing a single variable for multiple values. Additionally we can add an Assert to the final else branch to ensure that all possible values have been handled.

This now looks very similar to RoundTemporalInstant, RoundTime, and RoundDuration.


ptomato (approved):
Thanks!

PR #1920, opened 2021-11-14
Editorial: Restrict IsValidTimeZoneName case folding to ASCII

### LINKED ISSUE #1919: IsValidTimeZoneName "UTC" checking should limit case folding to ASCII
[IsValidTimeZoneName](https://tc39.es/proposal-temporal/#sec-isvalidtimezonename) currently uppercases the input "according to the Unicode Default Case Conversion algorithm", but it would be inappropriate for a hypothetical non-ASCII string that case-folds to "UTC" to be recognized as the UTC time zone.

gibson042:
Fixes #1919

### DIFF
--- a/spec/intl.html
+++ b/spec/intl.html
@@ -49,7 +49,8 @@
       </emu-alg>
     </emu-clause>
 
-    <emu-clause id="sup-canonicalizetimezonename" type="abstract operation">
+    <emu-clause id="sup-canonicalizetimezonename" aoid="CanonicalizeTimeZoneName402">
+      <h1><a href="https://tc39.es/ecma402/#sec-canonicalizetimezonename">CanonicalizeTimeZoneName</a> ( _timeZone_ )</h1>
       <h1>
         CanonicalizeTimeZoneName (
           _timeZone_: a String value that is a valid time zone name as verified by IsValidTimeZoneName,
@@ -1304,7 +1305,7 @@
       <emu-clause id="sup-temporal-calendar-abstract-ops">
         <h1>Abstract Operations for Temporal.Calendar Objects</h1>
 
-        <emu-clause id="sup-temporal-isbuiltincalendar">
+        <emu-clause id="sup-temporal-isbuiltincalendar" aoid="IsBuiltinCalendar402">
           <h1>IsBuiltinCalendar ( _id_ )</h1>
           <p>This definition supersedes the definition provided in <emu-xref href="#sec-temporal-isbuiltincalendar"></emu-xref>.</p>
           <emu-alg>
--- a/spec/intl.html
+++ b/spec/intl.html
@@ -25,11 +25,11 @@
     </ins>
 
     <p>
-      The ECMAScript 2021 Internationalization API Specification identifies time zones using the Zone and Link names of the IANA Time Zone Database. Their canonical form is the corresponding Zone name in the casing used in the IANA Time Zone Database.
+      The ECMAScript 2021 Internationalization API Specification identifies time zones using the Zone and Link names of the IANA Time Zone Database. Their canonical form is the corresponding Zone name in the casing used in the IANA Time Zone Database except as specifically overridden by CanonicalizeTimeZoneName.
     </p>
 
     <p>
-      All registered Zone and Link names are allowed. Implementations must recognize all such names, and use best available current and historical information about their offsets from UTC and their daylight saving time rules in calculations. However, the set of combinations of time zone name and language tag for which localized time zone names are available is implementation dependent.
+      A conforming implementation must recognize *"UTC"* and all other Zone and Link names (and <strong>only</strong> such names), and use best available current and historical information about their offsets from UTC and their daylight saving time rules in calculations. However, the set of combinations of time zone name and language tag for which localized time zone names are available is implementation dependent.
     </p>
 
     <emu-clause id="sup-isvalidtimezonename">
@@ -39,36 +39,37 @@
         <p>This definition supersedes the definition provided in <emu-xref href="#sec-isvalidtimezonename"></emu-xref>.</p>
       </ins>
       <p>
-        The IsValidTimeZoneName abstract operation verifies that the _timeZone_ argument (which must be a String value) represents a valid Zone or Link name of the IANA Time Zone Database.
+        The abstract operation IsValidTimeZoneName takes argument _timeZone_, a String value, and verifies that it represents a valid Zone or Link name of the IANA Time Zone Database.
       </p>
-      <p>
-        The abstract operation returns true if _timeZone_, converted to upper case as described in <emu-xref href="#sec-case-sensitivity-and-case-mapping"></emu-xref>, is equal to one of the Zone or Link names of the IANA Time Zone Database, converted to upper case as described in <emu-xref href="#sec-case-sensitivity-and-case-mapping"></emu-xref>. It returns false otherwise.
-      </p>
+
+      <emu-alg>
+        1. If one of the Zone or Link names of the IANA Time Zone Database is an ASCII-case-insensitive match of _timeZone_ as described in <emu-xref href="#sec-case-sensitivity-and-case-mapping"></emu-xref>, return *true*.
+        1. If _timeZone_ is an ASCII-case-insensitive match of *"UTC"*, return *true*.
+        1. Return *false*.
+      </emu-alg>
     </emu-clause>
 
-    <emu-clause id="sup-canonicalizetimezonename">
-      <h1><a href="https://tc39.es/ecma402/#sec-canonicalizetimezonename">CanonicalizeTimeZoneName</a> <ins>( _timeZone_ )</ins></h1>
+    <emu-clause id="sup-canonicalizetimezonename" type="abstract operation">
+      <h1>
+        CanonicalizeTimeZoneName (
+          _timeZone_: a String value that is a valid time zone name as verified by IsValidTimeZoneName,
+        )
+      </h1>
+      <dl class="header">
+        <dt>description</dt>
+        <dd>It returns the canonical and case-regularized form of _timeZone_.</dd>
+      </dl>
 
       <ins class="block">
         <p>This definition supersedes the definition provided in <emu-xref href="#sec-canonicalizetimezonename"></emu-xref>.</p>
       </ins>
-      <p>
-        The CanonicalizeTimeZoneName abstract operation returns the canonical and case-regularized form of the _timeZone_ argument (which must be a String value that is a valid time zone name as verified by the IsValidTimeZoneName abstract operation).
-        The following steps are taken:
-      </p>
 
       <emu-alg>
-        1. <ins>Assert: Type(_timeZone_) is String.</ins>
-        1. <ins>Assert: ! IsValidTimeZoneName(_timeZone_) is *true*.</ins>
-        1. Let _ianaTimeZone_ be the Zone or Link name of the IANA Time Zone Database such that _timeZone_, converted to upper case as described in <emu-xref href="#sec-case-sensitivity-and-case-mapping"></emu-xref>, is equal to _ianaTimeZone_, converted to upper case as described in <emu-xref href="#sec-case-sensitivity-and-case-mapping"></emu-xref>.
-        1. If _ianaTimeZone_ is a Link name, let _ianaTimeZone_ be the corresponding Zone name as specified in the *"backward"* file of the IANA Time Zone Database.
+        1. Let _ianaTimeZone_ be the String value of the Zone or Link name of the IANA Time Zone Database that is an ASCII-case-insensitive match of _timeZone_ as described in <emu-xref href="#sec-case-sensitivity-and-case-mapping"></emu-xref>.
+        1. If _ianaTimeZone_ is a Link name, let _ianaTimeZone_ be the String value of the corresponding Zone name as specified in the file <code>backward</code> of the IANA Time Zone Database.
         1. If _ianaTimeZone_ is *"Etc/UTC"* or *"Etc/GMT"*, return *"UTC"*.
         1. Return _ianaTimeZone_.
       </emu-alg>
-
-      <p>
-        The Intl.DateTimeFormat constructor allows this time zone name; if the time zone is not specified, the host environment's current time zone is used. Implementations shall support UTC and the <del>host environment's current</del> time zone <ins>returned from DefaultTimeZone</ins> (if different from UTC) in formatting.
-      </p>
     </emu-clause>
 
     <emu-clause id="sup-defaulttimezone">
--- a/spec/timezone.html
+++ b/spec/timezone.html
@@ -71,10 +71,7 @@
         If an ECMAScript implementation does not include the ECMA-402 API the following specification of the CanonicalizeTimeZoneName abstract operation is used.
       </p>
       <p>
-        The abstract operation CanonicalizeTimeZoneName takes argument _timeZone_, a String value, and returns the canonical and case-regularized form of _timeZone_.
-      </p>
-      <p>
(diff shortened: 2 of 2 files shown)


spec/intl.html
```diff
@@ -25,26 +25,25 @@ <h1><a href="https://tc39.es/ecma402/#sec-time-zone-names">Time Zone Names</a></
     </ins>
 
     <p>
-      The ECMAScript 2021 Internationalization API Specification identifies time zones using the Zone and Link names of the IANA Time Zone Database. Their canonical form is the corresponding Zone name in the casing used in the IANA Time Zone Database.
+      The ECMAScript 2021 Internationalization API Specification identifies time zones using the Zone and Link names of the IANA Time Zone Database. Their canonical form is the corresponding Zone name in the casing used in the IANA Time Zone Database <ins>except as specifically overridden by CanonicalizeTimeZoneName</ins>.
```
gibson042, line 28: This corrects a current mistake in ECMA-402, which overrides "Etc/UTC" and "Etc/GMT" (which are canonical names for distinct time zones) with "UTC" (which is an alias of the former).

spec/intl.html
```diff
@@ -25,26 +25,25 @@ <h1><a href="https://tc39.es/ecma402/#sec-time-zone-names">Time Zone Names</a></
...
     </p>
 
     <p>
-      All registered Zone and Link names are allowed. Implementations must recognize all such names, and use best available current and historical information about their offsets from UTC and their daylight saving time rules in calculations. However, the set of combinations of time zone name and language tag for which localized time zone names are available is implementation dependent.
+      All registered Zone and Link names are allowed. Implementations must recognize all such names <ins>(and <strong>only</strong> such names) as built-in</ins>, and use best available current and historical information about their offsets from UTC and their daylight saving time rules in calculations. However, the set of combinations of time zone name and language tag for which localized time zone names are available is implementation dependent.
```
gibson042, line 32: It is implicit in the algorithms of ECMA-402 IsValidTimeZoneName and CanonicalizeTimeZoneName that only IANA time zone names are valid. This makes it explicit.

Whether this is appropriate, and whether a _non_-402 implementation should be allowed to support a non-IANA time zone name (as in the current state of Temporal) is not answered here.

spec/intl.html
```diff
@@ -25,26 +25,25 @@ <h1><a href="https://tc39.es/ecma402/#sec-time-zone-names">Time Zone Names</a></
...
     </p>
 
+    <del class="block">
```
gibson042, line 35: Looking up built-in time zones in ECMA-262 IsValidTimeZoneName removes the need to define that algorithm in ECMA-402.

spec/timezone.html
```diff
@@ -71,19 +61,15 @@ <h1>CanonicalizeTimeZoneName ( _timeZone_ )</h1>
...
-        The minimum implementation of CanonicalizeTimeZoneName, supporting only the *"UTC"* time zone, performs the following steps when called:
+        The abstract operation CanonicalizeTimeZoneName takes argument _timeZone_, a String value that is a valid time zone name as verified by IsValidTimeZoneName, and returns the canonical and case-regularized form of _timeZone_.
       </p>
 
       <emu-alg>
         1. Assert: Type(_timeZone_) is String.
         1. Assert: ! IsValidTimeZoneName(_timeZone_) is *true*.
-        1. Return *"UTC"*.
+        1. Let _tzUpper_ be the String value that is the same as _timeZone_ except that each occurrence of a lowercase ASCII letter code unit (0x0061 through 0x007A, inclusive) has been replaced with its uppercase analog (0x0041 through 0x005A, inclusive).
+        1. If _tzUpper_ is *"UTC"*, return *"UTC"*.
```
gibson042, line 71: It would also be possible to remove the ECMA-402 override of CanonicalizeTimeZoneName by including the tzdb steps and the "UTC" overrides here in ECMA-262, but I don't know if that is considered appropriate.

ptomato (approved):
This looks like a good change. Although I'm not so confident about my mental model of how things should be divided between ECMA-262 and 402.

@ryzokuken @sffc It would be good to get your review on this as well, for that reason. @FrankYFTang also your perspective as implementor and ECMA-402 contributor.

sffc (commented):
Seems fine.  Procedurally, I think it would be slightly cleaner to make this a standalone normative PR against ECMA-402.

gibson042:
> I think it would be slightly cleaner to make this a standalone normative PR against ECMA-402.

@sffc Done at https://github.com/tc39/ecma402/pull/624, which ended up being editorial. This PR can be simplified after that lands, but will look different based on whether or not a non-ECMA-402 ECMA-262 Temporal implementation is allowed to support a built-in time zone that is absent from the IANA database (which is not allowed of an ECMA-402 implementation).

ptomato:
https://github.com/tc39/ecma402/pull/624 is merged now, so I think this can be edited somewhat.

spec/timezone.html
```diff
@@ -30,73 +30,76 @@ <h1>Time Zone Names</h1>
...
+
       <p>
-        The abstract operation CanonicalizeTimeZoneName takes argument _timeZone_, a String value, and returns the canonical and case-regularized form of _timeZone_.
-      </p>
-      <p>
-        CanonicalizeTimeZoneName may only be called with a _timeZone_ parameter that is a valid time zone name according to IsValidTimeZoneName.
-      </p>
-      <p>
-        The minimum implementation of CanonicalizeTimeZoneName, supporting only the *"UTC"* time zone, performs the following steps when called:
+        In the descriptions below, String value _A_ is an ASCII-case-insensitive match for String value _B_ if the String value derived from _A_ by replacing each occurrence of a lowercase ASCII letter code unit (0x0061 through 0x007A, inclusive) with the corresponding uppercase ASCII letter code unit (0x0041 through 0x005A, inclusive) while preserving all other code units is exactly the same sequence of code units as the String value that is derived from _B_ in the same way.
```
gibson042, line 78: There's admittedly a lot of duplication here. What would you think of removing IsValidTimeZoneName from both ECMA-262 and ECMA-402 by merging it into CanonicalizeTimeZoneName (which every invocation is already followed by) such that the latter would return undefined when given an invalid time zone name rather than requiring that the name has already been validated (and if there's approval, would that be in scope or out of scope for this PR)?

spec/timezone.html
```diff
@@ -30,73 +30,76 @@ <h1>Time Zone Names</h1>
...
-
       <emu-alg>
-        1. Assert: Type(_timeZone_) is String.
-        1. Assert: ! IsValidTimeZoneName(_timeZone_) is *true*.
-        1. Return *"UTC"*.
+        1. If _timeZone_ is an ASCII-case-insensitive match for *"UTC"*, return *"UTC"*.
+        1. Return the String value containing the implementation-defined canonical and case-regularized name of the built-in time zone whose name is an ASCII-case-insensitive match for _timeZone_.
       </emu-alg>
+      <emu-note>
+        If _timeZone_ is an ASCII-case-insensitive match for the name of a Link in the IANA Time Zone Database, the value returned from CanonicalizeTimeZoneName(_timeZone_) should be derived from the corresponding Zone name rather than from the Link name. And if either that Zone name or _timeZone_ are *"Etc/UTC"* or *"Etc/GMT"*, the value returned should be *"UTC"*. Both behaviours are required of an ECMAScript implementation that includes the ECMA-402 Internationalization API.
```
gibson042, line 85: I'd love to remove _all_ the ECMA-402-specific operation overrides by slightly rewording the already-existing constraints that limit built-in time zone names to those published by IANA and incorporating the Link following into ECMA-262 CanonicalizeTimeZoneName (presumably Normative Optional), but I'm not quite sure how to pull it off. Maybe it will be more clear if IsValidTimeZoneName is removed from both as I suggested above.

gibson042 (commented):
Done.

spec/timezone.html
```diff
@@ -30,10 +30,8 @@ <h1>Time Zone Names</h1>
       When called with any other string, it will throw a *RangeError* exception.
     </p>
     <p>
-      Implementations are encouraged to support the Zone and Link names of the IANA Time Zone Database, and to use the best available current and historical information about their offsets from UTC and their daylight saving time rules in calculations.
-    </p>
-    <p>
-      An ECMAScript implementation that includes the ECMA-402 Internationalization API has additional requirements for the built-in time zones that must be supported, as specified in the ECMA-402 specification.
+      An ECMAScript implementation that includes the ECMA-402 Internationalization API must define built-in named time zones in correspondence with the Zone and Link names of the IANA Time Zone Database and use best available current and historical information about their offsets from UTC and their daylight saving time rules in calculations as specified in the ECMA-402 specification.
+      Other implementations are encouraged to do the same.
```
Ms2ger, line 34: Hmm, I thought we were assuming non-402 implementations would only support UTC and maybe offsets? I might be misremembering, though.

spec/timezone.html
```diff
@@ -30,59 +30,76 @@ <h1>Time Zone Names</h1>
       When called with any other string, it will throw a *RangeError* exception.
     </p>
     <p>
-      An ECMAScript implementation that includes the ECMA-402 Internationalization API must define built-in named time zones in correspondence with the Zone and Link names of the IANA Time Zone Database and use best available current and historical information about their offsets from UTC and their daylight saving time rules in calculations as specified in the ECMA-402 specification.
+      An ECMAScript implementation that includes the ECMA-402 Internationalization API must define built-in named time zones in correspondence with the Zone and Link names of the IANA Time Zone Database (and <strong>only</strong> such names, particularly excluding offset time zones) and use best available current and historical information about their offsets from UTC and their daylight saving time rules in calculations as specified in the ECMA-402 specification.
```
Ms2ger, line 33: Is that because the offset time zones are handled separately? It might be worth calling that out.

spec/timezone.html
```diff
@@ -30,59 +30,76 @@ <h1>Time Zone Names</h1>
...
-        1. If _tzUpper_ is *"UTC"*, return *"UTC"*.
-        1. Return the String value containing the implementation-defined canonical and case-regularized name of the built-in time zone whose name is an ASCII case-insensitive match for _tzUpper_.
+        1. If _timeZone_ is an ASCII-case-insensitive match for *"UTC"*, return *"UTC"*.
+        1. Return the String value containing the implementation-defined canonical and case-regularized name of the built-in time zone whose name is an ASCII-case-insensitive match for _timeZone_.
       </emu-alg>
+      <emu-note>
+        If _timeZone_ is an ASCII-case-insensitive match for the name of a Link in the IANA Time Zone Database, the value returned from CanonicalizeTimeZoneName(_timeZone_) should be derived from the corresponding Zone name rather than from the Link name. And if either that Zone name or _timeZone_ are *"Etc/UTC"* or *"Etc/GMT"*, the value returned should be *"UTC"*. Both behaviours are required of an ECMAScript implementation that includes the ECMA-402 Internationalization API.
```
Ms2ger, line 85: Do we want to use "should" in a note?

FrankYFTang:
I have hard time w/ https://github.com/tc39/test262/pull/3346 . I am not sure what is the right fix in the spec needed for those tests to pass but right now what I see is the name is not taking TimeZoneUTCOffsetName and the grammar could be changed by https://github.com/tc39/proposal-temporal/pull/1941 but that is not enough because step 7-a and 7-b of ParseTemporalTimeZoneString will throw. I have not carefully read all detail of this PR and I am not sure this PR will solve all the remaining issues but surely either the calling place in 7a or inside IsValidTimeZoneName need to be changed to solve that.

spec/timezone.html
gibson042: It's just inherent in the behavior specified at https://tc39.es/ecma402/#sec-time-zone-names that any purported time zone name not present in the IANA data is invalid. So e.g. Intl.DateTimeFormat accepts "Etc/GMT+5" as `timeZone` but rejects "-05:00" with a RangeError.

spec/timezone.html
gibson042: I'm definitely open to other ideas. Are you looking for a rephrase like "implementations are encouraged to…"?

spec/timezone.html
ptomato: I think the idea was more like, non-402 implementations support at least UTC and offsets (which don't go through this path, though) and are free to support any additional subset of IANA zones or any other wacky zones that they please, although they are encouraged to do the same as 402. We affirmed this in yesterday's Temporal champions meeting.

spec/timezone.html
```diff
@@ -30,73 +30,76 @@ <h1>Time Zone Names</h1>
...
+        For the purposes of this section, a String value _A_ is an ASCII-case-insensitive match for String value _B_ if the String value derived from _A_ by replacing each occurrence of a lowercase ASCII letter code unit (0x0061 through 0x007A, inclusive) with the corresponding uppercase ASCII letter code unit (0x0041 through 0x005A, inclusive) while preserving all other code units is exactly the same sequence of code units as the String value that is derived from _B_ in the same way.
       </p>
-
       <emu-alg>
-        1. Assert: Type(_timeZone_) is String.
-        1. Let _tzText_ be ! StringToCodePoints(_timeZone_).
-        1. Let _tzUpperText_ be the result of toUppercase(_tzText_), according to the Unicode Default Case Conversion algorithm.
-        1. Let _tzUpper_ be ! CodePointsToString(_tzUpperText_).
-        1. If _tzUpper_ and *"UTC"* are the same sequence of code points, return *true*.
+        1. If _timeZone_ is an ASCII-case-insensitive match for *"UTC"*, return *true*.
```
ptomato, line 55: Are you sure this can be editorial? It seems to me that this would be normative if there were, as you described in #1919, "a hypothetical non-ASCII string that case-folds to `"UTC"`", but maybe no such string exists?

ptomato (commented):
Is it the case that anything needs to change in this PR according to the rough consensus we reached in yesterday's champions meeting? For example, do we now need to decouple IsValidTimeZoneName from 402's IsValidTimeZoneName?

spec/timezone.html
gibson042: Correct, that's exactly why I believe this to be editorial.

gibson042:
> Is it the case that anything needs to change in this PR according to the rough consensus we reached in yesterday's champions meeting? For example, do we now need to decouple IsValidTimeZoneName from 402's IsValidTimeZoneName?

Yes, I think so.

ptomato:
> > Is it the case that anything needs to change in this PR according to the rough consensus we reached in yesterday's champions meeting? For example, do we now need to decouple IsValidTimeZoneName from 402's IsValidTimeZoneName?
> 
> Yes, I think so.

On second thought, I think it's probably better to put that into its own issue (#1996) so that we don't keep this PR up in the air longer than necessary. I can rebase this today.

ptomato:
I guess it would be best in that case to remove "Editorial: Add implementation-defined built-in time zone lookup steps" since that takes things more in the direction of unifying 402 and 262 IsValidTimeZone, rather than splitting them. It's a bit difficult to separate but I'll continue working on this tomorrow.

spec/timezone.html
ptomato: I think we should consider that out of scope for this PR, but it seems like a good idea.

spec/timezone.html
ptomato: If we are going in the direction of decoupling them, then this seems no longer applicable.

spec/timezone.html
ptomato: We agreed at the champions meeting that we would like ECMA-402 to support offset time zones, so I removed this for now.

spec/timezone.html
ptomato: I removed this for now, since we are going to decouple the 262 and 402 operations.

ptomato:
@gibson042 I've kept the changes that I removed from this PR in a separate branch, https://github.com/tc39/proposal-temporal/tree/gibson042/implementation-defined in case you want to reuse some parts later.

PR #1934, opened 2021-11-19
Initialize post as empty String in FormatTimeZoneOffsetString

FrankYFTang:
otherwise step 13  may have string-concatenationof undefined in FormatTimeZoneOffsetString
```
13. Return the string-concatenation of sign, h, the code unit 0x003A (COLON), m, and post.
```

@ptomato @ljharb @justingrant @ryzokuken @Ms2ger


ptomato (approved):
I believe this is an editorial change, because the previous text was invalid as written. Thanks.

PR #1937, opened 2021-11-22
Editorial: Simplified FormatSecondsStringPart

FrankYFTang:
The first two line in the If/Else are exactly the same and have no reason to be put inside If/Else blocks

### DIFF
--- a/spec/abstractops.html
+++ b/spec/abstractops.html
@@ -627,13 +627,11 @@
       1. If _precision_ is *"minute"*, return *""*.
       1. Let _secondsString_ be the string-concatenation of the code unit 0x003A (COLON) and _second_ formatted as a two-digit decimal number, padded to the left with zeroes if necessary.
       1. Let _fraction_ be _millisecond_ × 10<sup>6</sup> + _microsecond_ × 10<sup>3</sup> + _nanosecond_.
+      1. If _fraction_ is 0, return _secondsString_.
+      1. Set _fraction_ to _fraction_ formatted as a nine-digit decimal number, padded to the left with zeroes if necessary.
       1. If _precision_ is *"auto"*, then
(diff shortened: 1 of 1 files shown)


spec/abstractops.html
```diff
@@ -627,13 +627,11 @@ <h1>FormatSecondsStringPart ( _second_, _millisecond_, _microsecond_, _nanosecon
...
       1. Let _secondsString_ be the string-concatenation of the code unit 0x003A (COLON) and _second_ formatted as a two-digit decimal number, padded to the left with zeroes if necessary.
       1. Let _fraction_ be _millisecond_ × 10<sup>6</sup> + _microsecond_ × 10<sup>3</sup> + _nanosecond_.
+      1. If _fraction_ is 0, return _secondsString_.
+      1. Set _fraction_ to _fraction_ formatted as a nine-digit decimal number, padded to the left with zeroes if necessary.
       1. If _precision_ is *"auto"*, then
-        1. If _fraction_ is 0, return _secondsString_.
-        1. Set _fraction_ to _fraction_ formatted as a nine-digit decimal number, padded to the left with zeroes if necessary.
         1. Set _fraction_ to the longest possible substring of _fraction_ starting at position 0 and not ending with the code unit 0x0030 (DIGIT ZERO).
       1. Else,
-        1. If _precision_ is 0, return _secondsString_.
```
ptomato, line 635: This isn't the same as line 631 (here it's _precision_ and there it's _fraction_ being compared to 0).

ptomato (changes requested):
I'm not convinced this change is correct, I think the two pieces of code are similar but not the same.

FrankYFTang:
sorry. You are right, I got confused.

PR #1938, opened 2021-11-23
Editoral: Mark use of TotalDurationNanoseconds inside AdjustRoundedDurationDays as infallible

Lubrsi:
Current definition (https://tc39.es/proposal-temporal/#sec-temporal-totaldurationnanoseconds):
```
1. Assert: offsetShift is an integer.
2. Set nanoseconds to ℝ(nanoseconds).
3. If days ≠ 0, then
    a. Set nanoseconds to nanoseconds − offsetShift.
4. Set hours to ℝ(hours) + ℝ(days) × 24.
5. Set minutes to ℝ(minutes) + hours × 60.
6. Set seconds to ℝ(seconds) + minutes × 60.
7. Set milliseconds to ℝ(milliseconds) + seconds × 1000.
8. Set microseconds to ℝ(microseconds) + milliseconds × 1000.
9. Return nanoseconds + microseconds × 1000.
```

### DIFF
--- a/spec/duration.html
+++ b/spec/duration.html
@@ -1413,7 +1413,7 @@
             [[Microseconds]]: _microseconds_,
             [[Nanoseconds]]: _nanoseconds_
             }.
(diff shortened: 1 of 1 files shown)


ptomato (approved):
Agreed, there can be no abrupt completion in this step! Thanks.

PR #1945, opened 2021-11-30
Fix TS types of RoundTo and TotalOf to allow plural unit names

justingrant:
This PR fixes bugs in the TS types RoundTo and TotalOf where plural unit names were not being accepted in a few places.


ptomato (approved):
:+1:

PR #1954, opened 2021-12-03
Editorial: Move calendar validation out of ParseTemporalCalendarString

ptomato:
Similar to #1897, we want to move any validation that is not purely syntactic out of ParseTemporal___String abstract operations. This accomplishes that for Calendar. However, it affects nothing observable, just reorganizes code.

Closes: #1901


ptomato:
cc @FrankYFTang

Ms2ger (approved):

PR #1959, opened 2021-12-04
Support newer DateTimeFormat output

justingrant:
Update GetFormatterParts implementation in polyfill to avoid failed tests on the latest Intl implementations in newer JS engines (e.g. FF nightly).

Port of https://github.com/js-temporal/temporal-polyfill/pull/97


ptomato:
I would've preferred to use formatToParts here anyway since this polyfill cares much less about older clients, preferring robustness as long as someone with a moderately recent browser can run the playground... but if this works for now, I don't feel like going back and changing it unless we run into a new instance of the output being changed.

PR #1964, opened 2021-12-08
Fix TS types for required CalendarProtocol methods

justingrant:
The [docs](https://tc39.es/proposal-temporal/docs/calendar.html#custom-calendars) say: 

> The other, more difficult, way to create a custom calendar is to create a plain object implementing the `Temporal.Calendar` protocol, without subclassing. The object must implement all of the `Temporal.Calendar` properties and methods except for `id`, `fields()`, `mergeFields()`, and `toJSON()`.

The TS types currently allow many CalendarProtocol methods to be undefined other than the 4 above. This PR aligns the TS types to the docs.

@sffc @ptomato holler if the docs are wrong.


ptomato (approved):
LGTM, in order to have a fully functional plain-object calendar all these methods are required.

To be clear, you can leave them out and nothing will complain so long as you don't use any features that require them, which we do in some of our tests. But that doesn't seem like the use case that TypeScript bindings should address.

PR #1966, opened 2021-12-09
Editorial: Use set instead of let when updating moveResult on BalanceDurationRelative step 9.e

Lubrsi:
It was previously instantiated on step 9.a.


ptomato (approved):
Thanks!

PR #1980, opened 2021-12-14
Polyfill: remove unnecessary runtime assertions in PrepareTemporalFields and ToPartialRecord

justingrant:
Per @ptomato's suggestion in https://github.com/js-temporal/temporal-polyfill/pull/109/files#r768187244, these two AOs can never have non-object inputs, so these runtime assertions shouldn't be necessary.


Ms2ger (approved):
Too bad we don't have actual assertions.

PR #1982, opened 2021-12-15
Update Test262 to 04cd6da0

justingrant:
* Catch up with latest Test262
* Remove test from expected failures list because its PR was merged.


ptomato (approved):
Thanks! :+1:

PR #1995, opened 2022-01-07
Add missing branding checks

jessealama:
References:

+ https://github.com/tc39/proposal-temporal/pull/1693

### DIFF
--- a/polyfill/test/usertimezone.mjs
+++ b/polyfill/test/usertimezone.mjs
@@ -95,13 +95,6 @@ describe('Userland time zone', () => {
 
     const inst = Temporal.Instant.fromEpochNanoseconds(0n);
 
-    it('converts to DateTime', () => {
-      equal(`${Temporal.TimeZone.prototype.getPlainDateTimeFor.call(obj, inst)}`, '1970-01-01T00:00:00');
-      equal(
-        `${Temporal.TimeZone.prototype.getPlainDateTimeFor.call(obj, inst, 'gregory')}`,
-        '1970-01-01T00:00:00[u-ca=gregory]'
-      );
-    });
     it('offset prints in instant.toString', () => equal(inst.toString({ timeZone: obj }), '1970-01-01T00:00:00+00:00'));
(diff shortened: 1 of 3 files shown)


ptomato:
I don't think the brand checks in the `id()` accessors are ill-conceived, I think they need to be there... they are there in the specification.

Looks like some of the legacy tests will need to be adjusted or deleted, and we should add the failing test262 tests to the expected failure list (at least, until we get another test262 PR merged to update the test262 tests upstream.)

polyfill/test/usertimezone.mjs
```diff
@@ -77,21 +77,29 @@ describe('Userland time zone', () => {
     });
   });
   describe('Trivial protocol implementation', () => {
-    const obj = {
+    class CustomTimeZone extends Temporal.TimeZone {
```
ptomato, line 80: I don't think this change is correct. We still want to test a plain-object implementation of the TimeZone protocol, and not an instance of TimeZone. I guess the "it converts to DateTime" test below fails, but that test should be removed because that's exactly the capability we're removing with this change.

ptomato (changes requested):

polyfill/test/usertimezone.mjs
jessealama: ah, OK, I see. I wasn't sure if we wanted to allow plain objects for `TimeZone` (there were some other tests that also use this pattern of introducing a fake subclass, and they have branding checks, too, so they misled me). I'll revert the commit.

Ms2ger (approved):

Ms2ger:
I started writing tests at https://github.com/tc39/test262/compare/main...Igalia:branding?expand=1, but it seems like some existing ones need changes as well, like `built-ins/Temporal/TimeZone/prototype/toJSON/tostring-call.js`.

ptomato (approved):

ptomato:
Yeah, looks like we will need to add expected failures for those tests until we can get the fix merged into test262.

PR #1998, opened 2022-01-12
Fix `ZonedDateTime` example

SamyCookie:
Hi,

if have correctly understood the API, I think the example should be more like my proposed change.

Thank you for this awesome API !


justingrant (approved):

ptomato (approved):
This is correct. Thanks!

PR #1999, opened 2022-01-12
Normative: Fix fraction substring in ParseTemporalTimeZoneString and ParseTimeZoneOffsetString

linusg:
See commit messages for details.


spec/abstractops.html
```diff
@@ -1492,7 +1492,7 @@ <h1>ParseTemporalTimeZoneString ( _isoString_ )</h1>
         1. Set _seconds_ to ! ToIntegerOrInfinity(_seconds_).
         1. If _fraction_ is not *undefined*, then
           1. Set _fraction_ to the string-concatenation of the previous value of _fraction_ and the string *"000000000"*.
-          1. Let _nanoseconds_ be the String value equal to the substring of _fraction_ from 1 to 10.
+          1. Let _nanoseconds_ be the String value equal to the substring of _fraction_ consisting of the code units with indices 0 (inclusive) through 9 (exclusive).
```
linusg, line 1495: Wording here is taken from `ParseTimeZoneOffsetString`:

https://github.com/tc39/proposal-temporal/blob/2a81fbca981eb684f1dc10ec479b5a8e872d8c37/spec/timezone.html#L516

ptomato: As per the conversation in https://github.com/tc39/proposal-temporal/pull/1830#discussion_r710402655, we determined that the wordier variant was unnecessary, so this can just be "the substring of _fraction_ from 0 to 9".

ptomato (commented):
I thought for sure that this had been fixed already, but I see what happened: #1830 fixed the problem by adjusting the substring indices, and #1796 fixed it again by examining a different parse node that didn't include the decimal separator. Apparently two rights make a wrong in this case...

At least for ParseTemporalTimeZoneString, I would guess that we don't need to get approval for the normative change since the intent of what this operation should do was clearly approved by TC39 (twice, in fact!) But I'm open to have my mind changed on that, so I'll leave this open for a while to give anyone a chance to object.

cc @gibson042, I seem to remember you planned to unify this so that the parse nodes including the decimal separators were the only ones used, and always take the substring from 1 to 10 everywhere?

cc @FrankYFTang as well, as the author of the previous fixes

spec/abstractops.html
linusg: Got it, updated :+1:

linusg:
Rebased & fixed a conflict; the first commit is now purely editorial as the bug that was being fixed is no longer present as of https://github.com/tc39/proposal-temporal/commit/6abe1465ec0993babf123e76adaf601b2388478c. Second commit remains untouched.

ptomato:
Feel free to submit the editorial fix separately to get it merged faster (though it's also fine if you want to leave it in this PR.)

As for the normative fix I'd like to defer to @gibson042, since I thought his plan was to standardize on taking the substring from 1 to 10 of the |...Fraction| productions, so I want to make sure we're not moving in opposite directions simultaneously.

gibson042:
The most important part is "standardize", so that different operations extract similar data from parse nodes in the same way. But yes, my preference is "from 1 to 10" of text that starts with a decimal separator, in accordance with a general goal of limiting grammar productions to those with semantic value (and in particular eliminating |FractionalPart| in favor of a restriction on |Fraction| like e.g. "_DecimalSeparator DecimalDigits<sub>[~Sep]</sub>_" with "_but only if DecimalDigits is not longer than nine code points_" or "_It is a Syntax Error if DecimalDigits is longer than nine code points_").

ptomato:
@linusg Would you have a chance to address Richard's comment soon? I am still assuming that we don't need to present this normative change for approval in a TC39 plenary, since the intent of what this operation should do was clearly approved in a previous plenary, so no hurry (otherwise the deadline would be this Friday). But still it would be good to resolve this.

linusg:
Ah, sorry, missed this. Updated!

ptomato (approved):
Thanks. Although this is a normative change, I'll merge it because I believe it clearly implements a change that already gained consensus in the October 2021 plenary.

If there are any objections within a reasonable time, let's say a week, I'll be happy to revert it.

PR #2001, opened 2022-01-12
Editorial: Mark CreateTemporalTimeZone("UTC") as infallible

linusg:
When `newTarget` is not provided, it falls back to `%Temporal.TimeZone%` & the only fallible operation in `CreateTemporalTimeZone`,  the call of `OrdinaryCreateFromConstructor`, becomes infallible.

See also:

https://github.com/tc39/proposal-temporal/blob/2a81fbca981eb684f1dc10ec479b5a8e872d8c37/spec/timezone.html#L584


ptomato (approved):
Thanks!

PR #2017, opened 2022-01-25
Include the valueOf function in Temporal.Duration.

12wrigja:
This was missed in https://github.com/tc39/proposal-temporal/commit/bc0b188c0e07657319f03489d7b8caa89c420a5d, and is a port of https://github.com/js-temporal/temporal-polyfill/pull/121


ptomato (approved):
Thanks!

PR #2019, opened 2022-01-26
Move tests for PlainYearMonth constructor, from.

Ms2ger:


ptomato (approved):
Looks good, waiting on https://github.com/tc39/test262/pull/3386 to merge (since that will likely get rebased rather than merged, so the commit hash will change)

PR #2021, opened 2022-01-28
Remove bad examples

jessealama:
Using a `Z` designator when creating an instance of the these `PlainX` types is incorrect.


Ms2ger (approved):
Thanks!

Issue #1984, opened 2021-12-17
Use early errors to simplify ISO 8601 grammar


ptomato:
Follow up from https://github.com/tc39/proposal-temporal/pull/1952#discussion_r764327901

Use early errors in the ISO 8601 grammar to avoid having to create such bizarre productions like TimeHourNotThirtyOneDayMonth, etc. This is an editorial change.

cc @gibson042

PR #2036, opened 2022-02-04
Editorial: Adjust grammar for years to exclude -000000

jessealama:
~~Also, remove redundant check in `ParseISODateTime`.~~

### DIFF
--- a/polyfill/test/validStrings.mjs
+++ b/polyfill/test/validStrings.mjs
@@ -63,6 +63,37 @@ function withCode(productionLike, func) {
   return new FromFunction(func, productionLike);
 }
 
+// Return a productionLike that defers to another productionLike for generation
+// but filters out any result that a validator function
+// rejects with SyntaxError.
+function withSyntaxConstraints(productionLike, validatorFunc) {
+  return { generate: filtered };
+
+  function filtered(data) {
+    let dataClone = { ...data };
+    let result = productionLike.generate(dataClone);
+    try {
+      validatorFunc(result);
+    } catch (e) {
+      if (e instanceof SyntaxError) {
+        // Generated result violated a constraint; try again
+        // (but allow crashing if the call stack gets too deep).
+        return filtered(data);
+      } else {
+        // Propagate any other error.
+        throw e;
+      }
+    }
+
+    // Copy properties that were modified on the cloned input data.
+    for (let key of Reflect.ownKeys(dataClone)) {
+      data[key] = dataClone[key];
+    }
+
+    return result;
+  }
+}
+
 class Literal {
   constructor(str) {
     this.str = str;
@@ -195,7 +226,12 @@ const timeFractionalPart = between(1, 9, digit());
 const fraction = seq(decimalSeparator, timeFractionalPart);
 
 const dateFourDigitYear = repeat(4, digit());
-const dateExtendedYear = seq(sign, repeat(6, digit()));
+
+const dateExtendedYear = withSyntaxConstraints(seq(sign, repeat(6, digit())), (result) => {
+  if (result === '-000000' || result === '−000000') {
+    throw new SyntaxError('Negative zero extended year');
+  }
+});
 const dateYear = withCode(
   choice(dateFourDigitYear, dateExtendedYear),
   (data, result) => (data.year = +result.replace('\u2212', '-'))
--- a/spec/abstractops.html
+++ b/spec/abstractops.html
@@ -1145,6 +1145,11 @@
       TemporalRelativeToString :
           TemporalDateTimeString
     </emu-grammar>
+    <ul>
+      <li>
+        It is a Syntax Error if |DateExtendedYear| is *"-000000"* or *"−000000"* (U+2212 MINUS SIGN followed by `000000`).
+      </li>
+    </ul>
   </emu-clause>
 
   <emu-clause id="sec-temporal-parseisodatetime" aoid="ParseISODateTime">
@@ -1160,7 +1165,6 @@
       1. Let _minute_ be the part of _isoString_ produced by the |TimeMinute|, |TimeMinuteNotValidDay|, |TimeMinuteThirtyOnly|, or |TimeMinuteThirtyOneOnly| productions, or *undefined* if none of those are present.
       1. Let _second_ be the part of _isoString_ produced by the |TimeSecond| or |TimeSecondNotValidMonth| productions, or *undefined* if neither of those are present.
       1. If the first code unit of _year_ is 0x2212 (MINUS SIGN), replace it with the code unit 0x002D (HYPHEN-MINUS).
-      1. If SameValue(_year_, *"-000000"*) is *true*, throw a *RangeError* exception.
       1. Set _year_ to ! ToIntegerOrInfinity(_year_).
       1. If _month_ is *undefined*, then
         1. Set _month_ to 1.


ptomato:
A couple of notes on this:
- I would actually not change the polyfill code here. Since the polyfill parsing using regexp is already different from the spec, it doesn't matter so much whether the implementation resembles the spec, as long as the result conforms. (And I'd hesitate to encode the nonzero digits in regexp, as the regexes are already complicated enough.) I would, however, update the grammar in `test/validStrings.mjs`.
- This seems like a good opportunity to introduce a section with early error semantics for the ISO string grammar, like @gibson042 proposed in https://github.com/tc39/proposal-temporal/issues/1984. The new _DateExtendedYear_ is pretty complicated and I think Richard's idea of using early errors would express the intention better.

linusg (commented):
> Also, remove redundant check in `ParseISODateTime`.

This doesn't appear to be actually included in the commit that mentions it.

jessealama:
> > Also, remove redundant check in `ParseISODateTime`.
> 
> This doesn't appear to be actually included in the commit that mentions it.

Quite right -- I edited the commit. That comment has become obsolete.

jessealama:
I've implemented (what I think is) an early error approach to `-000000`. I'm not sure if this is really sufficient. @gibson042 Do you think this a reasonable way to handle this special case?

ptomato:
I wonder if we can make a mechanism in `test/validStrings.mjs` that is more similar to the early error mechanism in the spec text?

spec/abstractops.html
```diff
@@ -1146,6 +1146,11 @@ <h1>ISO 8601 grammar</h1>
       TemporalRelativeToString :
           TemporalDateTimeString
     </emu-grammar>
+    <ul>
+      <li>
+        It is a Syntax Error if |DateExtendedYear| is *"-000000"*.
```
ptomato, line 1151: My other comment reminded me that **"−000000"** (U+2212 MINUS SIGN followed by 000000) should also be a Syntax Error.

jessealama:
Added some new functionality for mimicking early errors in the string generator.

ptomato (approved):
Nice. If you can make the linter happy then this is good to go from me.

@gibson042 what do you think about it?

linusg (changes requested):
This still needs a removal of the following, now redundant, line:

https://github.com/tc39/proposal-temporal/blob/e19f98150f57f54a712917b10c049159d1ba6928/spec/abstractops.html#L1164

polyfill/test/validStrings.mjs
```diff
@@ -63,6 +63,22 @@ function withCode(productionLike, func) {
...
+    validatorFunc(result);
+    return result;
+  } catch (e) {
+    if (e instanceof SyntaxError) {
+      return withEarlyError(productionLike, validatorFunc);
+    } else {
+      throw e;
+    }
+  }
+}
```
gibson042, line 80: Wouldn't it be more consistent and comprehensible to return an object _with_ a `generate` method rather than the results _of_ a `generate()` invocation?

```suggestion
// Return a productionLike that defers to another productionLike for generation
// but uses a validator function to filter out any result that a validator function
// rejects with SyntaxError.
function withSyntaxConstraints(productionLike, validatorFunc) {
  return { generate: filtered };

  function filtered(data) {
    let dataClone = { ...data };
    let result = productionLike.generate(dataClone);
    try {
      validatorFunc(result);
    } catch (e) {
      if (e instanceof SyntaxError) {
        // Generated result violated a constraint; try again
        // (but allow crashing if the call stack gets too deep).
        return filtered(data);
      } else {
        // Propagate any other error.
        throw e;
      }
    }

    // Copy properties that were modified on the cloned input data.
    for (let key of Reflect.ownKeys(dataClone)) {
      data[key] = dataClone[key];
    }

    return result;
  };
}
```

polyfill/test/validStrings.mjs
```diff
@@ -196,7 +214,11 @@ const timeFractionalPart = between(1, 9, digit());
 const fraction = seq(decimalSeparator, timeFractionalPart);
 
 const dateFourDigitYear = repeat(4, digit());
-const dateExtendedYear = seq(sign, repeat(6, digit()));
+
+const dateExtendedYear = withEarlyError(
+  seq(sign, repeat(6, digit())),
+  (result) => { if ('-000000' === result) throw new SyntaxError('Negative zero extended year'); }
```
gibson042, line 220: This fails to reject "−000000" (a non-ASCII minus sign followed by six zeroes).
```suggestion
  (result) => {
    if (result === '-000000' || result === '−000000') {
      throw new SyntaxError('Negative zero extended year');
    }
  }
```

gibson042 (changes requested):

jessealama:
> This still needs a removal of the following, now redundant, line:
> 
> https://github.com/tc39/proposal-temporal/blob/e19f98150f57f54a712917b10c049159d1ba6928/spec/abstractops.html#L1164

fixed

ptomato (changes requested):
What we have here now looks good.

~~However, I realized now that catching the Unicode minus sign with six zeroes is a normative change, because an implementation that conformed to the previous text would have to change its behaviour. I've opened #2069 for this. Here's what I propose: remove the check for Unicode minus sign with six zeroes from this PR (both from the early error and from validStrings) so that it does not require any behaviour change from implementations, merge it, and then create a new PR that will have to be presented at the March TC39 meeting to close #2069.~~

Edit: see below

jessealama:
> What we have here now looks good.
> 
> However, I realized now that catching the Unicode minus sign with six zeroes is a normative change, because an implementation that conformed to the previous text would have to change its behaviour. I've opened #2069 for this. Here's what I propose: remove the check for Unicode minus sign with six zeroes from this PR (both from the early error and from validStrings) so that it does not require any behaviour change from implementations, merge it, and then create a new PR that will have to be presented at the March TC39 meeting to close #2069.

Sounds good. I've removed support for minus sign six zeros, both in the spec and in the fuzzer.

polyfill/test/validStrings.mjs
```diff
@@ -174,7 +205,9 @@ function seq(...productions) {
 // Grammar productions, based on the grammar in RFC 3339
 
 // characters
-const sign = character('+-−');
```
ptomato, line 177: This change isn't correct - Unicode minus signs should still be accepted everywhere ASCII minus signs are accepted. It's just that this PR should preserve the current observable behaviour of rejecting `-000000` but allowing `\u2212000000`. I think if you just revert this hunk, the PR will be correct.

jessealama: Ah, I see -- OK, fixed

ptomato (approved):
:+1:

ptomato:
Is it marked as draft for any reason or is it ready to merge now? All good from my side.

polyfill/test/validStrings.mjs
```diff
@@ -63,6 +63,37 @@ function withCode(productionLike, func) {
   return new FromFunction(func, productionLike);
 }
 
+// Return a productionLike that defers to another productionLike for generation
+// but uses a validator function to filter out any result that a validator function
```
gibson042, line 67:
```suggestion
// but filters out any result that a validator function
```

gibson042 (approved):

ptomato (approved):

ptomato:
> However, I realized now that catching the Unicode minus sign with six zeroes is a normative change, because an implementation that conformed to the previous text would have to change its behaviour. I've opened #2069 for this. Here's what I propose: remove the check for Unicode minus sign with six zeroes from this PR (both from the early error and from validStrings) so that it does not require any behaviour change from implementations, merge it, and then create a new PR that will have to be presented at the March TC39 meeting to close #2069.

I just realized I got this wrong. The current observable behaviour is to reject `\u2212000000`, because we have these steps:

> 1. If the first code unit of _year_ is 0x2212 (MINUS SIGN), replace it with the code unit 0x002D (HYPHEN-MINUS).
> 1. If SameValue(_year_, *"-000000"*) is *true*, throw a *RangeError* exception.

Sorry for the bad advice. I was rebasing this anyway so I'll put it back the way you had it before merging it.

PR #2047, opened 2022-02-09
Move PlainYearMonth#{add,subtract} tests.

Ms2ger:


Ms2ger:
Submodule update to add the ported tests already landed.

ptomato (approved):
I checked the disappearing lines of coverage and they are just due to the different ways that V8 and NYC count whether a line is covered.

Issue #1993, opened 2021-12-28
TemporalTimeZoneIdentifier and TimeZoneIANAName exclude some valid IANA time zone names


gibson042:
The IANA time zone database includes zones "Etc/GMT+0" through "Etc/GMT+12", but these are not matched by the [_TimeZoneIANAName_](https://tc39.es/proposal-temporal/#prod-TimeZoneIANAName) production (and therefore not by the [_TemporalTimeZoneIdentifier_](https://tc39.es/proposal-temporal/#prod-TemporalTimeZoneIdentifier) production either) because "+" is not matched by [_TZChar_](https://tc39.es/proposal-temporal/#prod-TZChar) (while the sibling zones "Etc/GMT-0" through "Etc/GMT-14" _are_ matched because "-" is matched by TZChar). [_TimeZoneBracketedName_ ](https://tc39.es/proposal-temporal/#prod-TimeZoneBracketedName) specifically expands to **Etc/GMT** _ASCIISign_ _Hour_, but even that doesn't really work because _Hour_ only matches two-digit sequences 00 through 23 (resulting in incorrect acceptance of "Etc/GMT+00" through "Etc/GMT+09" and incorrect rejection of "Etc/GMT+0" through "Etc/GMT+9").

I think the special TimeZoneBracketedName expansion should be moved to TimeZoneIANAName and fixed to something like **Etc/GMT** _ASCIISign_ _Hour_[+Unpadded] or **Etc/GMT** _ASCIISign_ _UnpaddedHour_.

PR #2050, opened 2022-02-11
Normative: Define unpadded hours and fix offset Etc/GMT timezones

jessealama:
fixes https://github.com/tc39/proposal-temporal/issues/1993

### DIFF
--- a/spec/abstractops.html
+++ b/spec/abstractops.html
@@ -722,6 +722,14 @@
           ASCIISign
           U+2212
 
+      UnpaddedHour :
+          DecimalDigit
+          `1` DecimalDigit
+          `20`
+          `21`
+          `22`
+          `23`
+
       Hour :
           `0` DecimalDigit
           `1` DecimalDigit
@@ -941,11 +949,11 @@
           TimeZoneIANANameComponent `/` TimeZoneIANANameTail
 
       TimeZoneIANAName :
-          TimeZoneIANANameTail
+          `Etc/GMT` ASCIISign UnpaddedHour
+          TimeZoneIANANameTail but not `Etc/GMT` ASCIISign UnpaddedHour
 
       TimeZoneBracketedName :
           TimeZoneIANAName
-          `Etc/GMT` ASCIISign Hour
           TimeZoneUTCOffsetName
 
       TimeZoneBracketedAnnotation :


jessealama:
cc @gibson042

jessealama:
closes #1993

spec/abstractops.html
```diff
@@ -731,6 +731,11 @@ <h1>ISO 8601 grammar</h1>
           ASCIISign
           U+2212
 
+      UnpaddedHour :
+          DecimalDigit
+          `11`
+          `12`
```
ptomato, line 737: This isn't going to be sufficient to match `Etc/GMT-14`.

spec/abstractops.html
```diff
@@ -950,12 +955,12 @@ <h1>ISO 8601 grammar</h1>
           TimeZoneIANANameComponent `/` TimeZoneIANANameTail
 
       TimeZoneIANAName :
+          `Etc/GMT` ASCIISign UnpaddedHour
+          TimeZoneUTCOffsetName
           TimeZoneIANANameTail
 
       TimeZoneBracketedName :
           TimeZoneIANAName
-          `Etc/GMT` ASCIISign Hour
```
ptomato, line 957: I think we can remove TimeZoneBracketedName altogether now?

This change also affects step 4 of [ParseTemporalTimeZoneString](https://tc39.es/proposal-temporal/#_ref_2536) but I believe that is also a correct change.

ptomato (changes requested):

ptomato:
(A corresponding change to `test/validStrings.mjs` to verify the grammar changes would be helpful, as well)

spec/abstractops.html
jessealama: right, oops

fixed

spec/abstractops.html
jessealama: right, we can remove `TimeZoneBracketedName` -- done

I'm not sure I understand the comment about `ParseTemporalTimeZoneString`. Can you elaborate?

jessealama:
> (A corresponding change to `test/validStrings.mjs` to verify the grammar changes would be helpful, as well)

good idea -- done

spec/abstractops.html
ptomato: Sure! Here's ParseTemporalTimeZoneString step 4:

> Let each of _z_, _offsetString_, and _name_ be the source text matched by the respective _UTCDesignator_, _TimeZoneNumericUTCOffset_, and _TimeZoneIANAName_ Parse Node enclosed by _parseResult_, or an empty sequence of code points if not present.

This change affects that step because the change in _TimeZoneIANAName_ means that now _name_ can end up being one of the `Etc/GMT±n` strings as well as a `±hh:mm` offset string, which it could not before. But I believe that was a mistake, which this change now fixes.

polyfill/test/validStrings.mjs
```diff
@@ -293,20 +293,11 @@ const timeZoneNumericUTCOffsetNotAmbiguousAllowedNegativeHour = withCode(
   saveOffset
 );
 const timeZoneUTCOffset = choice(utcDesignator, timeZoneNumericUTCOffset);
-const timeZoneUTCOffsetName = seq(
```
ptomato, line 296: Doesn't `timeZoneUTCOffsetName` still have to be included in `timeZoneIANAName` to match the grammar in the spec?

polyfill/test/validStrings.mjs
```diff
@@ -293,20 +293,11 @@ const timeZoneNumericUTCOffsetNotAmbiguousAllowedNegativeHour = withCode(
...
-  sign,
-  hour,
-  choice([minuteSecond, [minuteSecond, [fraction]]], seq(':', minuteSecond, [':', minuteSecond, [fraction]]))
-);
-const timeZoneBracketedName = withCode(
-  choice(timeZoneUTCOffsetName, ...timezoneNames),
```
ptomato, line 302: Am I correct in assuming the `Etc/GMT±n` names are included in `timezoneNames` so we don't have to model them separately here?

spec/abstractops.html
gibson042: A syntax-level bound to 14 doesn't seem right to me either (e.g., what if an Etc/GMT-15 time zone is introduced in the future?). I think I would instead expect a bound of 23, corresponding with allowed UTC offsets (and more similarity in the definitions of _UnpaddedHour_ and _Hour_).

spec/abstractops.html
```diff
@@ -950,15 +957,12 @@ <h1>ISO 8601 grammar</h1>
...
 
       TimeZoneIANAName :
-          TimeZoneIANANameTail
-
-      TimeZoneBracketedName :
-          TimeZoneIANAName
-          `Etc/GMT` ASCIISign Hour
+          `Etc/GMT` ASCIISign UnpaddedHour
           TimeZoneUTCOffsetName
+          TimeZoneIANANameTail
```
gibson042, line 962: While this grammar is changing, we should resolve its ambiguity as well by prohibiting TimeZoneIANANameTail from matching an Etc/GMT… zone (it might not matter in practice, but allowing such ambiguity just seems like bad form).
```suggestion
          TimeZoneUTCOffsetName
          `Etc/GMT` ASCIISign UnpaddedHour
          TimeZoneIANANameTail but not `Etc/GMT` ASCIISign UnpaddedHour
```

spec/abstractops.html
```diff
@@ -950,15 +957,12 @@ <h1>ISO 8601 grammar</h1>
           TimeZoneIANANameComponent `/` TimeZoneIANANameTail
 
       TimeZoneIANAName :
```
gibson042, line 959: Hmm, it seems misleading to use _TimeZoneIANAName_ for describing a nonterminal that can also match _TimeZoneUTCOffsetName_ (which are **not** IANA time zone names). Can this be collapsed into _TemporalTimeZoneIdentifier_ (whose _TimeZoneNumericUTCOffset_ is just a duplication of _TimeZoneUTCOffsetName_ anyway)?

gibson042 (changes requested):

polyfill/test/validStrings.mjs
jessealama: right, that's my intention

spec/abstractops.html
jessealama: oops, quite right

fixed

polyfill/test/validStrings.mjs
jessealama: Or at least, my intention was to remove definitions in validStrings.mjs that have been removed from the grammar (the aim being to make sure that those the two mirror each other as closely as possible).

spec/abstractops.html
```diff
@@ -731,6 +731,22 @@ <h1>ISO 8601 grammar</h1>
...
+          DecimalDigit
+          `11`
+          `12`
+          `13`
+          `14`
+          `15`
+          `16`
+          `17`
+          `18`
+          `19`
```
ptomato, line 744: This list still omits `10`

To reduce the chance of errors like that, consider writing it like this:
```suggestion
          `1` DecimalDigit
```

spec/abstractops.html
ptomato: It looks like according to this grammar, now a string like `2022-02-21T09:25-08:00[-08:00]` would be disallowed, because _TimeZoneBracketedAnnotation_ only surrounds _TimeZoneIANAName_ whereas it used to surround _TimeZoneBracketedName_. Can you check this?

spec/abstractops.html
jessealama: I've cleaned things up by reverting the proposed grammar cleanup. (`validStrings.mjs` was cleaned up, too.)

polyfill/test/validStrings.mjs
```diff
@@ -298,15 +298,19 @@ const timeZoneUTCOffsetName = seq(
...
-);
-const timeZoneBracketedAnnotation = seq('[', timeZoneBracketedName, ']');
+const unpaddedHour = choice(
+  digit(),
+  seq(character('1'), digit()),
+  seq(character('2'), choice(character('0'), character('1'), character('2'), character('3')))
+)
 const timeZoneIANAName = withCode(
-  choice(...timezoneNames),
+  choice(
```
ptomato, line 307: I checked and actually the `Etc/GMT...` time zones are already present in the timezones.json package, so they'll be included here in `choice(...timezoneNames)`. So maybe we can just leave this out and it'll be sufficient to add a comment here explaining why the validStrings deviates from the grammar in the spec in this case.

spec/abstractops.html
ptomato: Was this still the case?

polyfill/test/validStrings.mjs
jessealama: I guess we can drop it here because it gets imported thanks to `timezones.json`?

spec/abstractops.html
jessealama: it seems that that was already invalid? I tried to parse that using the previous version of the grammar but it seems that it doesn't work. Don't we need a name of some kind ("Etc/GMT+1" or "America/Kentucky/Louisville") instead of a numeric offset like "-08:00"?

ptomato: I disagree that it was already invalid (TimeZoneUTCOffsetName was part of TimeZoneBracketedName) but I think this is now fixed in the current version that you have.

Richard's comment about collapsing TimeZoneIANAName into TemporalTimeZoneIdentifier should still be addressed.

polyfill/test/validStrings.mjs
ptomato: The offset names such as `-08:00` are not part of `timezones.json`, so they still need to be included separately.

spec/abstractops.html
jessealama: done

polyfill/test/validStrings.mjs
jessealama: ah, OK, I think I understand now; there was some confusion on my end. The use of "name" here made me think that we weren't talking about numbers, but rather natural language names. It seems weird to me to call "-08:00" a name (I'd just call it a number or numeric or numerical offset).

production restored

spec/abstractops.html
ptomato: I don't think this is completely done yet, now TimeZoneIANAName duplicates TemporalTimeZoneIdentifier, I'd recommend either removing TimeZoneIANAName, or just making it equal to TimeZoneIANANameTail as it was before.

Additionally as Richard mentioned, the two alternatives inside TemporalTimeZoneIdentifier of TimeZoneUTCOffsetName and TimeZoneNumericUTCOffset expand to the same things.

Finally, if you remove TimeZoneIANAName, you'll have to search for any references to it in the rest of the spec text and replace those with TemporalTimeZoneIdentifier.

jessealama: I've restored `TimeZoneIANAName`. It wasn't my intention to do such an invasive change. How do things look now?

polyfill/test/validStrings.mjs
jessealama: comment added

spec/abstractops.html
ptomato: Better. I think this is good enough to fix #1993. There is still something wrong with this grammar, but we have #1805 open for that, so I would be OK to consider it out of scope, especially since the slides I've prepared for plenary only mention the problem in #1993. @gibson042 are you OK with that?

ptomato (approved):

spec/abstractops.html
```diff
@@ -944,9 +952,9 @@ <h1>ISO 8601 grammar</h1>
           TimeZoneIANANameTail
 
       TimeZoneBracketedName :
-          TimeZoneIANAName
-          `Etc/GMT` ASCIISign Hour
           TimeZoneUTCOffsetName
+          `Etc/GMT` ASCIISign UnpaddedHour
+          TimeZoneIANANameTail but not `Etc/GMT` ASCIISign UnpaddedHour
```
gibson042, line 957: I'm concerned about the "Etc/GMT±\<hour>" text being excluded from |TemporalTimeZoneIdentifier|, which includes |TimeZoneIANAName| but not |TimeZoneBracketedName|. It would be better to move ``` `Etc/GMT` ASCIISign UnpaddedHour``` to the former, as suggested in #1993:
```
      TimeZoneIANAName :
          `Etc/GMT` ASCIISign UnpaddedHour
          TimeZoneIANANameTail but not `Etc/GMT` ASCIISign UnpaddedHour

      TimeZoneBracketedName :
          TimeZoneIANAName
          TimeZoneUTCOffsetName
```

gibson042 (changes requested):
The grammar appears to incompatible with operations such as ToRelativeTemporalObject (which attempts to parse a string that matched |TimeZoneIANAName| in ParseTemporalTimeZoneString against |TimeZoneNumericUTCOffset|, an operation that is guaranteed to fail because |TimeZoneIANAName| can never start with a sign). However, that's a preexisting and mostly unrelated issue (#1805?). I'm willing to :+1: this one after making the fix to recognize "Etc/GMT±\<hour>" strings as naming IANA time zones.

linusg (commented):
I don't know if there's a general policy around this, but from an implementor's perspective I'd kindly like to ask for these commits to be squashed on merge, we usually refer to individual commits for spec related implementation changes ([see here for an example](https://github.com/SerenityOS/serenity/pull/12990/commits)), and in the current form of this PR that's impossible to do. Thanks :)

jessealama:
> I don't know if there's a general policy around this, but from an implementor's perspective I'd kindly like to ask for these commits to be squashed on merge, we usually refer to individual commits for spec related implementation changes ([see here for an example](https://github.com/SerenityOS/serenity/pull/12990/commits)), and in the current form of this PR that's impossible to do. Thanks :)

Sure, no problem; I can squash it. (I myself am not sure what the policy is about that.)

ptomato:
This achieved consensus at the March 2022 TC39 plenary meeting.

PR #2063, opened 2022-02-16
Port some PlainTime tests.

Ms2ger:


ptomato (approved):
Most of the coverage changes that codecov reports are just due to different reporters as usual, except one: https://app.codecov.io/gh/tc39/proposal-temporal/compare/2063/tree/polyfill/lib/plaindatetime.mjs#D1L645
Looks like this removes coverage for PlainDateTime.prototype.toPlainTime.

(Not that a PlainTime method was the best place for that coverage, though? Maybe that just means we should note that we should be writing a test for that method elsewhere.)

To get the CI green, the linter has an issue and I guess we need to adapt 3f2b7586f2893c569f213f751cf5c5693b1ea7a9 to cover Node 16.x as well.

Ms2ger:
Nice catch, I checked that we had a test for `PlainTime.from(PlainDateTime)`, but didn't foresee us losing coverage from the expected result. I restored that test and will cover it in my next batch.

PR #2065, opened 2022-02-18
Port some PlainTime tests.

Ms2ger:


Ms2ger:
Ref https://github.com/tc39/test262/pull/3415

ptomato (approved):
:+1:, once we get the correct commit hash from merging the test262 PR

PR #2071, opened 2022-02-22
Port PlainTime tests.

Ms2ger:


Ms2ger:
https://github.com/tc39/test262/pull/3417

ptomato (approved):
:+1:, just waiting on the commit hash when the test262 PR is merged.

The drop in coverage looks alarming, but I verified it's only due to the way the two different coverage tools count line coverage.

PR #2072, opened 2022-02-22
Remove incorrect text from `withCalendar` docs

justingrant:
This PR removes leftover text in the docs for `withCalendar` in PlainDate and PlainDateTime that is no longer true after we disallowed `.with({calendar})`.


ptomato (approved):
Absolutely right, thanks.

PR #2084, opened 2022-03-10
Editorial: Add missing check in ParseTemporalInstant.

Ms2ger:
Without this check, we hit an assertion failure in CreateTemporalInstant, so I'm considering the change to be editorial.


ptomato (approved):
Looks correct, thanks!

PR #2085, opened 2022-03-10
Editorial: Replace dead code by an assertion.

Ms2ger:


ptomato (approved):
I believe this is correct because all alternatives for [_TimeZoneNumericUTCOffset_](https://tc39.es/proposal-temporal/#prod-TimeZoneNumericUTCOffset) contain _TimeZoneUTCOffsetSign_ and _TimeZoneUTCOffsetHour_.

PR #2086, opened 2022-03-10
Editorial: Use CreateTemporalCalendar in GetBuiltinCalendar.

Ms2ger:


ptomato (approved):
I believe this is OK, but note that it switches from using Construct to using OrdinaryCreateFromConstructor. To the best of my knowledge, these are equivalent because Construct calls the function object's internal [[Construct]] method, which calls the Temporal.Calendar constructor function, which in the end also calls CreateTemporalCalendar and OrdinaryCreateFromConstructor. But I seem to remember @FrankYFTang had some doubts about this. If there is inadvertently an observable effect of this change, we can revert this.

PR #2087, opened 2022-03-11
Remove tests ported to test262

jessealama:
References:

+ https://github.com/tc39/test262/pull/3430


ptomato:
Works for me, but should wait until the test262 PR is merged upstream and update the test262 submodule as well.

Ms2ger (approved):

PR #2088, opened 2022-03-11
Port Duration tests.

Ms2ger:


jessealama:
Looks like coverage went down, despite updating the test262 submodule.

Ms2ger:
Coverage changes seem to be caused by the different coverage tools

ptomato:
Yeah, it's caused by things like this
Screenshot from 2022-03-11 11-40-47
where the one tool marks all lines spanned by a single statement as covered, and the other tool marks only the beginning line. I can't remember exactly why we couldn't use the same tool for both, but I did try for a long time to get it to work and eventually gave up.

PR #2089, opened 2022-03-11
Editorial: Get on the result of PrepareTemporalFields can't fail.

Ms2ger:


spec/calendar.html
```diff
@@ -459,8 +459,9 @@ <h1>ResolveISOMonth ( _fields_ )</h1>
         It returns the integer month.
       </p>
       <emu-alg>
-        1. Let _month_ be ? Get(_fields_, *"month"*).
-        1. Let _monthCode_ be ? Get(_fields_, *"monthCode"*).
+        1. Note: _fields_ is created in PrepareTemporalFields, so the Get calls below can not fail.
```
ljharb, line 462: that's not something that's possible for this AO to know for certain, since anything could call it. It may be true that currently the only callers of ResolveISOMonth meet this contract, but that doesn't mean it's something this AO can assert.

ptomato: I get this concern, but on the other hand it would mean keeping a `?` in a place where it would be untestable. It would be impossible to write a test262 test that could observe an exception being thrown by these steps.

Would it address your concern if it was "Assert: _fields_ was created by PrepareTemporalFields" or "Assert: _fields_ is an ordinary object with no more and no less than the own data properties listed in Table 13" or something like that?

Ms2ger: @bakkot since this was in response to your review, do you have thoughts on this?

bakkot: I disagree with Jordan on this question; this note seems perfectly fine to me.

@ljharb would you be satisfied if this line were an assertion instead of a note? We have existing assertions along the lines of "x is an ordinary object" which justify the use of `!` in an algorithm; this seems no stranger.

ptomato: I made two suggestions above, would either of those work? Particularly "Assert: _fields_ is an ordinary object with no more and no less than the own data properties listed in [Table 13](https://tc39.es/proposal-temporal/#table-temporal-field-requirements)" seems to me like it explicitly mentions the reason why the Get wouldn't throw.

bakkot: Either of those suggestions seem fine to me, as does simply replacing the Note with an Assert with the current wording.

spec/calendar.html
```diff
@@ -459,8 +459,9 @@ <h1>ResolveISOMonth ( _fields_ )</h1>
         It returns the integer month.
       </p>
       <emu-alg>
-        1. Let _month_ be ? Get(_fields_, *"month"*).
-        1. Let _monthCode_ be ? Get(_fields_, *"monthCode"*).
+        1. Assert: _fields_ is created in PrepareTemporalFields, so the Get calls below can not fail.
```
ljharb, line 462:
```suggestion
        1. Assert: _fields_ is an ordinary object with no more and no less than the own data properties listed in [Table 13](https://tc39.es/proposal-temporal/#table-temporal-field-requirements).
```
this seems clearer about why it can't throw, without needlessly coupling it to the name of its calling AO.

ljharb (approved):

ptomato (approved):

spec/calendar.html
```diff
@@ -459,8 +459,9 @@ <h1>ResolveISOMonth ( _fields_ )</h1>
         It returns the integer month.
       </p>
       <emu-alg>
-        1. Let _month_ be ? Get(_fields_, *"month"*).
-        1. Let _monthCode_ be ? Get(_fields_, *"monthCode"*).
+        1. Assert: _fields_ is an ordinary object with no more and no less than the own data properties listed in [Table 13](https://tc39.es/proposal-temporal/#table-temporal-field-requirements).
```
ryzokuken, line 462: I'm not sure this is correct, I think you should use `emu-xref` and let it autofill the link title for you.

ryzokuken (commented):
I cannot request changes anymore, but perhaps we could do this fix in a follow-up if it's acceptable?

spec/calendar.html
ptomato: Indeed, Linus spotted it: #2101

PR #2091, opened 2022-03-14
Remove tests ported to test262

ptomato:
See https://github.com/tc39/test262/pull/3431


ptomato:
The test-polyfill-node16 job failed to upload the codecov results, which is why the coverage is so off. I'm not sure how to rerun only one job in the GitHub Actions UI. I only see a button to rerun all of them.

jessealama (approved):

PR #2095, opened 2022-03-15
Some minor editorial changes

ptomato:


spec/abstractops.html
```diff
@@ -1319,14 +1319,18 @@ <h1>ParseTemporalDateTimeString ( _isoString_ )</h1>
...
-  <emu-clause id="sec-temporal-parsetemporaldurationstring" aoid="ParseTemporalDurationString">
-    <h1>ParseTemporalDurationString ( _isoString_ )</h1>
-    <p>
-      The ParseTemporalDurationString abstract operation parses an ISO 8601 string and returns the information needed to construct a Temporal.Duration instance.
-    </p>
+  <emu-clause id="sec-temporal-parsetemporaldurationstring" type="abstract operation">
+    <h1>
+      ParseTemporalDurationString (
+        _isoString_: a String,
+      )
```
ljharb, line 1326:
```suggestion
      ): a Duration Record
```

Ms2ger (approved):

spec/abstractops.html
ptomato: My understanding from skimming https://github.com/tc39/ecma262/pull/2547 is that it would have to be "either a normal completion containing a Duration Record, or an abrupt completion"? Regardless, I haven't yet put in the time to understand that change very well, so I'd prefer to leave this for now until I understand it better.

PR #2101, opened 2022-03-16
Editorial: Fix rendering of table reference in ResolveISOMonth

linusg:
The markdown link is not understood by ecmarkup, and the table number might change in the future.

image


ptomato (approved):
Thanks for catching this.

PR #2102, opened 2022-03-17
More editorial Duration stuff

ptomato:
Follow-ups from #2094.


spec/duration.html
```diff
@@ -1489,7 +1489,7 @@ <h1>
         1. If _largestUnit_ is not one of *"year"*, *"month"*, *"week"*, or *"day"*, then
           1. Let _diffNs_ be ! DifferenceInstant(_relativeTo_.[[Nanoseconds]], _endNs_, 1, *"nanosecond"*, *"halfExpand"*).
```
Ms2ger, line 1490: Maybe it would be helpful to document your reasoning in-spec, like with "Assert: abs(_diffNs_) ≤ 1.728 × 10<sup>22</sup>."

Ms2ger (approved):

jessealama (approved):

spec/duration.html
ptomato: Good idea.

Ms2ger:
Filed https://github.com/tc39/ecmarkup/issues/421 for the lint failure

ptomato:
Oh, I missed the lint failure - thanks for rephrasing the asserts.

PR #2122, opened 2022-03-25
Use canonical spelling for EditorConfig file

Alhadis:
This PR renames `.editorConfig` to the [canonical lowercase spelling](https://editorconfig.org/#file-location), `.editorconfig`. The camelCase filename may be handled inconsistently between editors and tooling on case-sensitive filesystems.


ptomato (approved):
Thanks!

PR #2123, opened 2022-03-25
Drop PlainDateTime tests ported to test262

jessealama:
This is another step in the direction of having the test suite in test262, rather than in this repo.


ptomato (approved):
I've pulled in the new test262 commit.

PR #2133, opened 2022-04-06
Editorial: Handle undefined options in MergeLargestUnitOption

linusg:
Also converted the AO to use a structured header.

image


ptomato (approved):
Looks great, and thanks for helping to incrementally introduce structured headers.

PR #2142, opened 2022-04-15
Drop `from` tests ported to Test262

jessealama:
Goes along with [this](https://github.com/tc39/test262/pull/3487) PR on the Test262 side.


ptomato:
I updated the test262 commit.

ptomato (approved):

PR #2144, opened 2022-04-15
Polyfill: disallow calendar in Instant strings.

Ms2ger:
Fixes #2143.

### DIFF
--- a/polyfill/lib/ecmascript.mjs
+++ b/polyfill/lib/ecmascript.mjs
@@ -314,6 +314,8 @@ export const ES = ObjectAssign({}, ES2020, {
     };
   },
   ParseTemporalInstantString: (isoString) => {
+    const match = PARSE.instant.exec(isoString);
+    if (!match) throw new RangeError(`invalid ISO 8601 string: ${isoString}`);
     const result = ES.ParseISODateTime(isoString);
     if (!result.z && !result.offset) throw new RangeError('Temporal.Instant requires a time zone offset');
     return result;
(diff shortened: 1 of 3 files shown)


polyfill/lib/ecmascript.mjs
```diff
@@ -314,6 +314,8 @@ export const ES = ObjectAssign({}, ES2020, {
     };
   },
   ParseTemporalInstantString: (isoString) => {
+    const match = PARSE.instant.exec(isoString);
+    if (!match) throw new RangeError(`invalid ISO 8601 string: ${isoString}`);
```
ptomato, line 318: It might be easier to check after ParseISODateTime that `result.calendar` is undefined? That way there would be a better error message and additionally we wouldn't have to define yet another complicated regex.

polyfill/test/instant.mjs
```diff
@@ -217,7 +217,7 @@ describe('Instant', () => {
       equal(`${Instant.from('1976-11-18T15Z')}`, '1976-11-18T15:00:00Z');
     });
     it('ignores any specified calendar', () =>
```
ptomato, line 219: The description of this test should probably change.

ptomato (changes requested):

polyfill/lib/ecmascript.mjs
justingrant: Is this actually a polyfill bug not a spec bug?  See https://github.com/tc39/proposal-temporal/issues/2143#issuecomment-1100295870

ptomato:
We decided that #2143 is a spec bug that should be fixed instead.

PR #2147, opened 2022-04-21
Remove tests from Demitasse that landed in test262

ptomato:
Landed in tc39/test262#3490.


ptomato:
I checked all the changed coverage lines and they seem to be artifacts of the different coverage instrumentation between Demitasse and test262, as usual.

Ms2ger (approved):

PR #2148, opened 2022-04-22
Drop tests ported to Test262

jessealama:
Goes with [this](https://github.com/tc39/test262/pull/3497) PR on the Test262 side.


ptomato:
It looks like everything that these tests cover is already covered by existing test262 tests, so I think we can just merge this.

ptomato (approved):

PR #2163, opened 2022-04-28
Editorial: Fix errors introduced by pull/1928

FrankYFTang:
see my comments in https://github.com/tc39/proposal-temporal/pull/1928#issuecomment-1112599130
@ryzokuken @ptomato

### DIFF
--- a/spec/intl.html
+++ b/spec/intl.html
@@ -1703,7 +1703,8 @@
             1. If _calendar_.[[Identifier]] is *"iso8601"*, then
               1. Let _result_ be ? AddISODate(_date_.[[ISOYear]], _date_.[[ISOMonth]], _date_.[[ISODay]], _duration_.[[Years]], _duration_.[[Months]], _duration_.[[Weeks]], _balanceResult_.[[Days]], _overflow_).
             1. Else,
-              1. Let _result_ be ! CalendarDateAddition(_calendar_, _date_, _balanceResult_, _overflow_).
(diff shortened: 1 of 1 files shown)


spec/intl.html
```diff
@@ -1703,7 +1703,8 @@ <h1>Temporal.Calendar.prototype.dateAdd ( _date_, _duration_ [ , _options_ ] )</
             1. If _calendar_.[[Identifier]] is *"iso8601"*, then
               1. Let _result_ be ? AddISODate(_date_.[[ISOYear]], _date_.[[ISOMonth]], _date_.[[ISODay]], _duration_.[[Years]], _duration_.[[Months]], _duration_.[[Weeks]], _balanceResult_.[[Days]], _overflow_).
             1. Else,
-              1. Let _result_ be ! CalendarDateAddition(_calendar_, _date_, _balanceResult_, _overflow_).
+              1. Let _balanceDuration_ be a new Record { [[Years]]: _duration_.[[Years]], [[Months]]: _duration_.[[Months]], [[Weeks]]: _duration_.[[Weeks]], [[Days]]: _balanceResult_.[[Days]], [[Hours]]: _balanceResult_.[[Hours]], [[Minutes]]: _balanceResult_.[[Minutes]], [[Seconds]]: _balanceResult_.[[Seconds]], [[Milliseconds]]: _balanceResult_.[[Milliseconds]], [[Microseconds]]: _balanceResult_.[[Microseconds]], [[Nanoseconds]]: _balanceResult_.[[Nanoseconds]] }.
```
ptomato, line 1706:
```suggestion
              1. Let _balanceDuration_ be ! CreateDateDurationRecord(_duration_.[[Years]], _duration_.[[Months]], _duration_.[[Weeks]], _balanceResult_.[[Days]]).
```
(I think the error I overlooked is actually that CalendarDateAddition should take a Date Duration Record, not a Duration Record)

ptomato (approved):

spec/intl.html
FrankYFTang: That require changing the definition of CalendarDateAddition(). I will leave that for @ryzokuken to decide.

ryzokuken: I'd copied this from the existing AOs for the most part, but I agree with @ptomato here, his proposed solution seems quite close to the original text while fixing the bug.

ryzokuken (approved):
Thanks for this, check my comment on the previous review.

ryzokuken (approved):

PR #2168, opened 2022-05-02
Port some Duration.compare tests.

Ms2ger:


ptomato (approved):
Checked the coverage, it seems fine; as usual the discrepancies are caused by the difference between v8's internal coverage and `nyc instrument`

PR #2170, opened 2022-05-03
Update test262.

Ms2ger:


ptomato:
Looks like I broke some tests. I've opened https://github.com/tc39/test262/pull/3518 to fix it, and in the meantime will rewind this to point to an earlier test262 commit.

ptomato (approved):

PR #2181, opened 2022-05-06
Normative typo: Fix off-by-one in Temporal.Calendar.prototype.dayOfYear()

linusg:
Unlike the removed ToISODayOfYear AO, ECMA-262's DayWithinYear AO is zero-based.

Regression introduced in 6117d90.

---

If this is considered editorial again, please just reword the commit while squashing if necessary - still learning the rules there :)


ptomato (approved):
Oh, thanks for catching this. I took it into account with WeekDay but not with DayWithinYear.

Indeed, this is also not a normative change. It just preserves the status quo.

PR #2182, opened 2022-05-07
Editorial: use 2 × nsMaxInstant for the maximum difference of two Instants

ptomato:
This replaces more magic numbers with meaningful constants. (I missed this when I was grepping for "864" and "8.64")


linusg (commented):
@ptomato these could also be used in ParseTemporalInstant for the range check

gibson042 (approved):

Ms2ger (approved):

ptomato:
> @ptomato these could also be used in ParseTemporalInstant for the range check

Good catch, but that check is actually incorrect and I'm preparing a separate normative change to fix it.

PR #2185, opened 2022-05-09
Editorial: Compare Parse Node positions using "contained within" rather than "enclosed by"

gibson042:
cf. https://tc39.es/ecma262/multipage/text-processing.html#sec-countleftcapturingparenswithin


ptomato (approved):
The precedent you linked to in ECMA-262 makes sense, :+1:

PR #2189, opened 2022-05-10
Normative: Remove check for Instant range before subtracting UTC offset

ptomato:
This check was erroneous. It would throw if the wall-clock time
corresponding to the Instant, taken as if it were UTC, would be out of
range. We don't care about that, we only care if the actual Instant is out
of range.

This would throw when parsing the following valid Instant strings:

    Temporal.Instant.from('-271821-04-19T23:00-01:00').epochSeconds === -86400e8
    Temporal.Instant.from('+275760-09-13T01:00+01:00').epochSeconds === 86400e8

The reference polyfill requires a slightly different fix, because it uses
Date math in GetEpochFromISOParts. In this regard, it doesn't follow the
spec exactly, but the difference should not be observable.


Ms2ger (approved):
Makes sense. Do you have tests as well?

PR #2194, opened 2022-05-12
Editorial: Correct mistake made in pull/1928

FrankYFTang:
Overflow should be a String in https://github.com/tc39/proposal-temporal/pull/1928
@ryzokuken


ptomato (approved):
Thanks!

PR #2211, opened 2022-05-20
Polyfill: Lunisolar math fixes

justingrant:
Port of https://github.com/js-temporal/temporal-polyfill/pull/153


Ms2ger (approved):
I'll have to trust you on this. Feel free to land unless @ptomato objects

ptomato (approved):
The newly added tests for the Chinese calendar are failing in Node 12 because of lack of `relatedYear`. But, I still think we have no need to support Node 12 with this code because it's only being used to develop the proposal and run the test262 tests. So, no objection from me.

justingrant:
> The newly added tests for the Chinese calendar are failing in Node 12 because of lack of `relatedYear`.

Oops!  I just pushed a new commit that skips those tests for old Node. All green now. Merging.

PR #2243, opened 2022-05-29
Editorial: Align GetOption with ECMA-402

### LINKED ISSUE #2242: Temporal GetOption is incompatible with ECMA-402
[ECMA-402 GetOption](https://tc39.es/ecma402/#sec-getoption) expects a singular string-valued _type_ and \*undefined\* as the special "ignore" value for _values_, while the current [Temporal GetOption](https://tc39.es/proposal-temporal/#sec-getoption) expects _types_ to be a List of Type values and \~empty~ to be the special "ignore" value for _values_. As a result, the invocations in [Temporal InitializeDateTimeFormat](https://tc39.es/proposal-temporal/#sec-temporal-initializedatetimeformat), which still use the ECMA-402-style arguments, are incoherent.

The replacement of \*undefined\* with \~empty~ is easy to correct, and it turns out that the expansion of _type_ to _types_ is only used by [GetStringOrNumberOption](https://tc39.es/proposal-temporal/#sec-getstringornumberoption), which itself is only used by [ ToSecondsStringPrecision](https://tc39.es/proposal-temporal/#sec-temporal-tosecondsstringprecision).

gibson042:
Fixes #2242

### DIFF
--- a/spec/abstractops.html
+++ b/spec/abstractops.html
@@ -45,7 +45,7 @@
     </emu-alg>
   </emu-clause>
 
-  <!-- Copied from ECMA-402 GetOptions -->
+  <!-- Copied from ECMA-402 GetOption -->
   <emu-clause id="sec-getoption" type="abstract operation">
     <h1>
       GetOption (
--- a/spec/abstractops.html
+++ b/spec/abstractops.html
@@ -282,8 +282,8 @@
             [[Unit]]: *"nanosecond"*,
             [[Increment]]: 1
           }.
-      1. If _fractionalDigitsVal_ is *NaN*, *+&infin;*<sub>𝔽</sub>, or *-&infin;*<sub>𝔽</sub>, throw a *RangeError* exception.
-      1. Set _fractionalDigitCount_ to ! ToIntegerOrInfinity(_fractionalDigitsVal_).
+      1. If _fractionalDigitsVal_ is *NaN*, throw a *RangeError* exception.
+      1. Set _fractionalDigitCount_ to ? ToIntegerThrowOnInfinity(_fractionalDigitsVal_).
       1. If _fractionalDigitCount_ &lt; 0 or _fractionalDigitCount_ &gt; 9, throw a *RangeError* exception.
       1. If _fractionalDigitCount_ is 0, then
         1. Return the Record {
--- a/spec/abstractops.html
+++ b/spec/abstractops.html
@@ -289,13 +289,18 @@
             [[Increment]]: 1
           }.
       1. Assert: _smallestUnit_ is *undefined*.
-      1. Let _fractionalDigitCount_ be ? GetStringOrNumberOption(_normalizedOptions_, *"fractionalSecondDigits"*, « *"auto"* », 0, 9, *"auto"*).
-      1. If _fractionalDigitCount_ is *"auto"*, then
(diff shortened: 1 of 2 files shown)


ljharb:
Why this direction, instead of updating 402?

gibson042:
This proposal is introducing into ECMA-262 an operation that currently exists in ECMA-402 and is also modifying an algorithm in ECMA-402 that already uses the existing operation, so this minimizes necessary churn across the three repositories.

We could update ECMA-402 to change GetOption inputs from language values to spec values and then rebase Temporal atop it, but that seems like more effort for comparable benefit so I instead opted for the approach that doesn't require coordination.

ljharb:
We've already updated 402's GetOption once, iirc, to make it match the preferred one from Temporal - that seems the better approach this time as well.

gibson042:
Fair enough: https://github.com/tc39/ecma402/pull/687

This PR is still good now, but GetOption can be further updated after that PR lands.

gibson042:
Updated to adopt the new [ECMA-402 GetOption](https://tc39.es/ecma402/#sec-getoption) directly and move it into ECMA-262.

spec/abstractops.html
```diff
@@ -192,7 +176,7 @@ <h1>ToShowOffsetOption ( _normalizedOptions_ )</h1>
       It is different from the `offset` property passed to `Temporal.ZonedDateTime.from()`.
     </emu-note>
     <emu-alg>
-      1. Return ? GetOption(_normalizedOptions_, *"offset"*, « String », « *"auto"*, *"never"* », *"auto"*).
+      1. Return ? GetOption(_normalizedOptions_, *"offset"*, *"string"*, « *"auto"*, *"never"* », *"auto"*).
```
ljharb, line 179: why are these strings and not spec enums?

spec/abstractops.html
```diff
@@ -288,36 +272,41 @@ <h1>ToSecondsStringPrecision ( _normalizedOptions_ )</h1>
...
+        1. If _fractionalDigitsVal_ is not *undefined*, then
+          1. If ? ToString(_fractionalDigitsVal_) is not *"auto"*, throw a *RangeError* exception.
         1. Return the Record {
             [[Precision]]: *"auto"*,
             [[Unit]]: *"nanosecond"*,
             [[Increment]]: 1
           }.
-      1. If _digits_ is 0, then
+      1. If _fractionalDigitsVal_ is *NaN*, *+&infin;*<sub>𝔽</sub>, or *-&infin;*<sub>𝔽</sub>, throw a *RangeError* exception.
+      1. Set _fractionalDigitCount_ to ? ToIntegerOrInfinity(_fractionalDigitsVal_).
```
ptomato, line 285:
```suggestion
      1. Set _fractionalDigitCount_ to ! ToIntegerOrInfinity(_fractionalDigitsVal_).
```

ptomato (approved):
This is all good with me. I slightly prefer the explicitness of GetStringOrNumberOption for handling fractionalSecondDigits rather than inlining it, but I don't feel strongly about it.

spec/abstractops.html
gibson042: Because that is the state of ECMA-402 until https://github.com/tc39/ecma402/pull/689 lands.

PR #2250, opened 2022-06-04
Editorial: Improve the definition of calendar type canonical form

### LINKED ISSUE #2249: The canonical form of a calendar type is inaccurately described
[Calendar Types](https://tc39.es/proposal-temporal/#sec-calendar-types) includes the following:
> canonical form is a string containing all lower case letters with zero or more hyphens

However, "iso8601" (which is not only valid but _required_) includes characters that are digits rather than lower case letters. Further, note the imprecision of "letter" and "hyphen" (which are actually limited to the ASCII range by [UTS #35](https://unicode.org/reports/tr35/tr35.html#Unicode_Language_and_Locale_Identifiers)).

gibson042:
Fixes #2249

### DIFF
--- a/spec/calendar.html
+++ b/spec/calendar.html
@@ -20,7 +20,7 @@
     </p>
     <p>
       ECMAScript implementations identify built-in calendars using a <dfn variants="calendar types">calendar type</dfn> as defined by <a href="https://unicode.org/reports/tr35/tr35-dates.html#Calendar_Elements">Unicode Technical Standard #35, Part 4, Section 2</a>.
-      Their canonical form is a string containing all lower case letters with zero or more hyphens.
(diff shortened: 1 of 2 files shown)


ptomato (approved):
This is already an improvement, so fine to merge.

Optionally, should we mention here that they consist of components from 2 to 8 lowercase letters and digits separated by medial hyphens, or is it fine to leave that to the grammar to define? The current wording does not exclude `a-b-c-d-e`, for example.

gibson042:
> Optionally, should we mention here that they consist of components from 2 to 8 lowercase letters and digits separated by medial hyphens, or is it fine to leave that to the grammar to define? The current wording does not exclude `a-b-c-d-e`, for example.

I think that's sufficiently esoteric to stay in the grammar.

PR #2251, opened 2022-06-05
Editorial: Rename DefaultMergeFields to DefaultMergeCalendarFields

gibson042:
For clarity in the resulting state of ECMA-262.

### DIFF
--- a/spec/calendar.html
+++ b/spec/calendar.html
@@ -615,11 +615,17 @@
       </emu-alg>
     </emu-clause>
 
-    <emu-clause id="sec-temporal-defaultmergefields" aoid="DefaultMergeCalendarFields">
-      <h1>DefaultMergeCalendarFields ( _fields_, _additionalFields_ )</h1>
-      <p>
-        The abstract operation DefaultMergeCalendarFields implements the default logic for the `Temporal.Calendar.mergeFields` method, which is used as a fallback if the method is not present on a calendar, and also for the ISO 8601 calendar.
-      </p>
+    <emu-clause id="sec-temporal-defaultmergefields" type="abstract operation">
+      <h1>
+        DefaultMergeCalendarFields (
+          _fields_: an Object,
+          _additionalFields_: an Object,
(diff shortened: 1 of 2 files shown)


spec/calendar.html
```diff
@@ -615,11 +615,17 @@ <h1>ISODay ( _temporalObject_ )</h1>
       </emu-alg>
     </emu-clause>
 
-    <emu-clause id="sec-temporal-defaultmergefields" aoid="DefaultMergeFields">
-      <h1>DefaultMergeFields ( _fields_, _additionalFields_ )</h1>
-      <p>
-        The abstract operation DefaultMergeFields implements the default logic for the `Temporal.Calendar.mergeFields` method, which is used as a fallback if the method is not present on a calendar, and also for the ISO 8601 calendar.
-      </p>
+    <emu-clause id="sec-temporal-defaultmergefields" type="abstract operation">
```
ptomato, line 618:
```suggestion
    <emu-clause id="sec-temporal-defaultmergecalendarfields" type="abstract operation">
```

spec/calendar.html
```diff
@@ -615,11 +615,17 @@ <h1>ISODay ( _temporalObject_ )</h1>
...
+      <h1>
+        DefaultMergeCalendarFields (
+          _fields_: an Object,
+          _additionalFields_: an Object,
+        ): either a normal completion containing an Object or an abrupt completion
+      </h1>
+      <dl class="header">
+        <dt>description</dt>
+        <dd>It implements the default logic for the `Temporal.Calendar.mergeFields` method, which is used for the ISO 8601 calendar and as a fallback if the method is not present on a calendar.</dd>
```
ptomato, line 627:
```suggestion
        <dd>It implements the default logic for the `Temporal.Calendar.prototype.mergeFields` method, which is used for the ISO 8601 calendar and as a fallback if the method is not present on a calendar.</dd>
```

ptomato (approved):

PR #2253, opened 2022-06-06
Editorial: Remove unnecessary ToLimitedTemporalDuration

FrankYFTang:
There is only one caller to ToLimitedTemporalDuration so there are no need to make it general and pass in a _disallowedFields_


FrankYFTang:
@ptomato

gibson042 (approved):
I like this simplification.

Ms2ger (approved):

PR #2255, opened 2022-06-08
Editorial: Consolidate partial operations

gibson042:
There was a lot of redundancy between operations that construct "partial" vs. complete objects/records. These changes consolidate them with argument-based control, and leverage a single base PrepareTemporalFields wherever possible.


ptomato (approved):
Thanks!

PR #2260, opened 2022-06-08
Normative: Disallow one day long time zone offsets

anba:


ljharb:
Is this a timezone offset from UTC? If so, it can only range from -12 to +14 hours. If it's the difference between any two timezones, it could be as much as 26 hours.

anba:
It is the offset for a single time zone at a specific instant. The built-in time zone object supports up-to `nsPerDay - 1` long offsets:
```js
js> new Temporal.TimeZone("+23:59:59.999999999").getOffsetNanosecondsFor(new Temporal.Instant(0n))
86399999999999
```

ljharb:
Is the design intending to support possible future timezones with larger offsets than 14 hours? If so, why wouldn't "24 hours" be a valid value?

anba:
> If so, it can only range from -12 to +14 hours.

The supported range is actually even longer because of LMT (local mean time):
- https://github.com/eggert/tz/blob/0e8f0b06ac2f65d0c54a91121da09c8662cd88e6/asia#L3518
- https://github.com/eggert/tz/blob/0e8f0b06ac2f65d0c54a91121da09c8662cd88e6/northamerica#L689

<details>

```js
const min = new Temporal.Instant(-86_40000_00000_00000_00000n);

let minOffset = +Infinity;
let maxOffset = -Infinity;
let minTimeZone;
let maxTimeZone;

for (let id of Intl.supportedValuesOf("timeZone")) {
  console.log(id);
  let tz = new Temporal.TimeZone(id);

  let instant = min;
  for (let i = 0; i < 1000; ++i) {
    let offset = tz.getOffsetNanosecondsFor(instant);

    if (offset < minOffset) {
      minOffset = offset;
      minTimeZone = id;
    }
    if (offset > maxOffset) {
      maxOffset = offset;
      maxTimeZone = id;
    }

    let next = tz.getNextTransition(instant);
    if (next === null) {
      break;
    }
    instant = next;
  }
}
console.log("minOffset:", minOffset, minTimeZone);
console.log("maxOffset:", maxOffset, maxTimeZone);
```
</details>

ljharb:
Is it up to 23 hours and 59 minutes, or a smaller value?

I'm trying to understand why 24 hours is a magic illegal number but "24 hours - 1 minute" is fine.

anba:
> Is it up to 23 hours and 59 minutes, or a smaller value?

The maximum for built-in time zone objects is "+23:59:59.999999999" (and "-23:59:59.999999999" in the other direction). This can happen only when the time zone is created from an [offset string](https://tc39.es/proposal-temporal/#sec-temporal-parsetimezoneoffsetstring).

ljharb:
Thanks, it all makes sense now :-)

ptomato (approved):
Thanks, good catch.

Would you mind splitting out the editorial change into a separate pull request so we can merge it ahead of presenting the normative change?

ptomato:
Tests are in https://github.com/tc39/test262/pull/3669.

PR #2264, opened 2022-06-08
Editorial: Simplify ApplyUnsignedRoundingMode

FrankYFTang:
We know from the caller of ApplyUnsignedRoundingMode that x is >= 0 , and therefore r1 >= 0 and r2 = r1 + 1
so r2-r1 is always 1. Therefore, it is always true that  r1 / (r2-r1) = r1 / 1 = r1. The "/ (r2-r1)" is unnecessary.

@ptomato


linusg:
> EDITOR'S NOTE: This operation is intended to be the same one as in the [Intl.NumberFormat v3](https://tc39.es/proposal-intl-numberformat-v3/out/numberformat/diff.html#sec-applyunsignedroundingmode) proposal.

PR #2265, opened 2022-06-09
Normative: Validate results of Calendar operations

anba:
The validation was already mentioned in the preamble of each operation, but it wasn't actually performed.

### DIFF
--- a/spec/calendar.html
+++ b/spec/calendar.html
@@ -207,6 +207,7 @@
         1. Let _result_ be ? Invoke(_calendar_, *"dayOfWeek"*, « _dateLike_ »).
         1. Return ? ToPositiveInteger(_result_).
       </emu-alg>
+      <emu-note>ToPositiveInteger(*undefined*) will throw a *RangeError*.</emu-note>
     </emu-clause>
 
     <emu-clause id="sec-temporal-calendardayofyear" aoid="CalendarDayOfYear">
@@ -219,6 +220,7 @@
         1. Let _result_ be ? Invoke(_calendar_, *"dayOfYear"*, « _dateLike_ »).
         1. Return ? ToPositiveInteger(_result_).
       </emu-alg>
+      <emu-note>ToPositiveInteger(*undefined*) will throw a *RangeError*.</emu-note>
     </emu-clause>
 
     <emu-clause id="sec-temporal-calendarweekofyear" aoid="CalendarWeekOfYear">
@@ -231,6 +233,7 @@
         1. Let _result_ be ? Invoke(_calendar_, *"weekOfYear"*, « _dateLike_ »).
         1. Return ? ToPositiveInteger(_result_).
       </emu-alg>
+      <emu-note>ToPositiveInteger(*undefined*) will throw a *RangeError*.</emu-note>
     </emu-clause>
 
     <emu-clause id="sec-temporal-calendardaysinweek" aoid="CalendarDaysInWeek">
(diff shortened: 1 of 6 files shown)


spec/calendar.html
```diff
@@ -167,7 +167,6 @@ <h1>CalendarMonth ( _calendar_, _dateLike_ )</h1>
       <emu-alg>
         1. Assert: Type(_calendar_) is Object.
         1. Let _result_ be ? Invoke(_calendar_, *"month"*, « _dateLike_ »).
-        1. If _result_ is *undefined*, throw a *RangeError* exception.
```
ljharb, line 170: ToPositiveInteger won’t throw if the result is undefined - why remove this check?

anba: The commit message for that change has a more detailed explanation. Basically it's:
- `ToNumber(undefined)` -> NaN
- `ToIntegerOrInfinity(NaN)` -> 0
- `ToPositiveInteger(0)` -> throws a RangeError

ljharb: ah, i wouldn’t expect ToPositiveInteger to throw on zero.

gibson042: I agree, and this is in line with my suggestions at https://github.com/tc39/proposal-temporal/issues/2112#issuecomment-1148678997 .
> * **ToIntegerThrowOnInfinity** → **TruncateToInteger**
> * **ToIntegerWithoutRounding** → **InsistInteger**
> * **ToPositiveInteger** → **TruncateToPositiveInteger**

anba: > ah, i wouldn’t expect ToPositiveInteger to throw on zero.

I guess "positive" was used because in ECMA-262 "non-negative" is typically used when zero is included, whereas "positive" is (often) used when zero isn't included.

gibson042: I was also surprised, but not because zero is out of range—my surprise was because "To…" operations generally don't throw at all except when conversion is deemed impossible at a type-to-type level or to propagate exceptions from ECMAScript code, neither of which apply to ToPositiveInteger(0). Why should the behavior diverge from similarly-named operations, e.g. ToIntegerOrInfinity("foo") or ToIntegerOrInfinity(3.14) or ToLength(Infinity) or ToIndex(-1)?

That's why I want to make the names _dissimilar_ and more indicative of behavior.

ljharb: Since zeros have a sign in JS, I assumed +0 was included in "positive integers".

ptomato, line 170: Elsewhere when this is left implicit, we often add an emu-note saying something like "NOTE: ToPositiveInteger(undefined) will throw a RangeError" or something like that.

ptomato (approved):
Thanks! This can go on the list to present at the next plenary. I'll mark it as draft so it is not merged accidentally before then.

Would you mind splitting the second commit (removing the checks for undefined) into a separate editorial pull request?

spec/calendar.html
```diff
@@ -204,7 +204,8 @@ <h1>CalendarDayOfWeek ( _calendar_, _dateLike_ )</h1>
       </p>
       <emu-alg>
         1. Assert: Type(_calendar_) is Object.
-        1. Return ? Invoke(_calendar_, *"dayOfWeek"*, « _dateLike_ »).
+        1. Let _result_ be ? Invoke(_calendar_, *"dayOfWeek"*, « _dateLike_ »).
+        1. Return ? ToPositiveInteger(_result_).
```
ptomato, line 233: I realized that this changes the return type of these operations, so we need to add an extra 𝔽() around the places where we return this value to JS.

ptomato:
This reached consensus in July 2022. I've rebased it and addressed my comment about 𝔽(), and added an implementation of the normative change to the reference code. Tests are in https://github.com/tc39/test262/pull/3681, we'll need to update test262 once that's merged, since this invalidates some existing test262 tests.

PR #2274, opened 2022-06-09
Only call CanonicalizeTimeZoneName on valid time zone ids 

### LINKED ISSUE #2232: ToTemporalTimeZone: CanonicalizeTimeZoneName called with offset string
https://tc39.es/proposal-temporal/#sec-temporal-totemporaltimezone
> 4.c. Return ! [CreateTemporalTimeZone](https://tc39.es/proposal-temporal/#sec-temporal-createtemporaltimezone)(! [CanonicalizeTimeZoneName](https://tc39.es/proposal-temporal/#sec-canonicalizetimezonename)(parseResult.[[Name]])).

But `parseResult.[[Name]]` can be an offset string.

anba:
Fixes #2232


ptomato (approved):
Thanks, good catch.

PR #2275, opened 2022-06-09
Editorial: Add missing conversion to integer values

### LINKED ISSUE #2233: ResolveISOMonth: Missing conversion to integer
https://tc39.es/proposal-temporal/#sec-temporal-resolveisomonth
> 4.b. Return month.

Object properties are `Number` values, but this operation is spec'ed to return an integer. So it should instead be:
> 4.b. Return [ℝ](https://tc39.es/ecma262/#%E2%84%9D)(month).

anba:
Fixes #2233

Also added extra assertions, so it's clearer why the `ℝ(...)` conversions are valid.


gibson042 (approved):

ptomato (approved):
Thanks!

PR #2287, opened 2022-06-10
Normative: Reject an ambiguous time string even with a calendar

gibson042:
Fixes #2285

### DIFF
--- a/polyfill/lib/ecmascript.mjs
+++ b/polyfill/lib/ecmascript.mjs
@@ -18,9 +18,9 @@ const ObjectDefineProperty = Object.defineProperty;
 const ObjectGetOwnPropertyDescriptor = Object.getOwnPropertyDescriptor;
 const ObjectIs = Object.is;
 const ObjectEntries = Object.entries;
-const StringPrototypeSlice = String.prototype.slice;
 
 import bigInt from 'big-integer';
+import callBound from 'call-bind/callBound';
 import Call from 'es-abstract/2020/Call.js';
 import GetMethod from 'es-abstract/2020/GetMethod.js';
 import IsInteger from 'es-abstract/2020/IsInteger.js';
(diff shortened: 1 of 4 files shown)


polyfill/lib/ecmascript.mjs
```diff
@@ -375,7 +373,7 @@ export const ES = ObjectAssign({}, ES2020, {
     // and must be stripped so presence of a calendar doesn't result in interpretation
     // of otherwise ambiguous input as a time.
     const isoStringWithoutCalendar = calendar
-      ? $stringSlice(isoString, 0, isoString.length - calendar.length - 7)
+      ? ES.Call(StringPrototypeSlice, isoString, [0, -(calendar.length + 7)])
```
ljharb, line 376: why is this an improvement? this means the intrinsic cache isn't maximally primed by the dep graph.

gibson042: It's for consistency with the rest of the polyfill, which makes use of ES.Call for close alignment with the spec. This PR is not the right place to introduce callBind, although a PR dedicated to that might be appropriate.

ljharb: oh sorry, i missed that your PR was adding it in the first place, nvm.

gibson042: It also occurs to me that #2288 or an immediate followup is quite possibly the place.

ptomato (approved):
Agreed with the reasoning in #2285. Let's present this at the next TC39, marking as draft until then.

ptomato:
This gained consensus at the July 2022 TC39 plenary. Tests are now available in https://github.com/tc39/test262/pull/3641. I had some last minute doubts about whether this one was actually correct, but I'm convinced by the discussion in #2285.

PR #2291, opened 2022-06-13
Editorial: Remove RegulateISODate after BalanceISODate and inline one call to AddISODate

anba:


ptomato (approved):
Thanks. I had some trouble figuring out whether the change in AddISODate really preserved the same behaviour in all cases, but I think I convinced myself sufficiently.

PR #2293, opened 2022-06-13
Editorial: Revert normative changes from #2243

anba:
Ensure inputs like `-0.5` and `9.1` are still throwing a RangeError instead of being treated as `0`.

### DIFF
--- a/spec/abstractops.html
+++ b/spec/abstractops.html
@@ -282,9 +282,9 @@
             [[Unit]]: *"nanosecond"*,
             [[Increment]]: 1
           }.
-      1. If _fractionalDigitsVal_ is *NaN*, throw a *RangeError* exception.
-      1. Set _fractionalDigitCount_ to ? ToIntegerThrowOnInfinity(_fractionalDigitsVal_).
-      1. If _fractionalDigitCount_ &lt; 0 or _fractionalDigitCount_ &gt; 9, throw a *RangeError* exception.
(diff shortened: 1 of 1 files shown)


ptomato (changes requested):
Thanks for noticing this! But wouldn't it be better to make a clean revert of https://github.com/tc39/proposal-temporal/pull/2243/commits/541d19ed0352e2fac0d6c311e2f9f6fb5e3edb03? I'm not sure this does the same thing as ToIntegerOrInfinity.

anba:
A clean revert isn't really possible, because <https://github.com/tc39/proposal-temporal/commit/4dd6713dbbe8ccca58f6422d9e99cc912857dce5> also needs to be partially reverted. That commit added:
```diff
+ 1. Set _fractionalDigitCount_ to floor(ℝ(_fractionalDigitsVal_)).
+ 1. If _fractionalDigitCount_ &lt; 0 or _fractionalDigitCount_ &gt; 9, throw a *RangeError* exception.
```

whereas `GetStringOrNumberOption` performed:
> a. If value < minimum or value > maximum, throw a RangeError exception.
> b. Return floor(ℝ(value)).


Notice how both steps are switched.

anba:
The "range check followed by floor" pattern was copied from <https://tc39.es/ecma402/#sec-defaultnumberoption>.

ptomato (approved):
OK, got it. Thanks.

PR #2299, opened 2022-06-14
Drop some ported tests for Instant.

Ms2ger:


ptomato (approved):
:+1:, the code coverage change is an artifact of the different code coverage tools between the old tests and the test262 tests.

PR #2301, opened 2022-06-14
Editorial: MergeLargestUnitOption is always called with an Object options value

anba:
`MergeLargestUnitOption` is no longer called with `options == undefined`.


anba:
Oh, wait. `DifferenceZonedDateTime` still needs to be updated, too.

anba:
Okay, `DifferenceZonedDateTime` only needed its first two steps removed.

ptomato (approved):
Correct, thanks!

PR #2306, opened 2022-06-16
Editorial: Suppress linking to "remainder" in two cases

anba:


ptomato (approved):
Thanks! I never noticed these.

gibson042:
This seems like a general risk, so I've opened https://github.com/tc39/ecmarkup/pull/460 to fix it upstream.

PR #2316, opened 2022-06-17
Normative: Refactor CalendarFields for better linearity

gibson042:
...and to prevent unnecessarily-observable calls to iterator methods for purely internal purposes.


Ms2ger (commented):
I have no strong feelings on this either way. If you have a moment to change the polyfill, that's always appreciated.

ptomato (approved):
I don't feel strongly about this either. Happy to present it in July if none of the other champions have concerns about it.

ptomato:
Tests are in https://github.com/tc39/test262/pull/3682.

ptomato:
This reached consensus at the July 2022 TC39 meeting. I've added the corresponding change in the reference polyfill.

PR #2323, opened 2022-06-22
Editorial: Mark CreateTemporalTime in PlainTime.prototype.round as infallible

anba:
`RoundTime` always returns a valid time, so `CreateTemporalTime` can't fail.


ptomato (approved):
Agreed.

PR #2324, opened 2022-06-22
Editorial: Assert inputs are within the valid range for AddTime/RoundISODateTime/RoundTime

anba:


ptomato (approved):
More assert is more better, thanks!

PR #2325, opened 2022-06-22
Editorial: Add missing Number resp. Mathematical value conversions

anba:


ptomato (approved):
All agreed, thanks for catching these.

PR #2327, opened 2022-06-22
fix CreateTemporalDuration in RoundDuration

FrankYFTang:
should be ! instead of ?

### DIFF
--- a/spec/duration.html
+++ b/spec/duration.html
@@ -1592,7 +1592,7 @@
           1. Let _monthsWeeksInDays_ be DaysUntil(_yearsLater_, _yearsMonthsWeeksLater_).
           1. Set _relativeTo_ to _yearsLater_.
           1. Let _days_ be _days_ + _monthsWeeksInDays_.
-          1. Let _daysDuration_ be ? CreateTemporalDuration(0, 0, 0, _days_, 0, 0, 0, 0, 0, 0).
+          1. Let _daysDuration_ be ! CreateTemporalDuration(0, 0, 0, _days_, 0, 0, 0, 0, 0, 0).
           1. Let _daysLater_ be ? CalendarDateAdd(_calendar_, _relativeTo_, _daysDuration_, *undefined*, _dateAdd_).
           1. Let _untilOptions_ be OrdinaryObjectCreate(*null*).
           1. Perform ! CreateDataPropertyOrThrow(_untilOptions_, *"largestUnit"*, *"year"*).


FrankYFTang:
Comparing with other invokation of CreateTemporalDuration in RoundDuration I see no reason why this one is ? instead of !

FrankYFTang:
@gibson042 @ptomato

spec/duration.html
```diff
@@ -1592,7 +1592,7 @@ <h1>
           1. Let _monthsWeeksInDays_ be DaysUntil(_yearsLater_, _yearsMonthsWeeksLater_).
           1. Set _relativeTo_ to _yearsLater_.
           1. Let _days_ be _days_ + _monthsWeeksInDays_.
-          1. Let _daysDuration_ be ? CreateTemporalDuration(0, 0, 0, _days_, 0, 0, 0, 0, 0, 0).
+          1. Let _daysDuration_ be ! CreateTemporalDuration(0, 0, 0, _days_, 0, 0, 0, 0, 0, 0).
```
anba, line 1595: `monthsWeeksInDays` is guaranteed to be a small number, but `days` can be arbitrarily large, can't it? 

Step 6.e sets `days` to:
> 6.e. Set days to days + result.[[Days]] + result.[[Nanoseconds]] / result.[[DayLength]].

And `result.[[Days]]` can be larger than `Number.MAX_VALUE`, because [NanosecondsToDays](https://tc39.es/proposal-temporal/#sec-temporal-nanosecondstodays) can produce large mathemetical numbers in the loop in step 18:
1. A user-controlled time zone can return `oneDayFartherNs = nanoseconds + intermediateNs` in step 18.a.
1. For `sign = 1`, this will then lead to adding `days = days + 1`:
```
  (nanoseconds - dayLengthNs) × sign
= (nanoseconds - (oneDayFartherNs - intermediateNs)) × sign
= (nanoseconds - ((nanoseconds + intermediateNs) - intermediateNs)) × sign
= (nanoseconds - (nanoseconds)) × sign
= 0 × sign
= 0 × 1
= 0
```
3. Rinse and repeat until `days` can no longer be represented as a Number value.

And when `days` is too large, `CreateTemporalDuration` will throw a RangeError.

FrankYFTang: > And when `days` is too large, `CreateTemporalDuration` will throw a RangeError.

Could you point out which step in CreateTemporalDuration will throw such RangeError?

anba: The first step:
> 1. If ! IsValidDuration(years, months, weeks, days, ...) is false, throw a RangeError exception.

If `days` is larger than `Number.MAX_VALUE`, then step 2.a. in IsValidDuration will return `false`:
> 2.a. If 𝔽(v) is not finite, return false.

PR #2328, opened 2022-06-23
Use qualified names in demitasse tests.

Ms2ger:


ptomato (approved):
:+1:, but Ms2ger and I discussed not merging this unless we work out a plan to submit these files to test262 directly (this is preparation for a tool for autogenerating test262 tests from Demitasse tests)

ptomato:
So, we are going to try to do this.

PR #2330, opened 2022-06-24
Editorial: Mark multiple operations as infallible in DifferenceISODateTime

anba:


ptomato (approved):
The explanation makes sense, thanks.

PR #2332, opened 2022-06-27
Editorial: Remove unnecessary checks for undefined before ToPositiveInteger

anba:
`ToPositiveInteger(undefined)` calls `ToIntegerThrowOnInfinity(undefined)`, which in turn calls `ToIntegerOrInfinity(undefined)`, which in turn calls `ToNumber(undefined)`. `ToNumber` returns `NaN`, `ToIntegerOrInfinity` then returns `0`, `ToIntegerThrowOnInfinity` returns `0` unchanged, and finally `ToPositiveInteger` will throw a RangeError for `0`. That means it's not necessary to handle `undefined` before calling `ToPositiveInteger`.

----

Split from #2265.

### DIFF
--- a/spec/calendar.html
+++ b/spec/calendar.html
@@ -166,7 +166,6 @@
       <emu-alg>
         1. Assert: Type(_calendar_) is Object.
         1. Let _result_ be ? Invoke(_calendar_, *"month"*, « _dateLike_ »).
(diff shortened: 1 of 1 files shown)


ptomato (approved):
Thanks!

PR #2334, opened 2022-06-27
Add note about "timestamp" in ambiguity.md

sffc:


docs/ambiguity.md
```diff
@@ -28,6 +28,8 @@ Temporal has two types that store exact time: `Temporal.Instant` (which only sto
...
 For example, `Temporal.Instant` (an exact-time type) can be constructed using only a `BigInt` value of nanoseconds since epoch.
 
+Another term developers often encounter is "timestamp". This most often refers to an exact time represented by the number of seconds since Unix epoch. Temporal avoids using this terminology, however, because of historical ambiguity surrounding the term "timestamp". For example, many databases have a type called `TIMESTAMP`, but its meaning varies: in [MySQL](https://dev.mysql.com/doc/refman/8.0/en/datetime.html), it is an exact time; in [Oracle Database](https://docs.oracle.com/cd/B19306_01/server.102/b14225/ch4datetime.htm#i1006050), it is the number of seconds since the *wall-clock time* January 1, 1970 (a quantity one might call a "local timestamp"); and in [Microsoft SQL Server](https://docs.microsoft.com/en-us/answers/questions/238819/purpose-to-use-timestamp-datatype-in-sql-server.html), it is a monotonically increasing value unrelated to date and time.
```
justingrant, line 31: This looks great. Good content about different databases.  I'd only suggest to split one sentence per line to make future diffs easier.

```suggestion
Another term developers often encounter is "timestamp".
This most often refers to an exact time represented by the number of seconds since Unix epoch.
Temporal avoids using this terminology, however, because of historical ambiguity surrounding the term "timestamp".
For example, many databases have a type called `TIMESTAMP`, but its meaning varies: in [MySQL](https://dev.mysql.com/doc/refman/8.0/en/datetime.html), it is an exact time; in [Oracle Database](https://docs.oracle.com/cd/B19306_01/server.102/b14225/ch4datetime.htm#i1006050), it is the number of seconds since the *wall-clock time* January 1, 1970 (a quantity one might call a "local timestamp"); and in [Microsoft SQL Server](https://docs.microsoft.com/en-us/answers/questions/238819/purpose-to-use-timestamp-datatype-in-sql-server.html), it is a monotonically increasing value unrelated to date and time.
```

justingrant (approved):

ptomato (approved):

PR #2337, opened 2022-07-01
Editorial: Improve the "NotAmbiguous" grammar

gibson042:
This is an extraction of the initial editorial commits from #2284.


ptomato (approved):
Thanks!

PR #2340, opened 2022-07-05
Polyfill: Update PrepareTemporalFields to match the spec.

Ms2ger:


ptomato:
Seems reasonable; what's the reason it's marked as a draft? Is it the TODO comment about using MergeLists?

Ms2ger:
> I noticed this PR does add some TODO notes, are you planning to tack more commits on to resolve those, or are they coming in other PRs?

At some point if I have time to see if I can write tests that fail with the current implementation.

PR #2347, opened 2022-07-07
Fix typo and broken links

tars0x9752:
Fix broken links in `docs/ja` and a minor typo fix in `docs/zoneddatetime.md`.


docs/ja/README.md
```diff
@@ -224,7 +224,7 @@ date.daysInYear; // => 365
 
 ## 文字列による永続性
 
-すべての`Temporal`タイプは、永続性や相互運用性のために文字列による表現を持っています。各タイプと文字列表現の対応を以下に示します。Temporal で使用されている ISO 8601 や RFC 3339 に関する、より詳細な情報と標準化に向けた取り組みに関しては[ECMAScript 拡張の ISO-8601 と RFC 3339](./strings.md)を参照してください。
+すべての`Temporal`タイプは、永続性や相互運用性のために文字列による表現を持っています。各タイプと文字列表現の対応を以下に示します。Temporal で使用されている ISO 8601 や RFC 3339 に関する、より詳細な情報と標準化に向けた取り組みに関しては[ECMAScript 拡張の ISO-8601 と RFC 3339](./iso-string-ext.md)を参照してください。
```
tars0x9752, line 227: Perhaps this could be the `../string.md` which is the original English one. But I guess the translator intended to use `./iso-string-ext.md` here. In the context of the translation, `./iso-string-ext.md` seems to be fine to me. (I'm new to here so I'm not completely sure which one it should be though.)

ptomato: Either one is fine, I trust your judgement about what makes the most sense in this translation.

ptomato (approved):
Thanks!

docs/ja/README.md
tars0x9752: Ok, I just opened another PR #2353 to add a note to docs/ja about the latest content and leave the Japanese `./iso-string-ext.md` page as it is.

PR #2349, opened 2022-07-07
bump ecmarkup and fix typo caught thereby

bakkot:
This would also have caught the two typos in https://github.com/tc39/proposal-temporal/pull/2348 but @gibson042 was too quick.


ptomato (approved):
Thanks!

PR #2353, opened 2022-07-08
Add note about docs/ja

tars0x9752:
The added note means as follows:

> NOTE: This documentation is a partial translation of the original into Japanese. If you'd like to read the complete documentation and the most up-to-date information, please refer to the original.


ptomato (approved):
Thanks, makes sense.

PR #2362, opened 2022-07-25
Editorial: Correct HandleDateTimeValue from ! to ?

FrankYFTang:
@ptomato @ryzokuken @sffc


ptomato (approved):
Thanks. I agree that each of these can throw, for example when the calendar's `toString()` throws.

PR #2381, opened 2022-08-18
Return positive or negative overflow

FrankYFTang:
If any input to BalancePossiblyInfiniteDuration is already not finite, return positive overflow or negative overflow before any other operations

Fix https://github.com/tc39/proposal-temporal/issues/2380


FrankYFTang:
see https://github.com/tc39/proposal-temporal/issues/2380#issuecomment-1219194995
and https://chromium-review.googlesource.com/c/v8/v8/+/3837852

PR #2386, opened 2022-08-25
Editorial: Allow undefined for calendar in MaybeFormatCalendarAnnotation

linusg:
TemporalInstantToString calls TemporalDateTimeToString with undefined
for calendar and "never" for showCalendar, which then forwards these
values to MaybeFormatCalendarAnnotation as is.
This means that the current type description is inaccurate; correct this
by also considering undefined a valid value and add an assertion to
ensure undefined is never passed if showCalendar is "auto" or "always".


ptomato (approved):
Oh, thanks for catching this!

PR #2402, opened 2022-09-10
Fix CalendarDateMergeFields param and return to Object

FrankYFTang:
fix https://github.com/tc39/proposal-temporal/issues/2401

### DIFF
--- a/spec/intl.html
+++ b/spec/intl.html
@@ -1620,7 +1620,7 @@
               _calendar_: a String,
               _fields_: an Object,
               _additionalFields_: an Object,
-            ): either a normal completion containing an Object or an abrupt completion
+            ): an Object
           </h1>
           <dl class="header">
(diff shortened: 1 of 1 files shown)


ptomato (approved):
I'd like to keep #2401 open because I want to see if there's another way to rewrite the caller to pass a List, but this seems fine for correcting the immediate mistake. Can you please fix the ecmarkup linter error though?

FrankYFTang:
> if there's another way to rewrite the caller to pass a List

You can surely rewrite it to pass in a List of Record (of key + value) , but then we need to define the Record and you cannot avoid the part which it need to return the Object, not returning a list, unless you want to add logic outside it to transform that list of record into Object (which doable but I don't see a good motivation for such unnecessary work)
What is your motivaton to do so?

PR #2406, opened 2022-09-16
Z designator should be rejected in PlainTime

Aditi-1400:


ptomato (approved):
Tested in https://github.com/tc39/test262/pull/3667
Thanks!

PR #2411, opened 2022-09-30
Normative: Align {Calendar,TimeZone} id getters with toString

gibson042:
Avoid observable [[Get]]s
Fixes #2410


ptomato (approved):
I don't see any problem with this, as the toString call is likely an unintentional leftover from the change in #1692. I don't assume it'll be controversial so we probably don't need to discuss it in the next champions meeting.

ptomato:
Tests are in https://github.com/tc39/test262/pull/3694.

ptomato:
This change achieved consensus in the TC39 plenary meeting today.

PR #2412, opened 2022-09-30
Editorial: Introduce ToTimeZoneIdentifier to handle offset-or-name processing

gibson042:
Does not currently include polyfill updates.

### DIFF
--- a/spec/abstractops.html
+++ b/spec/abstractops.html
@@ -469,10 +469,8 @@
         1. If _timeZoneName_ is *undefined*, then
           1. Let _timeZone_ be *undefined*.
         1. Else,
-          1. If ParseText(StringToCodePoints(_timeZoneName_), |TimeZoneNumericUTCOffset|) is a List of errors, then
-            1. If ! IsValidTimeZoneName(_timeZoneName_) is *false*, throw a *RangeError* exception.
-            1. Set _timeZoneName_ to ! CanonicalizeTimeZoneName(_timeZoneName_).
-          1. Let _timeZone_ be ! CreateTemporalTimeZone(_timeZoneName_).
+          1. Assert: _timeZoneName_ is a String.
+          1. Let _timeZone_ be ? CreateTemporalTimeZone(? ToTimeZoneIdentifier(_timeZoneName_)).
           1. If _result_.[[TimeZone]].[[Z]] is *true*, then
             1. Set _offsetBehaviour_ to ~exact~.
           1. Else if _offsetString_ is *undefined*, then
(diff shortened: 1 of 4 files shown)


spec/abstractops.html
```diff
@@ -490,6 +488,24 @@ <h1>ToRelativeTemporalObject ( _options_ )</h1>
...
+        _argument_: an ECMAScript language value,
+      ): either a normal completion containing a String or a throw completion
+    </h1>
+    <dl class="header">
+      <dt>description</dt>
+      <dd>It validates that the argument identifies a built-in offset time zone or built-in named time zone and returns its canonical form.</dd>
+    </dl>
+    <emu-alg>
+      1. Let _S_ be ? ToString(_argument_).
+      1. If ParseText(StringToCodePoints(_S_), |TimeZoneNumericUTCOffset|) is not a List of errors, return _S_.
```
ptomato, line 503: It's debatable whether returning _S_ directly, fulfills the requirement for the "canonical form" described in the header, since _S_ could be a string like "-07:00:00.00000" and a "canonical" way of writing that would be "-07:00". (Although not "canonical" in a TZDB sense.)

This canonicalization of offset strings is done in CreateTemporalTimeZone anyway, since all call sites of ToTimeZoneIdentifier pass the result to CreateTemporalTimeZone, so you could consider moving it into here. But I'm not sure if there are any other callers of CreateTemporalTimeZone that rely on it being done there.

spec/intl.html
```diff
@@ -199,15 +199,14 @@ <h1>InitializeDateTimeFormat ( _dateTimeFormat_, _locales_, _options_ )</h1>
...
         1. Else,
-          1. Set _timeZone_ to ? ToString(_timeZone_).
-          1. If the result of ! IsValidTimeZoneName(_timeZone_) is *false*, then
-            1. Throw a *RangeError* exception.
-          1. Set _timeZone_ to ! CanonicalizeTimeZoneName(_timeZone_).
-        1. Set _dateTimeFormat_.[[TimeZone]] to _timeZone_.
+          1. Set _timeZoneIdentifier_ to ? ToTimeZoneIdentifier(_timeZoneIdentifier_).
+          1. NOTE: _timeZoneIdentifier_ may identify either an offset time zone or a named time zone.
+          1. If ! IsValidTimeZoneName(_timeZoneIdentifier_) is *false*, throw a *RangeError* exception.
+          1. Set _dateTimeFormat_.[[TimeZone]] to ! CanonicalizeTimeZoneName(_timeZoneIdentifier_).
```
ptomato, line 209: These should have `<ins>` and `<del>` tags.

ptomato (commented):
Thanks, this seems like a good refactor.

ptomato:
@gibson042 Do you plan to continue this one?

justingrant:
Is this PR still relevant? Or has it been superseded by more recent changes?

gibson042:
ToTemporalTimeZoneSlotValue and ParseTimeZoneIdentifier cover everything this PR was attempting to accomplish.

PR #2413, opened 2022-10-03
Editorial: Rename balanceResult to result on step 5.2 of DifferenceTemporalZonedDateTime

Lubrsi:
Step 5.2 uses the non-existent "balanceResult", which was forgotten to be updated in e192f2c3a870c29be4d586d423f7f47361d88543.

This changes "balanceResult" to "result" to match all the other uses of DifferenceInstant.


ptomato (approved):
Thanks!

PR #2417, opened 2022-10-05
Editorial: Add explicit assertion for days in RoundDuration

FrankYFTang:
These 2 assertions are currently not always true in the spec caused by bug in upper part of the spec but they need to be true before calling CreateTemporalDuration the line after.  Step 6-e currently cause the violation. We need days to be integer to pass into CreateTemporalDuration
```
e. Set days to days + result.[[Days]] + result.[[Nanoseconds]] / result.[[DayLength]].
```

@anba @ptomato @ryzokuken 
It does not fix https://github.com/tc39/proposal-temporal/issues/2246 but make the fix of https://github.com/tc39/proposal-temporal/issues/2246 easier to spot the issue


ptomato (approved):
I don't have any opinion on this, but if it makes things clearer for implementations then let's do it.

PR #2420, opened 2022-10-11
Editorial: Fix missing variable in CreateTemporalTimezone

mattco98:
The bug was introduced in 654e300db9882fc985840e11f35525864848f8fd


ptomato (approved):
Thanks for catching this!

PR #2421, opened 2022-10-11
Polyfill: Implement early return in UnbalanceDurationRelative as per spec

ptomato:
This early return is present in the spec text but was missing from the reference code. No change in the spec text, this just brings the code in line with the specification.

A test for this is in https://github.com/tc39/test262/pull/3686, we'll need to update the pinned test262 version before merging this, since otherwise the (faulty) existing test will break.


Aditi-1400 (approved):
Thanks! :)

Issue #2384, opened 2022-08-24
Fix assertions in non-402 Calendar methods


ptomato:
#2192 updated the spec so that non-402 implementations are _allowed_ to support other calendars than `iso8601`, but the assertions in e.g. [Temporal.Calendar.prototype.dateFromFields()](https://tc39.es/proposal-temporal/#sec-temporal.calendar.prototype.datefromfields) remained: "Assert: _calendar_.[[Identifier]] is **"iso8601"**." These assertions are no longer correct.

We can probably fix this with a note like "The minimum implementation of the `Temporal.Calendar.prototype.dateFromFields` method for ECMAScript implementations that do not include the ECMA-402 API, supporting only the **"iso8601"** calendar, performs the following steps when called:"

ptomato:
cc @rkirsling

PR #2431, opened 2022-11-02
A few editorial improvements

ptomato:
- Fix a bug in the reference code so that it's compliant with the spec text
- Fix faulty assertions in Calendar methods (#2384) (cc @rkirsling)
- Give a couple of AOs clearer names (#2223)


Aditi-1400 (approved):
These look good to me, Thanks :)

PR #2434, opened 2022-11-16
Various editorial changes

ptomato:
- Remove user code annotations in cases where user code cannot be called
- Remove a redundant and unobservable ToString


spec/zoneddatetime.html
```diff
@@ -1105,8 +1105,6 @@ <h1>
           1. Let _offsetString_ be ! Get(_fields_, *"offset"*).
           1. If _offsetString_ is *undefined*, then
             1. Set _offsetBehaviour_ to ~wall~.
-          1. Else,
-            1. Set _offsetString_ to ? ToString(_offsetString_).
```
Ms2ger, line 1109: I would prefer having an assertion here for clarity

Ms2ger (approved):

PR #2438, opened 2022-11-18
Normative: Fix generating durations from non-numeric inputs 

Aditi-1400:
Fixes: #2112 

Thanks @ptomato for help!

### DIFF
--- a/polyfill/lib/plaindatetime.mjs
+++ b/polyfill/lib/plaindatetime.mjs
@@ -35,21 +35,13 @@ export class PlainDateTime {
     isoYear = ES.ToIntegerWithTruncation(isoYear);
     isoMonth = ES.ToIntegerWithTruncation(isoMonth);
     isoDay = ES.ToIntegerWithTruncation(isoDay);
-    hour = ES.ToIntegerWithTruncation(hour);
-    minute = ES.ToIntegerWithTruncation(minute);
-    second = ES.ToIntegerWithTruncation(second);
-    millisecond = ES.ToIntegerWithTruncation(millisecond);
-    microsecond = ES.ToIntegerWithTruncation(microsecond);
-    nanosecond = ES.ToIntegerWithTruncation(nanosecond);
+    hour = hour === undefined ? 0 : ES.ToIntegerWithTruncation(hour);
+    minute = minute === undefined ? 0 : ES.ToIntegerWithTruncation(minute);
+    second = second === undefined ? 0 : ES.ToIntegerWithTruncation(second);
+    millisecond = millisecond === undefined ? 0 : ES.ToIntegerWithTruncation(millisecond);
+    microsecond = microsecond === undefined ? 0 : ES.ToIntegerWithTruncation(microsecond);
+    nanosecond = nanosecond === undefined ? 0 : ES.ToIntegerWithTruncation(nanosecond);
     calendar = ES.ToTemporalCalendar(calendar);
-
-    // Note: if the arguments are not passed,
-    //       ToIntegerWithTruncation(undefined) will have returned 0, which will
-    //       be rejected by RejectDateTime in CreateTemporalDateTimeSlots. This
-    //       check exists only to improve the error message.
-    if (arguments.length < 3) {
-      throw new RangeError('missing argument: isoYear, isoMonth and isoDay are required');
-    }
 
     ES.CreateTemporalDateTimeSlots(
       this,
--- a/spec/abstractops.html
+++ b/spec/abstractops.html
@@ -1649,7 +1649,6 @@
     </dl>
     <emu-alg>
       1. Let _number_ be ? ToNumber(_argument_).
-      1. If _number_ is *NaN*, *+0*<sub>𝔽</sub>, or *-0*<sub>𝔽</sub>, return 0.
       1. If IsIntegralNumber(_number_) is *false*, throw a *RangeError* exception.
       1. Return ℝ(_number_).
     </emu-alg>
--- a/spec/abstractops.html
+++ b/spec/abstractops.html
@@ -1611,7 +1611,7 @@
     </h1>
     <dl class="header">
       <dt>description</dt>
-      <dd>It converts _argument_ to an integer representing its Number value with fractional part truncated (replacing *NaN* with 0), or throws a *RangeError* when that Number value is not finite or not positive.</dd>
+      <dd>It converts _argument_ to an integer representing its Number value with fractional part truncated, or throws a *RangeError* when the result is *NaN*, not finite or not positive.</dd>
     </dl>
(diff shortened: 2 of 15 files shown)


ptomato (approved):

gibson042 (approved):
:+1:

ptomato:
This change achieved consensus at the TC39 plenary meeting today.

We still need to make the corresponding changes to test262, so that should happen next.

ptomato:
Rebased to include a change to CalendarYearOfWeek, which was merged in the meantime. I found that some of the changes to the Duration constructor had gotten dropped, probably in a rebase, so I fixed that while I was at it.

spec/plaindatetime.html
```diff
@@ -28,19 +28,18 @@ <h1>Temporal.PlainDateTime ( _isoYear_, _isoMonth_, _isoDay_ [ , _hour_ [ , _min
...
-        1. Let _isoDay_ be ? ToIntegerWithTruncation(_isoDay_).
-        1. Let _hour_ be ? ToIntegerWithTruncation(_hour_).
-        1. Let _minute_ be ? ToIntegerWithTruncation(_minute_).
-        1. Let _second_ be ? ToIntegerWithTruncation(_second_).
-        1. Let _millisecond_ be ? ToIntegerWithTruncation(_millisecond_).
-        1. Let _microsecond_ be ? ToIntegerWithTruncation(_microsecond_).
-        1. Let _nanosecond_ be ? ToIntegerWithTruncation(_nanosecond_).
+        1. Set _isoYear_ to ? ToIntegerWithTruncation(_isoYear_).
+        1. Set _isoMonth_ to ? ToIntegerWithTruncation(_isoMonth_).
+        1. Set _isoDay_ to ? ToIntegerWithTruncation(_isoDay_).
```
gibson042, line 36: Nit: We should be more consistent about overwriting argument aliases (like this) vs. introducing new ones for mathematical values (as in Temporal.Duration two files up), and appropriate names for those new aliases in the latter case. I'll open an ECMA-262 issue.

gibson042 (approved):
Spec changes look good modulo some editorial tweaks. Should this PR also update the polyfill?

spec/plaindatetime.html
gibson042: https://github.com/tc39/ecma262/issues/2967

Aditi-1400:
Test 262 updates are in: https://github.com/tc39/test262/pull/3759

spec/abstractops.html
```diff
@@ -1440,24 +1440,23 @@ <h1>
...
+      1. If _days_ is *undefined*, let _daysMV_ be 0; else let _daysMV_ be ? ToIntegerWithTruncation(CodePointsToString(_days_)).
+      1. If _hours_ is *undefined*, let _hoursMV_ be 0; else let _hoursMV_ be ? ToIntegerWithTruncation(CodePointsToString(_hours_)).
       1. If _fHours_ is not empty, then
         1. If any of _minutes_, _fMinutes_, _seconds_, _fSeconds_ is not empty, throw a *RangeError* exception.
         1. Let _fHoursDigits_ be the substring of CodePointsToString(_fHours_) from 1.
         1. Let _fHoursScale_ be the length of _fHoursDigits_.
         1. Let _minutesMV_ be ? ToIntegerWithTruncation(_fHoursDigits_) / 10<sup>_fHoursScale_</sup> &times; 60.
       1. Else,
-        1. Let _minutesMV_ be ? ToIntegerWithTruncation(CodePointsToString(_minutes_)).
+        1. If _minutes_ is *undefined*, let _minutesMV_ be 0; else let _minutesMV_ be ? ToIntegerWithTruncation(CodePointsToString(_minutes_)).
```
ptomato, line 1459: I think we can keep ParseTemporalDurationString as it was. In the case where one of the parse nodes is not present (for example, let's say _years_):
- _years_ becomes an empty sequence of code points
- CodePointsToString(_years_) is the empty string **""**
- [ToIntegerWithTruncation](https://tc39.es/proposal-temporal/#sec-tointegerwithtruncation)(**""**) calls [ToNumber](https://tc39.es/ecma262/#sec-tonumber)(**""**)
- ToNumber step 6, calls [StringToNumber](https://tc39.es/ecma262/#sec-stringtonumber)(**""**)
- StringToNumber parses an empty sequence of code points against [_StringNumericLiteral_](https://tc39.es/ecma262/#prod-StringNumericLiteral), which matches _StrWhiteSpace_(opt)
- [StringNumericValue](https://tc39.es/ecma262/#sec-runtime-semantics-stringnumericvalue) of _StrWhiteSpace_(opt) is **+0**<sub>𝔽</sub>
- ToIntegerWithTruncation returns truncate(ℝ(**+0**<sub>𝔽</sub>)) which is 0.

(Unlike the polyfill code where we do need to check for undefined because we're using a regexp)

polyfill/lib/plaindatetime.mjs
```diff
@@ -35,22 +35,14 @@ export class PlainDateTime {
...
     calendar = ES.ToTemporalCalendar(calendar);
 
-    // Note: if the arguments are not passed,
-    //       ToIntegerWithTruncation(undefined) will have returned 0, which will
-    //       be rejected by RejectDateTime in CreateTemporalDateTimeSlots. This
-    //       check exists only to improve the error message.
-    if (arguments.length < 3) {
-      throw new RangeError('missing argument: isoYear, isoMonth and isoDay are required');
-    }
-
```
ptomato, line 53: You're right that this message is now dead code, and the comment would need to be changed, but maybe we could move it above the ToIntegerWithTruncation calls instead? (Let's do the same thing for the other similar messages in other types' constructors.) The RangeError would be thrown anyway by ToIntegerWithTruncation, so the difference is not observable, but this extra check improves the experience in the browser playground for a user who tries `new Temporal.PlainDateTime()` with no arguments.

ptomato (changes requested):

polyfill/lib/plaindatetime.mjs
Aditi-1400: Um, If I move this check before the ToIntegerWithTruncation call, then we'll have to change the order-of-operations in missing-arguments tests.🤔

ptomato: Makes sense, I didn't think of that.

ptomato: We should probably remove the check from PlainDate, PlainYearMonth, and PlainMonthDay as well.

ptomato (approved):
Thanks! I think this is ready to merge now.

ptomato:
(For completeness, this achieved consensus at the November 2022 TC39 meeting.)

PR #2440, opened 2022-11-30
Editorial: Replace "is not zero" language

ptomato:
With mathematical values, the correct spec language is "≠ 0".


gibson042 (approved):
Thanks!

PR #2445, opened 2022-12-08
Remove Unix Epoch references from Temporal.Now.instant()

mlemoine-clgx:
Fixes #2418. Related to #2175.

Creating a PR as suggested.


ptomato (approved):
Thanks!

PR #2446, opened 2022-12-08
Editorial: Clarify that DaysInMonth operations return a count of days

gibson042:
...rather than e.g. the number of the last day.

Fixes #1315

### DIFF
--- a/spec/calendar.html
+++ b/spec/calendar.html
@@ -568,7 +568,7 @@
       <h1>
         ISODaysInMonth (
           _year_: an integer,
-          _month_: an integer between 1 and 12 inclusive,
+          _month_: an integer in the inclusive interval from 1 to 12,
         ): a positive integer
       </h1>
       <dl class="header">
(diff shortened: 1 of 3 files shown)


ljharb (approved):

spec/calendar.html
```diff
@@ -564,13 +564,21 @@ <h1>ConsolidateCalendars ( _one_, _two_ )</h1>
       </emu-alg>
     </emu-clause>
 
-    <emu-clause id="sec-temporal-isodaysinmonth" aoid="ISODaysInMonth">
-      <h1>ISODaysInMonth ( _year_, _month_ )</h1>
+    <emu-clause id="sec-temporal-isodaysinmonth" type="abstract operation">
+      <h1>
+        ISODaysInMonth (
+          _year_: an integer,
+          _month_: an integer between 1 and 12 inclusive,
```
ptomato, line 571:
```suggestion
          _month_: an integer in the inclusive interval from 1 to 12,
```
...is the current terminology in ECMA-262.

ptomato (approved):
I don't think this is sufficient to close #1315 since there's still an open discussion about the semantics of `daysInMonth` but I think this change is an improvement nonetheless.

PR #2450, opened 2022-12-16
WIP - Please ignore - Node 19 intl repro

justingrant:
This is a test PR to try to repro a Node 19 issue that shows up in CI running Node 19.2.0 but not on my local Node 19.2.0 install.

For now, please ignore.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -7,18 +7,18 @@ jobs:
   deploy:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


justingrant:
The reason why I wasn't seeing this on my local machine but was seeing it in CI turned out to be that GH Actions' Node 19 installs use ICU 72, while the versions available for download from nodejs's website and via homebrew use ICU 71. The root cause of the failures was this CLDR bug: https://unicode-org.atlassian.net/browse/CLDR-16243.  

Closing in favor of the original PR #2448.

PR #2453, opened 2023-01-03
Use new `temporal-test262-runner` library

justingrant:
This PR moves our optimized Test262-executing code into a library package [`@js-temporal/temporal-test262-runner`](https://github.com/js-temporal/temporal-test262-runner) that can be shared with other polyfills.


ptomato (approved):
Great!

Issue #2457, opened 2023-01-06
Typo in ToTemporalMonthDay


ptomato:
Spotted by @anba in #1502, split out into a separate issue.

I believe step 10 of [ToTemporalMonthDay](https://tc39.es/proposal-temporal/#sec-temporal-totemporalmonthday) should pass _result_.[[Year]] as the last argument, not _referenceISOYear_.

Justification: assume ToTemporalMonthDay is called with the string `2023-01-05[u-ca=hebrew]` which should give a PlainMonthDay of today's date in the Hebrew calendar, 12 Tevet. After step 7, we get a Record with:
- [[Year]]: 2023
- [[Month]]: 1
- [[Day]]: 5
- [[Hour]], ..., [[Nanosecond]]: all 0
- [[TimeZone]]: { [[Z]]: false, [[OffsetString]]: undefined, [[Name]]: undefined }
- [[Calendar]]: "hebrew"
The early return in step 9 is skipped. Step 10 creates a PlainYearMonth with the ISO reference date of 1972-01-05, which is 18 Tevet, not 12 Tevet. This result is wrong. Instead, the reference date should be 2023-01-05, which gets passed to CalendarMonthDayFromFields in step 11 to canonicalize the reference date to 1971-12-30 (12 Tevet in the reference year.)

PR #2460, opened 2023-01-06
Normative: Fix typo in ToTemporalMonthDay

Aditi-1400:
Fixes: https://github.com/tc39/proposal-temporal/issues/2457


ptomato (approved):
:+1:

I'll mark this as draft until presented at TC39.

The reference code already works this way, but we should make sure this is covered by test262 tests as well.

ptomato:
This normative change reached consensus in the March 2023 TC39 meeting. Test262 tests are in https://github.com/tc39/test262/pull/3818

PR #2479, opened 2023-01-17
Normative: Prevent loss of TimeZone information

Aditi-1400:
Modified `FormatDateTimePattern` to include TimeZone information.
Fixes: #2013

### DIFF
--- a/spec/intl.html
+++ b/spec/intl.html
@@ -396,9 +396,9 @@
     </emu-clause>
 
     <emu-clause id="sec-formatdatetimepattern" aoid="FormatDateTimePattern">
-      <h1>FormatDateTimePattern ( _dateTimeFormat_, <ins>_pattern_</ins>, _patternParts_, <del>_x_</del><ins>_epochNanoseconds_</ins>, _rangeFormatOptions_ <ins>[ , _timeZone_ ]</ins> )</h1>
+      <h1>FormatDateTimePattern ( _dateTimeFormat_, <ins>_pattern_</ins>, _patternParts_, <del>_x_</del><ins>_epochNanoseconds_</ins>, _rangeFormatOptions_<ins>, _timeZone_</ins> )</h1>
       <p>
-      The FormatDateTimePattern abstract operation is called with arguments _dateTimeFormat_ (which must be an object initialized as a DateTimeFormat), <ins>_pattern_ (which is a Record of the type contained by the %DateTimeFormat%.[[LocaleData]].[[<_locale_>]].[[formats]].[[<_calendar_>]] List as described in <emu-xref href="#sec-intl.datetimeformat-internal-slots"></emu-xref>),</ins> _patternParts_ (which is a list of Records as returned by PartitionPattern), <del>_x_</del><ins>_epochNanoseconds_</ins> (which must be a <del>Number</del><ins>BigInt</ins> value), and _rangeFormatOptions_ (which is a range pattern Record as used in [[rangePattern]] or *undefined*), interprets _x_ as a time value as specified in ES2023, <emu-xref href="#sec-time-values-and-time-range"></emu-xref>, and creates the corresponding parts according _pattern_ and to the effective locale and the formatting options of _dateTimeFormat_ and _rangeFormatOptions_. The following steps are taken:
+      The FormatDateTimePattern abstract operation is called with arguments _dateTimeFormat_ (which must be an object initialized as a DateTimeFormat), <ins>_pattern_ (which is a Record of the type contained by the %DateTimeFormat%.[[LocaleData]].[[<_locale_>]].[[formats]].[[<_calendar_>]] List as described in <emu-xref href="#sec-intl.datetimeformat-internal-slots"></emu-xref>),</ins> _patternParts_ (which is a list of Records as returned by PartitionPattern), <del>_x_</del><ins>_epochNanoseconds_</ins> (which must be a <del>Number</del><ins>BigInt</ins> value), _rangeFormatOptions_ (which is a range pattern Record as used in [[rangePattern]] or *undefined*) <ins>and _timeZone_</ins>, interprets _x_ as a time value as specified in ES2023, <emu-xref href="#sec-time-values-and-time-range"></emu-xref>, and creates the corresponding parts according _pattern_ and to the effective locale and the formatting options of _dateTimeFormat_ and _rangeFormatOptions_. The following steps are taken:
       </p>
 
       <emu-alg>
@@ -418,10 +418,9 @@
           1. Perform ! CreateDataPropertyOrThrow(_nf3Options_, *"minimumIntegerDigits"*, _fractionalSecondDigits_).
           1. Perform ! CreateDataPropertyOrThrow(_nf3Options_, *"useGrouping"*, *false*).
           1. Let _nf3_ be ? Construct(%NumberFormat%, &laquo; _locale_, _nf3Options_ &raquo;).
-        1. <ins>If _timeZone_ is present, then</ins>
-          1. <ins>Let _tm_ be ToLocalTime(<ins>_epochNanoseconds_</ins>, _dateTimeFormat_.[[Calendar]], _timeZone_).</ins>
-        1. <ins>Else,</ins>
-          1. Let _tm_ be ToLocalTime(<del>_x_</del><ins>_epochNanoseconds_</ins>, _dateTimeFormat_.[[Calendar]], _dateTimeFormat_.[[TimeZone]]).
+        1. <ins>If _dateTimeFormat_.[[TimeZone]] is not *undefined*, then</ins>
+          1. <ins>Throw a *RangeError* exception.</ins>
+        1. <ins>Let _tm_ be ToLocalTime(_epochNanoseconds_, _dateTimeFormat_.[[Calendar]], _timeZone_).</ins>
         1. Let _result_ be a new empty List.
         1. For each Record { [[Type]], [[Value]] } _patternPart_ in _patternParts_, do
           1. Let _p_ be _patternPart_.[[Type]].
@@ -438,7 +437,7 @@
             1. Append a new Record { [[Type]]: _p_, [[Value]]: _fv_ } as the last element of the list _result_.
           1. Else if _p_ is equal to *"timeZoneName"*, then
             1. Let _f_ be _dateTimeFormat_.[[TimeZoneName]].
-            1. Let _v_ be _dateTimeFormat_.[[TimeZone]].
+            1. Let _v_ be <del>_dateTimeFormat_.[[TimeZone]]</del><ins>_timeZone_</ins>.
             1. Let _fv_ be a String value representing _v_ in the form given by _f_; the String value depends upon the implementation and the effective locale of _dateTimeFormat_. The String value may also depend on the value of the [[InDST]] field of _tm_ if _f_ is *"short"*, *"long"*, *"shortOffset"*, or *"longOffset"*. If the implementation does not have a localized representation of _f_, then use the String value of _v_ itself.
             1. Append a new Record { [[Type]]: _p_, [[Value]]: _fv_ } as the last element of the list _result_.
           1. Else if _p_ matches a Property column of the row in <emu-xref href="#table-datetimeformat-components"></emu-xref>, then
@@ -564,9 +563,15 @@
           1. <ins>If ! SameTemporalType(_x_, _y_) is *false*, throw a *TypeError* exception.</ins>
         1. <ins>Let _x_ be ? HandleDateTimeValue(_dateTimeFormat_, _x_).</ins>
         1. <ins>Let _y_ be ? HandleDateTimeValue(_dateTimeFormat_, _y_).</ins>
+        1. <ins>If x.[[timeZone]] is not equal to y.[[timeZone]], then</ins>
+          1. <ins>Throw a *RangeError* exception.</ins>
+        1. <ins>Else if x.[[timeZone]] is not *undefined*, then</ins>
+          1. <ins>Let _timeZone_ be x.[[timeZone]].</ins>
+        1. <ins>Else,</ins>
+          1. <ins>Let _timeZone_ be _dateTimeFormat_.[[TimeZone]].</ins>
         1. If _x_<ins>.[[epochNanoseconds]]</ins> is greater than _y_<ins>.[[epochNanoseconds]]</ins>, throw a *RangeError* exception.
-        1. Let _tm1_ be ToLocalTime(_x_<ins>.[[epochNanoseconds]]</ins>, _dateTimeFormat_.[[Calendar]], _dateTimeFormat_.[[TimeZone]]).
-        1. Let _tm2_ be ToLocalTime(_y_<ins>.[[epochNanoseconds]]</ins>, _dateTimeFormat_.[[Calendar]], _dateTimeFormat_.[[TimeZone]]).
+        1. Let _tm1_ be ToLocalTime(_x_<ins>.[[epochNanoseconds]]</ins>, _dateTimeFormat_.[[Calendar]], <del>_dateTimeFormat_.[[TimeZone]]</del> <ins>_timeZone_</ins>).
+        1. Let _tm2_ be ToLocalTime(_y_<ins>.[[epochNanoseconds]]</ins>, _dateTimeFormat_.[[Calendar]], <del>_dateTimeFormat_.[[TimeZone]]</del><ins>_timeZone_</ins>).
         1. Let _rangePatterns_ be <del>_dateTimeFormat_.[[RangePatterns]]</del><ins>_x_.[[rangePatterns]]</ins>.
         1. <ins>Assert: _rangePatterns_ is equal to _y_.[[rangePatterns]].</ins>
         1. Let _rangePattern_ be *undefined*.
@@ -612,7 +617,7 @@
         1. If _dateFieldsPracticallyEqual_ is *true*, then
           1. Let _pattern_ be <del>_dateTimeFormat_.[[Pattern]]</del><ins>_x_.[[pattern]]</ins>.
           1. Let _patternParts_ be PartitionPattern(_pattern_).
-          1. Let _result_ be ? FormatDateTimePattern(_dateTimeFormat_, <ins>_pattern_,</ins> _patternParts_, _x_<ins>.[[epochNanoseconds]]</ins>, *undefined*).
+          1. Let _result_ be ? FormatDateTimePattern(_dateTimeFormat_, <ins>_pattern_,</ins> _patternParts_, _x_<ins>.[[epochNanoseconds]]</ins>, *undefined*, <ins>_timeZone_</ins>).
           1. For each element _r_ in _result_, do
             1. Set _r_.[[Source]] to *"shared"*.
           1. Return _result_.
@@ -627,7 +632,7 @@
           1. Else,
             1. Let _z_ be _y_<ins>.[[epochNanoseconds]]</ins>.
           1. Let _patternParts_ be PartitionPattern(_pattern_).
-          1. Let _partResult_ be ? FormatDateTimePattern(_dateTimeFormat_, <ins>_pattern_,</ins> _patternParts_, _z_, _rangePattern_).
+          1. Let _partResult_ be ? FormatDateTimePattern(_dateTimeFormat_, <ins>_pattern_,</ins> _patternParts_, _z_, _rangePattern_, <ins>_timeZone_</ins>).
           1. For each element _r_ in _partResult_, do
             1. Set _r_.[[Source]] to _source_.
           1. Add all elements in _partResult_ to _result_ in order.
@@ -772,7 +777,8 @@
       1. Return the Record {
           [[pattern]]: _pattern_.[[pattern]],
           [[rangePatterns]]: _pattern_.[[rangePatterns]],
-          [[epochNanoseconds]]: _instant_.[[Nanoseconds]]
+          [[epochNanoseconds]]: _instant_.[[Nanoseconds]],
+          [[timeZone]]: *undefined*
         }.
       </emu-alg>
     </emu-clause>
@@ -798,7 +804,8 @@
       1. Return the Record {
           [[pattern]]: _pattern_.[[pattern]],
           [[rangePatterns]]: _pattern_.[[rangePatterns]],
-          [[epochNanoseconds]]: _instant_.[[Nanoseconds]]
+          [[epochNanoseconds]]: _instant_.[[Nanoseconds]],
+          [[timeZone]]: *undefined*
         }.
       </emu-alg>
     </emu-clause>
@@ -825,7 +832,8 @@
       1. Return the Record {
           [[pattern]]: _pattern_.[[pattern]],
           [[rangePatterns]]: _pattern_.[[rangePatterns]],
-          [[epochNanoseconds]]: _instant_.[[Nanoseconds]]
+          [[epochNanoseconds]]: _instant_.[[Nanoseconds]],
+          [[timeZone]]: *undefined*
         }.
       </emu-alg>
     </emu-clause>
@@ -850,7 +858,8 @@
       1. Return the Record {
           [[pattern]]: _pattern_.[[pattern]],
           [[rangePatterns]]: _pattern_.[[rangePatterns]],
-          [[epochNanoseconds]]: _instant_.[[Nanoseconds]]
+          [[epochNanoseconds]]: _instant_.[[Nanoseconds]],
+          [[timeZone]]: *undefined*
         }.
       </emu-alg>
     </emu-clause>
@@ -876,7 +885,8 @@
       1. Return the Record {
           [[pattern]]: _pattern_.[[pattern]],
           [[rangePatterns]]: _pattern_.[[rangePatterns]],
-          [[epochNanoseconds]]: _instant_.[[Nanoseconds]]
+          [[epochNanoseconds]]: _instant_.[[Nanoseconds]],
+          [[timeZone]]: *undefined*
         }.
       </emu-alg>
     </emu-clause>
@@ -897,7 +907,8 @@
       1. Return the Record {
           [[pattern]]: _pattern_.[[pattern]],
           [[rangePatterns]]: _pattern_.[[rangePatterns]],
-          [[epochNanoseconds]]: _instant_.[[Nanoseconds]]
+          [[epochNanoseconds]]: _instant_.[[Nanoseconds]],
+          [[timeZone]]: *undefined*
         }.
       </emu-alg>
     </emu-clause>
@@ -907,7 +918,7 @@
       <h1>HandleDateTimeTemporalZonedDateTime ( _dateTimeFormat_, _zonedDateTime_ )</h1>
 
       <p>
-        The abstract operation HandleDateTimeTemporalZonedDateTime accepts the arguments _dateTimeFormat_ (which must be an object initialized as a DateTimeFormat) and _zonedDateTime_ (which must be an ECMAScript value has an [[InitializedTemporalDateTime]] internal slot). It returns a record which contains the appropriate pattern and epochNanoseconds values for the input. This abstract operation functions as follows:
+        The abstract operation HandleDateTimeTemporalZonedDateTime accepts the arguments _dateTimeFormat_ (which must be an object initialized as a DateTimeFormat) and _zonedDateTime_ (which must be an ECMAScript value has an [[InitializedTemporalDateTime]] internal slot). It returns a record which contains the appropriate pattern, epochNanoseconds and timezone values for the input. This abstract operation functions as follows:
       </p>
 
       <emu-alg>
@@ -924,7 +935,8 @@
       1. Return the Record {
           [[pattern]]: _pattern_.[[pattern]],
           [[rangePatterns]]: _pattern_.[[rangePatterns]],
-          [[epochNanoseconds]]: _instant_.[[Nanoseconds]]
+          [[epochNanoseconds]]: _instant_.[[Nanoseconds]],
+          [[timeZone]]: _timeZone_,
         }.
       </emu-alg>
     </emu-clause>
@@ -952,7 +964,8 @@
       1. Return the Record {
           [[pattern]]: _pattern_,
           [[rangePatterns]]: _rangePatterns_,
-          [[epochNanoseconds]]: _epochNanoseconds_
+          [[epochNanoseconds]]: _epochNanoseconds_,
+          [[timeZone]]: *undefined*
         }.
       </emu-alg>
     </emu-clause>
--- a/spec/intl.html
+++ b/spec/intl.html
@@ -396,8 +396,7 @@
     </emu-clause>
 
     <emu-clause id="sec-formatdatetimepattern" aoid="FormatDateTimePattern">
-      <h1>FormatDateTimePattern ( _dateTimeFormat_, <ins>_pattern_</ins>, _patternParts_, <del>_x_</del><ins>_epochNanoseconds_</ins>, _rangeFormatOptions_ )</h1>
-
+      <h1>FormatDateTimePattern ( _dateTimeFormat_, <ins>_pattern_</ins>, _patternParts_, <del>_x_</del><ins>_epochNanoseconds_</ins>, _rangeFormatOptions_ <ins>[ , _timeZone_ ]</ins> )</h1>
       <p>
       The FormatDateTimePattern abstract operation is called with arguments _dateTimeFormat_ (which must be an object initialized as a DateTimeFormat), <ins>_pattern_ (which is a Record of the type contained by the %DateTimeFormat%.[[LocaleData]].[[<_locale_>]].[[formats]].[[<_calendar_>]] List as described in <emu-xref href="#sec-intl.datetimeformat-internal-slots"></emu-xref>),</ins> _patternParts_ (which is a list of Records as returned by PartitionPattern), <del>_x_</del><ins>_epochNanoseconds_</ins> (which must be a <del>Number</del><ins>BigInt</ins> value), and _rangeFormatOptions_ (which is a range pattern Record as used in [[rangePattern]] or *undefined*), interprets _x_ as a time value as specified in ES2023, <emu-xref href="#sec-time-values-and-time-range"></emu-xref>, and creates the corresponding parts according _pattern_ and to the effective locale and the formatting options of _dateTimeFormat_ and _rangeFormatOptions_. The following steps are taken:
       </p>
@@ -419,7 +418,10 @@
           1. Perform ! CreateDataPropertyOrThrow(_nf3Options_, *"minimumIntegerDigits"*, _fractionalSecondDigits_).
           1. Perform ! CreateDataPropertyOrThrow(_nf3Options_, *"useGrouping"*, *false*).
           1. Let _nf3_ be ? Construct(%NumberFormat%, &laquo; _locale_, _nf3Options_ &raquo;).
-        1. Let _tm_ be ToLocalTime(<del>_x_</del><ins>_epochNanoseconds_</ins>, _dateTimeFormat_.[[Calendar]], _dateTimeFormat_.[[TimeZone]]).
+        1. <ins>If _timeZone_ is present, then</ins>
+          1. <ins>Let _tm_ be ToLocalTime(<ins>_epochNanoseconds_</ins>, _dateTimeFormat_.[[Calendar]], _timeZone_).</ins>
+        1. <ins>Else,</ins>
+          1. Let _tm_ be ToLocalTime(<del>_x_</del><ins>_epochNanoseconds_</ins>, _dateTimeFormat_.[[Calendar]], _dateTimeFormat_.[[TimeZone]]).
         1. Let _result_ be a new empty List.
         1. For each Record { [[Type]], [[Value]] } _patternPart_ in _patternParts_, do
           1. Let _p_ be _patternPart_.[[Type]].
@@ -497,7 +499,7 @@
       <emu-alg>
         1. <ins>Let _x_ be ? HandleDateTimeValue(_dateTimeFormat_, _x_).</ins>
         1. Let _patternParts_ be PartitionPattern(<del>_dateTimeFormat_.[[Pattern]]</del><ins>_x_.[[pattern]]</ins>).
-        1. Let _result_ be ? FormatDateTimePattern(_dateTimeFormat_, <ins>_x_.[[pattern]]</ins>, _patternParts_, <del>_x_</del><ins>_x_.[[epochNanoseconds]]</ins>, *undefined*).
+        1. Let _result_ be ? FormatDateTimePattern(_dateTimeFormat_, <ins>_x_.[[pattern]]</ins>, _patternParts_, <del>_x_</del><ins>_x_.[[epochNanoseconds]]</ins>, *undefined*, <ins>x.[[timeZone]]</ins>).
         1. Return _result_.
       </emu-alg>
 


spec/intl.html
```diff
@@ -419,7 +418,10 @@ <h1>FormatDateTimePattern ( _dateTimeFormat_, <ins>_pattern_</ins>, _patternPart
           1. Perform ! CreateDataPropertyOrThrow(_nf3Options_, *"minimumIntegerDigits"*, _fractionalSecondDigits_).
           1. Perform ! CreateDataPropertyOrThrow(_nf3Options_, *"useGrouping"*, *false*).
           1. Let _nf3_ be ? Construct(%NumberFormat%, &laquo; _locale_, _nf3Options_ &raquo;).
-        1. Let _tm_ be ToLocalTime(<del>_x_</del><ins>_epochNanoseconds_</ins>, _dateTimeFormat_.[[Calendar]], _dateTimeFormat_.[[TimeZone]]).
+        1. <ins>If _timeZone_ is present, then</ins>
+          1. <ins>Let _tm_ be ToLocalTime(<ins>_epochNanoseconds_</ins>, _dateTimeFormat_.[[Calendar]], _timeZone_).</ins>
```
ptomato, line 422:
```suggestion
          1. <ins>Let _tm_ be ToLocalTime(_epochNanoseconds_, _dateTimeFormat_.[[Calendar]], _timeZone_).</ins>
```

spec/intl.html
```diff
@@ -396,8 +396,7 @@ <h1>DateTime Format Functions</h1>
     </emu-clause>
 
     <emu-clause id="sec-formatdatetimepattern" aoid="FormatDateTimePattern">
-      <h1>FormatDateTimePattern ( _dateTimeFormat_, <ins>_pattern_</ins>, _patternParts_, <del>_x_</del><ins>_epochNanoseconds_</ins>, _rangeFormatOptions_ )</h1>
-
+      <h1>FormatDateTimePattern ( _dateTimeFormat_, <ins>_pattern_</ins>, _patternParts_, <del>_x_</del><ins>_epochNanoseconds_</ins>, _rangeFormatOptions_ <ins>[ , _timeZone_ ]</ins> )</h1>
```
ptomato, line 399: We should describe this new parameter in the paragraph below. I'd also consider calling it "timeZoneOverride" to clarify that it's meant to override the DateTimeFormat's internal time zone if it's present, but on the other hand that's less concise.

ptomato (changes requested):
Good start, but there's more needed, I think. First, we need to make a similar change in PartitionDateTimeRangePattern.

The _x_.[[timeZone]] has to come from somewhere, so we need to make sure that HandleDateTimeValue provides that field in the Record that it returns. I'd guess it should be set to undefined in every case except HandleDateTimeTemporalZonedDateTime. (That also suggests to me that we shouldn't check whether it is present in FormatDateTimePattern but instead check whether it is undefined.)

spec/intl.html
Aditi-1400: I’ve added it in description, but didn’t change the parameter name, for conciseness. Let me know if you’d like me to change that.

Aditi-1400:
> Good start, but there's more needed, I think. First, we need to make a similar change in PartitionDateTimeRangePattern.
> 
> The _x_.[[timeZone]] has to come from somewhere, so we need to make sure that HandleDateTimeValue provides that field in the Record that it returns. I'd guess it should be set to undefined in every case except HandleDateTimeTemporalZonedDateTime. (That also suggests to me that we shouldn't check whether it is present in FormatDateTimePattern but instead check whether it is undefined.)

Since, `HandleDateTimeValue` directly returns the records from function, I’ve made the changes to the respective functions, because that’s consistent with `HandleDateTimeTemporalZonedDateTime` but it might be better to add `[[timeZone]]:undefined` to `HandleDateTimeValue`, rather than returning it from other methods, what do you think?

spec/intl.html
```diff
@@ -615,7 +617,7 @@ <h1>PartitionDateTimeRangePattern ( _dateTimeFormat_, _x_, _y_ )</h1>
         1. If _dateFieldsPracticallyEqual_ is *true*, then
           1. Let _pattern_ be <del>_dateTimeFormat_.[[Pattern]]</del><ins>_x_.[[pattern]]</ins>.
           1. Let _patternParts_ be PartitionPattern(_pattern_).
-          1. Let _result_ be ? FormatDateTimePattern(_dateTimeFormat_, <ins>_pattern_,</ins> _patternParts_, _x_<ins>.[[epochNanoseconds]]</ins>, *undefined*).
+          1. Let _result_ be ? FormatDateTimePattern(_dateTimeFormat_, <ins>_pattern_,</ins> _patternParts_, _x_<ins>.[[epochNanoseconds]]</ins>, *undefined*, <ins>_x_.[[timeZone]]</ins> ).
```
ptomato, line 620: _x_.[[TimeZone]] may be undefined here if _x_ and _y_ were not ZonedDateTime instances.

spec/intl.html
```diff
@@ -627,10 +629,12 @@ <h1>PartitionDateTimeRangePattern ( _dateTimeFormat_, _x_, _y_ )</h1>
           1. Let _source_ be _rangePatternPart_.[[Source]].
           1. If _source_ is *"startRange"* or *"shared"*, then
             1. Let _z_ be _x_<ins>.[[epochNanoseconds]]</ins>.
+            1. <ins>Let _timeZone_ be _x_.[[timeZone]]</ins>.
           1. Else,
             1. Let _z_ be _y_<ins>.[[epochNanoseconds]]</ins>.
+            1. <ins>Let _timeZone_ be _y_.[[timeZone]]</ins>.
```
ptomato, line 635: Same here.

ptomato (commented):
I think PartitionDateTimeRangePattern is not quite there, everything else looks good.

In `formatRange` and `formatRangeToParts`, the time zones need to be both either present or undefined (this is already guaranteed by SameTemporalType, near the beginning of the operation) but after that we need to check that they are the same time zone, and throw a RangeError if not. After that, we can just use one variable _timeZone_ which is either the time zone of _x_ and _y_ or the time zone of _dateTimeFormat_ if _x_ and _y_ weren't ZonedDateTime instances. We should do this before calculating _tm1_ and _tm2_ because they should use _timeZone_, not _dateTimeFormat_.[[TimeZone]].

spec/intl.html
ptomato: This is fine, the parameter name can always be updated some other time.

ptomato:
> Since, `HandleDateTimeValue` directly returns the records from function, I’ve made the changes to the respective functions, because that’s consistent with `HandleDateTimeTemporalZonedDateTime` but it might be better to add `[[timeZone]]:undefined` to `HandleDateTimeValue`, rather than returning it from other methods, what do you think?

I think the way you did it is fine, but probably an equivalent way would also be fine. The current way seems concise, I'd imagine that HandleDateTimeValue would get longer if we added the time zone field after each Handle... call.

spec/intl.html
```diff
@@ -615,7 +623,7 @@ <h1>PartitionDateTimeRangePattern ( _dateTimeFormat_, _x_, _y_ )</h1>
         1. If _dateFieldsPracticallyEqual_ is *true*, then
           1. Let _pattern_ be <del>_dateTimeFormat_.[[Pattern]]</del><ins>_x_.[[pattern]]</ins>.
           1. Let _patternParts_ be PartitionPattern(_pattern_).
-          1. Let _result_ be ? FormatDateTimePattern(_dateTimeFormat_, <ins>_pattern_,</ins> _patternParts_, _x_<ins>.[[epochNanoseconds]]</ins>, *undefined*).
+          1. Let _result_ be ? FormatDateTimePattern(_dateTimeFormat_, <ins>_pattern_,</ins> _patternParts_, _x_<ins>.[[epochNanoseconds]]</ins>, *undefined*, <ins>_timeZone_</ins> ).
```
ptomato, line 626: Formatting nitpick:
```suggestion
          1. Let _result_ be ? FormatDateTimePattern(_dateTimeFormat_, <ins>_pattern_,</ins> _patternParts_, _x_<ins>.[[epochNanoseconds]]</ins>, *undefined*, <ins>_timeZone_</ins>).
```

ptomato (approved):
Great, thanks!

I did notice one more place where I think we need to make a change from _dateTimeFormat_.[[TimeZone]] to _timeZone_: step 16.e.ii of FormatDateTimePattern (the one starting "Else if p is equal to "timeZoneName", then"). There's also 16.f.viii but when I was looking at it, I actually came to believe that that is dead code because it's already covered by 16.e.ii. I'll remove it in an editorial PR which I'm working on.

Other than that, I think this is ready to merge.

spec/intl.html
```diff
@@ -436,7 +438,7 @@ <h1>FormatDateTimePattern ( _dateTimeFormat_, <ins>_pattern_</ins>, _patternPart
             1. Append a new Record { [[Type]]: _p_, [[Value]]: _fv_ } as the last element of the list _result_.
           1. Else if _p_ is equal to *"timeZoneName"*, then
             1. Let _f_ be _dateTimeFormat_.[[TimeZoneName]].
-            1. Let _v_ be _dateTimeFormat_.[[TimeZone]].
+            1. Let _v_ be <del>_dateTimeFormat_.[[TimeZone]]</del><ins>_timeZone_</ins>.
```
gibson042, line 441: Isn't this incorrect for cases where _timeZone_ is undefined?

spec/intl.html
```diff
@@ -567,9 +569,15 @@ <h1>PartitionDateTimeRangePattern ( _dateTimeFormat_, _x_, _y_ )</h1>
           1. <ins>If ! SameTemporalType(_x_, _y_) is *false*, throw a *TypeError* exception.</ins>
         1. <ins>Let _x_ be ? HandleDateTimeValue(_dateTimeFormat_, _x_).</ins>
         1. <ins>Let _y_ be ? HandleDateTimeValue(_dateTimeFormat_, _y_).</ins>
+        1. <ins>If x.[[timeZone]] is not equal to y.[[timeZone]], then</ins>
+          1. <ins>Throw a *RangeError* exception.</ins>
+        1. <ins>Else if x.[[timeZone]] is not *undefined*, then</ins>
+          1. <ins>Let _timeZone_ be x.[[timeZone]].</ins>
```
gibson042, line 575:
```suggestion
        1. <ins>If _x_.[[timeZone]] &neq; to _y_.[[timeZone]], then</ins>
          1. <ins>Throw a *RangeError* exception.</ins>
        1. <ins>Else if _x_.[[timeZone]] is not *undefined*, then</ins>
          1. <ins>Let _timeZone_ be _x_.[[timeZone]].</ins>
```

spec/intl.html
```diff
@@ -502,7 +504,7 @@ <h1>PartitionDateTimePattern ( _dateTimeFormat_, _x_ )</h1>
       <emu-alg>
         1. <ins>Let _x_ be ? HandleDateTimeValue(_dateTimeFormat_, _x_).</ins>
         1. Let _patternParts_ be PartitionPattern(<del>_dateTimeFormat_.[[Pattern]]</del><ins>_x_.[[pattern]]</ins>).
-        1. Let _result_ be ? FormatDateTimePattern(_dateTimeFormat_, <ins>_x_.[[pattern]]</ins>, _patternParts_, <del>_x_</del><ins>_x_.[[epochNanoseconds]]</ins>, *undefined*).
+        1. Let _result_ be ? FormatDateTimePattern(_dateTimeFormat_, <ins>_x_.[[pattern]]</ins>, _patternParts_, <del>_x_</del><ins>_x_.[[epochNanoseconds]]</ins>, *undefined*, <ins>x.[[timeZone]]</ins>).
```
gibson042, line 507:
```suggestion
        1. Let _result_ be ? FormatDateTimePattern(_dateTimeFormat_, <ins>_x_.[[pattern]]</ins>, _patternParts_, <del>_x_</del><ins>_x_.[[epochNanoseconds]]</ins>, *undefined*, <ins>_x_.[[timeZone]]</ins>).
```

spec/intl.html
```diff
@@ -910,7 +924,7 @@ <h1>HandleDateTimeTemporalInstant ( _dateTimeFormat_, _instant_ )</h1>
       <h1>HandleDateTimeTemporalZonedDateTime ( _dateTimeFormat_, _zonedDateTime_ )</h1>
 
       <p>
-        The abstract operation HandleDateTimeTemporalZonedDateTime accepts the arguments _dateTimeFormat_ (which must be an object initialized as a DateTimeFormat) and _zonedDateTime_ (which must be an ECMAScript value has an [[InitializedTemporalDateTime]] internal slot). It returns a record which contains the appropriate pattern and epochNanoseconds values for the input. This abstract operation functions as follows:
+        The abstract operation HandleDateTimeTemporalZonedDateTime accepts the arguments _dateTimeFormat_ (which must be an object initialized as a DateTimeFormat) and _zonedDateTime_ (which must be an ECMAScript value has an [[InitializedTemporalDateTime]] internal slot). It returns a record which contains the appropriate pattern, epochNanoseconds and timezone values for the input. This abstract operation functions as follows:
```
gibson042, line 927:
```suggestion
        The abstract operation HandleDateTimeTemporalZonedDateTime accepts the arguments _dateTimeFormat_ (which must be an object initialized as a DateTimeFormat) and _zonedDateTime_ (which must be an ECMAScript value has an [[InitializedTemporalDateTime]] internal slot). It returns a Record which contains the appropriate [[pattern]], [[epochNanoseconds]], and [[timeZone]] values for the input. This abstract operation functions as follows:
```

spec/intl.html
```diff
@@ -502,7 +504,7 @@ <h1>PartitionDateTimePattern ( _dateTimeFormat_, _x_ )</h1>
       <emu-alg>
         1. <ins>Let _x_ be ? HandleDateTimeValue(_dateTimeFormat_, _x_).</ins>
```
gibson042, line 505: I think you forgot to add `[[timeZone]]: *undefined*` to the Record returned by HandleDateTimeOthers.

gibson042 (changes requested):

spec/intl.html
Aditi-1400: Oh yes! I'll fix this.

spec/intl.html
Aditi-1400: Oops!

FrankYFTang:
I have a major concern about this change and like other members from TG2 to consider this.
Up to this point (before Temporal), Intl.DateTimeFormat format Date() which is basically an instance in time. And the choice of timeZone is decided and remember inside Intl.DateTimeFormat constructor so the same Intl.DateTimeFormat always format the same timezone if the timeZone is configure to be outputted. With this change, this mean if I set the Intl.DateTimeFormat to timeZone A, it will format all Date and other Temporal object to timeZone A but for the Temporal.ZonedDateTime it will format that to timeZone B from Temporal.ZonedDateTime . This significantly change the design contract of Intl.DateTimeFormat .

justingrant:
I've found it helpful to think about this PR as addressing three related but not identical questions:
1) Should optionless `ZonedDateTime.p.toLocaleString()` use the ZDT's time zone or the system time zone?
2) Should optionless `new Intl.DateTimeFormat().format(zdt)` use the ZDT's time zone or the system time zone?
3) How should conflicts be handled (in either API) when a `timeZone` option is provided and it conflicts with the ZDT's `timeZone` property?

It may be easier to handle these questions in the order above, because I suspect that it will be easier to get consensus on (1) than (2), and because resolving (3) may be hard until we decide on (2). 

@FrankYFTang @sffc (or anyone else) - do you object to (1) using the ZDT's time zone? If not, then it'd be great to consider (1) settled and scope the discussion to only (2) and (3).

> This significantly change the design contract of Intl.DateTimeFormat

Not sure if it affects this discussion, but \*all\* Temporal objects change the design contract of `Intl.DateTimeFormat`, in the sense that the type of the Temporal object being formatted will change the behavior of one or more DTF options:
* Formatting a PlainYearMonth will cause all options related to times and day-of-month to be ignored
* Formatting a PlainMonthDay will cause all options related to times and years to be ignored
* Formatting a PlainTime will cause all date-related options to be ignored
* Formatting a PlainDate will cause all time-related options to be ignored
* Formatting a PlainDateTime will cause the `timeZone` option to be interpreted differently. Instead of changing the value displayed, the `timeZone` option will only control the time zone shown when the `timeZoneName` option is present.

In other words, the behavior of `Intl.DateTimeFormat` is only unchanged for `Temporal.Instant` because its data model matches `Date`. All other Temporal types cause changes to `Intl.DateTimeFormat` behavior. We may decide that the ZDT changes are different and deserve different treatment than the others, but we should have a theory for why ZDTs changes are not OK while the others are OK.

FrankYFTang:
one possibility is 
1. Intl.DateTimeFormat.prototype.format always throw error if the timeZone in the Intl.DateFormat.Time is different from the one from the ZDT and 
2. the ZDT.prototype.toLocaleString always pass it's timeZone own into the options bag and pass into 
```
3. Let dateFormat be ? Construct(%DateTimeFormat%, « locales, options »).
```
in
https://tc39.es/proposal-temporal/#sup-temporal.zoneddatetime.prototype.tolocalestring
so the dateFormat used inside  ZDT.prototype.toLocaleString  is always the same as the timeZone in ZDT

sffc:
TG2 discussion 2023-02-09: https://github.com/tc39/ecma402/blob/master/meetings/notes-2023-02-09.md#temporal-issues-2479-prevent-loss-of-timezone-information

In the meeting, I proposed 7 options; I am adding number 8 that is similar to number 3 except slightly stricter to fully prevent programmer errors:

1. Don't accept ZDT in DTF.p.format
2. Accept ZDT; throw if the ZDT's time zone does not match the DTF's time zone (including the implicit system time zone)
3. Accept ZDT; if the DTF was constructed with an undefined time zone, use the ZDT's time zone; otherwise, if the DTF's explicit time zone differs from the ZDT's time zone, throw; otherwise, the time zone matches, and use it
4. Accept ZDT; use the ZDT's time zone, ignoring the DTF's time zone
5. Accept ZDT; use the DTF's time zone, ignoring the ZDT time zone
6. Add some option to the DTF constructor that tweaks this behavior
7. Add Intl.ZonedDateTimeFormat
8. Accept ZDT; if the DTF was constructed with an undefined time zone, use the ZDT's time zone; otherwise (if the DTF has an explicit time zone), throw (even if the time zone matches the ZDT)

sffc:
I like option 8 because it makes clear what is going on. `new Intl.DateTimeFormat("en", { timeZone: "x" })` will never work, and `new Intl.DateTimeFormat("en")` will always work. Implementation-wise, this requires storing exactly 1 bit of extra information in the Intl.DateTimeFormat: whether the timeZone is explicit or not. The format calls for all types except ZDT remain the same, and for ZDT, we have a check at the top if the bit is set to throw, and otherwise to do the runtime lookup of the time zone name.

justingrant:
Here's what I think are requirements of a solution:
* a) ZDT can be formatted using existing DateTimeFormat API.
* b) Doesn't ignore or override the time zone of a ZDT. (If you want another time zone, call `withTimeZone` and format the result.)
* c) Option-less `new Intl.DateTimeFormat().format(zdt)` matches agreed-on behavior of option-less `ZDT.p.toLocaleString`, which is to use the ZDT's time zone not the system time zone.
* d) Conflict between ZDT time zone and `timeZone` option should throw an exception.

With those in mind, here's how I'd group the proposed solutions.

**Preferred** _(whichever we choose, I'd want `toLocaleString` to share the same behavior)_
* <span>3.</span> Accept ZDT; if the DTF was constructed with an undefined time zone, use the ZDT's time zone; otherwise, if the DTF's explicit time zone differs from the ZDT's time zone, throw; otherwise, the time zone matches, and use it
* <span>8.</span> Accept ZDT; if the DTF was constructed with an undefined time zone, use the ZDT's time zone; otherwise (if the DTF has an explicit time zone), throw (even if the time zone matches the ZDT)

**Meh**
- <span>6.</span> Add some option to the DTF constructor that tweaks this behavior (my assumption is that this new option would be to opt out of using the ZDT time zone e.g. `ignoreZonedDateTimeTimeZone: true`, not to opt into it e.g. `timeZone: 'auto'`) 

**Not OK**
- <span>1.</span> Don't accept ZDT in DTF.p.format
- <span>2.</span> Accept ZDT; throw if the ZDT's time zone does not match the DTF's time zone (including the implicit system time zone)
- <span>4.</span> Accept ZDT; use the ZDT's time zone, ignoring the DTF's time zone
- <span>5.</span> Accept ZDT; use the DTF's time zone, ignoring the ZDT time zone
- <span>7.</span> Add Intl.ZonedDateTimeFormat

FrankYFTang:
> In the meeting, I proposed 7 options; I am adding number 8 that is similar to number 3 except slightly stricter to fully prevent programmer errors [**EDIT**: and also a further variation in number 9]:

Option 3, 8, and 9 all consider a concept of “if the DTF was constructed with an undefined time zone” and I would like to clarify 1) WHAT does “the DTF was constructed with an undefined time zone” mean and 2) WHEN would developer write code for such case.

I think the Temporal champion discussion seems to treat the undefined value as a value which the developers “do NOT specify a time zone” not realizing that actually is “the ONLY WAY for developers to SPECIFY to use the host environment's current time zone”.  

According to step 29-30 of 11.1.2 InitializeDateTimeFormat ( dateTimeFormat, locales, options ) in
https://tc39.es/ecma402/#sec-initializedatetimeformat
“29. Let timeZone be ? Get(options, "timeZone").
30. If timeZone is undefined, then
a. Set timeZone to DefaultTimeZone().“ 
And  “6.4.3 DefaultTimeZone ( )” https://tc39.es/ecma402/#sup-defaulttimezone
“The implementation-defined abstract operation DefaultTimeZone takes no arguments and returns a String. It returns a String value representing **the host environment's current time zone**, which is a valid (6.4.1) and canonicalized (6.4.2) time zone name.”

And also 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat
“timeZone The time zone to use. The only value implementations must recognize is "UTC"; the default is **the runtime's default time zone**. Implementations may also recognize the time zone names of the IANA time zone database, such as "Asia/Shanghai", "Asia/Kolkata", "America/New_York".”

Notice, this is the ONLY WAY a developer could SPECIFY their intent to construct an Intl.DateTimeFormat under “the host environment's current time zone”. In other words, we have to understand, when the developer write not to include a timeZone key with a string value, either by A) not including an option bag, B) an option bag without a timeZone key, or C) an option bag with a timeZone key and an undefined value , the developers are not telling the engine they do not care about what the timezone would be use (so the cannot just call a random() to pick one of the valid timezone from the IANA db), but instead, they are SPECIFYING that timezone MUST be the one which match the user’s “host environment's current time zone”. And this (all three A) - C) lead to the condition “30. If timeZone is undefined” to true)  is the ONLY WAY the developers to specify such intent. Also, it is the most common usage of Intl.DateTimeFormat- showing the formatted result according to the “host environment's current time zone”. The specified semantic is ONLY THE “host environment's current time zone” at the time of the construction (not 1 ms before, nor 1 ms after)  could be used for such a time zone. 

The behavior of Option 3, 8, and 9 violates such intent and provides the developers NO WAY to format the given ZDT under the “host environment's current time zone” with that DTF object.

Therefore, I strongly object to Option 3, 8 and 9. It gave the developers no option to use DTF w/ ZDT for the most common use case so far existing on the internet. 

Also, consider the following code snippet

```
let host_dft = new Intl.DateTimeFormat(undefined, {timeStyle: "full"});
host_dft.resolvedOptions().timeZone;
// “America/Los_Angeles”
host_dft.format(new Date());
// "11:47:58 AM Pacific Standard Time"
host_dft.format(Temporal.Now.plainDateTime("iso8601"));
// "2/23/2023, 11:47:58 AM"
host_dft.format(Temporal.Now.plainDateTimeISO());
// "2/23/2023, 11:47:58 AM"
host_dft.format(Temporal.Now.plainTimeISO());
// "11:47:58 AM"

host_dft.format(Temporal.Now.plainDateTime("iso8601", "Asia/Taipei"));
// "2/24/2023, 3:47:58 AM"
host_dft.format(Temporal.Now.plainDateTimeISO("Asia/Taipei"));
// "2/24/2023, 3:47:58 AM"
host_dft.format(Temporal.Now.plainTimeISO("Asia/Taipei"));
// "3:47:58 AM"

host_dft.format(Temporal.Now.zonedDateTime("iso8601"));
// "2/23/2023, 11:47:58 AM Pacific Standard Time"
host_dft.format(Temporal.Now.zonedDateTimeISO());
// "2/23/2023, 11:47:58 AM Pacific Standard Time"
```



Now, the question is what should be the output of the following two lines? Notice all the results above were either no timezone displayed since the value is not bound to a specific timezone (for the cases of Temporal PT and PDT) or under the “host environment's current time zone”. 
```
host_dft.format(Temporal.Now.zonedDateTime("iso8601", "Asia/Taipei"));
// ???
host_dft.format(Temporal.Now.zonedDateTimeISO("Asia/Taipei"));
// ???
```
I believe it should be 
Output I:
```
host_dft.format(Temporal.Now.zonedDateTime("iso8601", "Asia/Taipei"));
// "2/23/2023, 11:47:58 AM Pacific Standard Time"
host_dft.format(Temporal.Now.zonedDateTimeISO("Asia/Taipei"));
// "2/23/2023, 11:47:58 AM Pacific Standard Time"
```
Not Output II:
host_dft.format(Temporal.Now.zonedDateTime("iso8601", "Asia/Taipei"));
// "3:47:58 AM Taipei Standard Time"
host_dft.format(Temporal.Now.zonedDateTimeISO("Asia/Taipei"));
// "3:47:58 AM Taipei Standard Time"
```

Because 
1) the host_dft is constructed to use the “host environment's current time zone” , not "Asia/Taipei". Also because 
2) host_dft.resolvedOptions().timeZone clearly stated it will be used to format the string under “America/Los_Angeles” but  "3:47:58 AM Taipei Standard Time" is not under “America/Los_Angeles”.

I would object Option 3, 4, 8 and 9 because they do not allow a DFT constructed for the purpose of formatting under “host environment's current time zone” to be used for ZDT.
I would prefer Option 5 since it allows the developer to use a DFT constructed for the purpose of formatting under “host environment's current time zone” to do its SPECIFIED job. And if the developer wants to display the timezone under the timezone of the ZDT, they could simply construct another DFT for that. 
I am open to Option 2 too.
I think 1 and 7 (not mutually exclusive) may be a good way to go since we then have the opportunity to not call Get(options, "timeZone") on its constructor.

**Comment by FrankYFTang** [comment]
> Here's what I think are requirements of a solution:

I need to add another most important requirement for the solution:

e) The ZDT need to be able to be formatted under "the host environment's current time zone" by the DFT.

**Comment by gibson042** [comment]
@FrankYFTang I think you're overly hung up on "_DTF was constructed with an undefined time zone_", which is understandable because that's what appears in the option descriptions. But as I attempted to clarify in the February 15 meeting, there could be a different way to create a DateTimeFormat instance in a mode where it uses the time zone of an input ZonedDateTime instance—the key question is **whether such a mode should exist**.  If so, then the mechanism of expressing it with respect to backwards compatibility of details like «_`timeZone: undefined` already has an interpretation of "host time zone" and is the only way to specify that_» is a secondary concern.

**Comment by FrankYFTang** [comment]
Let's consider if the host time zone is “America/Los_Angeles” and we have the following code
```
let zdt = Temporal.Now.zonedDateTimeISO("Asia/Taipei");
```
under Option 5, we can do the following
```
let dft = new Intl.DateTimeFormat(undefined, {timeStyle: "full"});
let formatted_in_host_timezone = dft.format(zdt);
// "2/23/2023, 11:47:58 AM Pacific Standard Time"
let dft2 = new Intl.DateTimeFormat(undefined, {timeStyle: "full", timeZone: zdt.timeZone});
let formatted_in_zdt_timezone = dft2.format(zdt);
// "3:47:58 AM Taipei Standard Time"
```
for either choice, you create only 1 DFT to format the string.
for formatted_in_host_timezone, we need to create one DFT only
for formatted_in_zdt_timezone, we also need to create one DFT only 

Under Option 3, 8, or 9, how would you get the same result as the one under "the host environment's current time zone" ? and how many object do you need to create?
For those who think Option 3, 8 or 9 are good choice, could you write down how the code would looks like to get

formatted_in_host_timezone as "2/23/2023, 11:47:58 AM Pacific Standard Time" 

and 

formatted_in_zdt_timezone as "3:47:58 AM Taipei Standard Time"

and tell me how many object would need to be created to get each string?

**Comment by FrankYFTang** [comment]
> there could be a different way to create a DateTimeFormat instance in a mode where it uses the time zone of an input ZonedDateTime instance

It existed already
```
new Intl.DateTimeFormat(undefined, {timeStyle: "full", timeZone: zdt.timeZone});
```

**Comment by sffc** [comment]
If you really want to format using the local time zone, you can be explicit, like this:

```javascript
let dtf = new Intl.DateTimeFormat(undefined, {timeStyle: "full", timeZone: Temporal.Now.timeZone()});
dtf.format(zdt.toInstant());
```

---

P.S. If I got to choose by myself, my preference is still for option 1, just don't ever accept ZDT into a DTF. Users migrating from legacy Date can use `.toInstant()` to get the behavior they are accustomed to.

FrankYFTang:
> ```js
> let dtf = new Intl.DateTimeFormat(undefined, {timeStyle: "full", timeZone: Temporal.Now.timeZone()});
> dtf.format(zdt.toInstant());
> ```

In the above code, you need to create 3 observable object to do the job- 
1. one T.TimeZone: Temporal.Now.timeZone() , which internally always call SystemTimeZone() then always call CreateTemporalTimeZone().
2. one Intl.DateTimeFormat
3. one T.Instant :   Notice in https://tc39.es/proposal-temporal/#sec-temporal.zoneddatetime.prototype.toinstant it always call  CreateTemporalInstant() internally 

but ZDT.p.timeZone does NOT create a new T.TimeZone, it only return the already created TimeZone stored inside zdt.  see https://tc39.es/proposal-temporal/#sec-get-temporal.zoneddatetime.prototype.timezone

so this way you will need to crate two more (one T.TimeZone and another T.Instant) to do what option 2 or 5 could be done in only one observable object. 

BTW, If you are not passing the ZDT itself to the DTF here anyway, then why we need to  to construct the DTF with timeZone: Temporal.Now.timeZone() here? instead of just

```js
let dtf = new Intl.DateTimeFormat(undefined, {timeStyle: "full"});
dtf.format(zdt.toInstant());
```
In that way, you will still need to create one extra observable object - one T.Instant

FrankYFTang:
Another thing I would like to point out, as the spec stand as today, T.ZoneDateTime can only be used with the Intl.DTF which the timeZone is in the DefaultTimeZone() or otherwise it will throw RangeError during the format. And I do not believe this PR in the current shape address that.

1. all the T.*.p.toLocaleString() and the Intl.DTF.p.format() call FormatDateTime()
2. FormatDateTime() call PartitionDateTimePattern()
3. PartitionDateTimePattern() call  HandleDateTimeValue()
4. in HandleDateTimeValue() if the object is T.ZDT call HandleDateTimeTemporalZonedDateTime()
5. In HandleDateTimeTemporalZonedDateTime() step 6

```
5. Let timeZone be ? ToString(zonedDateTime.[[TimeZone]]).
6. If dateTimeFormat.[[TimeZone]] is not equal to DefaultTimeZone(), and timeZone is not equal to dateTimeFormat.[[TimeZone]], then
  a. Throw a RangeError exception.
```

So... as the Temporal spec which passed stage 3 and stand today, the Intl.DTF (and also T.ZDT.toLocaleString() ) can only be used if the ZDT is in the timezone equal to Temporal.Now.timeZome() and the Intl.DTF is in the same timezone. Otherwise the format will throw RangeError on the ZDT. And that limitation is not addressed by this PR which sent to TC39 several weeks ago.

spec/intl.html
```diff
@@ -927,7 +941,8 @@ <h1>HandleDateTimeTemporalZonedDateTime ( _dateTimeFormat_, _zonedDateTime_ )</h
       1. Return the Record {
           [[pattern]]: _pattern_.[[pattern]],
           [[rangePatterns]]: _pattern_.[[rangePatterns]],
-          [[epochNanoseconds]]: _instant_.[[Nanoseconds]]
+          [[epochNanoseconds]]: _instant_.[[Nanoseconds]],
+          [[timeZone]]: _timeZone_,
```
FrankYFTang, line 945: the "[[timeZone]]: _timeZone_" here is always equal to "_dateTimeFormat_.[[TimeZone]]" and equal to "DefaultTimeZone()" because otherwise, a RangeError would already throw 3 steps above.

spec/intl.html
FrankYFTang, line 507: notice the only possible value for _x_.[[timeZone]] here is either *undefined* or the same value of DefaultTimeZone() but nothing else. The value DefaultTimeZone() came from ZDT and the value *undefined* came from other data type. There is impossible to have any other values at this point Jan 31, 2023  655e3ef .

spec/intl.html
```diff
@@ -419,7 +418,10 @@ <h1>FormatDateTimePattern ( _dateTimeFormat_, <ins>_pattern_</ins>, _patternPart
           1. Perform ! CreateDataPropertyOrThrow(_nf3Options_, *"minimumIntegerDigits"*, _fractionalSecondDigits_).
           1. Perform ! CreateDataPropertyOrThrow(_nf3Options_, *"useGrouping"*, *false*).
           1. Let _nf3_ be ? Construct(%NumberFormat%, &laquo; _locale_, _nf3Options_ &raquo;).
-        1. Let _tm_ be ToLocalTime(<del>_x_</del><ins>_epochNanoseconds_</ins>, _dateTimeFormat_.[[Calendar]], _dateTimeFormat_.[[TimeZone]]).
+        1. <ins>If _timeZone_ is not *undefined*, then</ins>
+          1. <ins>Let _tm_ be ToLocalTime(_epochNanoseconds_, _dateTimeFormat_.[[Calendar]], _timeZone_).</ins>
```
FrankYFTang, line 422: I do not understand this change, at this point, the only possible values for _timeZone_ is either *undefined* or the same as _dateTimeFormat_.[[TimeZone]] and this change prodced exactly as 
```
Let _tm_ be ToLocalTime(<del>_x_</del><ins>_epochNanoseconds_</ins>, _dateTimeFormat_.[[Calendar]], _dateTimeFormat_.[[TimeZone]]).
```
all the time .

spec/intl.html
FrankYFTang, line 441: Notice the _timeZone_ here could only possible be *undefined* or the same as _dateTimeFormat_.[[TimeZone]] . there are no other possible values for _timeZone_ , guarded by the  RangeError throw in step 6 of  HandleDateTimeTemporalZonedDateTime see https://tc39.es/proposal-temporal/#sec-temporal-handledatetimevaluetemporalzoneddatetime

justingrant:
> And that limitation is not addressed by this PR which sent to TC39 several weeks ago.

Thanks for your review! Once we decide what we want the behavior to be, we can change the spec text to match that expected behavior, and also to fix any bugs found in the PR's spec text.

> ```js
> ```js
> let dtf = new Intl.DateTimeFormat(undefined, {timeStyle: "full"});
> dtf.format(zdt.toInstant());
> ```
> In that way, you will still need to create one extra observable object - one T.Instant

If object creation is a problem, then here's a zero-object alternative:

```js
let dtf = new Intl.DateTimeFormat(undefined, {timeStyle: "full"});
dtf.format(zdt.epochMilliseconds);
```

> 1. the host_dtf is constructed to use the “host environment's current time zone” , not "Asia/Taipei"
> 2. host_dtf.resolvedOptions().timeZone clearly stated it will be used to format the string under “America/Los_Angeles” but  "2/24/2023, 3:47:58 AM Taipei Standard Time" is not under “America/Los_Angeles”.

Because the data model of each Temporal type (other than Instant) varies from `Date`, the type passed to `format` will necessarily cause some DTF options to be overridden. For example, `Temporal.PlainDate` lacks time units and time zones in its data model, so even if `.resolvedOptions().hour === '2-digit'` and `.resolvedOptions().timeZoneName === 'long'`, the output will act as if both options were `undefined`.

 Just like the behavior of the `hour` and `timeZoneName` options are overridden in the `Temporal.PlainDate` case, it's reasonable that the behavior of other options like `timeZone` or `timeZoneName` may change in the `Temporal.ZonedDateTime` case, especially in the case where those options are `undefined`.

I agree that this type-specific behavior of `format` makes reasoning about `resolvedOptions()` more complicated. But unless we wanted to have different variations of resolvedOptions, e.g. `resolvedOptions('plainDate')` or  `resolvedOptions('zonedDateTime')` then I'm not sure how we can avoid that complexity. Also, `resolvedOptions` is more of an advanced API than `new DTF()` and `format`. If we have to add complexity somewhere, it's usually better to add it to already-advanced APIs if it means we can keep more popular APIs simpler and more ergonomic.

One other thing to note is that the no-options DTF output for `Date` and ZDT will be quite different: `'2/23/2023'` for `Date` vs. `'2/23/2023, 10:16:29 PM PST'` for ZDT. This may help developers figure out why they're getting different results, and may clarify that ZDT's data model includes a time zone while `Date`'s does not. I realize this won't help in the case where non-default options are used, but it will make it easier in simpler cases and when debugging.

My general point is that we shouldn't assume that passing Temporal objects to `format` must behave exactly like `Date`. Instead we should try to have `format` behave appropriately for the data model of its argument.

> e) The ZDT need to be able to be formatted under "the host environment's current time zone" by the DTF.
> 

My understanding is that (e) is solvable in all of the 1-9 solutions we're discussing, by passing a non-DTF argument to `format`. Correct? If so, then I'd prefer to treat this requirement as unrelated to our 1-9 choice. Sound OK?

> f) The formatting behavior of the DTF need to consistently within the same object and should not preduce conflicting output.

It kinda depends on what kind of consistency is more important:

* **i) Consistency between `zdt.toLocaleString(args)` and `new DTF(args).format(zdt)`**, where "consistency" means getting the same results back given the same `args`.
* **ii) Consistency between `zdt.toLocaleString()` and `.toLocaleString()` of all other ECMAScript built-ins**, where "consistency" means, at a minimum, that you can call it with no parameters (or a single locale parameter) and can get a default localized format of the receiver without requiring any additional options.
* **iii) Consistency between `ZDT.p.toLocaleString` and `new Intl.DateTimeFormat().format(zdt.epochMilliseconds)`**, where "consistency" means that both calls return the same output.

Unless I'm missing something, all three cannot be satisfied with one solution. We must pick one to break.

If we break (i), then porting from `toLocaleString` to DTF gets more difficult and buggy. This may dissuade developers from using DTF.

If we break (ii), then developers can't just call `zdt.toLocaleString()` without it throwing, which is an ergonomic blocker for the most basic localized code. This may dissuade developers from using `toLocaleString`.

If we break (iii), then localized formatting of two objects with different data models won't display the same output. For developers who are mistakenly assuming that `Date` and `Temporal.ZonedDateTime` are the same thing, this may cause confusion until they realize that they are different types with different data models and different formatting behavior. 

Of the three, I suspect that breaking (iii) will have the least negative impact on our overall shared goal, which is encouraging more developers to successfully use `toLocaleString` and `Intl` for localization instead of rolling their own hacks like `${month}/${day}/${year}`.

I think you're correct that breaking (iii) will confuse some developers. But I think those problems are smaller in scope (and will impact fewer developers) than breaking (i) or (ii).

spec/intl.html
```diff
@@ -922,7 +935,8 @@ <h1>HandleDateTimeTemporalZonedDateTime ( _dateTimeFormat_, _zonedDateTime_ )</h
       1. Return the Record {
           [[pattern]]: _pattern_.[[pattern]],
           [[rangePatterns]]: _pattern_.[[rangePatterns]],
-          [[epochNanoseconds]]: _instant_.[[Nanoseconds]]
+          [[epochNanoseconds]]: _instant_.[[Nanoseconds]],
```
justingrant, line 938: _(leaving a comment here because I can't leave it on line 918 above)_ 

I think this spec text above may be a problem: 

> 1. If _dateTimeFormat_.[[TimeZone]] is not equal to DefaultTimeZone(), and _timeZone_ is not equal to _dateTimeFormat_.[[TimeZone]], then
>   1. Throw a *RangeError* exception.

If I'm reading this correctly, the code below would not throw when the DTF's `timeZone` option matches the system's time zone.  Based on my understanding, this is not desired behavior. @ptomato @sffc is this a correct reading?

```js
// run this code on a computer with system time zone America/Los_Angeles
zdt = Temporal.ZonedDateTime.from('2020-01-01[Asia/Tokyo');

new Intl.DateTimeFormat('en').format(zdt);
// => as expected, formats using ZDT time zone Asia/Tokyo

new Intl.DateTimeFormat('en', { timeZone: 'America/Denver' }).format(zdt);
// => as expected, RangeError because timeZone option is disallowed when formatting ZDT

new Intl.DateTimeFormat('en', { timeZone: 'America/Los_Angeles' }).format(zdt);
// expected: RangeError because timeZone option is disallowed when formatting ZDT
// actual: no exception, uses ZDT's time zone because time zone option matches DefaultTimeZone()
```

justingrant (changes requested):

spec/intl.html
```diff
@@ -419,7 +418,9 @@ <h1>FormatDateTimePattern ( _dateTimeFormat_, <ins>_pattern_</ins>, _patternPart
           1. Perform ! CreateDataPropertyOrThrow(_nf3Options_, *"minimumIntegerDigits"*, _fractionalSecondDigits_).
           1. Perform ! CreateDataPropertyOrThrow(_nf3Options_, *"useGrouping"*, *false*).
           1. Let _nf3_ be ? Construct(%NumberFormat%, &laquo; _locale_, _nf3Options_ &raquo;).
-        1. Let _tm_ be ToLocalTime(<del>_x_</del><ins>_epochNanoseconds_</ins>, _dateTimeFormat_.[[Calendar]], _dateTimeFormat_.[[TimeZone]]).
+        1. <ins>If _dateTimeFormat_.[[TimeZone]] is not *undefined*, then</ins>
+          1. <ins>Throw a *RangeError* exception.</ins>
```
ptomato, line 422: To implement option 8, I think the logic we want here is:
- If both _dateTimeFormat_.[[TimeZone]] and _timeZone_ are not undefined, throw a RangeError exception.
- If _timeZone_ is undefined, set _timeZone_ to _dateTimeFormat_.[[TimeZone]].
- If _timeZone_ is undefined, set _timeZone_ to DefaultTimeZone().

That means we also have to leave _dateTimeFormat_.[[TimeZone]] undefined if it's not explicitly specified in InitializeDateTimeFormat and not call DefaultTimeZone there. Probably in practice, implementations would still call DefaultTimeZone in InitializeDateTimeFormat, and keep track of whether the time zone was explicitly specified using a separate flag. We could choose to do that in the spec as well. It might look something like this:
- If _timeZone_ is not undefined and _dateTimeFormat_.[[TimeZoneSpecified]] is true, throw a RangeError exception.
- If _timeZone_ is undefined, set _timeZone_ to _dateTimeFormat_.[[TimeZone]].

spec/intl.html
ptomato: If we go with the suggestion from my previous comment, then _timeZone_ can't be undefined here anymore.

spec/intl.html
```diff
@@ -562,9 +563,15 @@ <h1>PartitionDateTimeRangePattern ( _dateTimeFormat_, _x_, _y_ )</h1>
           1. <ins>If ! SameTemporalType(_x_, _y_) is *false*, throw a *TypeError* exception.</ins>
         1. <ins>Let _x_ be ? HandleDateTimeValue(_dateTimeFormat_, _x_).</ins>
         1. <ins>Let _y_ be ? HandleDateTimeValue(_dateTimeFormat_, _y_).</ins>
+        1. <ins>If x.[[timeZone]] is not equal to y.[[timeZone]], then</ins>
+          1. <ins>Throw a *RangeError* exception.</ins>
+        1. <ins>Else if x.[[timeZone]] is not *undefined*, then</ins>
```
ptomato, line 568: This would need a similar change to the logic, either to include a call to DefaultTimeZone or checking whether the _dateTimeFormat_.[[TimeZone]] was specified explicitly, whichever way we decide to go with.

spec/intl.html
ptomato: Yeah, we'd need to change those lines above. If we go with my suggestion from a previous comment, I think we can remove them altogether because the check would be done in FormatDateTimePattern?

sffc:
TG2 discussion: https://github.com/tc39/ecma402/blob/master/meetings/notes-2023-03-09.md#zoneddatetime--datetimeformat

Conclusion: Not all stakeholders were present, but the direction of "option 8" was okay with everyone who was present.

FrankYFTang:
Option 8 is unacceptable
"8. Accept ZDT. If the DTF was constructed with an undefined time zone, use the ZDT's time zone. Otherwise (if the DTF has an explicit time zone), throw (even if the time zone matches the ZDT)."

Because when the DTF was constructed withe a *undefined* value in the option bag as time zone, it is SPECIFIED to use the  DefaultTimeZone() as the timeZone.  The *undefined* value is a SPECIFIED value, and also the only value to be use for the developer to contstruct the Intl.DateTimeFormat to use the  DefaultTimeZone() as the timeZone in the constructor. It is not an UNSPECIFIED value, it is a special SPECIFIED value using the JavaScript *undefined* as a dedicated SPECIFIED value. The Intl.DateTimeFormat ALWAYS has a timeZone, either string value to indiate the one in the IANA TZ DB, or the IANA TZ DB value of the DefaultTimeZone() if the option bag use *undefined* as a value to indicate that. That is the semantic from the day one of the Intl.DateTimeFormat. There are no such thing "the DateTime object has no timezone" , a *undefined* value in the option bag is still a timeZone- the defaultTimeZone() and there are no other way currently to indicate the developer want to use defaultTimeZone() other than that.  

The concept that if option bag has timeZone: *undefined* or no timeZone in the option bag represent "the DateTime object has no timezone" is incorrect. When that happen, it indicate the developer want to use the defaultTimeZone() , not any time zone (so engine could use a rand() to select a random one from TimeZoneDB)  or no time zone.

FrankYFTang:
in https://github.com/tc39/ecma402/blob/master/meetings/notes-2023-03-09.md#zoneddatetime--datetimeformat
"The goal for solving this is to recognize that a very large percentage of Intl.DateTime objects have no timezone at all. "

this is a totally flaw statement. The reality is 100% of Intl.DateTime object has timeZone. 

(new Intl.DateTimeFormat()).resolvedOptions().timeZone is never *undefined*.

There are 0% of the Intl.DateTime object has no timezone, not a single one.

justingrant:
> in https://github.com/tc39/ecma402/blob/master/meetings/notes-2023-03-09.md#zoneddatetime--datetimeformat "The goal for solving this is to recognize that a very large percentage of Intl.DateTime objects have no timezone at all. "
> 
> this is a totally flaw statement. The reality is 100% of Intl.DateTime object has timeZone.

Hi Frank - Meeting notes are often inexact. What was meant (and what was understood by attendees during the meeting) was more like: "a very large percentage of Intl.DateTime objects were not created with an explicit `timeZone` option."  Thanks for bringing this to my attention. I'll send a PR to correct the notes to match the discussion in the meeting.

FrankYFTang:
> > And that limitation is not addressed by this PR which sent to TC39 several weeks ago.
> 
> Thanks for your review! Once we decide what we want the behavior to be, we can change the spec text to match that expected behavior, and also to fix any bugs found in the PR's spec text.

My point is when you present this PR to Jan TC39, it was an PR not going to address anything. I am happy I blocked it and request more time to review this. 

Here is my view 
1. T.ZoneDateTime object always has a timeZone
2. Intl.DateTimeFormat object always has a timeZone

Think about this, if you call the following code 

```
let z1 = Temporal.Now.zonedDateTime('iso8601');
let z2 = Temporal.Now.zonedDateTime('iso8601', undefined);
let z3= Temporal.Now.zonedDateTimeISO();
let z4= Temporal.Now.zonedDateTimeISO(undefeind);
let tz1 = Temporal.Now.timeZone ( );
let dft1 = Intl.DateTimeFormat();
let dft2 = Intl.DateTimeFormat("en");
let dft3 = Intl.DateTimeFormat("en", {});
let dft4 = Intl.DateTimeFormat("en", {timeZone: undefined});
```
which object has timeZone? which one does not have timeZone? Why?

My answer is, all of them has timeZone even none of them use a string to specify a timeZone. The very same reason z1-z4 has a timeZone above is exactly the same reason why dft1 - dft4 also has a timeZone. If you argue dft1 - dft4 has no timezone then I would argue z1-z4 also has no timeZone by your same argument, but you will disagree with that, right? for the very same reason you believe z1-z4 and tz1 always has a timeZone, I would argue dft1 - dft4 also always has a timezone.

justingrant:
> > in https://github.com/tc39/ecma402/blob/master/meetings/notes-2023-03-09.md#zoneddatetime--datetimeformat "The goal for solving this is to recognize that a very large percentage of Intl.DateTime objects have no timezone at all. "
> > this is a totally flaw statement. The reality is 100% of Intl.DateTime object has timeZone.
> 
> Hi Frank - Meeting notes are often inexact. What was meant (and what was understood by attendees during the meeting) was more like: "a very large percentage of Intl.DateTime objects were not created with an explicit `timeZone` option." Thanks for bringing this to my attention. I'll send a PR to correct the notes to match the discussion in the meeting.

I filed https://github.com/tc39/ecma402/pull/759 to update the notes to match the discussion.

FrankYFTang:
> "a very large percentage of Intl.DateTime objects were not created with an explicit timeZone option." 

ok, could you estimate after you ship Temporal and users start to use Temporal, how many % of the ZoneDateTime would be created by calling either of the following?

```
let z1 = Temporal.Now.zonedDateTime('iso8601');
let z2 = Temporal.Now.zonedDateTime('iso8601', undefined);
let z3= Temporal.Now.zonedDateTimeISO();
let z4= Temporal.Now.zonedDateTimeISO(undefined);
let z5 = Temporal.Now.zonedDateTime('iso8601', Temporal.Now.timeZone ( ));
let z6= Temporal.Now.zonedDateTimeISO(Temporal.Now.timeZone ( ));
```
If I follow your definition of "not created with an explicit timeZone option" I would argue all z1-z6 are created also  "not created with an explicit timeZone option" since no timeZone String were specified in the code. 
Therefore, I could also argue "a very large percentage of Temporal.ZoneDateTime objects will not created with an explicit timeZone option."  but one of the calling pattern above, would you agree?

FrankYFTang:
"a very large percentage of Intl.DateTime objects were not created with an explicit timeZone option."

"created with an explicit timeZone option." or "created without an explicit timeZone option." is unimportant to Intl.DateTimeFormat because  "created without an explicit timeZone option" is actually the only way to create the Intl.DateTimeFormat with DefaultTimeZone  and there are no other ways to do that. there are no string value we can use for timeZone key in the option bag to indicate the developer desire to construct the Intl.DateTimeFormat with DefaultTimeZone therefore the only way to explicit indicate it want to use the DefaultTimeZone is actually explicitly not to set the timeZone to a string value, but not define it in the option bag, or map it to *undefined* . But the current spec text has not observable operation to distinguish the timeZone is set to undefiend or the option bag has no timeZone key (you may change the spec to check the present of the key, but not in the current spec) so all the code wrote till this point, either there are no option bag, or option bag has no timeZone key or timeZone key explicit set to *undefined* all indicate the developer want  to use the DefaultTimeZone. You cannot think in these case the developer does not set the timeZone, they set, set it to DefaultTimeZone and there are no other ways (before Temporal) to set that.

FrankYFTang:
I check the ECMA402 history
ECMA-402, 1st edition, December 2012
would take either "UTC" or *undefined* as timeZone in option bag of Intl.DateTimeFormat to distinguish the Intl.DateTimeFormat is constructed for UTC or the local time zone.
and the resolvedOptions().timeZone could be *undefined* 

But starting ECMA-402 2nd edition, June 2015 
Intl.DateTimeFormat will take either a valid IANA TimeZone db name in string form or *undefined* as timeZone in option bag of Intl.DateTimeFormat  and always return the resolved timeZone string in the resolvedOptions().timeZone  always.

FrankYFTang:
"8. Accept ZDT. If the DTF was constructed with an undefined time zone, use the ZDT's time zone. Otherwise (if the DTF has an explicit time zone), throw (even if the time zone matches the ZDT)."

How about the ZDT which was  constructed with an undefined time zone?  (as z1 - z4 in my above code example) should we use the DFT timezone for that case instead? (why not?)

FrankYFTang:
If you believe Intl.DateTimeFormat should have an internal flag to indicate it was 
1) "constructed without explicit stated timeZone string" vs 
2) "constructed with explicit timeZone string", 

then I will argue you will also need to add such an internal flag to T.ZoneDateTime and T.TimeZone because T.TimeZone could also be created by calling Temporal.Now.timeZone ( ) without explicit stating timeZone string  and T.ZonedDateTime  could be created by calling  Temporal.Now.zonedDateTime () or Temporal.Now.zonedDateTimeISO() without explicit stating timeZone string. 

If you are not willing to handle that as a mode in T.TimeZone and T.ZonedDateTime to complicate the Temporal API for unnecessary overhead, then I am neither willing to handle that as a mode in Intl.DateTimeFormat to complicate the Intl API.

littledan:
> The undefined value is a SPECIFIED value, and also the only value to be use for the developer to contstruct the Intl.DateTimeFormat to use the DefaultTimeZone() as the timeZone in the constructor. It is not an UNSPECIFIED value, it is a special SPECIFIED value using the JavaScript undefined as a dedicated SPECIFIED value. 

Although these are the semantics of `Intl.DateTimeFormat` today, I don't think it's how most developers think of it. Temporal adds an explicit API to get the current timezone, which I think developers will find a lot more intuitive than the current method with `new Intl.DateTimeFormat().resolvedOptions.timeZone`.

Given how much else in Date is a calculation based on the "current time zone" (at the time of the calculation, but not attached to any operator object), a reasonable mental model that gets you almost there (modulo timezone changes during JS execution, but how many developers know about that?) is that the current time zone is "in" the Date, rather than the DateTimeFormat object.

Given this, and given how implicit and tricky a missing property on an options bag is as an explicit signal, I think option 8 makes sense. I hope we can continue discussing this question at the next Temporal champions meeting on Thursday.

sffc:
> "created without an explicit timeZone option" is actually the only way to create the Intl.DateTimeFormat with DefaultTimeZone and there are no other ways to do that

Temporal adds a new, explicit way to do it:

`new Intl.DateTimeFormat("en", { timeZone: Temporal.Now.timeZone() })`

> I will argue you will also need to add such an internal flag to T.ZoneDateTime and T.TimeZone

The fundamental difference here is that ZonedDateTime and TimeZone are clearly defined to carry time zone information. It is part of their name.

The perspective that I think many people in this issue are taking is based on how Intl.DateTimeFormat reads in code. `{ timeZone: undefined }` _reads_ as though it means "please automatically use the most appropriate time zone". The spec currently uses the DefaultTimeZone in all cases. What we're proposing is that when a ZonedDateTime is being formatted, that time zone is a better choice than DefaultTimeZone.

justingrant:
I find it helpful when trying to resolve disagreements like this to try to ensure that everyone understands everyone else's position. Not necessarily to agree (that can come later) but at least to have a common understanding about why we believe what we do. 

With that in mind, Frank I'll try to summarize below what I understand about your positions. Please let us know if I'm understanding them correctly. 

The time zone used by `dtf.p.format(zdt)` should never ignore the time zone in the DTF instance, even if the DTF instance was created with an undefined `timeZone`, because:
* `dtf.p.format(new Date())` uses the DTF's time zone (and has always done this for 10+ years), so users will expect this behavior when DTF is used to format a non-`Date` type too.
* Users will have a similar expectation from `new Intl.DateTimeFormat()` and `Temporal.Now.zonedDateTimeISO()`, because neither API explicitly provides a time zone name. Therefore, any solution should treat those two cases the same way. 
* It is confusing if the value returned by `resolvedOptions().timeZone` differs from the timezone used in the localized string output of `format()`.
* There are potential performance and/or memory concerns if `DateTimeFormat` is forced to store an additional bit to track whether timeZone was `undefined` or not.
* Degrading the performance of `Date.p.toLocaleString` (which is prominently featured in benchmarks) is not acceptable.

Is this correct?  Please clarify (with short bullet points if possible!) the things I got wrong. Thanks!

There's also one additional concern that Shane raised that I think supports your position, but I don't remember if you agreed or not. If you do agree, then should we lump it in with the points above in our discussion?
* DTF achieves performance gains by caching expensive-to-create ICU objects. If the localized name of the time zone varies based on the parameter to `format`, then we may not be able to get as much (or any?) performance gains from DTF compared to `toLocaleString`.

gibson042:
> * Degrading the performance of `Date.p.toLocaleString` (which is prominently featured in benchmarks) is not acceptable.

I think it is important to note that degradation is only possible for code that works today, i.e. with input values that are not part of Temporal (`undefined`, Date instance, or other number-coercible).

FrankYFTang:
> , a reasonable mental model that gets you almost there (modulo timezone changes during JS execution, but how many developers know about that?) is that the current time zone is "in" the Date, rather than the DateTimeFormat object.

Totally disagree. 
Consider the following code in your developer console:
```js
// The current timezone is in 'America/Los_Angeles'
let d = new Date();
let dft1 = new Intl.DateTimeFormat("en", {timeStyle: "short"}); 
// the timezone of dft1 is 'America/Los_Angeles'
```
now, go to your system setting set the timezone to 'America/New_York'

```
let dft2 = new Intl.DateTimeFormat("en", {timeStyle: "short"}); 
// the timezone of dft2 is 'America/New_York'

let result1 = dft1.format(d);
let result2 = dft2.format(d);
```
IF "the current time zone is "in" the Date, rather than the DateTimeFormat object." is true, then result1 and result2 will be exactly the same, right?

But no, that is not the case,  (try it yourself). Therefore I prove your statement above is FALSE.

FrankYFTang:
> > "created without an explicit timeZone option" is actually the only way to create the Intl.DateTimeFormat with DefaultTimeZone and there are no other ways to do that
> 
> Temporal adds a new, explicit way to do it:
> 
> `new Intl.DateTimeFormat("en", { timeZone: Temporal.Now.timeZone() })`

The qeustion is NOT how could  developers in the future create Intl.DateTimeFormat with explicit timezone , but rather to address Justin's statement in the TG2 meeting note:

"JGT: ... The goal for solving this is to recognize that a very large percentage of Intl.DateTimeFormat objects were not created with an explicit `timeZone` option. "

JGT claimed up to this point "a very large percentage of Intl.DateTimeFormat objects were not created with an explicit `timeZone` option. "

Temporal is not yet shipped with any browser yet, so no web page ever can use the new Intl.DateTimeFormat("en", { timeZone: Temporal.Now.timeZone() }) 
to create Intl.DateTimeFormat to specify the timeZone for the default timezone. So the only way to explicit the use of default timezone TODAY (2023-03-15) are
```js  
Intl.DateTimeFormat("en", { timeZone: undefined }) 
Intl.DateTimeFormat("en", {} )
Intl.DateTimeFormat("en") 
```
and no other way AS TODAY (and ever since the creation of the Intl.DateTimeFormat)

FrankYFTang:
> The fundamental difference here is that ZonedDateTime and TimeZone are clearly defined to carry time zone information. It is part of their name.

I find the "part of the name" argument very strange. Since the name is ZonedDateTime, not ZoneDateTime. The Zoned part is a modifier with a suffix 'd' and make it an adjustive in English to modify DateTime. It only mean the DateTime is zoned, just like no body willl think Ice Cream carry Ice, Hot Dog carry hot, French Fries carry France, or Italaian Saussage carry Italy.

FrankYFTang:
> The perspective that I think many people in this issue are taking is based on how Intl.DateTimeFormat reads in code. `{ timeZone: undefined }` _reads_ as though it means "please automatically use the most appropriate time zone".

The problem is such definition violate what was clearly understand by the developers on the internet last 8-10 years. It does NOT mean "the most appropriate time zone". It specificlly mean "the timezone the user is in when the object is created". Why? 
A. Because dft.resolovedOptions() has a timeZone property.
If it means "use the most appropriate time zone" then it should not have a timeZone property but it does has, and the document stated what that value is "the user default timezone" (not "the most appropriate time zone")

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat

"timeZone
The time zone to use. The only value implementations must recognize is "UTC"; the default is the runtime's default time zone. Implementations may also recognize the time zone names of the [IANA time zone database](https://www.iana.org/time-zones), such as "Asia/Shanghai", "Asia/Kolkata", "America/New_York".
"
Notice it clearly states " the default is THE runtime's default time zone." NOT "the most appropriate timezone" NOR "a random timezone".

```
console.log(usedOptions.timeZone);
// "America/New_York" (the users default timezone)
```
also
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/resolvedOptions
```
timeZone
The value provided for this property in the options argument; [undefined](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined) (representing the runtime's default time zone) if none was provided. Warning: Applications should not rely on [undefined](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined) being returned, as future versions may return a [String](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String) value identifying the runtime's default time zone instead.
```
It said clearly "the runtime's default time zone ". 


B. Because dft.resolvedOptions().timeZone always match the system time zone while dft is created.


The concept that default represent  anything other than " the system default time zone" is simply against the reality of the web and against all the documentation on the web about this. 

Let me phrase it again, up until now, the one and only way to specify the Intl.DateTimeFormat to use the default timezone is by using the default (in the 3 ways I specified) and no other ways. I understand in the future after Temporal ship you can use Temporal.Now.timeZone() but that is not yet available to the users. And all the code currently written for Intl.DateTimeFormat  can only use the default to create such. 

Other printed sources to show the timeZone default value is clearly defined as "the default system time zone" NOT a non-existing concept of "the most appropriate time zone".

[1] Phang, Chong Lip, "ECMAScript 2022 The Definitive Guide to Modern JavaScript", 2022, p124 
https://www.google.com/books/edition/ECMAScript_2022/K_uZEAAAQBAJ?hl=en&gbpv=1&dq=Intl.DateTimeFormat+default+timezone&pg=PA124&printsec=frontcover
[2]  Phang, Chong Lip, "An Effective Guide to Modern JavaScript (ECMAScript 2017 / ES8)", 2017, p102
https://www.google.com/books/edition/An_Effective_Guide_to_Modern_JavaScript/--gvDwAAQBAJ?hl=en&gbpv=1&dq=Intl.DateTimeFormat+default+timezone&pg=PA102&printsec=frontcover
[3] Ferguson and  Cirkel "JavaScript Recipes A Problem-Solution Approach", 2016, Apress,  p179
https://www.google.com/books/edition/JavaScript_Recipes/hefIDQAAQBAJ?hl=en&gbpv=1&dq=Intl.DateTimeFormat+timezone&pg=PA179&printsec=frontcover
[4] Casado, Pablo Fernandez, "Domine Javascript 4a Edición", 2022, Ediciones de la U, p79
https://www.google.com/books/edition/Domine_Javascript_4a_Edici%C3%B3n/lANcEAAAQBAJ?hl=en&gbpv=1&dq=Intl.DateTimeFormat+timezone&pg=PA79&printsec=frontcover

FrankYFTang:
> Is this correct? Please clarify (with short bullet points if possible!) the things I got wrong. Thanks!

You summarized it very well. But there are additional points.
* Intl.DateTimeFormat always carry a timeZone. (Therefore, in the TG2 meeting note your statement "Today in ECMAScript no ECMAScript object carries a timezone with it." is FALSE). 
* Vast majority of the developer want to create an Intl.DateTimeFormat with the the user's default time zone.
* As today, which represent all the pre-existing code, the ONLY WAY to "create an Intl.DateTimeFormat with the the user's default time zone" is to use default but there are no other way (for example, developer cannot pass in {timeZone: "system"} or  {timeZone:"default" } as string to indicate that.
* timeZone inside the Intl.DateTimeFormat is an immutable property, once it is created, the value is not changed, even if the user use the UI to chagne the system timeZone to other value, or the user travel to a different place and his cell phone pick up the timeZone change and automatically change the default timezone to a different value. A Intl.DateTimeFormat created after  that with default will have a different timeZone than the one which created before that . Therefore, this timeZone is immutable inside the Intl.DateTimeFormat object, it does NOT represent "the user system default at the time calling the format", it represent "the user system default at the time the object being created"
* All objects created of the 3 classes - Intl.DateTimeFormat and Temporal.ZonedDateTime and Temporal.TimeZone carried a "resolved" timeZone in them (yes, even Temporal.TimeZone, because the timeZone created by Temporal.Now.timeZone() resolved to different timeZone if user change the system timezone between the calls. You have to treat the timeZone carried by these three object the same weight because internally it was resolved by the same routine. For example, if you believe new Intl.DateTimeFormat() carray the timeZone which represent "the most appropriate timeZone" (which I strongly against, but just for argument sake, let's say that is what it represent), then you have to also believe the timeZone created by Temporal.Now.timeZone()  or the one carried inside Temporal.Now.zoneDateTimeISO()  or Temporal.Now.zoneDateTime()  also means "the most appropriate timeZone" because internally they call into the same routine. Then, we will end up have no routine in the JavaScript to represent creating the "default user timeZone", right?
* Since the timeZone carried by objects of these 3 classes (assuming all created by new Intl.DateTimeFormat() without any argument, created by Temporal.Now.timeZone() which does not expect any argument, created by Temporal.Now.zonedDateTime() while without argument or with undefined, or created by Temporal.Now.zonedDateTimeISO("iso8601") without the 2nd argument or undefined as the 2nd one.) all created by the same underline routine, we have to treat them as equaliy important. Notice the value could be different because user may change the system timezone between the calls. 

Another thing I want to point out is, even after we ship Temporal, the following lines still could be different
```js
let tz = Temporal.Now.timeZone();
let dft1 = new Intl.DateTimeFormat('en', {timeZone: undefined});
let opt = {timeZone: Temporal.Now.timeZone()};
let dft2 = new Intl.DateTimeFormat('en', opt);
let dft3 = new Intl.DateTimeFormat('en', {timeZone: tz);
```

dft1 carry the timeZone of the value "user default timeZone" WHILE the Intl.DateTimeFormat is constructed.
dft2 carry the timeZone of the value "user default timeZone" BEFORE the Intl.DateTimeFormat is constructed.
dft3 carry the timeZone of the value "user default timeZone" WHILE tz is constructed.

If we run the following code and change timezone in between, for example

```js
// set timezone to "Asia/Tapiei"
let tz = Temporal.Now.timeZone();
// set timezone to "Asia/Tokyo"
let dft1 = new Intl.DateTimeFormat('en', {timeZone: undefined});
// set timezone to "Europe/London"
let opt = {timeZone: Temporal.Now.timeZone()};
// set timezone to "Asia/Tokyo"
let dft2 = new Intl.DateTimeFormat('en', opt);
let dft3 = new Intl.DateTimeFormat('en', {timeZone: tz);
```
(notice I set the system timeZone back to "Asia/Tokyo" before all the calls to new Intl.DateTimeFormat() )
dft1 will set to "Asia/Tokyo" 
dft2 will set to "Europe/London"
dft3 will set to "Asia/Tapiei"
passing undefined in, instead of passing Temporal.Now.timeZone() is "an atomic operation" to create that object with the timeZone of the "current user timezone" of its creation, which is not provided by passing in Temporal.Now.timeZone()

justingrant:
> Intl.DateTimeFormat always carry a timeZone. (Therefore, in the TG2 meeting note your statement "Today in ECMAScript no ECMAScript object carries a timezone with it." is FALSE).

Sorry, the notes were not transcribed correctly. I filed yet another PR to correct them. Here's what was discussed and what was intended to be in the notes:

```
JGT: Today in ECMAScript no ECMAScript object that can be formatted by `Intl.DateTimeFormat`
carries a timezone with it. With Temporal.ZonedDateTime, this object carries a timezone, and there’s
a potential conflict. 
```

> Vast majority of the developer want to create an Intl.DateTimeFormat with the the user's default time zone.

Agreed. 

> As today, which represent all the pre-existing code, the ONLY WAY to "create an Intl.DateTimeFormat with the the user's default time zone" is to use default because there are no other way (for example, developer cannot pass in {timeZone: "system"} or {timeZone:"default" } as string to indicate that.

Agreed. 

> All objects created of the 3 classes - Intl.DateTimeFormat and Temporal.ZonedDateTime and Temporal.TimeZone carried a "resolved" timeZone in them 

Agreed. 

> Then, we will end up have no routine in the JavaScript to represent creating the "default user timeZone", right?

Agreed. All three classes fetch the current system time zone at time of construction, store it immutably, and use it later.

> If we run the following code and change timezone in between, for example

Agree. We all understand the behavior when the system time zone changes and AFAIK are not suggesting to change that behavior.

> You have to treat the timeZone carried by these three object the same weight because internally it was resolved by the same routine.

I think this might be where we may not be in alignment yet. Even though the time zone is stored the same way in all three classes, the use cases for each class are different. I'd like to collaborate in tomorrow's meeting to understand those use cases and to figure out how best to make them successful.

justingrant:
> > The fundamental difference here is that ZonedDateTime and TimeZone are clearly defined to carry time zone information. It is part of their name.
> 
> I find the "part of the name" argument very strange. Since the name is ZonedDateTime, not ZoneDateTime. The Zoned part is a modifier with a suffix 'd' and make it an adjustive in English to modify DateTime. It only mean the DateTime is zoned, just like no body willl think Ice Cream carry Ice, Hot Dog carry hot, French Fries carry France, or Italaian Saussage carry Italy.

Another way to approach this would be to put yourself in the shoes of a novice developer who's heard of this new "Temporal" thing but has never used any Temporal types. So they open up Google, or VS Code, or browser devools and type "Temporal." and, like novice developers everywhere, look at what autocomplete shows them. Like this:

image

If a developer is looking for a type that has a date, a time, and a time zone, then there's only one obvious name in the list above. More than 10 years ago, Java named their type [`ZonedDateTime`](https://docs.oracle.com/javase/8/docs/api/java/time/ZonedDateTime.html). We can argue whether they should have dropped the "d" or not. But they didn't, and avoiding needless cross-platform variation is important so we chose the same name for Temporal.ZonedDateTime.

Same for `Temporal.Now.zonedDateTime`.  It's very clear which of those methods return something with a time zone. 

image

Can we PLEASE all agree that the presence of the letters "Zone" in the name (with or without a "d") provide a very clear hint to the user that there's a time zone in there?

justingrant:
As a starting point for discussing use cases in tomorrow's meeting, here's a few ZDT-formatting use cases that I'm familiar with. Feel free to add more that answer the question: why would a userland developer want to format a ZDT?

Here are a few cases that I expect to be used frequently. Feedback welcome!

1. Speeding up code that currently uses `ZonedDateTime.p..toLocaleString()`. I expect this to be the primary use case for formatting ZDT objects, because if perf isn't that important then the developer would probably use `toLocaleString` which is more ergonomic.

2. An app that currently formats a global list of data (e.g. a search engine of airplane flight departures) in the user's time zone wants to also format departures in other time zones (e.g. the time zone of the departing airport) side-by-side with the user's local time. 

3. Refactoring an implementation of (2) in a pre-Temporal codebase (where each data point is represented by a `{date, timeZone}` tuple) to instead use a ZonedDateTime to represent each event.

4. Showing a list of local dates or times where the source data may be spread across multiple time zones and each data point is represented by a ZDT. Example: an advertising server showing the local time of day or local day of week that users clicked on a particular ad, so that the advertiser can decide what time/day is the best time to pay for ads. 

Feel free to add other cases that you've I suspect we may think of some more in the meeting. But maybe we could start with those 4 cases?

justingrant:
Meeting 2023-03-16: We could not reach consensus on the issues described above, so we're going to instead build a backup option that will allow maximum flexibility in the future.

1. `DTP.p.format` and `DTF.p.formatToParts` will throw a TypeError when passed a ZDT instance.
2. If a non-undefined `timeZone` option is passed to `ZDT.p.toLocaleString`, throw a TypeError.
3. Otherwise, `ZDT.p.toLocaleString` will act as if it:
   a. Adds the ZDT's time zone and calendar to the options passed to the `%DateTimeFormat%` constructor.
   b. Sends the ZDT's instant to be formatted using those options
   c. mimics the observable behavior for how the `Intl.DateTimeFormat` constructor accesses options bags.

@sffc's suggestion was that we'd add an optional `timeZone` parameter to the `InitializeDateTimeFormat` AO, and presumably make upstream changes too so that that `timeZone` param can get into `InitializeDateTimeFormat` .

FrankYFTang:
I just want to point out one thing here, not an opposition:
"a. Adds the ZDT's time zone and calendar to the options passed to the %DateTimeFormat% constructor."

If you look at the current spec text (step 4 of HandleDateTimeTemporalZonedDateTime) , the toLocaleString will throw RangeError if the calendar of the ZDT is not "iso8601" so the step you describe above will change that (I am not oppose you to change that but just want to point out that is a change)

FrankYFTang:
Also, probably unrelated to this bug, the current spec also mandate
1)  if the  calendar of T.PlainDateTime or T.PlainDate is not "iso8601", the calling to toLocaleString()  will always throw RangeError (see step 4 of HandleDateTimeTemporalZonedDateTime and step 6 of HandleDateTimeTemporalDate) 
2) unless you call toLocaleString() with an explicit {calendar: zdt.calendar} , the toLocaleString will always throw RangeError for T.PlaneDate, T.PlainDateTime, T.PlainYearMonth, T.PlainMonthDay

so.... I want to point out, I just realize, not only T.ZDT is a new object which carry a TimeZone and request to format by Intl.DTF which Date does not carray a TimeZone. Similarly 

T.PlainDateTime, T.PlainDate, T.PlainYearMonth, T.PlainMonthDay (maybe T.PlainTime too???) T.ZonedDateTime are also a new object which carray a calendar and may conflict (and since it is by default iso8601, it will most likely almost ALWAYS conflict ) with the calendar inside the Intl.DTF.

ptomato:
I've implemented the backup solution in https://github.com/tc39/proposal-temporal/pull/2522

A few things to note:
- formatRange and formatRangeToParts also do not work
- I've kept the calendar behaviour as it was. If we want to change that (along with the other Temporal objects' calendar behaviour as suggested in #2521) then I think it should be a separate change that we discuss, and I'd prefer to have that happen in a follow-up proposal, which should be possible since it would be changing behaviour that currently throws to be permissible.

FrankYFTang:
About TimeZone and ZonedDateTime - I put down a Stage 0 in https://github.com/FrankYFTang/intl-zoneddatetimeformat

ptomato:
Closing this for now, since it is superseded by #2522 which was approved at the March 2023 TC39 meeting.

PR #2489, opened 2023-01-25
Update d.ts file with new rounding modes.

12wrigja:
This was implemented in d450be195a81228d284777969bedd9e0475400bd, but the types were not updated to reflect the newly acceptable rounding modes.


ptomato (approved):
Thanks for catching this!

PR #2494, opened 2023-01-30
Editorial: Remove non-text content from `*` format wrappers

gibson042:
Elements are _technically_ allowed inside ecmarkdown `*` language value format wrappers, but it seems like bad form and slightly confusing (since there is no similar form within the language to express those values). This PR replaces such uses, all of which involve BigInt powers of 10.
```diff
-*10<sup>x</sup>*<sub>ℤ</sub>
+ℤ(10<sup>x</sup>)
```


ptomato (approved):
Thanks!

PR #2496, opened 2023-02-02
Fix broken code in `Plain{Date|Time}.p.toZonedDateTime` docs

justingrant:
There's some typos in code samples of docs of `toZonedDateTime` for both PlainDate and PlainTime.  While I was there, I moved around a sentence for clarity.


ptomato (approved):
Good catch, thanks!

PR #2502, opened 2023-02-15
Prevent auto-link of "safe" in "DST-safe"

justingrant:
Thanks to @bakkot for teaching me about [`<emu-not-ref>`](https://tc39.es/ecmarkup/#emu-not-ref) !


ptomato (approved):
Good catch, thanks!

PR #2505, opened 2023-02-18
Normative: Prevent mergeFields from reordering properties

### LINKED ISSUE #2504: mergeFields should not reorder input properties
This appears to have been unintentionally omitted from #2474: https://github.com/tc39/proposal-temporal/pull/2474#discussion_r1111056071
> For a simple example, consider how this algorithm requires `Reflect.ownKeys(Temporal.Calendar.from("iso8601").mergeFields({ year: 2023, month: 2, day: 18 }, { month: 3 }))` to return `["year", "day", "month"]` rather than `["year", "month", "day"]`.

gibson042:
Fixes #2504

test262 PR: https://github.com/tc39/test262/pull/3789

### DIFF
--- a/spec/calendar.html
+++ b/spec/calendar.html
@@ -1394,7 +1394,15 @@
         1. Assert: _calendar_.[[Identifier]] is *"iso8601"*.
         1. Let _overriddenKeys_ be ISOFieldKeysToIgnore(_additionalKeys_).
         1. Let _merged_ be OrdinaryObjectCreate(*null*).
-        1. Perform ! CopyDataProperties(_merged_, _fieldsCopy_, _overriddenKeys_, « *undefined* »).
+        1. NOTE: The following steps ensure that property iteration order of _merged_ matches that of _fields_ as modified by omitting overridden properties and appending non-overlapping properties from _additionalFields_ in iteration order.
+        1. Let _fieldsKeys_ be ! _fieldsCopy_.[[OwnPropertyKeys]]().
+        1. For each element _key_ of _fieldsKeys_, do
+          1. Let _propValue_ be *undefined*.
+          1. If _overriddenKeys_ contains _key_, then
+            1. Set _propValue_ to ! Get(_additionalFieldsCopy_, _key_).
+          1. Else,
+            1. Set _propValue_ to ! Get(_fieldsCopy_, _key_).
+          1. If _propValue_ is not *undefined*, perform ! CreateDataPropertyOrThrow(_merged_, _key_, _propValue_).
         1. Perform ! CopyDataProperties(_merged_, _additionalFieldsCopy_, « »).
         1. Return _merged_.
       </emu-alg>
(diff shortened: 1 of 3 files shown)


spec/calendar.html
```diff
@@ -1366,12 +1366,20 @@ <h1>Temporal.Calendar.prototype.mergeFields ( _fields_, _additionalFields_ )</h1
         1. Set _additionalFields_ to ? ToObject(_additionalFields_).
         1. Let _additionalFieldsCopy_ be OrdinaryObjectCreate(*null*).
         1. Perform ? CopyDataProperties(_additionalFieldsCopy_, _additionalFields_, « », « *undefined* »).
-        1. NOTE: Every property of _additionalFieldsCopy_ is a data property with non-*undefined* value, but some property keys may be Symbols.
+        1. NOTE: Every property of _fieldsCopy_ and _additionalFieldsCopy_ is an enumerable data property with non-*undefined* value, but some property keys may be Symbols.
```
ljharb, line 1369:
```suggestion
        1. NOTE: Every property of _fieldsCopy_ and _additionalFieldsCopy_ is an enumerable data property with a non-*undefined* value, but some property keys may be Symbols.
```

spec/calendar.html
```diff
@@ -1366,12 +1366,20 @@ <h1>Temporal.Calendar.prototype.mergeFields ( _fields_, _additionalFields_ )</h1
...
         1. Let _additionalKeys_ be ! _additionalFieldsCopy_.[[OwnPropertyKeys]]().
         1. Assert: _calendar_.[[Identifier]] is *"iso8601"*.
-        1. Let _ignoredKeys_ be ISOFieldKeysToIgnore(_additionalKeys_).
+        1. Let _overriddenKeys_ be ISOFieldKeysToIgnore(_additionalKeys_).
         1. Let _merged_ be OrdinaryObjectCreate(*null*).
-        1. Perform ! CopyDataProperties(_merged_, _fieldsCopy_, _ignoredKeys_, « *undefined* »).
+        1. NOTE: The following steps ensure that property iteration order of _merged_ matches that of _fields_ as modified by omitting overridden properties and appending non-overlapping properties from _additionalFields_ in iteration order.
```
ljharb, line 1374: is there a reason not to upstream this improvement into CopyDataProperties itself?

gibson042: This only comes up when multiple objects are being merged, whereas [CopyDataProperties](https://tc39.es/ecma262/multipage/abstract-operations.html#sec-copydataproperties) is concerned with a _single_ source. Are you aware of any other need for such a generalization?

ljharb: ah, i suppose not - Object.assign merges multiple objects but I'm not sure if the semantics match.

Either way tho, you're using this in two places in this PR - perhaps it should still be extracted to an AO?

gibson042: It's not actually two places, sup-temporal.calendar.prototype.mergefields overrides sec-temporal.calendar.prototype.mergefields and all such overrides will be going away after the current PRs settle and we non-ISO-8601 calendar operations move from ECMA-402 to ECMA-262.

ptomato (approved):
Ugh, I guess I merged #2474 too early - I thought this was sufficiently settled. In any case I hope we can consider it part of that change and don't have to take it to plenary again.

PR #2506, opened 2023-02-18
Editorial: Improve alignment between ISO 8601 vs. other-calendar operations

gibson042:
Extracted from #2500

### DIFF
--- a/spec/calendar.html
+++ b/spec/calendar.html
@@ -741,14 +741,13 @@
     </emu-clause>
 
     <emu-clause id="sec-temporal-isodatefromfields" aoid="ISODateFromFields">
-      <h1>ISODateFromFields ( _fields_, _options_ )</h1>
+      <h1>ISODateFromFields ( _fields_, _overflow_ )</h1>
       <p>
         The ISODateFromFields abstract operation implements the calendar-specific logic in the `Temporal.Calendar.prototype.dateFromFields` method for the ISO 8601 calendar.
       </p>
       <emu-alg>
         1. Assert: Type(_fields_) is Object.
-        1. Set _fields_ to ? PrepareTemporalFields(_fields_, « *"day"*, *"month"*, *"monthCode"*, *"year"* », « *"year"*, *"day"* »).
-        1. Let _overflow_ be ? ToTemporalOverflow(_options_).
+        1. Assert: _overflow_ is either *"constrain"* or *"reject"*.
         1. Let _year_ be ! Get(_fields_, *"year"*).
         1. Assert: Type(_year_) is Number.
         1. Let _month_ be ? ResolveISOMonth(_fields_).
@@ -759,15 +758,14 @@
     </emu-clause>
 
     <emu-clause id="sec-temporal-isoyearmonthfromfields" aoid="ISOYearMonthFromFields">
-      <h1>ISOYearMonthFromFields ( _fields_, _options_ )</h1>
+      <h1>ISOYearMonthFromFields ( _fields_, _overflow_ )</h1>
       <p>
         The ISOYearMonthFromFields abstract operation implements the calendar-specific logic in the `Temporal.Calendar.prototype.yearMonthFromFields` method for the ISO 8601 calendar.
         It returns a Record with three fields ([[Year]], [[Month]], and [[ReferenceISODay]]). All are integers.
       </p>
       <emu-alg>
         1. Assert: Type(_fields_) is Object.
-        1. Set _fields_ to ? PrepareTemporalFields(_fields_, « *"month"*, *"monthCode"*, *"year"* », « *"year"* »).
-        1. Let _overflow_ be ? ToTemporalOverflow(_options_).
+        1. Assert: _overflow_ is either *"constrain"* or *"reject"*.
         1. Let _year_ be ! Get(_fields_, *"year"*).
         1. Assert: Type(_year_) is Number.
         1. Let _month_ be ? ResolveISOMonth(_fields_).
@@ -781,14 +779,13 @@
     </emu-clause>
 
     <emu-clause id="sec-temporal-isomonthdayfromfields" aoid="ISOMonthDayFromFields">
-      <h1>ISOMonthDayFromFields ( _fields_, _options_ )</h1>
+      <h1>ISOMonthDayFromFields ( _fields_, _overflow_ )</h1>
       <p>
         The ISOMonthDayFromFields abstract operation implements the calendar-specific logic in the `Temporal.Calendar.prototype.monthDayFromFields` method for the ISO 8601 calendar.
       </p>
       <emu-alg>
         1. Assert: Type(_fields_) is Object.
-        1. Set _fields_ to ? PrepareTemporalFields(_fields_, « *"day"*, *"month"*, *"monthCode"*, *"year"* », « *"day"* »).
-        1. Let _overflow_ be ? ToTemporalOverflow(_options_).
+        1. Assert: _overflow_ is either *"constrain"* or *"reject"*.
         1. Let _month_ be ! Get(_fields_, *"month"*).
         1. Let _monthCode_ be ! Get(_fields_, *"monthCode"*).
         1. Let _year_ be ! Get(_fields_, *"year"*).
@@ -992,7 +989,9 @@
         1. Assert: _calendar_.[[Identifier]] is *"iso8601"*.
         1. If Type(_fields_) is not Object, throw a *TypeError* exception.
         1. Set _options_ to ? GetOptionsObject(_options_).
-        1. Let _result_ be ? ISODateFromFields(_fields_, _options_).
+        1. Set _fields_ to ? PrepareTemporalFields(_fields_, « *"day"*, *"month"*, *"monthCode"*, *"year"* », « *"year"*, *"day"* »).
+        1. Let _overflow_ be ? ToTemporalOverflow(_options_).
+        1. Let _result_ be ? ISODateFromFields(_fields_, _overflow_).
         1. Return ? CreateTemporalDate(_result_.[[Year]], _result_.[[Month]], _result_.[[Day]], _calendar_).
       </emu-alg>
     </emu-clause>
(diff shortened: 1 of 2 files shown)


spec/calendar.html
```diff
@@ -739,15 +758,14 @@ <h1>ISODateFromFields ( _fields_, _options_ )</h1>
     </emu-clause>
 
     <emu-clause id="sec-temporal-isoyearmonthfromfields" aoid="ISOYearMonthFromFields">
-      <h1>ISOYearMonthFromFields ( _fields_, _options_ )</h1>
+      <h1>ISOYearMonthFromFields ( _fields_, _overflow_ )</h1>
       <p>
         The ISOYearMonthFromFields abstract operation implements the calendar-specific logic in the `Temporal.Calendar.prototype.yearMonthFromFields` method for the ISO 8601 calendar.
         It returns a Record with three fields ([[Year]], [[Month]], and [[ReferenceISODay]]). All are integers.
```
justingrant, line 764: Should we be more specific about sign here?

```suggestion
        It returns a Record with three fields ([[Year]], [[Month]], and [[ReferenceISODay]]). [[Month]], and [[ReferenceISODay]] are positive integers, while [[Year]] is a signed integer.
```

spec/intl.html
```diff
@@ -1791,6 +1772,30 @@ <h1>
             In a _calendar_ such as *"japanese"* where eras do not start and end at year and/or month boundaries, note that the returned List should contain *"era"* and *"eraYear"* if _keys_ contains *"day"*, *"month"*, or *"monthCode"* (not only if it contains *"era"*, *"eraYear"*, or *"year"*, as in the example above) because it's possible for changing the day or month to cause a conflict with the era.
```
justingrant, line 1772: Another minor clarification that may be helpful while we're in the neighborhood:

```suggestion
            In a _calendar_ such as *"japanese"* where eras do not start and end at year and/or month boundaries, note that the returned List should contain *"era"* and *"eraYear"* if _keys_ contains *"day"*, *"month"*, or *"monthCode"* (not only if it contains *"era"*, *"eraYear"*, or *"year"*, as in the example above) because Japanese calendar eras can start in the middle of a year, which makes it possible that changing the day or month can cause a conflict with the era.
```

justingrant: Actually, ignore that suggestion: I skipped over the statement at the top of the paragraph that talked about eras starting midyear.

spec/intl.html
```diff
@@ -1791,6 +1772,30 @@ <h1>
...
+            CalendarResolveFields (
+              _calendar_: a String,
+              _fields_: an Object,
+              _type_: ~date~, ~year-month~, or ~month-day~,
+            ): either a normal completion containing ~unused~ or a throw completion
+          </h1>
+          <dl class="header">
+            <dt>description</dt>
+            <dd>
+              It performs implementation-defined processing to validate that _fields_ (which describes a date in the built-in calendar identified by _calendar_) is sufficiently complete to satisfy _type_ and not internally inconsistent, and merges data that can be represented in multiple forms into standard properties of _fields_ (for example, merging *"month"* and *"monthCode"* into *"month"* and merging *"era"* and *"eraYear"* into *"year"*).
```
justingrant, line 1787: > which describes a date

To me, "date" implies all YMD components, but can't the incoming object include only a part of a date, e.g. YM or MD? Is there another term that could be used here? "Date or part of a date" ?

> standard properties of _fields_

Is it worth defining "standard properties" precisely here?  

> merges data

Does this mean that it mutates `_fields_`?  Or something else?

spec/intl.html
```diff
@@ -1791,6 +1772,30 @@ <h1>
...
+            </dd>
+          </dl>
+          <p>
+            The operation throws a *TypeError* exception if the properties of _fields_ are internally inconsistent within the calendar or insufficient to identify a unique instance of _type_ in the calendar. For example:
+          </p>
+          <ul>
+            <li>If _type_ is ~date~ or ~month-day~ and *"day"* has the usual interpretation in the calendar and the corresponding value is *undefined*.</li>
```
justingrant, line 1794: > has the usual interpretation in the calendar

I found this phrase a bit confusing. Should we either clarify or somewhere else define what "usual interpretation in the calendar" means?

spec/intl.html
gibson042: > > which describes a date
> 
> To me, "date" implies all YMD components, but can't the incoming object include only a part of a date, e.g. YM or MD? Is there another term that could be used here? "Date or part of a date" ?

Done.

> > standard properties of _fields_
> 
> Is it worth defining "standard properties" precisely here?

I don't think so, because it's calendar-specific and not observable by ECMAScript code (output from CalendarResolveFields is always consumed completely by CalendarDateToISO or CalendarMonthDayToISOReferenceDate, both of which are also implementation-defined).

> > merges data
> 
> Does this mean that it mutates `_fields_`? Or something else?

Yes, it mutates `_fields_`. I've update the text to hopefully clarify that.

spec/intl.html
gibson042: Updated.

spec/intl.html
```diff
@@ -1716,11 +1716,11 @@ <h1>
             CalendarDateFields (
               _calendar_: a String,
               _fields_: a List of Strings,
-            )
+            ): a List of Strings
           </h1>
           <dl class="header">
             <dt>description</dt>
-            <dd>It takes a list of standard fields in _fields_ that are necessary for a given operation and returns a new list by adding relevant calendar-specific fields for the calendar represented by _calendar_. This is relevant for calendars which accept fields other than the standard set of built-in calendar fields.</dd>
+            <dd>It interprets _fields_ as the names of fields (<emu-xref href="#table-temporal-field-requirements"></emu-xref>) necessary for a given operation and returns a new List by appending the names of additional calendar-specific fields that are relevant for the built-in calendar identified by _calendar_. For example, *"era"* and *"eraYear"* are appended when _calendar_ is *"gregory"* or *"japanese"*.</dd>
```
ptomato, line 1704: This should mention that the appended fields must also be relevant to the supplied fields - for example, `era` and `eraYear` are only appended when _fields_ contains `year`. Otherwise, we'd unnecessarily try to read the `era` and `eraYear` properties of PlainMonthDay in era-calendars. (In the original text, I think the word "relevant" was intended to express this, but was ambiguous.)

spec/intl.html
```diff
@@ -1411,24 +1411,22 @@ <h1>
...
           </p>
           <ul>
-            <li>*"year"*, *"monthCode"*, *"day"*</li>
-            <li>*"year"*, *"month"*, *"day"*</li>
-            <li>*"era"*, *"eraYear"*, *"monthCode"*, *"day"*</li>
-            <li>*"era"*, *"eraYear"*, *"month"*, *"day"*</li>
+            <li>If *"day"* has the usual interpretation in the calendar and the corresponding value is *undefined*.</li>
+            <li>If *"month"* and *"monthCode"* have the usual interpretation in the calendar and either the corresponding values for both are *undefined* or neither value is *undefined* but they do not identify the same month.</li>
+            <li>If the calendar supports the usual partitioning of years into eras with their own year counting (as in the Gregorian or traditional Japanese calendars) as represented by *"year"*, *"era"*, and *"eraYear"* and either the corresponding values for all three are *undefined*, or the value for one but not both of *"era"* and *"eraYear"* is *undefined*, or none of the three values are *undefined* but the values for *"era"* and *"eraYear"* do not together identify the same year as the value for *"year"*.</li>
```
ptomato: I found the bulleted list in the previous version a bit easier to understand than this prose. But I'm not sure how to edit it to make it clearer. Maybe something like:

"...and any of the following cases:
  - none of "year", "era", "eraYear" are defined;
  - "year" is undefined and only one of "era" or "eraYear" is defined;
  - all three are defined, but the values for "era" and "eraYear" do not together identify the same year as the value for "year"."

spec/calendar.html
```diff
@@ -107,6 +107,7 @@ <h1>CalendarFields ( _calendar_, _fieldNames_ )</h1>
         The CalendarFields abstract operation transforms a List of String values _fieldNames_ into another List of String values by calling the `fields` method of the given _calendar_ Object.
       </p>
       <emu-alg>
+        1. NOTE: If _calendar_ is a built-in object, then the following step will usually invoke <emu-xref href="#sec-temporal.calendar.prototype.fields"></emu-xref> or <emu-xref href="#sup-temporal.calendar.prototype.fields"></emu-xref>.
```
ptomato, line 110: I'm not opposed to adding these notes, but they are contradicted by https://github.com/tc39/proposal-temporal/pull/2485 (which could use a code review so that we can merge it :smile:) so maybe we should omit this to minimize churn?

Specifically, commit https://github.com/tc39/proposal-temporal/pull/2485/commits/6c2ce352c0476852a3496208fc3694a9721da31b#diff-1e3b606bd8b3ef5c33eaf7b43afdfc240b0a926c8d0a66e02bfc9e1a7e8de8a2 adds a step here that explicitly invokes %Temporal.Calendar.prototype.fields% if _calendar_ is a string. The current Invoke step may still invoke %Temporal.Calendar.prototype.fields% if that's the method that's found by prototype lookup (e.g., if the custom calendar extends `Temporal.Calendar` and doesn't override `fields`) but I'd no longer call it "usual".

Same thing goes for the other notes in this commit.

ptomato (approved):
I think the prose for CalendarFields needs one correction, the rest are just optional suggestions for improvements. Thanks for working on this, I think it's an improvement!

spec/intl.html
gibson042: Done.

spec/intl.html
gibson042: Done.

spec/calendar.html
gibson042: Aside from the obvious benefit, this also supports convenient linking in the rendered spec. I'll remove/replace these notes after #2485 lands (and review it soon).

ptomato (approved):
Looks like everyone's review comments have been addressed. I'd like to merge this so we can keep other things moving along.

As for the notes that will need revision in #2485, I'll plan to rebase that on top of this when I merge it, and just remove the notes for now; if you have a suggestion for how they should be replaced instead, let me know over on that PR when you review it.

PR #2507, opened 2023-02-20
Editorial: Remove redundant redefinition of "ASCII-case-insensitive"

gibson042:
The definition is already being added to ECMA-262 by https://tc39.es/proposal-temporal/#sec-ecmascript-language-types-string-type .


ptomato (approved):
Thanks!

Issue #1985, opened 2021-12-17
Simplify FractionalPart grammar productions


ptomato:
Follow up from https://github.com/tc39/proposal-temporal/pull/1957#discussion_r766246004

The FractionalPart productions should be removed in favour of using the Fraction productions and chopping the decimal separator off the front of the string. This should be an editorial change.

PR #2511, opened 2023-03-01
Editorial: Replace FractionalPart by its definition and develop the d…

guijemont:
…efinition

After changes such as #1957, the FractionalPart production became redundant, and it was a suggestion left over from that change to remove it.

Closes: #1985


ptomato (approved):
I agree with Richard, but for now writing it all out explicitly is what the editors asked for in TimeZoneUTCOffsetString, so...

PR #2514, opened 2023-03-03
Docs: Fix incorrect explanations and examples about ZonedDateTime

gibson042:


docs/zoneddatetime.md
```diff
@@ -224,13 +224,12 @@ zdt = Temporal.ZonedDateTime.from({ timeZone: 'Europe/Paris', year: 2001, month:
...
 
-This function can be used to sort arrays of `Temporal.ZonedDateTime` objects.
-
-Comparison will use exact time, not clock time, because sorting is almost always based on when events happened in the real world.
-Note that during the hour before and after DST ends, sorting of clock time may not match the order the events actually occurred.
+Comparison uses exact time, not calendar date and clock time, because sorting is almost always based on when events happen in the real world (and note that sorting by clock time may not match the order of actual occurrence near discontinuities such as DST transitions).
 
-Note that this function will return `0` if the two objects have different `calendar` or `timeZone` properties, as long as their exact timestamps are equal.
+Calendars and time zones are also ignored in the comparison for the same reason.
+For example, this method returns `0` for instances that fall on the same day and time in the ISO 8601 calendar, even if their fields do not match due to use of different calendars and/or time zones.
```
justingrant, line 230:
```suggestion
For example, this method returns `0` for instances that fall on the same date and time in the UTC time zone and ISO 8601 calendar, even if fields like `minute` or `day` do not match due to use of different time zones and/or calendars.
```

docs/zoneddatetime.md
```diff
@@ -1224,7 +1223,7 @@ Compares two `Temporal.ZonedDateTime` objects for equality.
 This function exists because it's not possible to compare using `zonedDateTime == other` or `zonedDateTime === other`, due to ambiguity in the primitive representation and between Temporal types.
 
 If you don't need to know the order in which two events occur, then this function is easier to use than `Temporal.ZonedDateTime.compare`.
-But both methods do the same thing, so a `0` returned from `compare` implies a `true` result from `equals`, and vice-versa.
+However, there are subtle differences between the two methods—a `true` result from `equals` includes comparison of calendar and time zone, and is therefore stronger than a `0` result from compare (which ignores calendar and time zone).
```
justingrant, line 1226: Good catch. The old docs were how it used to work before we were required to switch to the current behavior as a condition of getting Stage 3!

justingrant (approved):
Looks good, only one minor suggestion. Thanks!

PR #2520, opened 2023-03-15
Introduce a host hook for the current time.

Ms2ger:
Fixes #1654.

### DIFF
--- a/spec/mainadditions.html
+++ b/spec/mainadditions.html
@@ -186,6 +186,77 @@
     </emu-note>
   </emu-clause>
 
+  <emu-clause id="sec-temporal-date-constructor">
+    <h1><a href="https://tc39.es/ecma262/#sec-date-constructor">The Date Constructor</a></h1>
+    <emu-clause id="sec-temporal-date">
+      <h1><a href="https://tc39.es/ecma262/#sec-date">Date ( ..._values_ )</a></h1>
+      <p>This function performs the following steps when called:</p>
+      <emu-alg>
+        1. If NewTarget is *undefined*, then
+          1. <del>Let _now_ be the time value (UTC) identifying the current time.</del>
+          1. <ins>Let _global_ be GetGlobalObject().</ins>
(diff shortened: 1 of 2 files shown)


justingrant (approved):

ljharb (approved):
is SystemUTCEpochNanoseconds already wired into Date() somehow, so they're always the same value?

littledan (approved):

littledan:
Because there's always JS on the stack when this algorithm is called, I guess the HTML implementation of the hook can say, "return the current high resolution time, given the current global object, asserting that it is within bounds".

Ms2ger:
> is SystemUTCEpochNanoseconds already wired into Date() somehow, so they're always the same value?

No, not yet, because the resolution is already low enough. We could do that, I guess.

> Because there's always JS on the stack when this algorithm is called, I guess the HTML implementation of the hook can say, "return the current high resolution time, given the current global object, asserting that it is within bounds".

Oh, good question. Maybe it'd be better to pass a global object explicitly, to reduce the action at a distance. WDYT, @annevk?

annevk (commented):
Yeah we need a global. (Using current global object on the HTML side seems okay though.)

(In particular we'd use https://w3c.github.io/hr-time/#dfn-current-high-resolution-time as underlying concept and that wants a global to know whether or not to clamp.)

It would be even simpler if we could just give you nanoseconds and not have to wrap them in a `BigInt` or whatever ℤ is.

annevk:
Is the idea to refactor https://tc39.es/ecma262/#sec-time-values-and-time-range on top of this hook? That seems workable, just want to make sure we don't have to revisit this anyone soon.

ptomato (approved):
I had to think hard about whether floor division was the correct operation here when converting from nanoseconds to milliseconds, but I believe it is: the millisecond number should only change when the exact millisecond ticks over. For example, at -999999999 ns we want -999 ms, and at 999999 ns we want 0 ms.

PR #2553, opened 2023-04-14
Fixes to builtins-as-strings change

ptomato:
These are all regressions or bugs introduced by #2482, spotted by @anba, eagle-eyed as always.

### DIFF
--- a/spec/abstractops.html
+++ b/spec/abstractops.html
@@ -501,8 +501,6 @@
       1. Else,
         1. Let _string_ be ? ToString(_value_).
         1. Let _result_ be ? ParseTemporalRelativeToString(_string_).
-        1. Let _calendar_ be _result_.[[Calendar]].
-        1. If _calendar_ is *undefined*, set _calendar_ to *"iso8601"*.
         1. Let _offsetString_ be _result_.[[TimeZone]].[[OffsetString]].
         1. Let _timeZoneName_ be _result_.[[TimeZone]].[[Name]].
         1. If _timeZoneName_ is *undefined*, then
@@ -514,6 +512,10 @@
           1. Else if _offsetString_ is *undefined*, then
             1. Set _offsetBehaviour_ to ~wall~.
           1. Set _matchBehaviour_ to ~match minutes~.
+        1. Let _calendar_ be _result_.[[Calendar]].
+        1. If _calendar_ is *undefined*, set _calendar_ to *"iso8601"*.
+        1. If IsBuiltinCalendar(_calendar_) is *false*, throw a *RangeError* exception.
+        1. Set _calendar_ to the ASCII-lowercase of _calendar_.
       1. If _timeZone_ is *undefined*, then
         1. Return ? CreateTemporalDate(_result_.[[Year]], _result_.[[Month]], _result_.[[Day]], _calendar_).
       1. If _offsetBehaviour_ is ~option~, then
--- a/spec/calendar.html
+++ b/spec/calendar.html
@@ -702,13 +702,17 @@
       </emu-alg>
     </emu-clause>
 
-    <emu-clause id="sec-temporal-calendarequals" aoid="CalendarEquals">
-      <h1>CalendarEquals ( _one_, _two_ )</h1>
-      <p>
-        The abstract operation CalendarEquals takes two arguments _one_ and _two_, which must be Objects.
-        It returns *true* if its arguments represent the same calendar.
-        It performs the following steps:
-      </p>
+    <emu-clause id="sec-temporal-calendarequals" type="abstract operation">
+      <h1>
+        CalendarEquals (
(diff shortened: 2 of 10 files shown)


justingrant (approved):
Great!

spec/calendar.html
```diff
@@ -700,13 +701,17 @@ <h1>
...
+    <emu-clause id="sec-temporal-calendarequals" type="abstract operation">
+      <h1>
+        CalendarEquals (
+          _one_: a String or Object,
+          _two_: a String or Object,
+        ): either a normal completion containing either *true* or *false*, or a throw completion
+      </h1>
+      <dl class="header">
+        <dt>description</dt>
+        <dd>It returns *true* if its arguments represent the same calendar.</dd>
```
gibson042, line 713:
```suggestion
        <dd>It returns *true* if its arguments represent calendars using the same identifier.</dd>
```

spec/timezone.html
```diff
@@ -789,13 +789,17 @@ <h1>GetPossibleInstantsFor ( _timeZone_, _dateTime_ )</h1>
...
+    <emu-clause id="sec-temporal-timezoneequals" type="abstract operation">
+      <h1>
+        TimeZoneEquals (
+          _one_: a String or Object,
+          _two_: a String or Object,
+        ): either a normal completion containing either *true* or *false*, or a throw completion
+      </h1>
+      <dl class="header">
+        <dt>description</dt>
+        <dd>It returns *true* if its arguments represent the same time zone.</dd>
```
gibson042, line 801:
```suggestion
        <dd>It returns *true* if its arguments represent time zones using the same identifier.</dd>
```

spec/calendar.html
```diff
@@ -716,13 +721,17 @@ <h1>CalendarEquals ( _one_, _two_ )</h1>
...
+          _two_: a String or Object,
+        ): either a normal completion containing either a String or Object, or a throw completion
+      </h1>
+      <dl class="header">
+        <dt>description</dt>
+        <dd>It returns the calendar that should take priority when combining two Temporal objects with different calendars, or throws an exception if the calendars cannot be combined.</dd>
+      </dl>
       <emu-alg>
         1. If _one_ and _two_ are the same Object value, return _two_.
         1. Let _calendarOne_ be ? ToTemporalCalendarIdentifier(_one_).
```
gibson042, line 737: Passing thought: I wonder if redundancy with CalendarEquals should be reduced? Probably not, since I think it would require some contortions like parameterizing "iso8601" here (e.g., `If Completion(ConsolidateCalendars(_one_, _two_)).[[Type]] is ~normal~, return *true*. Otherwise, return *false*.` in CalendarEquals vs. `Let _calendar_ be ? ConsolidateCalendars(_dateTime_.[[Calendar]], _plainDate_.[[Calendar]], *"iso8601"*).` in Temporal.PlainDateTime.prototype.withPlainDate).

gibson042 (approved):

spec/duration.html
```diff
@@ -1455,7 +1474,10 @@ <h1>
             1. Set _oneMonthDays_ to _moveResult_.[[Days]].
         1. Else,
           1. Assert: _largestUnit_ is *"week"*.
-          1. Let _dateAdd_ be ? GetMethod(_calendar_, *"dateAdd"*).
+          1. If _calendar_ is an Object, then
+            1. Let _dateAdd_ be ? GetMethod(_calendar_, *"dateAdd"*).
+          1. Else,
+            1. Let _dateAdd_ be ~unused~.
```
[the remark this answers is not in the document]
ptomato: This will be completely different in #2519 anyway, and I already have a note to myself on that topic: https://github.com/tc39/proposal-temporal/pull/2519#discussion_r1166146921
I'd prefer not to introduce an operation here, but instead focus on #2519.

spec/calendar.html
ptomato: I'm skeptical about that. For one thing, checking the completion type would mean that CalendarEquals would return false if ToTemporalCalendarIdentifier throws, instead of throwing as it currently does.

spec/abstractops.html
```diff
@@ -512,6 +510,10 @@ <h1>ToRelativeTemporalObject ( _options_ )</h1>
           1. Else if _offsetString_ is *undefined*, then
             1. Set _offsetBehaviour_ to ~wall~.
           1. Set _matchBehaviour_ to ~match minutes~.
+        1. Let _calendar_ be _result_.[[Calendar]].
+        1. If _calendar_ is *undefined*, set _calendar_ to *"iso8601"*.
+        1. If IsBuiltinCalendar(_calendar_) is *false*, throw a *RangeError* exception.
```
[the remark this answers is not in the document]
ptomato: Yes, my thought as well.

PR #2567, opened 2023-04-28
Editorial: Remove repeated words

ptomato:
A couple of modifications to PRs that have been merged recently.

_Edit:_ Since no one has had a chance to look at this yet, I've tacked a few more commits on, which are small fixes done while I was researching #2563.


justingrant (approved):
LGTM

PR #2597, opened 2023-06-08
Editorial: Move ToObject from SnapshotOwnProperties to Calendar.p.mergeFields

gibson042:
Other call sites always provide an Object, so this isolates the inconsistency.


justingrant (approved):
Yep this is clearer.

ptomato (approved):
Thanks!

PR #2599, opened 2023-06-10
Polyfill: Support Node 20 and work around ICU 73.1 bug

justingrant:
This PR adds support for running tests in Node 20 and adds a workaround for an [ICU 73.1 regression](https://unicode-org.atlassian.net/browse/ICU-22412) in the iso8601 calendar that breaks some tests in Node 20.

The fix was to stop using Intl.DateTimeFormat for Gregorian-based calendars like `roc` and `japanese`. As a side effect, a few tests that previously failed now work correctly, and tests run a little faster. 

This PR depends on https://github.com/tc39/test262/pull/3848, so don't please don't merge this PR until I have a chance to rebase this PR's submodule to the latest test262 after the test262 PR is merged.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -10,10 +10,10 @@ jobs:
       - uses: actions/checkout@v3
         with:
           persist-credentials: false
-      - name: use node.js v19.x
+      - name: use node.js v20.x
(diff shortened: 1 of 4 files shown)


ptomato (approved):
Thanks!

PR #2600, opened 2023-06-11
Editorial: Inline single-use grammar nonterminals

gibson042:


spec/abstractops.html
```diff
@@ -1071,11 +1061,9 @@ <h1>ISO 8601 grammar</h1>
       DateSpecYearMonth :
           DateYear `-`? DateMonth
 
-      TwoDashes :
-          `--`
-
       DateSpecMonthDay :
-          TwoDashes? DateMonth `-`? DateDay
+          `--` DateMonth `-`? DateDay
+          DateMonth `-`? DateDay
```
ptomato, line 1066: Note https://github.com/tc39/proposal-temporal/issues/1800; but I think this is still in line with that.

ptomato (approved):

PR #2606, opened 2023-06-12
Editorial: Introduce an SDO for checking legacy time zone names

gibson042:
I think I finally found a way to clean up the complexity of [|TimeZoneIANAName|](https://tc39.es/proposal-temporal/#prod-TimeZoneIANAName) needing to support ASCII-case-insensitive matches of legacy time zone names. This PR introduces an IsLegacyIANATimeZoneName syntax-directed operation that checks against both the static list and the Etc/GMT±\<UnpaddedHour> pattern, and also provides an explanatory note.

### DIFF
--- a/spec/abstractops.html
+++ b/spec/abstractops.html
@@ -977,9 +977,6 @@
           ASCIISign
           U+2212
 
-      UnpaddedHour :
-          Hour[~Padded]
-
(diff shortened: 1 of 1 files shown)


ptomato (approved):
Great find, this is much clearer. Thanks!

PR #2626, opened 2023-07-11
Typo in README's reference to `Temporal.Instant.timeZoneId()` function

itsknob:
It was referencing `timeZone` and not `timeZoneId`. When using the Documentation Playground, this readme is the page that loads first. It may be confusing to have an incorrect function name as one of the first things people might try in the Developer Console.


ptomato (approved):
Thank you! In fact my colleague @ryzokuken just pointed this out to me the other day, but I didn't get a chance to fix it yet. Much appreciated.

PR #2627, opened 2023-07-12
Editorial: Prefer returning static values after alias comparison

gibson042:
Better aligns with ECMA-262: https://github.com/tc39/ecma262/pull/3122


ptomato (approved):
Seems like a good improvement in clarity, thanks!

PR #2643, opened 2023-08-07
Fix cookbook examples

ptomato:
These examples are tiny web apps and don't have any assertions in them, so they are not run by test-cookbook. That's the reason nobody caught this earlier.


littledan (approved):
LGTM. As a follow-on, does someone want to add assertions?

Ms2ger (approved):

PR #2644, opened 2023-08-10
Editorial: Structured headers, AO for formatting UTC offset with nanoseconds

ptomato:
I know we just inlined FormatTimeZoneOffsetString into GetOffsetStringFor, but due to rebasing #2519 I'm having to un-inline it — it'll be used in more than one place when we pre-calculate the UTC offset nanoseconds in cases where it was calculated twice.

Also includes some commits adding more structured headers that I had queued up but never submitted.


justingrant (approved):
> we just inlined FormatTimeZoneOffsetString into GetOffsetStringFor, but due to rebasing #2519 I'm having to un-inline it

Haha, easy come, easy go!  Looks good.

PR #2646, opened 2023-08-14
Polyfill: Remove throw in ParseTemporalInstant

justingrant:
Found by codecov: ParseTemporalInstantString already throws if neither `z` nor `offset` are present, so there's no need to check again after ParseTemporalInstantString returns.


ptomato (approved):
:+1:

PR #2650, opened 2023-08-16
Polyfill: Be more robust against late-run primordial manipulation

gibson042:
Excludes fixes that are already included in #2500.


ptomato (approved):
Thanks!

PR #2652, opened 2023-08-17
Editorial: Improve the explanation of ISO 8601 grammar deviations

gibson042:
Fixes #2601

### DIFF
--- a/spec/abstractops.html
+++ b/spec/abstractops.html
@@ -978,22 +978,22 @@
       <li>Fractional parts may have 1 through 9 decimal places.</li>
       <li>In time representations, only seconds are allowed to have a fractional part.</li>
       <li>In duration representations, only hours, minutes, and seconds are allowed to have a fractional part.</li>
-      <li>Any number of conforming <a href="https://datatracker.ietf.org/doc/html/draft-ietf-sedate-datetime-extended#section-3.1">suffixes in square brackets</a> are allowed.</li>
       <li>
-        Time zone and <a href="https://tools.ietf.org/html/bcp47#section-2.1">BCP 47 calendar</a> suffixes are the only recognized ones.
-        Others are ignored, unless they are marked with a *!*, in which case they are rejected.
+        Any number of conforming <a href="https://datatracker.ietf.org/doc/html/draft-ietf-sedate-datetime-extended#section-3.1">suffixes in square brackets</a> are allowed.
+        However, the only recognized suffixes are time zone and <a href="https://tools.ietf.org/html/bcp47#section-2.1">BCP 47 calendar</a>.
+        Others are ignored, unless they are prefixed with `!`, in which case they are rejected.
       </li>
       <li>A space may be used to separate the date and time in a combined date / time representation, but not in a duration.</li>
       <li>Alphabetic designators may be in lower or upper case.</li>
       <li>Period or comma may be used as the decimal separator.</li>
-      <li>A time zone offset of *"-00:00"* is allowed, and means the same thing as *"+00:00"*.</li>
+      <li>UTC offsets of *"-00:00"* and *"-0000"* and *"-00"* are allowed, and all mean the same thing as *"+00:00"*.</li>
       <li>UTC offsets may have seconds and up to 9 sub-second fractional digits.</li>
       <li>The constituent date, time, and UTC offset parts of a combined representation may each independently use basic format (with no separator symbols) or extended format (with mandatory `-` or `:` separators), as long as each such part is itself in either basic format or extended format (e.g., *"1970-01-01T012345"* and *"19700101T01:23:45"* are valid but *"1970-0101T012345"* and *"1970-01-01T0123:45"* are not).</li>
       <li>
         When parsing a date representation for a Temporal.PlainMonthDay, the year may be omitted.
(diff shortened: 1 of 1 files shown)


spec/abstractops.html
```diff
@@ -978,26 +978,23 @@ <h1>ISO 8601 grammar</h1>
...
-        (The date, time, and time zone offset must each be fully in Basic format or Extended format.)
+        Any number of conforming <a href="https://datatracker.ietf.org/doc/html/draft-ietf-sedate-datetime-extended#section-3.1">suffixes in square brackets</a> are allowed.
+        However, the only recognized suffixes are time zone and <a href="https://tools.ietf.org/html/bcp47#section-2.1">BCP 47 calendar</a>.
+        Others are ignored, unless they are prefixed with `!`, in which case they are rejected.
       </li>
+      <li>A space may be used to separate the date and time in a combined date / time representation, but not in a duration (e.g., *"1970-01-01 00:00Z"* is valid but *"P1D 1H"* is not).</li>
+      <li>Alphabetic designators may be in lower or upper case (e.g., *"1970-01-01t00:00Z"* and *"1970-01-01T00:00z"* and *"pT1m"* are valid).</li>
+      <li>Period or comma may be used as the decimal separator (e.g., *"PT1,00H"* is a valid representation of a 1-hour duration).</li>
+      <li>A UTC offset of *"-00:00"* is allowed, and means the same thing as *"+00:00"*.</li>
```
justingrant, line 989: >  means the same thing as *"+00:00"*.

In some contexts, e.g. `Temporal.Instant.from`, this is true, but it may not be true everywhere in Temporal.

In RFC 3339 and IXDTF, "-00:00" means "local offset is unknown" which is the same as the IXDTF (but not RFC3339) meaning of "Z".  And Z is definitely treated differently from +00:00 in some contexts. For example, Z is disallowed when parsing Temporal.Plain* types, and strings with Z can never cause an offset vs. IANA conflict in `ZonedDateTime.from()`.

I don't remember whether  -00:00  is treated the same as Z though. @ptomato do you know?

ptomato: We don't treat `-00:00` the same as Z, so this statement is correct as far as I know. The reason is that we decided it was more important to follow ISO 8601 here. See #2113 and https://github.com/tc39/proposal-temporal/blob/main/meetings/agenda-minutes-2022-05-11.md#special-case--0000-offset-2113

ptomato (approved):

spec/abstractops.html
gibson042: [ParseISODateTime](https://tc39.es/proposal-temporal/#sec-temporal-parseisodatetime) captures any offset string and returns it in [[TimeZone]].[[OffsetString]], from which
* [ToTemporalZonedDateTime](https://tc39.es/proposal-temporal/#sec-temporal-totemporalzoneddatetime) and [ToRelativeTemporalObject](https://tc39.es/proposal-temporal/#sec-temporal-torelativetemporalobject) convert "±00[[:]00]" to to an offset of 0 nanoseconds via [ParseDateTimeUTCOffset](https://tc39.es/proposal-temporal/#sec-parsedatetimeutcoffset).
* [ParseTemporalTimeZoneString](https://tc39.es/proposal-temporal/#sec-temporal-parsetemporaltimezonestring) converts "±00[[:]00]" to { [[Name]]: \~empty~, [[OffsetMinutes]]: 0 } via [ParseTimeZoneIdentifier](https://tc39.es/proposal-temporal/#sec-parsetimezoneidentifier).
* [ParseTemporalInstantString](https://tc39.es/proposal-temporal/#sec-temporal-parsetemporalinstantstring) propagates it into [[TimeZoneOffsetString]].
  * Where it is consumed only by [ParseTemporalInstant](https://tc39.es/proposal-temporal/#sec-temporal-parsetemporalinstant), which converts "±00[[:]00]" to an offset of 0 nanoseconds via [ParseDateTimeUTCOffset](https://tc39.es/proposal-temporal/#sec-parsedatetimeutcoffset).

So I think it is fair to say that Temporal always applies the same interpretation to "-00:00" and "+00:00".

justingrant: > We don't treat `-00:00` the same as Z

Makes sense. In retrospect, we probably should have treated -00:00 the same as Z in all cases, including our prohibition on using it when parsing Plain* types, because -00:00 means the same as Z in IXDTF: it means that the local offset is unknown. 

But in practice, almost no one uses -00:00 (because it's incompatible with ISO 8601). So because no one uses it, I don't think it's worth proactively cracking open the spec to include -00:00 in the Z-blocking behavior of Plain* parsing. 

If we do have to make more normative changes as a result of IXDTF changes, then we may want to consider making this change, though.

What do you think?

gibson042: It would be better to align with IXDTF on this, but I agree that the opportunity to do so may have passed.

ptomato: I would be against relitigating this. We already had this discussion (I linked the notes above) and the outcome was that while ISO 8601 and IXDTF are in opposition here (IXDTF by necessity because RFC 3339 already deviated), it was better to align with ISO 8601. I don't see any new information that would justify reconsidering.

Anyway, it seems we agree about what this PR should do, so let's move ahead with it as is.

PR #2654, opened 2023-08-18
Editorial: Merge Temporal{Zoned,}DateTimeString nonterminals

gibson042:
They differ only by optionality of |TimeZoneAnnotation|, so ParseISODateTime was specifying unnecessary work.


ptomato (approved):
Nicely elegant, thanks.

PR #2659, opened 2023-08-19
Don't look for slots on maybe-undefined relativeTo

### LINKED ISSUE #2618: Temporal.Duration.prototype.round: Test relativeTo is an object before checking internal slots
Step 25 is:
> If relativeTo has an [[InitializedTemporalZonedDateTime]] internal slot, then

But `relativeTo` can be `undefined`, in which case it's not valid to test for internal slots.

justingrant:
Fixes #2618


gibson042 (approved):

ptomato:
I realized when rebasing that this was already solved in #2657. Sorry for the extra work!

PR #2660, opened 2023-08-19
Polyfill: Extend transition-finding lookahead

justingrant:
GetNamedTimeZonePreviousTransition currently has an optimization to avoid fruitlessly searching for a transition in far-future years in time zones that don't have any transitions. This optimization ignores cases where a time zone's offset changes are scheduled more than one year in advance.  This PR extends the lookahead period to three years.

Similarly, GetNamedTimeZoneNextTransition has an optimization where it assumes that no time zone has any transitions more than one year in the future, which IMO is a pretty bad bug because all dates in 2025+ report no transitions in any time zone. This PR extends this lookahead to three years later than either the current date or the input instant, whichever is later. This ensures that, for time zones that currently use recurring DST rules, a next transition will always be reported for any future date. (Admittedly, it seems unlikely that current DST rules will still be in place in 100,000 years, but we need to pick something and aligning transitions with what `Date` and `Intl.DateTimeFormat` reports seems like a good idea.)

When testing locally, these changes slow down these functions by 2%-4%, which IMO is an OK tradeoff for more accuracy.

Fixes #2564

### DIFF
--- a/polyfill/lib/ecmascript.mjs
+++ b/polyfill/lib/ecmascript.mjs
@@ -2830,7 +2830,13 @@ export function GetNamedTimeZoneNextTransition(id, epochNanoseconds) {
   if (epochNanoseconds.lesser(BEFORE_FIRST_DST)) {
     return GetNamedTimeZoneNextTransition(id, BEFORE_FIRST_DST);
   }
-  const uppercap = SystemUTCEpochNanoSeconds().plus(DAY_NANOS.multiply(366));
+  // Optimization: the farthest that we'll look for a next transition is 3 years
+  // after the later of epochNanoseconds or the current time. If there are no
+  // transitions found before then, we'll assume that there will not be any more
+  // transitions after that.
+  const now = SystemUTCEpochNanoSeconds();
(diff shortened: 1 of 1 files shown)


ptomato:
> Similarly, GetNamedTimeZoneNextTransition has an optimization where it assumes that no time zone has any transitions more than one year in the future, which IMO is a pretty bad bug because all dates in 2025+ report no transitions in any time zone. This PR extends this lookahead to three years later than either the current date or the input instant, whichever is later. This ensures that, for time zones that currently use recurring DST rules, a next transition will always be reported for any future date. (Admittedly, it seems unlikely that current DST rules will still be in place in 100,000 years, but we need to pick something and aligning transitions with what `Date` and `Intl.DateTimeFormat` reports seems like a good idea.)

I didn't realize this was the case, but sure enough it is! What I thought the status quo was, (and what my intention was with the optimization), was to assume that if a time zone didn't have any transitions in the coming year, it wouldn't have any after that. I agree this should be fixed, because this is clearly not the correct expectation:
```js
> Temporal.TimeZone.from('America/Vancouver').getNextTransition('2027-08-21T00Z')
null
```

PR #2674, opened 2023-09-14
Fix getOffsetStringFor docs for sub-minute offsets

### LINKED ISSUE #2673: Doc vs polyfill mismatch for timeZone.getOffsetStringFor(), expected string should not show seconds.
The [documentation on Temporal.TimeZone](https://tc39.es/proposal-temporal/docs/timezone.html#getOffsetStringFor) says:

> This method is similar to timeZone.getOffsetNanosecondsFor(), but returns the offset formatted as a string, with sign, hours, and minutes.

However, the polyfill may add seconds:
`mytz = new Temporal.TimeZone('Europe/Paris');`
`mytz.getOffsetStringFor('1900-02-01T12:00Z'); // '00:09:21' `

It should probably round to the nearest minute before computing the string (to be specified).

justingrant:
Fixes #2673 to clarify that sub-minute time zone offsets will be output by `Temporal.TimeZone.prototype.getOffsetStringFor` when the offset isn't aligned on a minute boundary.

Note that sub-minute offsets are only allowed for IANA and custom zones, and are disallowed for offset time zones per #2607.

### DIFF
--- a/docs/timezone.md
+++ b/docs/timezone.md
@@ -331,6 +331,7 @@ tz.getOffsetNanosecondsFor('2020-11-06T01:00Z'); // => 0
 **Returns**: a string indicating the UTC offset at the given time.
 
 This method is similar to `timeZone.getOffsetNanosecondsFor()`, but returns the offset formatted as a string, with sign, hours, and minutes.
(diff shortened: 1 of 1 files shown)


ptomato (approved):
Yep, you beat me to it! Thanks.

PR #2710, opened 2023-10-13
Fix handling of overflow returns to BalancePossiblyInfiniteDurationRelative

12wrigja:
The [current spec](https://github.com/tc39/proposal-temporal/blob/3f82e8d48d3c062d07a4ae77dc175d65cc3be8ae/spec/duration.html#L1365-L1367) for `BalancePossiblyInfinteTimeDurationRelative` includes early-returning the overflow values, but this is not present in the polyfill code.

This seems to have been introduced in 0547d29dbdf4c5aa9c2dca3c6072513dc52abe36.

### DIFF
--- a/polyfill/lib/ecmascript.mjs
+++ b/polyfill/lib/ecmascript.mjs
@@ -3544,17 +3544,11 @@ export function BalancePossiblyInfiniteTimeDurationRelative(
     days = 0;
   }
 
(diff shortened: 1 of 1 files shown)


ptomato (approved):
Thanks for spotting this, it's correct.

Normally I would try to add a test262 test, since apparently this is a coverage gap. But in #2612, hopefully landing soon, we are removing the possibility for a duration to overflow in this function (because you just won't be able to construct durations that could overflow here.) So I think in this case it's not necessary.

PR #2726, opened 2023-11-10
Remove conditional lookups

ptomato:
This addresses Richard's comments on #2519.

Tests are in https://github.com/tc39/test262/pull/3959

Closes: #2724

### DIFF
--- a/spec/calendar.html
+++ b/spec/calendar.html
@@ -186,6 +186,25 @@
       </emu-alg>
     </emu-clause>
 
+    <emu-clause id="sec-temporal-createcalendarmethodsrecordfromrelativeto" type="abstract operation">
+      <h1>
+        CreateCalendarMethodsRecordFromRelativeTo (
+          _plainRelativeTo_: a `Temporal.PlainDate` or *undefined*,
+          _zonedRelativeTo_: a `Temporal.ZonedDateTime` or *undefined*,
+          _methods_: a List of one of ~dateAdd~, ~dateFromFields~, ~dateUntil~, ~day~, ~fields~, ~mergeFields~, ~monthDayFromFields~, or ~yearMonthFromFields~,
+        ): either a normal completion containing either a Calendar Methods Record or *undefined*, or a throw completion
+      </h1>
+      <dl class="header">
+        <dt>description</dt>
+        <dd>It creates a Calendar Methods Record from the calendar of the given `relativeTo` object, looking up the methods given in _methods_. If no `relativeTo` object is given, it returns *undefined* instead of a Calendar Methods Record.</dd>
+      </dl>
+      <emu-alg>
+        1. If _zonedRelativeTo_ is not *undefined*, return ? CreateCalendarMethodsRecord(_zonedRelativeTo_.[[Calendar]], _methods_).
+        1. If _plainRelativeTo_ is not *undefined*, return ? CreateCalendarMethodsRecord(_plainRelativeTo_.[[Calendar]], _methods_).
+        1. Return *undefined*.
+      </emu-alg>
+    </emu-clause>
+
     <emu-clause id="sec-temporal-calendarmethodsrecordlookup" type="abstract operation">
       <h1>
         CalendarMethodsRecordLookup (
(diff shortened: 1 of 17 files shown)


justingrant (approved):
👍

spec/duration.html
```diff
@@ -105,9 +105,7 @@ <h1>Temporal.Duration.compare ( _one_, _two_ [ , _options_ ] )</h1>
         1. Let _calendarRec_ be *undefined*.
         1. If _zonedRelativeTo_ is not *undefined* or _plainRelativeTo_ is not *undefined*, then
           1. If _zonedRelativeTo_ is not *undefined*, let _calendar_ be _zonedRelativeTo_.[[Calendar]]; else let _calendar_ be _plainRelativeTo_.[[Calendar]].
-          1. Set _calendarRec_ to ! CreateCalendarMethodsRecord(_calendar_, « »).
-          1. If _calendarUnitsPresent_ is *true*, then
-            1. Perform ? CalendarMethodsRecordLookup(_calendarRec_, ~dateAdd~).
+          1. Set _calendarRec_ to ? CreateCalendarMethodsRecord(_calendar_, « ~dateAdd~ »).
```
gibson042, line 108: I wonder if the four instances of this _`If _zonedRelativeTo_ is not *undefined* or _plainRelativeTo_ is not *undefined*, … Set _calendarRec_ to ? CreateCalendarMethodsRecord(_calendar_, « … »)`_ pattern justify introducing a new operation.
```suggestion
        1. Let _calendarRec_ be CreateCalendarMethodsRecordFromField(« _zonedRelativeTo_, _plainRelativeTo_ », « ~dateAdd~ »).
```

gibson042 (approved):

spec/duration.html
ptomato: Thanks for the suggestion. Added MaybeCreateCalendarMethodsRecordFromRelativeTo.

ptomato:
@gibson042 Would you be able to check if you agree with https://github.com/tc39/test262/pull/3959 as well, so we can get this merged promptly?

spec/calendar.html
```diff
@@ -186,6 +186,25 @@ <h1>
       </emu-alg>
     </emu-clause>
 
+    <emu-clause id="sec-temporal-maybecreatecalendarmethodsrecordfromrelativeto" type="abstract operation">
+      <h1>
+        MaybeCreateCalendarMethodsRecordFromRelativeTo (
```
gibson042, line 191: I think "Maybe" can be dropped from the name, "CreateCalendarMethodsRecordFromRelativeTo" reads just as well without misleading w.r.t. undefined in my opinion.

spec/calendar.html
```diff
@@ -186,6 +186,25 @@ <h1>
...
+          _methods_: a List of one of ~dateAdd~, ~dateFromFields~, ~dateUntil~, ~day~, ~fields~, ~mergeFields~, ~monthDayFromFields~, or ~yearMonthFromFields~,
+        ): either a normal completion containing either a Calendar Methods Record or *undefined*, or a throw completion
+      </h1>
+      <dl class="header">
+        <dt>description</dt>
+        <dd>It creates a Calendar Methods Record from the calendar of the given `relativeTo` object, looking up the methods given in _methods_. If no `relativeTo` object is given, it returns *undefined* instead of a Calendar Methods Record.</dd>
+      </dl>
+      <emu-alg>
+        1. If _zonedRelativeTo_ is *undefined* and _plainRelativeTo_ is *undefined*, return *undefined*.
+        1. If _zonedRelativeTo_ is not *undefined*, let _calendar_ be _zonedRelativeTo_.[[Calendar]]; else let _calendar_ be _plainRelativeTo_.[[Calendar]].
```
gibson042, line 203: Nit: I think having a dedicated operation (and an alias for the list of field names) allows this logic to better highlight the similar treatment of plain vs. zoned input.
```
        1. If _zonedRelativeTo_ is not *undefined*, return ? CreateCalendarMethodsRecord(_zonedRelativeTo_.[[Calendar]], _methods_).
        1. If _plainRelativeTo_ is not *undefined*, return ? CreateCalendarMethodsRecord(_plainRelativeTo_.[[Calendar]], _methods_).
        1. Return *undefined*.
```

gibson042 (approved):
> @gibson042 Would you be able to check if you agree with [tc39/test262#3959](https://github.com/tc39/test262/pull/3959) as well, so we can get this merged promptly?

Done.

spec/calendar.html
ptomato: Agreed, I like this better.

ptomato:
Thanks for the reviews.

PR #2736, opened 2023-12-13
Editorial: Use consistent capitalization for Abstract Operations

arai-a:
ECMAScript spec uses "Abstract Operations" and "Abstract Operations for Something" as the capitalization in the heading.


ptomato (approved):
Thanks!

PR #2737, opened 2023-12-13
Editorial: Add missing constructor properties of the Temporal object

arai-a:
`Temporal.Calendar` and `Temporal.ZonedDateTime` are missing from the `Temporal` object's property


ptomato:
Thanks!

ptomato (approved):

PR #2739, opened 2023-12-13
Editorial: missing variable markup

Ms2ger:


ptomato (approved):
Thanks! Technically correct, but consider adding the markup to `_year_` and `_day_` in the same line? :smile:

ptomato (approved):

PR #2753, opened 2024-01-15
Improve roundToNearestMonth example

### LINKED ISSUE #2751: Example violates best practices for cross-calendar code
The [roundToNearestMonth](https://github.com/tc39/proposal-temporal/blob/main/docs/cookbook/roundToNearestMonth.mjs) example in the cookbook contains code like this:

```
if (month > 12) {
  month %= 12;
  year++;
}
```

which seems to violate the [best practices](https://tc39.es/proposal-temporal/docs/calendar.html#writing-cross-calendar-code) by assuming there's always 12 months in a year.

Seems to me like it should use `.add({ months: 1 })` instead of working with the month and year numbers directly. I can submit a PR fixing this if needed.

mjomble:
Fixes #2751 

In addition to removing the use of hardcoded 12, I found that using Duration to find the closest option was more readable than the day of month arithmetic.

### DIFF
--- a/docs/cookbook/roundToNearestMonth.mjs
+++ b/docs/cookbook/roundToNearestMonth.mjs
@@ -1,11 +1,12 @@
 const date = Temporal.PlainDate.from('2018-09-16');
 
-let { year, month } = date;
-if ((date.day - 1) / date.daysInMonth >= 0.5) month++;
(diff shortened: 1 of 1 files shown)


ptomato (approved):
Thanks!

PR #2755, opened 2024-01-16
Editorial: Add no separator text input samples

kwatanabe1050:
I update document about no-separator input samples of each from-function for this issue: https://github.com/js-temporal/temporal-polyfill/issues/273

The document explains it follows ISO 8601 but the specification explained the format more strictly.
(I'm not good at English so I may overlook in the document, sorry.)
Therefore I added sample codes to show no separator is in specification.

If there's no explains really, I may be able to add short grammar explaining page into `Other documentation` list.

### DIFF
--- a/docs/plaindate.md
+++ b/docs/plaindate.md
@@ -96,6 +96,7 @@ Example usage:
 <!-- prettier-ignore-start -->
 ```javascript
 date = Temporal.PlainDate.from('2006-08-24'); // => 2006-08-24
+date = Temporal.PlainDate.from('20060824'); // => 2006-08-24
 date = Temporal.PlainDate.from('2006-08-24T15:43:27'); // => 2006-08-24
 date = Temporal.PlainDate.from('2006-08-24T15:43:27+01:00[Europe/Brussels]');
   // => 2006-08-24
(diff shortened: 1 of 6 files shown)


docs/calendar.md
```diff
@@ -220,6 +220,7 @@ cal = Temporal.Calendar.from('gregory');
 cal = Temporal.Calendar.from('2020-01-13T16:31:00.065858086');
 cal = Temporal.Calendar.from('2020-01-13T16:31:00.065858086-08:00[America/Vancouver][u-ca=iso8601]');
 /* WRONG */ cal = Temporal.Calendar.from('[u-ca-iso8601]'); // => throws, lone annotation not a valid ISO 8601 string
+cal = Temporal.Calendar.from('202001'); // OK because YYYYMM is ISO format of YearMonth
```
justingrant, line 223: I would recommend omitting this line, because while it will not throw and will return the ISO 8601 calendar, it's also a confusing example. 

We don't show other formats for `Temporal.Calendar.from` for the same reason. 

Could you remove this line please?  Thanks!

Your other changes look great. Once this line is removed, we can approve and merge this PR. Thanks for the PR!

PR #2756, opened 2024-01-17
Normative: Make CalendarDateWeekOfYear return undefined 

Aditi-1400:
Make `CalendarDateWeekOfYear` return undefined for all calendars without well-defined week calendar system

Fixes: #2744

### DIFF
--- a/polyfill/lib/calendar.mjs
+++ b/polyfill/lib/calendar.mjs
@@ -230,15 +230,66 @@ export class Calendar {
     date = ES.ToTemporalDate(date);
     return impl[GetSlot(this, CALENDAR_ID)].dayOfYear(date);
   }
-  weekOfYear(date) {
+  calendarDateWeekOfYear(date) {
+    // Supports only Gregorian and ISO8601 calendar; can be updated to add support for other calendars.
+    // Returns undefined for calendars without a well-defined week calendar system.
+    // eslint-disable-next-line max-len
+    // Also see: https://github.com/unicode-org/icu/blob/ab72ab1d4a3c3f9beeb7d92b0c7817ca93dfdb04/icu4c/source/i18n/calendar.cpp#L1606
     if (!ES.IsTemporalCalendar(this)) throw new TypeError('invalid receiver');
     date = ES.ToTemporalDate(date);
-    return impl[GetSlot(this, CALENDAR_ID)].weekOfYear(date);
+    const id = GetSlot(this, CALENDAR_ID);
+    if (id !== 'gregory' && id !== 'iso8601') {
+      return { week: undefined, year: undefined };
+    }
+    const calendar = impl[id];
+    let yow = GetSlot(date, ISO_YEAR);
+    const dayOfWeek = this.dayOfWeek(date);
+    const dayOfYear = this.dayOfYear(date);
+    const fdow = id === 'iso8601' ? 1 : calendar.helper.getFirstDayOfWeek();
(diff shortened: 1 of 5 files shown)


spec/intl.html
```diff
@@ -1697,6 +1697,9 @@ <h1>
             Usually the [[Year]] field will contain the same value given by CalendarDateYear, but may contain the previous or next year if the week number in the [[Week]] field overlaps two different years.
             See also ToISOWeekOfYear.
           </p>
+          <p>
+            It returns *undefined* in [[Week]] and [[Year]] field for calendars that do not have a well-defined week calendar system.
```
ptomato, line 1701: In the Temporal champions meeting we discussed that this should have a link to Intl. I've opened an [issue](https://github.com/tc39/proposal-intl-era-monthcode/issues/15) in the Intl Era/Month Codes proposal to suggest moving week numbering into scope, so that issue seems like a natural thing to link to. How about an editor's note that mentions this will be further specified there?

ptomato (changes requested):
After initially reviewing this I thought of something else that needs to be done. We should change CalendarWeekOfYear and CalendarYearOfWeek so that a custom calendar's `weekOfYear` and `yearOfWeek` method can return undefined. (Similar to what we do in CalendarEra and CalendarEraYear.) This is also something we could write test262 tests for.

ptomato (approved):
:+1: Thanks!

polyfill/lib/calendar.mjs
```diff
@@ -230,15 +230,55 @@ export class Calendar {
...
+    if (id === 'gregory' || id === 'iso8601') {
+      const calendar = impl[GetSlot(this, CALENDAR_ID)];
+      let yow = GetSlot(date, ISO_YEAR);
+      const dayOfWeek = this.dayOfWeek(date);
+      const dayOfYear = this.dayOfYear(date);
+      const fdow = id === 'iso8601' ? 1 : calendar.helper.getFirstDayOfWeek();
+      const mdow = id === 'iso8601' ? 4 : calendar.helper.getMinimalDaysInFirstWeek();
+      var relDow = (dayOfWeek + 7 - fdow) % 7;
+      // Assuming the year length is less than 7000 days.
+      var relDowJan1 = (dayOfWeek - dayOfYear + 7001 - fdow) % 7;
```
gibson042, line 250: This is rather opaque; I can see that in iso8601 `relDowJan1` is 0 in iso8601 for a Monday January 1 and any later day of that year, 1 for a Tuesday January 1 and the rest of that year, etc. but reading meaning from that seems unnecessarily difficult.
```suggestion
      // For both the input date and the first day of its calendar year, calculate the day of week
      // relative to first day of week in the relevant calendar (e.g., in iso8601, relative to Monday).
      var relDow = (dayOfWeek + 7 - fdow) % 7;
      var relDowJan1 = (dayOfWeek - dayOfYear + 7001 - fdow) % 7;
```

polyfill/lib/calendar.mjs
```diff
@@ -230,15 +230,55 @@ export class Calendar {
...
+      var woy = MathFloor((dayOfYear - 1 + relDowJan1) / 7);
```
gibson042, line 251: The logic below this point is also opaque and would benefit from comments.

gibson042 (approved):

ptomato:
This normative change reached consensus in the 2023-02 TC39 plenary meeting. I've updated this PR to include the test262 tests that I've just merged.

PR #2759, opened 2024-01-27
Normative: Fix date difference for end-of-month edge cases

ptomato:
This adjusts the difference algorithm for Gregorian-year dates so that when an intermediate date occurs past the end of a month, it is not shifted to the end of that month.

Previously, in some edge cases where taking the difference in months or years would return a number of months and zero days, we now return one month less and 28, 29, or 30 days instead.

Example: 1970-01-29 until 1971-02-28, largestUnit years
Old result: 1 year, 1 month
New result: 1 year, 30 days

Note that largestUnit weeks and largestUnit days, the latter of which is the default, are not affected.

~~(**Note:** Stacked on top of an editorial change which I will eventually make a separate PR for.)~~

Closes: #2535


pipobscure (approved):

ptomato:
This change gained consensus at TC39 on 2024-02-06.

PR #2762, opened 2024-01-30
Editorial: Rewrite DifferenceISODate as a clearer algorithm

ptomato:
The DifferenceISODate AO was a port from optimized JS to spec language, but was difficult to understand. Replace it with an algorithm that is written in an easier to understand way, but produces identical results (verified by testing a 1:1 implementation of the old and new algorithms on every possible pair of dates in a 4-year period.)

The reference code is also changed to match the AO more closely and be easier to understand, although some optimizations are made.

(This is the editorial part of #2759.)


Ms2ger (approved):
Thanks!

PR #2764, opened 2024-01-31
Bump ecmarkup

bakkot:
Ecmarkup [now has](https://github.com/tc39/ecmarkup/pull/568) _extremely basic_, strictly local, very permissive typechecking, which can detect when you are calling an AO with a literal value which doesn't make sense given the declared type of that AO (for some literal values and declared types).

It catches a couple typos here, which I fixed. There's one issue I didn't fix because I don't know what it's intended to be, which is that [this line](https://github.com/tc39/proposal-temporal/blob/b918dc3b62ec2942683970412ce9671ee501d15b/spec/abstractops.html#L1101) passes `false` to a place which accepts either `~negative~` or `~positive~`.

### DIFF
--- a/spec/abstractops.html
+++ b/spec/abstractops.html
@@ -1098,7 +1098,7 @@
     </dl>
     <emu-alg>
       1. Let _quotient_ be _x_ / _increment_.
-      1. Let _unsignedRoundingMode_ be GetUnsignedRoundingMode(_roundingMode_, *false*).
+      1. Let _unsignedRoundingMode_ be GetUnsignedRoundingMode(_roundingMode_, ~positive~).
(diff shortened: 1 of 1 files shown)


ptomato (approved):
@bakkot Thanks, much appreciated!

That `false` [used to](https://github.com/tc39/proposal-temporal/commit/75279e5b1d721cc489d25a3aa431ed5308d7bf77#diff-3a221d4b1eb72b9134f63736e30ed2d2c11eccd7ee2fb34beb3c9ac8e1c3247cR658) be passed to a boolean parameter called "isNegative", so I'll assume that it should now be `~positive~`. I'll tack on a commit fixing that, and then merge this.

PR #2765, opened 2024-01-31
Editorial: use new IteratorStepValue AO

bakkot:
https://github.com/tc39/ecma262/pull/3268 added a new AO which lets us simplify use of the iteration protocol.

I didn't update the polyfill, sorry.


ptomato (approved):
Thanks!

Issue #2552, opened 2023-04-12
ObjectImplementsTemporalCalendarProtocol: Excessive checks of properties?


anba:
[ObjectImplementsTemporalCalendarProtocol](https://tc39.es/proposal-temporal/#sec-temporal-objectimplementstemporalcalendarprotocol) currently checks all properties which are called in the various Calendar* abstract operations. For example:
- `"dateAdd"` because of CalendarDateAdd
- `"dateFromFields"` because of CalendarDateFromFields
- `"dateUntil"` because of CalendarDateUntil
- `"day"` because of CalendarDay
- etc.

This seems (a) a bit excessive and (b) seems problematic for forward compatibility. When a future proposal adds, let's say [CalendarWeeksInYear](https://github.com/js-temporal/proposal-temporal-v2/issues/8) and a corresponding `Temporal.Calendar.prototype.weeksInYear` function, does that mean `ObjectImplementsTemporalCalendarProtocol` needs to check the property key `"weeksInYear"`, too? If yes, does that mean that custom calendars will suddenly stop working when an implementation adds support for `Temporal.Calendar.prototype.weeksInYear`, because [ToTemporalCalendarSlotValue](https://tc39.es/proposal-temporal/#sec-temporal-totemporalcalendarslotvalue) will start throwing a TypeError if `"weeksInYear"` is absent?

Also applies to [ObjectImplementsTemporalTimeZoneProtocol](https://tc39.es/proposal-temporal/#sec-temporal-objectimplementstemporaltimezoneprotocol).

---

Kind of related:
[IsRegExp](https://tc39.es/ecma262/#sec-isregexp) tests for a single property to classify an object as RegExp-like. This special behaviour of `Symbol.match` isn't liked by all (https://github.com/tc39/ecma262/pull/1318). There used to be a `Symbol.isRegExp` well-known symbol to mark an object as RegExp-like. It was replaced with `Symbol.match` in ES6 draft rev29 per the [resolution from the Nov 2014 meeting](https://github.com/tc39/notes/blob/main/meetings/2014-11/nov-18.md#410-regexp-subclassing-fixes).

justingrant:
> This seems (a) a bit excessive and (b) seems problematic for forward compatibility

[...]

Issue #2549, opened 2023-04-12
Temporal.Calendar.prototype.mergeFields has an unnecessary initialisation


anba:
[Temporal.Calendar.prototype.mergeFields](https://tc39.es/proposal-temporal/#sec-temporal.calendar.prototype.mergefields), step 16.a should be removed, because elsewhere this extra initialisation isn't used. (`Set` in steps 16.b-c then need to be `Let`.)

Issue #2548, opened 2023-04-12
CalendarFields: Assert "fieldNames" parameter is sorted and doesn't contain any duplicates?


anba:
Asserting the "fieldNames" parameter is sorted, doesn't contain any duplicates, and contains only valid calendar field names may make it more obvious that implementations can skip calling `Temporal.Calendar.prototype.fields` for the fast-path.

Issue #2534, opened 2023-03-24
RejectObjectWithCalendarOrTimeZone: Avoid "Reject" so there's no confusion about Promise rejection?


anba:
"Reject" is exclusively used for Promise rejection in ECMA-262. Should "Reject" be replaced with a different verb in "RejectObjectWithCalendarOrTimeZone"?

For example compare to [RejectPromise](https://tc39.es/ecma262/#sec-rejectpromise), which unconditionally rejects a Promise, whereas "RejectObjectWithCalendarOrTimeZone" conditionally throws an error.

gibson042:
One possibility is refactoring into something like HasCalendarOrTimeZone or IsTemporalLikeObject and constructing errors at the call site when it returns true, which would seem like an improvement.

PR #2770, opened 2024-02-13
Editorial: Address various issues raised by Anba

ptomato:
Closes: #2533, #2534, #2548, #2549, #2552

### DIFF
--- a/spec/abstractops.html
+++ b/spec/abstractops.html
@@ -843,26 +843,24 @@
     </emu-alg>
   </emu-clause>
 
-  <emu-clause id="sec-temporal-rejecttemporallikeobject" type="abstract operation">
+  <emu-clause id="sec-temporal-ispartialtemporalobject" type="abstract operation">
     <h1>
-      RejectTemporalLikeObject (
-        _object_: an Object,
-      ): either a normal completion containing ~unused~, or a throw completion
+      IsPartialTemporalObject (
+        _object_: an ECMAScript language value,
+      ): either a normal completion containing a Boolean, or a throw completion
     </h1>
     <dl class="header">
       <dt>description</dt>
-      <dd>It throws an exception if its argument _object_ is an instance of one of the time-related or date-related Temporal types, or is an object that has a `calendar` or `timeZone` property.</dd>
+      <dd>It determines whether _value_ is a suitable input for one of the Temporal objects' `with()` methods: it must be an Object, it must not be an instance of one of the time-related or date-related Temporal types, and it must not have a `calendar` or `timeZone` property.</dd>
     </dl>
     <emu-alg>
-      1. If _object_ has an [[InitializedTemporalDate]], [[InitializedTemporalDateTime]], [[InitializedTemporalMonthDay]], [[InitializedTemporalTime]], [[InitializedTemporalYearMonth]], or [[InitializedTemporalZonedDateTime]] internal slot, then
-        1. Throw a *TypeError* exception.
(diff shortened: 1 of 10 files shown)


ptomato:
cc @anba

spec/plaindate.html
```diff
@@ -406,9 +406,7 @@ <h1>Temporal.PlainDate.prototype.with ( _temporalDateLike_ [ , _options_ ] )</h1
       <emu-alg>
         1. Let _temporalDate_ be the *this* value.
         1. Perform ? RequireInternalSlot(_temporalDate_, [[InitializedTemporalDate]]).
-        1. If Type(_temporalDateLike_) is not Object, then
-          1. Throw a *TypeError* exception.
-        1. Perform ? RejectTemporalLikeObject(_temporalDateLike_).
+        1. If _temporalDateLike_ is not an Object, or ? IsTemporalLikeObject(_temporalDateLike_) is *true*, throw a *TypeError* exception.
```
anba, line 409: Suggestion: Maybe move the `Object` check into `IsTemporalLikeObject` to avoid checking it before all calls? Similar to how [IsArray](https://tc39.es/ecma262/#sec-isarray) or [IsRegExp](https://tc39.es/ecma262/#sec-isregexp) check for objects in their first step.

ptomato: I thought about that, but it's semantically very different from IsTemporalLikeObject - we'd have to return true from IsTemporalLikeObject if the object was not an Object! I guess we could rename it IsSuitableInputForTemporalWithMethod and reverse the sense of the check.

ptomato: Or, just reverse the sense of the check, if we clarify that IsTemporalLikeObject is not supposed to include actual Temporal objects!

ptomato: Maybe IsPartialTemporalObject

spec/calendar.html
```diff
@@ -778,6 +778,10 @@ <h1>
           1. If ? HasProperty(_object_, _key_) is *false*, return *false*.
         1. Return *true*.
       </emu-alg>
+      <emu-note>
+        This operation is a convenience which makes it easier to catch bugs in custom calendars.
+        The list of required property keys must not be extended in future editions, even if the Temporal.Calendar protocol gains more methods.
```
Ms2ger, line 783: Please rephrase to avoid "must"

spec/timezone.html
```diff
@@ -735,6 +735,10 @@ <h1>
           1. If ? HasProperty(_object_, _key_) is *false*, return *false*.
         1. Return *true*.
       </emu-alg>
+      <emu-note>
+        This operation is a convenience which makes it easier to catch bugs in custom time zones.
+        The list of required property keys must not be extended in future editions, even if the Temporal.TimeZone protocol gains more methods.
```
Ms2ger, line 740: Ditto

spec/abstractops.html
```diff
@@ -843,26 +843,24 @@ <h1>
     </emu-alg>
   </emu-clause>
 
-  <emu-clause id="sec-temporal-rejecttemporallikeobject" type="abstract operation">
+  <emu-clause id="sec-temporal-istemporallikeobject" type="abstract operation">
```
Ms2ger, line 846: Adjust the id to match the new name?

Ms2ger (approved):
lgtm % comments

spec/abstractops.html
ptomato: Oops, thanks for catching this.

PR #2780, opened 2024-02-21
Editorial: Fix doc build.js for localized page

kwatanabe1050:
Hi there, I'm Japanses user but now ja-localized document page can not use playground in a console.

Because localized page uses sub dir but those resource is written with relative path from doc root and all pages use same static header/footer.
Therefore I fixed doc build script to replace relative path in header/footer content for each sub directory output.

Thank you.


ptomato (approved):
Thanks for fixing this! Much appreciated.

PR #2806, opened 2024-03-20
Post-adjustment: Remove unnecessary object creation

ptomato:
This is to address belated feedback from Anba on #2718. The options object previously created here is unnecessary, because 'constrain' is always the default for this option. Not creating the options object here is consistent with the rest of the spec text.

Closes: #2803


ptomato:
cc @anba

ryzokuken (approved):
RSLGTM

Ms2ger (approved):

PR #2809, opened 2024-03-26
Editorial: Close a bunch of old minor issues

ptomato:
I went through the backlog and fixed a bunch of minor editorial issues that are unlikely to collide with any of the PRs currently in review.

Best reviewed commit by commit.

Hopefully none of these need much discussion. If discussion is needed about a particular commit, I'll instead remove it from here and put it in a separate PR so it doesn't hold up the rest.


justingrant (approved):
Looks great, thanks for making this!

PR #2819, opened 2024-04-15
Polyfill: Update d.ts

fabon-f:
This PR reflects a recent normative change on `weekOfYear` / `yearOfWeek` (#2756) in type definition.


ptomato (approved):
Thanks!

PR #2828, opened 2024-04-26
Editorial: More minor improvements

ptomato:
Put duration units in a table, and use that table to make operations easier to follow.

Address a couple of other minor editorial issues.

### DIFF
--- a/spec/instant.html
+++ b/spec/instant.html
@@ -589,7 +589,7 @@
           _ns2_: a BigInt,
           _roundingIncrement_: a positive integer,
           _smallestUnit_: a String,
-          _roundingMode_: a String from the Identifier column of <emu-xref href="#table-temporal-rounding-modes"></emu-xref>,
+          _roundingMode_: a String from the "Identifier" column of <emu-xref href="#table-temporal-rounding-modes"></emu-xref>,
         ): a Normalized Time Duration Record
       </h1>
       <dl class="header">
@@ -611,7 +611,7 @@
           _ns_: a BigInt,
           _increment_: a positive integer,
           _unit_: a String from the "Singular" column of <emu-xref href="#table-temporal-units"></emu-xref>,
-          _roundingMode_: a String from the Identifier column of <emu-xref href="#table-temporal-rounding-modes"></emu-xref>,
+          _roundingMode_: a String from the "Identifier" column of <emu-xref href="#table-temporal-rounding-modes"></emu-xref>,
         ): a BigInt
       </h1>
       <dl class="header">
--- a/spec/instant.html
+++ b/spec/instant.html
@@ -610,7 +610,7 @@
         RoundTemporalInstant (
           _ns_: a BigInt,
           _increment_: a positive integer,
-          _unit_: a String,
+          _unit_: a String from the "Singular" column of <emu-xref href="#table-temporal-units"></emu-xref>,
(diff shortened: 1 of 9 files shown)


spec/instant.html
```diff
@@ -619,19 +619,9 @@ <h1>
...
-          1. Let _incrementNs_ be _increment_ &times; 10<sup>9</sup>.
-        1. Else if _unit_ is *"millisecond"*, then
-          1. Let _incrementNs_ be _increment_ &times; 10<sup>6</sup>.
-        1. Else if _unit_ is *"microsecond"*, then
-          1. Let _incrementNs_ be _increment_ &times; 10<sup>3</sup>.
-        1. Else,
-          1. Assert: _unit_ is *"nanosecond"*.
-          1. Let _incrementNs_ be _increment_.
+        1. Assert: The category of _unit_ as given in the Category column of the corresponding row of <emu-xref href="#table-temporal-units"></emu-xref> is ~time~.
+        1. Let _unitLength_ be the length of _unit_ as given in the Length in Nanoseconds column of the corresponding row of <emu-xref href="#table-temporal-units"></emu-xref>.
```
Ms2ger, line 623: Is this esmeta-proof?

ptomato: I don't know whether ESMeta can deal with it, but I based it on language such as in [UnicodeMatchProperty](https://tc39.es/ecma262/#sec-runtime-semantics-unicodematchproperty-p):

> 1. Assert: _p_ is a Unicode property name or property alias listed in the “Property name and aliases” column of Table 66 or Table 67.
> 2. Let _c_ be the canonical property name of _p_ as given in the “Canonical property name” column of the corresponding row.

ptomato: I went down rather more of a rabbit hole than I intended to, but here's what ESMeta makes of UnicodeMatchProperty:
```
def UnicodeMatchProperty(
  rer: Unknown["RegExpRecord"],
  p: Unknown["ECMAScriptSourceText"],
): Unknown["UnicodePropertyName"] {
  (yet "If _rer_.[[UnicodeSets]] is *true* and _p_ is a Unicode <emu-not-ref>property name</emu-not-ref> listed in the “<emu-not-ref>Property name</emu-not-ref>” column of <emu-xref href=\"#table-binary-unicode-properties-of-strings\"></emu-xref>, then\n  1. Return the List of Unicode code points _p_.")
  assert (yet "_p_ is a Unicode <emu-not-ref>property name</emu-not-ref> or property alias listed in the “<emu-not-ref>Property name</emu-not-ref> and aliases” column of <emu-xref href=\"#table-nonbinary-unicode-properties\"></emu-xref> or <emu-xref href=\"#table-binary-unicode-properties\"></emu-xref>.")
  (yet "Let _c_ be the canonical <emu-not-ref>property name</emu-not-ref> of _p_ as given in the “Canonical <emu-not-ref>property name</emu-not-ref>” column of the corresponding row.")
  (yet "Return the List of Unicode code points _c_.")
}
```

(`yet` means "language that is not implemented yet" so it's probably something that would be handled with [manual rules](https://github.com/es-meta/esmeta/blob/38ca7770e6c64861735c6a02c902cf58f533bc73/src/main/resources/manuals/default/rule.json#L25).)

ptomato: (Also as a note to self, ESMeta cannot be built with JDK 21: https://github.com/sbt/sbt/issues/7235)

ptomato:
Rephrased slightly based on more examples from other proposals.

Also added a commit putting column names between quotes consistently.

PR #2867, opened 2024-05-28
Docs: Cleanup changes in "removals"

gibson042:


ptomato:
Thanks! I will incorporate these edits directly into the `removals` branch, and pull "Docs: Add missing time of day" and "Docs: Use consistent language when referencing the POSIX epoch" into my next editorial PR.

PR #2886, opened 2024-06-07
Editorial: Consistently mark increment as "positive integer"

anba:
This makes it easier to validate the math in NudgeToCalendarUnit when calling ApplyUnsignedRoundingMode. It also matches other operations which already require that `increment` is a positive integer.

(Cherry-picked from #2884.)


ptomato (approved):
Thanks!

PR #2904, opened 2024-06-24
Synchronise Intl updates with latest ECMA-402

anba:
Synchronise `Intl.DateTimeFormat` updates with latest ECMA-402, see individual commit messages for more in-depth explanations.


spec/biblio.json
```diff
@@ -1,111 +1,2 @@
 [
```
ptomato, line 1: Should we just get rid of `biblio.json` at this point? Now that 262 and 402 both have published bibliographies, maybe it's no longer necessary.

ptomato (commented):
I didn't go through the second commit in detail yet, but now that we have CanonicalizeUValue in ECMA-402 would you like to include your conclusion from https://github.com/tc39/ecma402/issues/828#issuecomment-2178192185 in this PR as well?

spec/biblio.json
anba: Sure. Added 9c02c45b154484bbaaa894d2d099ee32f3a3e6de to remove biblio.json.

anba:
> [...] now that we have CanonicalizeUValue in ECMA-402 would you like to include your conclusion from [tc39/ecma402#828 (comment)](https://github.com/tc39/ecma402/issues/828#issuecomment-2178192185) in this PR as well?

Added b0b9467cf0b9cddc16df2702426a65125015118f to replace ASCII lowercase normalisation for calendar identifiers with `CanonicalizeUValue`.

ptomato:
@anba It's labelled "Normative", but I'm not sure it is really normative — it seems to me we are simply adjusting to a normative change that was already made upstream in ECMA-402.

spec/intl.html
```diff
@@ -1577,122 +1607,102 @@ <h1>Intl.DateTimeFormat.prototype.resolvedOptions ( )</h1>
               <th>Internal Slot</th>
```
ptomato, line 1607: This table now has no `<del>` or `<ins>` markers any more. Can we remove it entirely and just let the links go to upstream ECMA-402?

ptomato (approved):
Thanks for doing this. It looks much cleaner and easier to understand now.

I don't think I fully understand the `hourCycle`/`hour12` issue but if you say that is what's needed to replicate existing behaviour, I believe that.

I'd consider this ready to merge, just would like to clear up your expectations for the CanonicalizeUValue commit — do you think this needs to be presented at plenary? I would say it does not.

Optionally, consider removing the resolvedOptions table and the definition of AvailableCalendars in favour of linking to the biblio versions of them. However I wouldn't want to delay the PR waiting for this, so I can also just do those later.

spec/intl.html
anba: Done in a76057d7193d4b73acc10d2cd56feb452decedf1

anba:
> I'd consider this ready to merge, just would like to clear up your expectations for the CanonicalizeUValue commit — do you think this needs to be presented at plenary? I would say it does not.

I don't think it's necessary to present this is at plenary. I've only marked this as "normative" because the polyfill needs to be changed, so there's at least one implementation where this change has a normative effect. 

> Optionally, consider removing the resolvedOptions table and the definition of AvailableCalendars in favour of linking to the biblio versions of them. However I wouldn't want to delay the PR waiting for this, so I can also just do those later.

I've removed the `resolvedOptions` table, but I wasn't sure about what to do with `AvailableCalendars`, because `AvailableCalendars` is called from the main spec of Temporal, but upstream `AvailableCalendars` is from ECMA-402.

ptomato:
Thanks.

I've looked at updating the reference code to also canonicalize calendars, to match the latest ECMA-402. However, there's one test262 test which treats `islamicc` and `islamic-civil` as different, so we'd need to update that first. I don't want to block this PR on that, so I'll open a follow up issue.

ptomato:
I'll also deal with AvailableCalendars in a follow up issue.

PR #2932, opened 2024-09-05
Bump ecmarkup and biblios

bakkot:
Since there's also been some relevant upstream changes which the linter can identify, it applies those changes as well.

Commits can be reviewed individually.


ptomato (approved):
Thanks, all looks good!

PR #2933, opened 2024-09-07
Polyfill: Update type definitions of `Intl`

fabon-f:
* `DateTimeFormatRangePart`: included in standard TypeScript definition as `Intl.DateTimeRangeFormatPart` (https://github.com/microsoft/TypeScript/blob/31afb988d1c6857ce4d1ab3b2558385e68a565a8/src/lib/es2021.intl.d.ts#L14)
* `DateTimeFormatOptions`: all defined options are already included in standard TypeScript definition (https://github.com/microsoft/TypeScript/blob/31afb988d1c6857ce4d1ab3b2558385e68a565a8/src/lib/es2021.intl.d.ts#L6-L12)

These types are supported in official type definition of ES2021, so it's possible to omit from this repository's `d.ts` file.


ptomato (approved):
Thanks!

PR #2935, opened 2024-09-11
Docs improvements

ptomato:
Various improvements to the docs, mostly taken from a sweep of the issue backlog. Also corrected other mistakes I ran into.

### DIFF
--- a/docs/duration.md
+++ b/docs/duration.md
@@ -73,11 +73,11 @@ new Temporal.Duration(); // => PT0S
 
 ## Static methods
 
-### Temporal.Duration.**from**(_thing_: any) : Temporal.Duration
+### Temporal.Duration.**from**(_item_: Temporal.Duration | object | string) : Temporal.Duration
 
 **Parameters:**
 
-- `thing`: A `Duration`-like object or a string from which to create a `Temporal.Duration`.
+- `item`: a value convertible to a `Temporal.Duration`.
 
 **Returns:** a new `Temporal.Duration` object.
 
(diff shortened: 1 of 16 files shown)


docs/duration.md
```diff
@@ -73,11 +73,11 @@ new Temporal.Duration(); // => PT0S
 
 ## Static methods
 
-### Temporal.Duration.**from**(_thing_: any) : Temporal.Duration
+### Temporal.Duration.**from**(_durationLike_: Temporal.Duration | object | string) : Temporal.Duration
 
 **Parameters:**
 
-- `thing`: A `Duration`-like object or a string from which to create a `Temporal.Duration`.
+- `durationLike`: a value convertible to a `Temporal.Duration`.
```
gibson042, line 80: These "_${type}Like_" changes don't seem quite right to me; I would say that (w.l.o.g.) "duration-like" refers to just an **object** having the right properties (analogous to "[array-likes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array#array-like_objects)" having `length` and array index properties). Generically, I guess the input is "_durationable_"/"_instantable_"/"_datetimeable_"/etc., although such names derived from already compound foundations aren't great. Maybe we should just tweak the "_thing_" convention and name them all "_item_" or "_input_" or "_coercible_" (and probably update the [spec](https://tc39.es/proposal-temporal/#sec-temporal.plaindate.from) to match)?

ptomato: Sure, if _durationLike_ suggests an object then it's probably less confusing to use a vaguer term such as _item_.

gibson042 (approved):
LGTM!

PR #2939, opened 2024-09-16
Normative: Never ignore a specified year in CalendarMonthDayToISOReferenceDate

gibson042:
Fixes #2863

Help wanted with tests (assuming there is in fact coverage for this code).


gibson042:
Replaced by #2940.

PR #2941, opened 2024-09-18
Polyfill: Be more robust against post-load environment changes

gibson042:
While working on #2940, I noticed that the polyfill was vulnerable to post-load manipulation of primordials. This PR addresses every such pattern I could find... the first PR is semi-mechanical replacement of dynamic-lookup method invocations (includes those hidden by iterable spread/destructuring or for..of), the second fully so.

The result is slightly less readable, but hews more closely to the spec and doubles down on the `ES.Call` protections that are already commonplace.

### DIFF
--- a/polyfill/lib/calendar.mjs
+++ b/polyfill/lib/calendar.mjs
@@ -1,34 +1,65 @@
+import {
+  // constructors and similar
+  Date as Date,
+  Map as Map,
+  Set as Set,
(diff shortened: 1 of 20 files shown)


ptomato (approved):
I haven't really considered it much of a priority do this, because it's only intended for testing - but I wouldn't turn this down either.

Let's merge it quickly, as it'd be a pain to rebase.

PR #2942, opened 2024-09-18
Polyfill: Make field references in error messages more consistent

gibson042:
The majority do not quote them, so this replaces the exceptions.


ptomato (approved):
:+1:

Issue #2951, opened 2024-09-24
Temporal.ZonedDateTime.prototype.getTimeZoneTransition: Unnecessary explicit check for "UTC"


anba:
Step 8:
> If IsOffsetTimeZoneIdentifier(timeZone) is true or timeZone is "UTC", return null.

The check `timeZone is "UTC"` isn't necessary, because both `GetNamedTimeZoneNextTransition` and `GetNamedTimeZonePreviousTransition` handle "UTC".

PR #2954, opened 2024-09-25
Editorial batch

ptomato:
A few more minor things, and a fix for correctly uploading to codecov.io so that viewing source works in the codecov reports.

Closes: #2951 (cc @anba)

### DIFF
--- a/spec/plaintime.html
+++ b/spec/plaintime.html
@@ -517,7 +517,7 @@
         1. Let _microseconds_ be _mus2_ - _mus1_.
         1. Let _nanoseconds_ be _ns2_ - _ns1_.
         1. Let _norm_ be NormalizeTimeDuration(_hours_, _minutes_, _seconds_, _milliseconds_, _microseconds_, _nanoseconds_).
-        1. Assert: NormalizedTimeDurationAbs(_norm_).[[TotalNanoseconds]] &lt; nsPerDay.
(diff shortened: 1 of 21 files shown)


ptomato:
Added a larger batch of @anba's requests onto here, since review hasn't started yet.

spec/plaintime.html
```diff
@@ -516,9 +516,7 @@ <h1>
         1. Let _milliseconds_ be _ms2_ - _ms1_.
         1. Let _microseconds_ be _mus2_ - _mus1_.
         1. Let _nanoseconds_ be _ns2_ - _ns1_.
-        1. Let _norm_ be NormalizeTimeDuration(_hours_, _minutes_, _seconds_, _milliseconds_, _microseconds_, _nanoseconds_).
-        1. Assert: NormalizedTimeDurationAbs(_norm_).[[TotalNanoseconds]] &lt; nsPerDay.
```
Ms2ger, line 520: I'm not sure this is correct; `NormalizeTimeDuration` checks `abs(nanoseconds) ≤ maxTimeDuration`, and this is a check against `nsPerDay`, which is much smaller.

Ms2ger (approved):
Thanks!

spec/plaintime.html
ptomato: Oops, you are totally right about this

PR #2964, opened 2024-09-25
Editorial: Avoid GetUTCEpochNanoseconds modification

anba:
`GetUTCEpochNanoseconds` was changed to include a subtraction. There are currently exactly three callers which require this change. All other callers, which are in the majority, don't use this optional offset parameter.

Instead of modifying `GetUTCEpochNanoseconds`, perform this subtraction in the caller. Also change `InterpretISODateTimeOffset` to perform a single call to `GetUTCEpochNanoseconds`.

Also don't bother to µ-optimise `InterpretISODateTimeOffset` to avoid calling `GetUTCEpochNanoseconds` when `possibleEpochNs` is empty. This case happens too rarely to bother optimising it.

### DIFF
--- a/polyfill/lib/ecmascript.mjs
+++ b/polyfill/lib/ecmascript.mjs
@@ -1383,9 +1383,8 @@ export function ToTemporalInstant(item) {
     second,
     millisecond,
     microsecond,
-    nanosecond,
-    offsetNanoseconds
-  );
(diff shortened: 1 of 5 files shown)


ptomato (approved):
Thanks!
I'll update the reference code to sync with this.

PR #2977, opened 2024-09-26
Editorial: Correct various value types in GetTemporalUnitValuedOption

anba:
Fixes various issues in GetTemporalUnitValuedOption:
- `allowedStrings` is a list of strings, so it shouldn't be initialised as a copy of `allowedValues`.
- `GetOption` was incorrectly called with `default = UNSET`.
- `default = REQUIRED` can be passed directly to `GetOption`.
- `allowedValues` doesn't contain `UNSET`, so the loop doesn't need to handle that case.


ptomato (approved):
Thanks!

PR #2994, opened 2024-10-03
Editorial: Clarify CalendarResolveFields TypeError vs. RangeError circumstances

### LINKED ISSUE #2993: CalendarResolveFields doesn't specify in which order TypeError and RangeError should be thrown
_(originally posted by @anba in [#2962](https://github.com/tc39/proposal-temporal/issues/2962#issuecomment-2376846508))_

We should probably specify that it generalizes the pattern of [ISOResolveMonth](https://tc39.es/proposal-temporal/#sec-temporal-isoresolvemonth) (first checking to see if unset field(s) warrant a TypeError, then checking values to maybe throw a RangeError).

gibson042:
RangeError is thrown only when non-unset fields can provide sufficient data, which implies that a TypeError was not thrown.

Fixes #2993


ptomato (approved):
LGTM!

PR #3008, opened 2024-10-08
Editorial: Clarify requirement for implementation-defined month-day validation

ptomato:
This is a speculative PR for what I think is probably the best way to address #3002.

Closes: #3002


justingrant (approved):
Love it!

ptomato:
Let's go ahead with this so that Anba can be unblocked and we have a clean slate for the plenary presentation.

PR #3017, opened 2024-10-23
Use `Intl.LocalesArgument` in  type definition of `toLocaleString` methods

fabon-f:
related: https://github.com/js-temporal/temporal-polyfill/pull/224 .

According to the spec of Temporal and ECMA-402, first argument of `toLocaleString` methods can be `Intl.Locale` object (or its array). But the type definition in this repository doesn't match to the spec.

The polyfill already accepts `Intl.Locale` object as locale so I think there is no need to modify it. (Perhaps tests should be added, but I'm not sure where and how to add them.)

### DIFF
--- a/polyfill/index.d.ts
+++ b/polyfill/index.d.ts
@@ -552,7 +552,7 @@ export namespace Temporal {
     subtract(other: Temporal.Duration | DurationLike | string): Temporal.Duration;
     round(roundTo: DurationRoundTo): Temporal.Duration;
     total(totalOf: DurationTotalOf): number;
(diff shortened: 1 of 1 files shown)


ptomato (approved):
Thanks! This seems fine. The TypeScript definition file is not normative, it's just for informational purposes, so it should agree with the spec text.

If you wanted to add tests I'd suggest adding files named something like `accepts-intl-locale.js` in https://github.com/tc39/test262/tree/main/test/intl402/Temporal/Instant/prototype/toLocaleString (and similarly for all the other toLocaleString methods.)

PR #3026, opened 2024-10-28
Editorial: Change TimeDuration to use time durations instead of internal durations

anba:
Changes:
- Move handling for `day` units from `RoundTimeDuration` to `Temporal.Duration.prototype.round`.
- Correct argument type for `DifferenceInstant` to only allow time units.
- `RoundTimeDuration` is now only called with time units and its input and output is a time duration.
- Remove the no rounding fast path from `DifferenceTemporalPlainTime` to match `DifferenceInstant`.

Fixes:
`TemporalDurationFromInternal` is fallible in `Temporal.Duration.prototype.toString`:
```js
Temporal.Duration.from({
  days: 1,
  seconds: 2**53 - 1 - (24*60*60),
  nanoseconds: 999_999_999
}).toString({
  roundingMode: "ceil",
  fractionalSecondDigits: 7
});
```

### DIFF
--- a/polyfill/lib/duration.mjs
+++ b/polyfill/lib/duration.mjs
@@ -38,6 +38,7 @@ import {
   SetSlot,
   TIME_ZONE
 } from './slots.mjs';
+import { TimeDuration } from './timeduration.mjs';
 
 export class Duration {
   constructor(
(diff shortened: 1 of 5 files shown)


ptomato (approved):
Thanks! I made the corresponding changes in the reference code. We should add test262 coverage for the case you identified, but I'll open a separate issue for that.

PR #3027, opened 2024-10-29
Use `Temporal.Now.timeZoneId()` instead of `timeZone()`

printfn:
Found this minor documentation bug, `Temporal.Now.timeZone()` does not exist so it should be `timeZoneId()`.


ptomato (approved):
Thanks!

PR #3037, opened 2024-11-05
Docs: Document hoursInDay and startOfDay when same day starts twice (#2938)

catamorphism:
See https://github.com/tc39/proposal-temporal/issues/2938


ptomato (approved):
Thanks!

PR #3040, opened 2024-11-08
Polyfill: Date.prototype.toTemporalInstant should not be a constructor

ptomato:
This PR updates test262 and adds a fix for a newly-added test that the reference code fails.

### DIFF
--- a/polyfill/lib/legacydate.mjs
+++ b/polyfill/lib/legacydate.mjs
@@ -5,7 +5,18 @@ import { Instant } from './instant.mjs';
 
 import bigInt from 'big-integer';
 
-export function toTemporalInstant() {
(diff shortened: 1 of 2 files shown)


jasonwilliams (approved):
Polyfill change LGTM
Could you explain the issue with the `staging/Intl402/Temporal/old/non-iso-calendars.js` test? Is it related to the polyfill change?

Ms2ger (approved):
Thanks!

Ms2ger:
> Polyfill change LGTM Could you explain the issue with the `staging/Intl402/Temporal/old/non-iso-calendars.js` test? Is it related to the polyfill change?

No, it's because of a change to test262 that's being pulled in along with the test for this polyfill change

PR #3042, opened 2024-11-12
Fix typos ('milliecond'/'millieconds')

catamorphism:


ptomato (approved):
Apparently I like to leave out the "S" here. Thanks!

PR #3046, opened 2024-11-21
Editorial: Fix IsPartialTemporalObject parameter name

trflynn89:
It is referred to as `value` in the subsequent description and steps.


ptomato (approved):
Thanks!

PR #3052, opened 2024-12-05
Editorial: Fix next batch of editorial issues

ptomato:
This closes several issues spotted by @anba and @trflynn89, and also brings in the latest commits from test262.


Ms2ger (approved):
Thanks!

PR #3054, opened 2024-12-09
Normative: Reject too large dates in ToTemporalMonthDay

anba:
Follow-up to #3008 / #3002 to reject too large dates.


ptomato:
Thanks. We'll discuss this in the next champions meeting.

ptomato (changes requested):
Could we make the diff to be presented as minimal as possible? Ideally just the two edits in plainmonthday.html. Since the PlainYearMonth change is unobservable, I'd be happy to accept a PR for it beforehand.

ptomato:
Tests are in https://github.com/tc39/test262/pull/4389.

ptomato:
Approved in the TC39 plenary of 2025-02-18. I'll pull in the test262 update before merging this.

PR #3058, opened 2024-12-09
Sync amendments to Intl-402

anba:


spec/intl.html
```diff
@@ -1160,7 +854,7 @@ <h1>
               1. Set _relevantFieldsEqual_ to *false*.
         1. If _relevantFieldsEqual_ is *true*, then
           1. Let _collapsedResult_ be a new empty List.
-          1. Let _resultParts_ be <del>!</del> FormatDateTimePattern(_dateTimeFormat_, _format_, _pattern_, _x_).
+          1. Let _resultParts_ be FormatDateTimePattern(_dateTimeFormat_, _format_, _pattern_, _x_).
```
ptomato, line 857:
```suggestion
          1. Let _resultParts_ be FormatDateTimePattern(_dateTimeFormat_, _format_, _pattern_, _xEpochNanoseconds_).
```

spec/intl.html
```diff
@@ -1174,7 +868,7 @@ <h1>
             1. Let _z_ be _x_.
           1. Else,
             1. Let _z_ be _y_.
```
ptomato, line 870:
```suggestion
            1. Let _z_ be _xEpochNanoseconds_.
          1. Else,
            1. Let _z_ be _yEpochNanoseconds_.
```

ptomato (changes requested):
Thanks! I believe you missed a couple of name changes, otherwise looks good.

spec/intl.html
```diff
@@ -1578,55 +1272,6 @@ <h1>Intl.DateTimeFormat.prototype.formatRangeToParts ( _startDate_, _endDate_ )<
...
-        1. <ins>Let _format_ be _dtf_.[[DateTimeFormat]].</ins>
-        1. Let _options_ be OrdinaryObjectCreate(%Object.prototype%).
-        1. For each row of <emu-xref href="#table-datetimeformat-resolvedoptions-properties"></emu-xref>, except the header row, in table order, do
-          1. Let _p_ be the Property value of the current row.
-          1. If there is an Internal Slot value in the current row, then
-            1. <del>Let _v_ be the value of _dtf_'s internal slot whose name is the Internal Slot value of the current row.</del>
-            1. <ins>If _p_ is neither *"hourCycle"* nor *"hour12"*, or if _format_ has a field [[hour]], then</ins>
-              1. <ins>Let _v_ be the value of _dtf_'s internal slot whose name is the Internal Slot value of the current row.</ins>
-            1. <ins>Else,</ins>
-              1. <ins>Let _v_ be *undefined*.</ins>
```
ptomato, line 1601: I took another look at this today, with an eye to merging it. But I noticed that this change hasn't been incorporated upstream yet: https://tc39.es/ecma402/#sec-intl.datetimeformat.prototype.resolvedoptions
Could you take another look at it?

anba: This case is handled in [CreateDateTimeFormat](https://tc39.es/ecma402/#sec-createdatetimeformat), step 44:
> 44. If bestFormat has a field [[hour]], then
>        a. Set dateTimeFormat.[[HourCycle]] to hc.

Which means the Temporal change to remove that step needs to be removed.

ptomato: Got it. Thanks!

ptomato (approved):
Approving based on the above comments.

PR #3073, opened 2025-01-14
Editorial and docs updates

ptomato:
- Use "RFC 9557 string" where appropriate
- Use "available time zone identifier" where appropriate
- Update test262


docs/plaindate.md
```diff
@@ -71,13 +71,13 @@ If the value is any other object, it:
 - Must have either a number `month` property or a string `monthCode` property.
 - May have a `calendar` property. If omitted, the [ISO 8601 calendar](https://en.wikipedia.org/wiki/ISO_8601#Dates) will be used by default.
 
-If the value is not an object, it must be a string, which is expected to be in ISO 8601 format.
+If the value is not an object, it must be a string, which is expected to be in RFC 9557 format; roughly a superset of ISO 8601 format.
```
justingrant, line 74: This is a little ambiguous. One way to read it is that a full RFC 9557 string is required, but only the date portion is required. 

Maybe this sentence could be clearer, something like this:

> ...which is expected to be in RFC 9557 format, or a subset of that format that includes at least the date portion. Note that ISO 8601 and RFC 3339 formats are also subsets of RFC 9557, so will also work, except as noted below.

docs/plaindate.md
```diff
@@ -71,13 +71,13 @@ If the value is any other object, it:
...
 Any time part is optional and will be ignored.
 Time zone or UTC offset information will also be ignored, with one exception: if a string contains a `Z` in place of a numeric UTC offset, then a `RangeError` will be thrown because interpreting these strings as a local date and time is usually a bug. `Temporal.Instant.from` should be used instead to parse these strings, and the result's `toZonedDateTimeISO` method can be used to obtain a timezone-local date and time.
 
 In unusual cases of needing date or time components of `Z`-terminated timestamp strings (e.g. daily rollover of a UTC-timestamped log file), use the time zone `'UTC'`. For example, the following code returns a "UTC date": `Temporal.Instant.from(item).toZonedDateTimeISO('UTC').toPlainDate()`.
 
-If the string isn't valid according to ISO 8601, then a `RangeError` will be thrown regardless of the value of `overflow`.
+If the string isn't valid according to RFC 9557, then a `RangeError` will be thrown regardless of the value of `overflow`.
```
justingrant, line 80: > If the string isn't a valid subset of the RFC 9557 format....

docs/plaindate.md
justingrant: Might also be nice to give an example here, e.g. 

> ...which is expected to be in RFC 9557 format, or a subset of that format that includes at least the date portion, for example `"2024-12-25"`. Note that ISO 8601 and RFC 3339 formats are also subsets of RFC 9557, so will also work, except as noted below.  Also note that Temporal accepts a few rare cases that RFC9557 does not, like the use of 6-digit years.  For more info, see {link here}.

docs/plaindate.md
```diff
@@ -641,13 +641,14 @@ date.equals(date); // => true
...
 
-**Returns:** a string in the ISO 8601 date format representing `date`.
+**Returns:** a string in the RFC 9557 date format representing `date`.
 
 This method overrides the `Object.prototype.toString()` method and provides a convenient, unambiguous string representation of `date`.
 The string can be passed to `Temporal.PlainDate.from()` to create a new `Temporal.PlainDate` object.
 
 Normally, a calendar annotation is shown when `date`'s calendar is not the ISO 8601 calendar.
 By setting the `calendarName` option to `'always'` or `'never'` this can be overridden to always or never show the annotation, respectively.
+With `calendarName: 'never'` the returned string will additionally be valid in the ISO 8601 date format.
```
justingrant, line 651: > ISO 8601 and RFC 3339 date formats

docs/timezone.md
```diff
@@ -21,7 +21,7 @@ Every wall-clock time is defined using a **UTC Offset**: the amount of exact tim
 For example, on January 19, 2020 in California, the UTC Offset (or "offset" for short) was `-08:00` which means that wall-clock time in San Francisco was 8 hours behind UTC, so 10:00AM locally on that day was 18:00 UTC.
 However the same calendar date and wall-clock time India would have an offset of `+05:30`: 5½ hours later than UTC.
 
-[ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) and [RFC 3339](https://tools.ietf.org/html/rfc3339) define standard representations for exact times as a date and time value, e.g. `2020-09-06T17:35:24.485Z`. The `Z` suffix indicates that this is an exact UTC time.
+[RFC 9557](https://datatracker.ietf.org/doc/html/rfc9557), [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601), and [RFC 3339](https://datatracker.ietf.org/doc/html/rfc3339) define standard representations for exact times as a date and time value, e.g. `2020-09-06T17:35:24.485Z`. The `Z` suffix indicates that this is an exact UTC time.
```
justingrant, line 24: Might be good to have two examples here, one with Z and another with a numeric offset.  Otherwise readers might assume that the Z is the only way to get an exact time.

spec.html
```diff
@@ -16,7 +16,7 @@ <h1>Introduction</h1>
     <li>Strongly-typed objects for dates, times, date/<emu-not-ref>time values</emu-not-ref>, year/month values, month/day values, "zoned" date/<emu-not-ref>time values</emu-not-ref>, and durations</li>
     <li>Immutability for all Temporal objects</li>
     <li>String serialization and interoperability via standardized formats</li>
-    <li>Compliance with industry standards like ISO 8601, RFC 3339, and RFC5545 (iCalendar)</li>
+    <li>Compliance with industry standards like ISO 8601, RFC 9557, RFC 3339, and RFC5545 (iCalendar)</li>
```
justingrant, line 19: Maybe put RFCs first here?

spec/abstractops.html
```diff
@@ -1010,10 +1010,10 @@ <h1>
   </emu-clause>
 
   <emu-clause id="sec-temporal-iso8601grammar">
-    <h1>ISO 8601 grammar</h1>
+    <h1>ISO 8601 / RFC 9557 grammar</h1>
```
justingrant, line 1013: Maybe reverse the order in this section?

spec/intl.html
```diff
@@ -1642,7 +1642,7 @@ <h1><a href="https://tc39.es/ecma402/#sec-properties-of-intl-datetimeformat-inst
       <li>[[Locale]] is a String value with the language tag of the locale whose localization is used for formatting.</li>
       <li>[[Calendar]] is a String value representing the <a href="https://unicode.org/reports/tr35/#UnicodeCalendarIdentifier">Unicode Calendar Identifier</a> used for formatting.</li>
       <li>[[NumberingSystem]] is a String value representing the <a href="https://unicode.org/reports/tr35/#UnicodeNumberSystemIdentifier">Unicode Number System Identifier</a> used for formatting.</li>
-      <li>[[TimeZone]] is a String value used for formatting that is either a time zone identifier from the IANA Time Zone Database or a UTC offset in ISO 8601 extended format.</li>
+      <li>[[TimeZone]] is a String value used for formatting that is either <del>a</del> <ins>an available</ins> time zone identifier from the IANA Time Zone Database or a UTC offset in ISO 8601 extended format.</li>
```
Ms2ger, line 1645: Shouldn't we cut everything after "an available time zone identifier" as redundant?

Ms2ger (approved):
Approve but please check Justin's comments

ptomato:
Comments addressed. I appreciate all the suggestions!

PR #3086, opened 2025-02-11
docs: Add note about MDN being the main API reference going forward

ptomato:
A note in the banner and also on the front page.


justingrant (approved):
Love it!  I just dropped another commit into this PR with a one-word text change and some CSS changes for the header. Feel free to incorporate or ignore!

image

PR #3094, opened 2025-02-18
docs: Update typo of string formatting output

kepae:
The actual output is: `28 février 22`

This example demonstrates date wrap around the era notation, but doesn't note the era. An alternative example might demonstrate subtraction without passing past 0 CE, but for now let's just fix the typo.


ptomato (approved):
Thanks! I'm guessing this was probably a copy-paste error.

PR #3098, opened 2025-02-28
Clarify strict inequality in `getTimeZoneTransition()`

leftmostcat:
While strict inequality of `t` and `epochNanoseconds` can be surmised from behavior (without it, `someDateTime.getTimeZoneTransition("previous") == someDateTime.getTimeZoneTransition("previous").getTimeZoneTransition("previous")`), I think this is a little clearer at a quick glance.

### DIFF
--- a/spec/timezone.html
+++ b/spec/timezone.html
@@ -88,7 +88,7 @@
       <dl class="header">
       </dl>
       <p>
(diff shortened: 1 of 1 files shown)


ptomato (approved):
Agreed, thanks for the wording improvement!

PR #3101, opened 2025-03-14
Polyfill: Fix type definition of Duration.toLocaleString

printfn:
The options argument of `Duration.toLocaleString` is incorrectly typed as `Intl.DateTimeFormatOptions` which is incorrect. For example this code fails typechecking:

```ts
duration.toLocaleString('en', { style: 'digital' })`
```

TypeScript's lib.d.ts does not yet include type definitions for Intl.DurationFormat, so I've added the relevant options manually. This can be removed once TypeScript has shipped support.

### DIFF
--- a/polyfill/index.d.ts
+++ b/polyfill/index.d.ts
@@ -508,6 +508,36 @@ export namespace Temporal {
   };
 
   /**
+   * Options for Intl.DurationFormat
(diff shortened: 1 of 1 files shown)


ptomato (approved):
Thanks!

PR #3108, opened 2025-05-02
Editorial: Clarify definition of "time zone transition"

ptomato:
Closes: #3105

### DIFF
--- a/spec/timezone.html
+++ b/spec/timezone.html
@@ -88,12 +88,13 @@
       <dl class="header">
       </dl>
       <p>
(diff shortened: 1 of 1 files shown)


gibson042 (approved):

justingrant (approved):
is UTC offset transition a `<dfn>` (or whatever ECMarkup calls them) anywhere? If not, should it be?

ptomato:
I probably wouldn't bother making a `<dfn>` as we'd have to do it across ECMA-262 and ECMA-402 as well...

However when I checked where it was used I realized I should have made the same change in GetNamedTimeZonePreviousTransition. Will update.

PR #3112, opened 2025-05-07
Polyfill: Adjust time zone transition search window

ptomato:
For a few time zones, the two-week search window for an interval to bisect is too large, because transitions occur less than two weeks apart. Special-case these time zones.

This allows increasing the default window to 19 days (a bunch of Argentinian time zones have transitions less than 20 days apart, so this seems like a good boundary where we don't have to add too many special cases.) This gives a small performance improvement in the common case.

Closes: #3110


gibson042 (approved):
LGTM

PR #3115, opened 2025-05-25
Editorial: Define and consistently use a "ISO 8601 / RFC 9557 string" term

gibson042:


ptomato:
I think we decided some time ago to refer to "RFC 9557" primarily, and call the grammar "ISO 8601 / RFC 9557". @justingrant Do you remember?

I don't have much of an opinion on this.

ptomato:
> RFC 9557 doesn't cover duration strings.

I am pretty sure that we still refer to any duration strings "ISO 8601" and not "RFC 9557".

justingrant (approved):
LGTM

ptomato:
OK, if there aren't any objections to the new term then let's just merge this.

PR #3117, opened 2025-05-26
Normative: Added Template Literal Type for the ISO8601 Duration strings

mscharl:
I thought it would be nice to add typechecking to the ISO8601 duration strings.

Only the typing has been extended so that the duration string must follow the standardized pattern. Nothing has changed functionally.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -1,8 +1,10 @@
 <!DOCTYPE html>
 <meta charset="utf8">
 <pre class="metadata">
-title: Temporal proposal
+title: Temporal
+status: proposal
 stage: 3
+shortname: &lt;a href="https://github.com/tc39/proposal-temporal"&gt;proposal-temporal&lt;/a&gt;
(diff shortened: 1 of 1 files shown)


justingrant (commented):
Thanks for the PR!  I don't think we're likely to accept this as-is though. The main challenge is that duration strings in userland apps will typically be typed as `string`, which would cause a compiler error in every call to `Duration.from`, requiring users to cast the input in every call.

As a general rule, the type declarations for platform APIs generally use literal types (for both strings and, rarely, for numbers) for enumerated types used for options passed to APIs, like `disambiguation` or `offset` in Temporal. But they typically don't use literal types for the data accepted by APIs.

Note that this decision isn't really ours, it's the TypeScript team that has ownership over the declarations for platform APIs. But my understanding is that they follow the approach described above. @DanielRosenwasser @rbuckton is this correct?

I think we might be open to utility type declarations for the various Temporal types' string formats, as long as `string` is also accepted. But like above, this decision would be up to the TS team so I'd defer to Dan and Ron above.

PR #3122, opened 2025-06-16
Editorial: Use consistent wording for throwing on undefined NewTarget

linusg:
The PlainDate constructor was the only one that already had it written this way, which seems to be the preferred style in upstream ECMA-262.


ptomato (approved):
Looks good, thanks!

PR #3135, opened 2025-07-18
Editorial: Split GetTemporalUnitValuedOption with ValidateTemporalUnitValue

sffc:
Editorial changes split from https://github.com/tc39/proposal-temporal/pull/3130.

### DIFF
--- a/spec/abstractops.html
+++ b/spec/abstractops.html
@@ -584,6 +584,7 @@
       <dd>It validates that the result of GetTemporalUnitValuedOption is <emu-not-ref>covered</emu-not-ref> by the union of _unitGroup_ and _extraValues_.</dd>
     </dl>
     <emu-alg>
+      1. If _value_ is ~unset~, return ~unused~.
       1. If _extraValues_ is present and _extraValues_ contains _value_, return ~unused~.
       1. Let _category_ be the value in the “Category” column of the row of <emu-xref href="#table-temporal-units"></emu-xref> whose “Value” column contains _value_. If there is no such row, throw a *RangeError* exception.
       1. If _category_ is ~date~ and _unitGroup_ is ~date~ or ~datetime~, return ~unused~.
@@ -1864,7 +1865,7 @@
     <emu-alg>
       1. NOTE: The following steps read options and perform independent validation in alphabetical order.
       1. Let _largestUnit_ be ? GetTemporalUnitValuedOption(_options_, *"largestUnit"*, ~unset~).
-      1. Perform ? ValidateTemporalUnitValue(_largestUnit_, _unitGroup_, « ~unset~ »).
+      1. Perform ? ValidateTemporalUnitValue(_largestUnit_, _unitGroup_).
       1. If _largestUnit_ is ~unset~, then
         1. Set _largestUnit_ to ~auto~.
       1. If _disallowedUnits_ contains _largestUnit_, throw a *RangeError* exception.
@@ -1873,7 +1874,7 @@
       1. If _operation_ is ~since~, then
         1. Set _roundingMode_ to NegateRoundingMode(_roundingMode_).
       1. Let _smallestUnit_ be ? GetTemporalUnitValuedOption(_options_, *"smallestUnit"*, ~unset~).
-      1. Perform ? ValidateTemporalUnitValue(_smallestUnit_, _unitGroup_, « ~unset~ »).
+      1. Perform ? ValidateTemporalUnitValue(_smallestUnit_, _unitGroup_).
       1. If _smallestUnit_ is ~unset~, then
         1. Set _smallestUnit_ to _fallbackSmallestUnit_.
       1. If _disallowedUnits_ contains _smallestUnit_, throw a *RangeError* exception.
--- a/spec/abstractops.html
+++ b/spec/abstractops.html
@@ -1864,7 +1864,7 @@
     <emu-alg>
       1. NOTE: The following steps read options and perform independent validation in alphabetical order.
       1. Let _largestUnit_ be ? GetTemporalUnitValuedOption(_options_, *"largestUnit"*, ~unset~).
-      1. Perform ? ValidateTemporalUnitValue(_largestUnit_, _unitGroup_).
+      1. Perform ? ValidateTemporalUnitValue(_largestUnit_, _unitGroup_, « ~unset~ »).
       1. If _largestUnit_ is ~unset~, then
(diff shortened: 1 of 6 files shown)


spec/abstractops.html
```diff
@@ -548,51 +548,54 @@ <h1>
...
-        1. Let _unit_ be the value in the "Value" column of the row.
-        1. If the "Category" column of the row is ~date~ and _unitGroup_ is ~date~ or ~datetime~, append _unit_ to _allowedValues_.
-        1. Else if the "Category" column of the row is ~time~ and _unitGroup_ is ~time~ or ~datetime~, append _unit_ to _allowedValues_.
-      1. If _extraValues_ is present, then
-        1. Set _allowedValues_ to the list-concatenation of _allowedValues_ and _extraValues_.
+        1. Let _singularName_ be the value in the "Singular property name" column of the row.
+        1. Append _singularName_ to _allowedStrings_.
+        1. Let _pluralName_ be the value in the "Plural property name" column of the row.
+        1. Append _pluralName_ to _allowedStrings_.
+      1. NOTE: For each singular Temporal unit name that is contained within _allowedStrings_, the corresponding plural name is also contained within it.
```
ptomato: Since we no longer need to build the list dynamically, it might be clearer to write something like

1. Let _allowedStrings_ be a List of all the values in the "Singular property name" and "Plural property name" columns of Table ___.
1. Append **"auto"** to _allowedStrings_.

Or even write it out in full:

1. Let _allowedStrings_ be « **"auto"**, **"year"**, **"years"**, **"month"**, **"months"**, ... ».

(Although writing it out is verbose and repetitive for not really any benefit)

spec/abstractops.html
```diff
@@ -548,51 +548,54 @@ <h1>
...
       1. Return the value in the "Value" column of <emu-xref href="#table-temporal-units"></emu-xref> corresponding to the row with _value_ in its "Singular property name" or "Plural property name" column.
     </emu-alg>
   </emu-clause>
 
+  <emu-clause id="sec-temporal-validatetemporalunitvaluedoption" type="abstract operation">
+    <h1>
+      ValidateTemporalUnitValue (
+        _value_: a Temporal unit, ~unset~, or ~auto~,
+        _unitGroup_: ~date~, ~time~, or ~datetime~,
+        optional _extraValues_: a List of either Temporal units or ~auto~,
```
ptomato, line 579:
```suggestion
        optional _extraValues_: a List of either Temporal units or ~auto~ or ~unset~,
```

spec/abstractops.html
```diff
@@ -548,51 +548,54 @@ <h1>
...
+      ): either a normal completion containing ~unused~ or a throw completion
+    </h1>
+    <dl class="header">
+      <dt>description</dt>
+      <dd>It validates that the result of GetTemporalUnitValuedOption is <emu-not-ref>covered</emu-not-ref> by the union of _unitGroup_ and _extraValues_.</dd>
+    </dl>
+    <emu-alg>
+      1. If _extraValues_ is present and _extraValues_ contains _value_, return ~unused~.
+      1. Let _category_ be the value in the “Category” column of the row of <emu-xref href="#table-temporal-units"></emu-xref> whose “Value” column contains _value_. If there is no such row, throw a RangeError exception.
```
ptomato: Markup nit:
```suggestion
      1. Let _category_ be the value in the “Category” column of the row of <emu-xref href="#table-temporal-units"></emu-xref> whose “Value” column contains _value_. If there is no such row, throw a *RangeError* exception.
```

ptomato: Also this throws if _value_ is UNSET, unless _extraValues_ explicitly contains UNSET as well. I'm not sure that is intentional... (see next comment)

spec/abstractops.html
```diff
@@ -1864,13 +1867,19 @@ <h1>
     </dl>
     <emu-alg>
       1. NOTE: The following steps read options and perform independent validation in alphabetical order.
-      1. Let _largestUnit_ be ? GetTemporalUnitValuedOption(_options_, *"largestUnit"*, _unitGroup_, ~auto~).
+      1. Let _largestUnit_ be ? GetTemporalUnitValuedOption(_options_, *"largestUnit"*, ~unset~).
+      1. Perform ? ValidateTemporalUnitValue(_largestUnit_, _unitGroup_).
+      1. If _largestUnit_ is ~unset~, then
+        1. Set _largestUnit_ to ~auto~.
```
ptomato, line 1870: ...because, for example, this "_largestUnit_ is UNSET" condition can never be reached. So either we should pass « UNSET » to ValidateTemporalUnitValue, or move the UNSET → AUTO check before ValidateTemporalUnitValue.

spec/abstractops.html
```diff
@@ -548,51 +548,54 @@ <h1>
...
+    <dl class="header">
+      <dt>description</dt>
+      <dd>It validates that the result of GetTemporalUnitValuedOption is <emu-not-ref>covered</emu-not-ref> by the union of _unitGroup_ and _extraValues_.</dd>
+    </dl>
+    <emu-alg>
+      1. If _extraValues_ is present and _extraValues_ contains _value_, return ~unused~.
+      1. Let _category_ be the value in the “Category” column of the row of <emu-xref href="#table-temporal-units"></emu-xref> whose “Value” column contains _value_. If there is no such row, throw a RangeError exception.
+      1. If _category_ is ~date~ and _unitGroup_ is ~date~ or ~datetime~, return ~unused~.
+      1. If _category_ is ~time~ and _unitGroup_ is ~time~ or ~datetime~, return ~unused~.
+      1. Throw a RangeError exception.
```
ptomato:
```suggestion
      1. Throw a *RangeError* exception.
```

ptomato (changes requested):

spec/abstractops.html
sffc: `ValidateTemporalUnitValue` does not mutate `_largestUnit_`.

sffc: Oh sorry I need to add `~unset~` as a valid value to the call to `ValidateTemporalUnitValue`.

sffc: Actually I think I can just always allow `~unset~` here because the option being required is still handled in GetTemporalUnitValuedOption.

spec/abstractops.html
sffc: Done

ptomato (approved):
:shipit: Thanks!

PR #3138, opened 2025-07-21
Editorial: Move regulating and balancing logic into ISODateSurpasses

sffc:
This is a big step toward making CalendarDateUntil be calendar-agnostic (#3136).

I have done my best to verify that the change is fully editorial. I wrote Rust code for the before and after and ran it for all date pairs between 2022 and 2026, and found no differences. You can find my code here:

https://gist.github.com/sffc/2c7811dab688a3d21c3e8a4478a8d1dc

The function named `CalendarDateUntil` is the current algorithm, ported line-by-line, and `CalendarDateUntil4` is the new one. (Versions 2 and 3 are a sneak-peak of what I want to do in follow-up PRs to the spec)

### DIFF
--- a/polyfill/lib/calendar.mjs
+++ b/polyfill/lib/calendar.mjs
@@ -131,7 +131,19 @@ function calendarDateWeekOfYear(id, isoDate) {
   return { week: woy, year: yow };
 }
 
-function ISODateSurpasses(sign, y1, m1, d1, isoDate2) {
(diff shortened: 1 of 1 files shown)


sffc:
Note: the new spec performs operations that may be redundant, like calling `BalanceISOYearMonth` again and again for every field, but it makes it much more clear what the spec logic is actually doing. Someone implementing this can easily inline the code back into CalendarDateUntil and get back the old spec. Specs should prioritize clarity.

PR #3139, opened 2025-07-22
Fix: Allow auto in ValidateTemporalUnitValue

### LINKED ISSUE #3137: How does `ValidateTemporalUnitValue` permit `AUTO`?
With the following script (taken from [test262](https://github.com/tc39/test262/blob/4dafd2158b5982730e49c6ed2abf55db0a4315ad/test/built-ins/Temporal/PlainDate/prototype/since/largestunit-default.js#L18)):
```js
const feb20 = Temporal.PlainDate.from("2020-02-01");
const feb21 = Temporal.PlainDate.from("2021-02-01");
feb21.since(feb20, { largestUnit: "auto" });
```

We will end up in [`GetDifferenceSettings`](https://tc39.es/proposal-temporal/#sec-temporal-getdifferencesettings) with `AUTO` as the `largestUnit`:
```
2. Let largestUnit be ? GetTemporalUnitValuedOption(options, "largestUnit", UNSET).
3. Perform ? ValidateTemporalUnitValue(largestUnit, unitGroup).
```

In [`ValidateTemporalUnitValue`](https://tc39.es/proposal-temporal/#sec-temporal-validatetemporalunitvaluedoption), aren't we then guaranteed to throw a `RangeError`?

```
1. If value is UNSET, return UNUSED.
2. If extraValues is present and extraValues contains value, return UNUSED.
3. Let category be the value in the “Category” column of the row of Table 21 whose “Value” column contains value.
   If there is no such row, throw a RangeError exception.
```

Because `extraValues` is not present, and [Table 21](https://tc39.es/proposal-temporal/#table-temporal-units) does not contain an "auto" row.

sffc:
Fixes #3137

### DIFF
--- a/spec/abstractops.html
+++ b/spec/abstractops.html
@@ -1866,7 +1866,7 @@
     <emu-alg>
       1. NOTE: The following steps read options and perform independent validation in alphabetical order.
       1. Let _largestUnit_ be ? GetTemporalUnitValuedOption(_options_, *"largestUnit"*, ~unset~).
-      1. Perform ? ValidateTemporalUnitValue(_largestUnit_, _unitGroup_).
+      1. Perform ? ValidateTemporalUnitValue(_largestUnit_, _unitGroup_, « ~auto~ »).
       1. If _largestUnit_ is ~unset~, then
         1. Set _largestUnit_ to ~auto~.
       1. If _disallowedUnits_ contains _largestUnit_, throw a *RangeError* exception.


ptomato (approved):
I basically came up with the same fix while in transit to a conference yesterday, so LGTM. Thanks!

(Ideally, we should avoid this in future by implementing any refactors in the reference code as well, at the same time — that would have caught this. But I wanted to get the editorial change merged quickly so that we'd have a clean normative PR to present to TC39. So as always, it's a tradeoff.)

PR #3154, opened 2025-09-19
Polyfill: Speed up largestUnit: month date difference for calendars without leap months

gibson042:
Fixes #3153

Note that checking against test262 will require locally adopting https://github.com/js-temporal/temporal-test262-runner/pull/20, manually creating the missing directory, or a similar action.


ptomato (approved):
Very nice. I had started looking for speedups at a lower level but this sidesteps the whole problem (for calendars with 12-month years, at least.)

PR #3155, opened 2025-09-19
Polyfill: Speed up many-months date addition for calendars without leap months

gibson042:
Ref #3153

This doesn't close the issue because it's limited to calendars with a fixed 12-month year, but I think it helps with those enough to justify.


polyfill/lib/calendar.mjs
```diff
@@ -984,6 +986,11 @@ const nonIsoHelperBase = {
   },
   addCalendar(calendarDate, { years = 0, months = 0, weeks = 0, days = 0 }, overflow, cache) {
     const { year, day, monthCode } = calendarDate;
+    if (Math.abs(months) > 12 && !monthCodeInfo[this.id]?.additionalMonths) {
+      const { quotient, remainder } = bigInt(months).divmod(12);
```
ptomato: Do we really need bigint here? The months should be guaranteed to be less than 2^32.

gibson042: I was reaching for the named `divmod`, but you're right that it's overkill. Replaced and force-pushed.

PR #3156, opened 2025-09-22
Polyfill: Speed up Chinese calendar getMonthList

gibson042:
Noticed in passing.

### DIFF
--- a/polyfill/lib/ecmascript.mjs
+++ b/polyfill/lib/ecmascript.mjs
@@ -30,7 +30,6 @@ import {
   ArrayPrototypeReduce,
   ArrayPrototypeSort,
   DateNow,
-  DatePrototypeGetTime,
   DatePrototypeGetUTCFullYear,
   DatePrototypeGetUTCMonth,
   DatePrototypeGetUTCDate,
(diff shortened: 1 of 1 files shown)


polyfill/lib/calendar.mjs
```diff
@@ -1863,69 +1864,77 @@ const helperChinese = ObjectAssign({}, nonIsoHelperBase, {
...
-      } else {
+    // `Date.UTC(year, monthIndex, days)` automatically rebalances, but maps year [0, 99] to [1900, 1999].
+    // `setUTCDate` also rebalances, but requires a Date instance to mutate.
+    const isoStringFeb1 = toUtcIsoDateString(calendarYear, 2, 1);
+    const getFormatterInput =
+      calendarYear >= 100 || calendarYear < 0
+        ? () => DateUTC(calendarYear, 1, daysPastJan31)
+        : () => Call(DatePrototypeSetUTCDate, new DateCtor(isoStringFeb1), [daysPastJan31]);
+    const updateCalendarFields = () => {
+      const formatterInput = getFormatterInput();
```
ptomato: We already have a workaround for this problem elsewhere in the codebase :smile: 

You could do something like this:

```suggestion
    const updateCalendarFields = () => {
      const formatterInput = ES.GetUTCEpochMilliseconds({ isoDate: { year: calendarYear, month: 1, day: daysPastJan31 }, time: ES.MidnightTimeRecord() });
```

PR #3161, opened 2025-09-26
Polyfill: Speed up months arithmetic in coptic, ethioaa, ethiopic, hebrew calendars

ptomato:
We have an optimization for months arithmetic in calendars where a year is always 12 months. Extend this to cover calendars with years that are not 12-months but that do have a mathematically predictable X-months-in- Y-years cycle.

The Coptic and Orthodox family of calendars always have 13 months per year. The Hebrew calendar always has 235 months per 19 years (the Metonic cycle).

Closes: #3153


gibson042 (approved):
:rocket:

kathleenmotley49-maker (approved):

PR #3162, opened 2025-10-03
Editorial: Add note about hard-coded overflow constrain

sffc:
This is intentional, but it looks like a bug when implementing, so it's worth calling out.

### DIFF
--- a/spec/plaindate.html
+++ b/spec/plaindate.html
@@ -730,7 +730,9 @@
           1. If _sign_ × (_d1_ - _isoDate2_.[[Day]]) > 0, return *true*.
         1. Return *false*.
       </emu-alg>
-      <emu-note>This operation intentionally uses overflow ~constrain~ when regulating the year-month.</emu-note>
+      <emu-note>
+        <p>This operation intentionally uses overflow ~constrain~ when regulating the year-month. As a result, adding the duration returned by `.since` or `.until` to _baseDate_ may cause a *RangeError* exception when _overflow_ is ~reject~.</p>
+      </emu-note>
     </emu-clause>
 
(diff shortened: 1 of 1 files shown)


spec/plaindate.html
```diff
@@ -730,6 +730,7 @@ <h1>
           1. If _sign_ × (_d1_ - _isoDate2_.[[Day]]) > 0, return *true*.
         1. Return *false*.
       </emu-alg>
+      <emu-note>This operation intentionally uses overflow ~constrain~ when regulating the year-month.</emu-note>
```
justingrant: Is it worth explaining why we use `~constrain~` here?

justingrant (approved):
One one minor (and optional) question.

spec/plaindate.html
sffc: I'm not sure I can clearly explain why. It's just a choice we made for how the algorithm works. There are other choices that could have been made.

Do you want to try?

justingrant: I suspect that @ptomato may have more recent knowledge than I do about this.

ptomato: Well it was actually added by @sffc here: https://github.com/tc39/proposal-temporal/pull/3138

It's been too long since I read Shane's proof in the comments of that PR, so I don't have context on it anymore.

sffc: Ok, I think the answer is that `.until` does not know anything about overflow options; only `.add` accepts such a parameter. As a result, `.until` returns a duration that subsequently works with `.add` in overflow constrain mode, but the duration might not work in overflow reject mode.

sffc: I wrote this up. Is it helpful?

PR #3165, opened 2025-10-10
Extend NonISOResolveFields examples to include RangeError for invalid MonthCode

### LINKED ISSUE #3163: Not clear that non-ISO calendars should reject month codes that don't exist
The ISO calendar rejects month codes that aren't M01 through M12 in CalendarResolveFields.

However, I don't see in NonISOResolveFields where this rejection should happen. There are examples of cases that should reject, but this is not one of them, as far as I can see.

I confirmed on Matrix that the champions believe that non-ISO calendars _should_ reject month codes that don't exist in their calendar (like "gregory" rejecting "M13" or "M01L").

gibson042:
Fixes #3163

### DIFF
--- a/spec/calendar.html
+++ b/spec/calendar.html
@@ -1202,8 +1202,8 @@
       <p>The operation throws a *TypeError* exception if the non-~unset~ fields of _fields_ are insufficient to identify a unique instance of _type_ in the calendar (e.g., when at least one field in each combination capable of determining some part of its data is ~unset~) or a *RangeError* exception if the fields are sufficient but their values are internally inconsistent within the calendar (e.g., when fields such as [[Month]] and [[MonthCode]] have conflicting non-~unset~ values). For example:</p>
       <ul>
         <li>If _type_ is ~date~ or ~month-day~ and "day" in the calendar has an interpretation similar to ISO 8601 and _fields_.[[Day]] is ~unset~.</li>
-        <li>If "month" in the calendar has an interpretation similar to ISO 8601 and _fields_.[[MonthCode]] identifies a month code that is not valid in any year of the calendar.</li>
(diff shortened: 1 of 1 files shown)


spec/calendar.html
```diff
@@ -1202,6 +1202,7 @@ <h1>
       <p>The operation throws a *TypeError* exception if the non-~unset~ fields of _fields_ are insufficient to identify a unique instance of _type_ in the calendar (e.g., when at least one field in each combination capable of determining some part of its data is ~unset~) or a *RangeError* exception if the fields are sufficient but their values are internally inconsistent within the calendar (e.g., when fields such as [[Month]] and [[MonthCode]] have conflicting non-~unset~ values). For example:</p>
       <ul>
         <li>If _type_ is ~date~ or ~month-day~ and "day" in the calendar has an interpretation similar to ISO 8601 and _fields_.[[Day]] is ~unset~.</li>
+        <li>If "month" in the calendar has an interpretation similar to ISO 8601 and _fields_.[[MonthCode]] identifies a month code that is not valid in any year of the calendar.</li>
```
sffc:
```suggestion
        <li>If _fields_.[[MonthCode]] identifies a month code that is not valid in any year of the calendar.</li>
```

spec/calendar.html
```diff
@@ -1202,6 +1202,7 @@ <h1>
...
         <li>If "month" in the calendar has an interpretation similar to ISO 8601 and either _fields_.[[Month]] and _fields_.[[MonthCode]] are both ~unset~ or neither value is ~unset~ but they do not identify the same month.</li>
```
ptomato:
```suggestion
        <li>If _fields_.[[Month]] and _fields_.[[MonthCode]] are both ~unset~ or neither value is ~unset~ but they do not identify the same month.</li>
```

PR #3167, opened 2025-10-20
Editorial: Fix some type assertions

ptomato:
A group at the University of Bergen has done some work using an [automated proof assistant](https://rocq-prover.org/) on the type assertions in the spec text, and brought several issues to my attention. This PR fixes those issues.

The proof code is available at https://github.com/bldl/temporal-rocq

I recommend reviewing each commit separately.

Summary of the changes:
- Adding type assertions to constrain UTC offset minutes and nanoseconds to their range of ±1 day (exclusive), and adding a Time Zone Identifier Parse Record to support this
- Fixing a mistake in the type of the [[Days]] field of Time Record
- Bringing the description of ISO Date Records in line with Time Records in terms of where the validity is described.

Thank you to @voiestad and @aria-eide.

### DIFF
--- a/spec/plaindate.html
+++ b/spec/plaindate.html
@@ -571,6 +571,7 @@
       <p>
         An <dfn variants="ISO Date Records">ISO Date Record</dfn> is a Record value used to represent a valid calendar date in the ISO 8601 calendar, although the year may be outside of the allowed range for Temporal.
         ISO Date Records are produced by the abstract operation CreateISODateRecord.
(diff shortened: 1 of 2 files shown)


spec/plaintime.html
aria-eide, line 1: On line 458, `optional _deltaDays_: a non-negative integer` should be changed to `optional _deltaDays_: an integer`

ptomato: Thanks for catching that!

PR #3169, opened 2025-10-22
Add exhaustive test for all date pairs in 4-year cycle

ptomato:
The ISO calendar repeats every 4 years due to leap years (ignoring the exceptions at 100 and 400 years). This test checks the difference between every possible pair of dates in one such 4-year cycle.

Based on the script I wrote in the comment at https://github.com/tc39/proposal-temporal/issues/2535#issuecomment-1806603482

### DIFF
--- a/polyfill/test/thorough/support.mjs
+++ b/polyfill/test/thorough/support.mjs
@@ -828,14 +828,15 @@ export function getProgressBar(start, total) {
       clear: true
     });
   }
+  let printInterval = 10;
+  while (total > 100 * printInterval) printInterval *= 10;
   return new (class FakeProgressBar {
     #done = 0;
 
     tick(delta = 1) {
       this.#done += delta;
-      // Do print _something_ every 1000 tests, so that there is something to
-      // look at while it is in progress.
-      if (delta && this.#done % 1000 === 0) {
+      // Do print _something_ periodically to demonstrate continued progress.
+      if (delta && this.#done % printInterval === 0) {
         const elapsed = Number(nowBigInt() - start) / 1_000_000_000;
         output(`${this.#done} tests completed in ${elapsed.toFixed(1)} seconds.`);
       }


ptomato:
The snapshot file is 85 MB uncompressed, 15 MB compressed with gzip.

polyfill/test/thorough/yearcycle.mjs
```diff
@@ -0,0 +1,62 @@
...
+    d.hours !== 0 ||
+    d.minutes !== 0 ||
+    d.seconds !== 0 ||
+    d.milliseconds !== 0 ||
+    d.microseconds !== 0 ||
+    d.nanoseconds !== 0
+  ) {
+    throw new Error(`Expected zero in all components but years, months, and days; got ${d}`);
+  }
+  return d.toString().replace(/[PD]/g, '');
```
gibson042: Omitting the "D"s seems like one step too far.
```suggestion
  // Omit the initial "P", resulting in strings like "1Y1M1D" and "40D".
  return d.toString().slice(1);
```

polyfill/test/thorough/yearcycle.mjs
```diff
@@ -0,0 +1,62 @@
...
+// Minimize the number of characters stored in each snapshot, since there are
+// three million snapshots. Pack dates and durations into as few ASCII chars as
+// possible while still preserving some sort of human legibility in case a
+// snapshot needs debugging.
+
+function dateToShortString(date) {
+  return `${date.year % 10}${date.monthCode.slice(1)}${`${date.day}`.padStart(2, '0')}`;
+}
+
+function durationToShortString(d) {
```
gibson042: Suggestion: Rename this function to `isoCalendarDurationToShortString`.

polyfill/test/thorough/support.mjs
```diff
@@ -274,14 +274,16 @@ export function getProgressBar(start, total) {
...
   return new (class FakeProgressBar {
     #done = 0;
 
     tick(delta = 1) {
       this.#done += delta;
-      // Do print _something_ every 1000 tests, so that there is something to
-      // look at while it is in progress.
-      if (delta && this.#done % 1000 === 0) {
+      // Do print _something_ periodically, so that there is something to look
+      // at while it is in progress.
```
gibson042:
```suggestion
      // Do print _something_ periodically to demonstrate continued progress.
```

polyfill/test/thorough/yearcycle.mjs
ptomato: Why too far? Without a D it should still be unambiguous, since a number without Y or M can only be a number of days ... and we're talking 4 megabytes of useless ASCII D characters :joy:

ptomato: Ehhh, as long as it's under the large file limit, it's probably fine. If we need to minimize the snapshot sizes, there are a _lot_ more improvements I could make to the other snapshot files.

Issue #3168, opened 2025-10-22
Assertion failure in year-and-a-bit durations relative to leap day


ptomato:
I observed this bug with the reference polyfill:
```js
const d = new Temporal.Duration(1, 0, 0, 0, 1);
const relativeTo = new Temporal.PlainDate(2020, 2, 29);
console.log(d.round({ smallestUnit: 'years', relativeTo }).toString());
console.log(d.total({ unit: 'years', relativeTo }));
```
This fails the following assertion:
https://github.com/tc39/proposal-temporal/blob/857d12db936170c62927aaef468a34dd6f3a0abe/polyfill/lib/ecmascript.mjs#L3261

It occurs because _destEpochNs_ (2021-02-28T01:00Z) > _endEpochNs_ (2021-02-28T00:00Z). If I understand the problem correctly, _startEpochNs_ ought to have the latter value and _endEpochNs_ ought to be 2022-02-28T00:00Z, probably because _r1_ and _r2_ should be 1 and 2 instead of 0 and 1.

I'm hopeful that this is a polyfill-only bug and not a problem with the spec text, because three other implementations (SpiderMonkey, temporal_rs, Fullcalendar polyfill) all print the correct results:
```
P1Y
1.0001141552511414
```
I caught this while writing snapshot tests. This means that it isn't currently covered by test262, or it would have been caught already.

catamorphism:
This triggers a similar assertion failure:

```
const d = new Temporal.Duration(0, 1, 0, 0, 1);
const relativeTo = new Temporal.PlainDateTime(2025, 3, 31);
d.round({smallestUnit: 'month', relativeTo})
```

Same result as the original, but with "month" instead of "year" in the assertion failure message.

PR #3172, opened 2025-10-28
Change NudgeToCalendarUnit to use relative date when comparing durations

catamorphism:
See https://github.com/tc39/proposal-temporal/issues/3168

### DIFF
--- a/polyfill/lib/ecmascript.mjs
+++ b/polyfill/lib/ecmascript.mjs
@@ -3169,40 +3169,34 @@ export function DifferenceZonedDateTime(ns1, ns2, timeZone, calendar, largestUni
   return CombineDateAndTimeDuration(dateDifference, timeDuration);
 }
 
-// Epoch-nanosecond bounding technique where the start/end of the calendar-unit
-// interval are converted to epoch-nanosecond times and destEpochNs is nudged to
-// either one.
-function NudgeToCalendarUnit(
+// Returns: { r1, r2, startEpochNs, endEpochNs, startDuration, endDuration }
+function ComputeNudgeWindow(
   sign,
   duration,
   originEpochNs,
-  destEpochNs,
   isoDateTime,
   timeZone,
   calendar,
   increment,
   unit,
-  roundingMode
+  additionalShift
 ) {
-  // unit must be day, week, month, or year
-  // timeZone may be undefined
-
   // Create a duration with smallestUnit trunc'd towards zero
   // Create a separate duration that incorporates roundingIncrement
   let r1, r2, startDuration, endDuration;
   switch (unit) {
     case 'year': {
       const years = RoundNumberToIncrement(duration.date.years, increment, 'trunc');
-      r1 = years;
-      r2 = years + increment * sign;
+      r1 = !additionalShift ? years : years + increment * sign;
+      r2 = r1 + increment * sign;
       startDuration = { years: r1, months: 0, weeks: 0, days: 0 };
       endDuration = { ...startDuration, years: r2 };
       break;
     }
     case 'month': {
       const months = RoundNumberToIncrement(duration.date.months, increment, 'trunc');
-      r1 = months;
-      r2 = months + increment * sign;
+      r1 = !additionalShift ? months : months + increment * sign;
+      r2 = r1 + increment * sign;
       startDuration = AdjustDateDurationRecord(duration.date, 0, 0, r1);
       endDuration = AdjustDateDurationRecord(duration.date, 0, 0, r2);
       break;
@@ -3240,7 +3234,7 @@ function NudgeToCalendarUnit(
     // If the start of the bound is the same as the "origin" (aka relativeTo),
     // use the origin's epoch-nanoseconds as-is instead of relying on isoDateTime,
     // which then gets zoned and converted back to epoch-nanoseconds,
-    // which looses precision and creates a distorted bounding window.
+    // which loses precision and creates a distorted bounding window.
     startEpochNs = originEpochNs;
   } else {
     const start = CalendarDateAdd(calendar, isoDateTime.isoDate, startDuration, 'constrain');
@@ -3257,9 +3251,85 @@ function NudgeToCalendarUnit(
     ? GetEpochNanosecondsFor(timeZone, endDateTime, 'compatible')
     : GetUTCEpochNanoseconds(endDateTime);
 
+  return { r1, r2, startEpochNs, endEpochNs, startDuration, endDuration };
+}
+
+// Epoch-nanosecond bounding technique where the start/end of the calendar-unit
+// interval are converted to epoch-nanosecond times and destEpochNs is nudged to
+// either one.
+function NudgeToCalendarUnit(
+  sign,
+  duration,
+  originEpochNs,
+  destEpochNs,
+  isoDateTime,
+  timeZone,
+  calendar,
+  increment,
+  unit,
+  roundingMode
+) {
+  // unit must be day, week, month, or year
+  // timeZone may be undefined
+
+  var didExpandCalendarUnit = false;
+  let nudgeWindow = ComputeNudgeWindow(
+    sign,
+    duration,
+    originEpochNs,
+    isoDateTime,
+    timeZone,
+    calendar,
+    increment,
+    unit,
+    false
+  );
+  var { r1, r2, startEpochNs, endEpochNs, startDuration, endDuration } = nudgeWindow;
+
   // Round the smallestUnit within the epoch-nanosecond span
-  if (sign === 1) assert(startEpochNs.leq(destEpochNs) && destEpochNs.leq(endEpochNs), `${unit} was 0 days long`);
-  if (sign === -1) assert(endEpochNs.leq(destEpochNs) && destEpochNs.leq(startEpochNs), `${unit} was 0 days long`);
+  if (sign === 1) {
+    if (!(nudgeWindow.startEpochNs.leq(destEpochNs) && destEpochNs.leq(nudgeWindow.endEpochNs))) {
+      // Retry nudge window if it's out of bounds
+      nudgeWindow = ComputeNudgeWindow(
+        sign,
+        duration,
+        originEpochNs,
+        isoDateTime,
+        timeZone,
+        calendar,
+        increment,
+        unit,
+        true
+      );
+      assert(
+        nudgeWindow.startEpochNs.leq(destEpochNs) && destEpochNs.leq(nudgeWindow.endEpochNs),
+        `${unit} was 0 days long`
+      );
+      didExpandCalendarUnit = true;
+    }
+  } else if (sign == -1) {
+    if (!(nudgeWindow.endEpochNs.leq(destEpochNs) && destEpochNs.leq(nudgeWindow.startEpochNs))) {
+      // Retry nudge window if it's out of bounds
+      nudgeWindow = ComputeNudgeWindow(
+        sign,
+        duration,
+        originEpochNs,
+        isoDateTime,
+        timeZone,
+        calendar,
+        increment,
+        unit,
+        true
+      );
+      assert(
+        nudgeWindow.endEpochNs.leq(destEpochNs) && destEpochNs.leq(nudgeWindow.startEpochNs),
+        `${unit} was 0 days long`
+      );
+      didExpandCalendarUnit = true;
+    }
+  }
+  ({ r1, r2, startEpochNs, endEpochNs, startDuration, endDuration } = nudgeWindow);
+
   assert(!endEpochNs.equals(startEpochNs), 'startEpochNs must ≠ endEpochNs');
   const numerator = TimeDuration.fromEpochNsDiff(destEpochNs, startEpochNs);
   const denominator = TimeDuration.fromEpochNsDiff(endEpochNs, startEpochNs);
@@ -3278,8 +3348,8 @@ function NudgeToCalendarUnit(
   assert(MathAbs(r1) <= MathAbs(total) && MathAbs(total) <= MathAbs(r2), 'r1 ≤ total ≤ r2');
 
   // Determine whether expanded or contracted
-  const didExpandCalendarUnit = roundedUnit === MathAbs(r2);
-  duration = { date: didExpandCalendarUnit ? endDuration : startDuration, time: TimeDuration.ZERO };
+  didExpandCalendarUnit ||= roundedUnit === MathAbs(r2);
+  duration = { date: roundedUnit == MathAbs(r2) ? endDuration : startDuration, time: TimeDuration.ZERO };
 
   const nudgeResult = {
     duration,
--- a/spec/duration.html
+++ b/spec/duration.html
@@ -1564,38 +1564,43 @@
       </emu-table>
     </emu-clause>
 
-    <emu-clause id="sec-temporal-nudgetocalendarunit" type="abstract operation">
+    <emu-clause id="sec-temporal-computenudgewindow" type="abstract operation">
       <h1>
-        NudgeToCalendarUnit (
+        ComputeNudgeWindow (
           _sign_: -1 or 1,
           _duration_: an Internal Duration Record,
           _originEpochNs_: a BigInt,
-          _destEpochNs_: a BigInt,
           _isoDateTime_: an ISO Date-Time Record,
           _timeZone_: an available time zone identifier or ~unset~,
           _calendar_: a calendar type,
           _increment_: a positive integer,
           _unit_: a date unit,
-          _roundingMode_: a rounding mode,
-        ): either a normal completion containing a Record with fields [[NudgeResult]] (a Duration Nudge Result Record) and [[Total]] (a mathematical value), or a throw completion
+          _additionalShift_: a Boolean,
+        ): either a normal completion containing a Record with fields [[R1]] (a mathematical value), [[R2]] (a mathematical value), [[StartEpochNs]] (a BigInt), [[EndEpochNs]] (a BigInt), [[StartDuration]] (an Internal Duration Record), and [[EndDuration]] (an Internal Duration Record), or a throw completion
       </h1>
       <dl class="header">
         <dt>description</dt>
         <dd>
-          It implements rounding a duration to an increment of a calendar unit, relative to a starting point, by calculating the upper and lower bounds of the starting point added to the duration in epoch nanoseconds, and rounding according to which one is closer to _destEpochNs_.
+          It implements calculating the upper and lower bounds of the starting point added to the duration in epoch nanoseconds.
         </dd>
       </dl>
       <emu-alg>
         1. If _unit_ is ~year~, then
           1. Let _years_ be RoundNumberToIncrement(_duration_.[[Date]].[[Years]], _increment_, ~trunc~).
-          1. Let _r1_ be _years_.
-          1. Let _r2_ be _years_ + _increment_ × _sign_.
+          1. If _additionalShift_ is *false*, then
+            1. Let _r1_ be _years_.
+          1. Else,
+            1. Let _r1_ be _years_ + _increment_ × _sign_.
+          1. Let _r2_ be _r1_ + _increment_ × _sign_.
           1. Let _startDuration_ be ? CreateDateDurationRecord(_r1_, 0, 0, 0).
           1. Let _endDuration_ be ? CreateDateDurationRecord(_r2_, 0, 0, 0).
         1. Else if _unit_ is ~month~, then
           1. Let _months_ be RoundNumberToIncrement(_duration_.[[Date]].[[Months]], _increment_, ~trunc~).
-          1. Let _r1_ be _months_.
-          1. Let _r2_ be _months_ + _increment_ × _sign_.
+          1. If _additionalShift_ is *false*, then
+            1. Let _r1_ be _months_.
+          1. Else,
+            1. Let _r1_ be _months_ + _increment_ × _sign_.
+          1. Let _r2_ be _r1_ + _increment_ × _sign_.
           1. Let _startDuration_ be ? AdjustDateDurationRecord(_duration_.[[Date]], 0, 0, _r1_).
           1. Let _endDuration_ be ? AdjustDateDurationRecord(_duration_.[[Date]], 0, 0, _r2_).
         1. Else if _unit_ is ~week~, then
@@ -1632,10 +1637,52 @@
           1. Let _endEpochNs_ be GetUTCEpochNanoseconds(_endDateTime_).
         1. Else,
           1. Let _endEpochNs_ be ? GetEpochNanosecondsFor(_timeZone_, _endDateTime_, ~compatible~).
+        1. Return the Record { [[R1]]: _r1_, [[R2]]: _r2_, [[StartEpochNs]]: _startEpochNs_, [[EndEpochNs]]: _endEpochNs_, [[StartDuration]]: _startDuration_, [[EndDuration]]: _endDuration_ }.
+      </emu-alg>
+    </emu-clause>
+
+    <emu-clause id="sec-temporal-nudgetocalendarunit" type="abstract operation">
+      <h1>
+        NudgeToCalendarUnit (
+          _sign_: -1 or 1,
+          _duration_: an Internal Duration Record,
+          _originEpochNs_: a BigInt,
+          _destEpochNs_: a BigInt,
+          _isoDateTime_: an ISO Date-Time Record,
+          _timeZone_: an available time zone identifier or ~unset~,
+          _calendar_: a calendar type,
+          _increment_: a positive integer,
+          _unit_: a date unit,
+          _roundingMode_: a rounding mode,
+        ): either a normal completion containing a Record with fields [[NudgeResult]] (a Duration Nudge Result Record) and [[Total]] (a mathematical value), or a throw completion
+      </h1>
+      <dl class="header">
+        <dt>description</dt>
+        <dd>
+          It implements rounding a duration to an increment of a calendar unit, relative to a starting point, by calculating the upper and lower bounds of the starting point added to the duration in epoch nanoseconds, and rounding according to which one is closer to _destEpochNs_.
+        </dd>
+      </dl>
+      <emu-alg>
+        1. Let _didExpandCalendarUnit_ be *false*.
+        1. Let _nudgeWindow_ be ? ComputeNudgeWindow(_sign_, _duration_, _originEpochNs_, _isoDateTime_, _timeZone_, _calendar_, _increment_, _unit_, *false*).
+        1. Let _startEpochNs_ be _nudgeWindow_.[[StartEpochNs]].
+        1. Let _endEpochNs_ be _nudgeWindow_.[[EndEpochNs]].
         1. If _sign_ is 1, then
-          1. Assert: _startEpochNs_ ≤ _destEpochNs_ ≤ _endEpochNs_.
+          1. If _startEpochNs_ ≤ _destEpochNs_ ≤ _endEpochNs_ is *false*, then
+            1. Set _nudgeWindow_ to ? ComputeNudgeWindow(_sign_, _duration_, _originEpochNs_, _isoDateTime_, _timeZone_, _calendar_, _increment_, _unit_, *true*).
+            1. Assert: _nudgeWindow_.[[StartEpochNs]] ≤ _destEpochNs_ ≤ _nudgeWindow_.[[EndEpochNs]].
+            1. Set _didExpandCalendarUnit_ to *true*.
         1. Else,
-          1. Assert: _endEpochNs_ ≤ _destEpochNs_ ≤ _startEpochNs_.
+          1. If _endEpochNs_ ≤ _destEpochNs_ ≤ _startEpochNs_ is *false*, then
+            1. Set _nudgeWindow_ to ? ComputeNudgeWindow(_sign_, _duration_, _originEpochNs_, _isoDateTime_, _timeZone_, _calendar_, _increment_, _unit_, *true*).
+            1. Assert: _nudgeWindow_.[[EndEpochNs]] ≤ _destEpochNs_ ≤ _nudgeWindow_.[[StartEpochNs]].
+            1. Set _didExpandCalendarUnit_ to *true*.
+        1. Let _r1_ be _nudgeWindow_.[[R1]].
+        1. Let _r2_ be _nudgeWindow_.[[R2]].
+        1. Set _startEpochNs_ to _nudgeWindow_.[[StartEpochNs]].
+        1. Set _endEpochNs_ to _nudgeWindow_.[[StartEpochNs]].
+        1. Let _startDuration_ be _nudgeWindow_.[[StartDuration]].
+        1. Let _endDuration_ be _nudgeWindow_.[[EndDuration]].
         1. Assert: _startEpochNs_ ≠ _endEpochNs_.
         1. Let _progress_ be (_destEpochNs_ - _startEpochNs_) / (_endEpochNs_ - _startEpochNs_).
         1. Let _total_ be _r1_ + _progress_ × _increment_ × _sign_.
@@ -1649,11 +1696,10 @@
           1. Assert: abs(_r1_) ≤ abs(_total_) &lt; abs(_r2_).
           1. Let _roundedUnit_ be ApplyUnsignedRoundingMode(abs(_total_), abs(_r1_), abs(_r2_), _unsignedRoundingMode_).
         1. If _roundedUnit_ is abs(_r2_), then
-          1. Let _didExpandCalendarUnit_ be *true*.
+          1. Set _didExpandCalendarUnit_ to *true*.
           1. Let _resultDuration_ be _endDuration_.
           1. Let _nudgedEpochNs_ be _endEpochNs_.
         1. Else,
-          1. Let _didExpandCalendarUnit_ be *false*.
           1. Let _resultDuration_ be _startDuration_.
           1. Let _nudgedEpochNs_ be _startEpochNs_.
         1. Set _resultDuration_ to CombineDateAndTimeDuration(_resultDuration_, 0).


ptomato:
So I haven't sat down yet to think through the code but I did run this on my in-progress snapshot tests. It definitely fixes the assertion failures, and the results seem correct:
```js
Temporal.Duration.from({ months: 1, hours: 10 }).round({ smallestUnit: 'months', roundingMode: 'expand', relativeTo: '2020-01-31' })
  // old: assertion failure
  // new: P2M (seems correct)

Temporal.Duration.from({ years: 2345, hours: 12 }).round({ smallestUnit: 'years', roundingMode: 'expand', relativeTo: '2020-02-29' })
  // old: assertion failure
  // new: P2346Y (seems correct)

Temporal.Duration.from({ months: 1, hours: 10 }).total({ unit: 'months', relativeTo: '2020-01-31' })
  // old: assertion failure
  // new: 1.0134408602150538
  // (seems correct: the bounding box is from 2020-02-29T00:00 (relativeTo + 1 month)
  // to 2020-03-31T00:00 (relativeTo + 2 months), which is 744 hours, and the result is
  // the floating point approximation of 1+10/744)
```
(Additionally, these cases produced wrong answers in a build of the existing code with assertions disabled)

But I also found one case that seems to produce a wrong result where the old code was correct:
```js
Temporal.Duration.from({ years: 1 }).round({ smallestUnit: 'months', relativeTo: '2020-02-29' })
  // old: P1Y
  // new: P12M (seems wrong, should balance back up to the implicit largestUnit in the original duration)
```

(I'd have liked to push an in-progress branch with the snapshot tests but currently the testing space is too large, and the snapshots run afoul of GitHub's file size limit. So here's what I've currently got, without snapshots: https://github.com/tc39/proposal-temporal/commit/bd34a53b2597d2f1516167f5bbd27d328748979c To recreate the snapshots, run `node polyfill/test/thorough/durationrounding.mjs -u` and `node polyfill/test/thorough/durationtotal.mjs -u`, then rebase your fixes on top of it and run them again without `-u` to compare)

polyfill/lib/ecmascript.mjs
```diff
@@ -3190,13 +3218,29 @@ function NudgeToCalendarUnit(
...
       const years = RoundNumberToIncrement(duration.date.years, increment, 'trunc');
       r1 = years;
       r2 = years + increment * sign;
       startDuration = { years: r1, months: 0, weeks: 0, days: 0 };
       endDuration = { ...startDuration, years: r2 };
+      cmpResult = compare(CombineDateAndTimeDuration(endDuration, TimeDuration.ZERO), duration);
+      if ((sign > 0 && cmpResult != 1) || (sign < 0 && cmpResult != -1)) {
+        didExpandCalendarUnit = true;
+        r1 = r2;
+        r2 = years + increment * 2 * sign;
```
ptomato: Why is the ×2 necessary here and in the months case below? Previously r1 and r2 would be exactly one rounding increment apart.

ptomato (commented):
@arshaw If you have a moment to weigh in on this, I'd really appreciate it — you probably have a deeper understanding of the NudgeCalendarUnit algorithm than I do.

polyfill/lib/ecmascript.mjs
catamorphism: Because `r1` after assigning `r2` to it is now equal to `years + increment * sign` -- so the `*2` is so that `r1` and `r2` will still be exactly one rounding increment apart.

ptomato: Ah, got it. maybe it'd be clearer to do `r2 = r1 + increment * sign`?

catamorphism: Good point, changed in [ac93d0c](https://github.com/tc39/proposal-temporal/pull/3172/commits/ac93d0c40968ab7ab19d208a5056ac399cd81907)

spec/duration.html
```diff
@@ -1489,6 +1470,43 @@ <h1>
...
+    <emu-clause id="sec-temporal-comparedurations" type="abstract operation">
+      <h1>
+        CompareDurations (
+          _duration1_: an Internal Duration Record,
+          _duration2_: an Internal Duration Record,
+          _zonedRelativeTo_: a Temporal.ZonedDateTime or *undefined*,
+          _plainRelativeTo_: a Temporal.PlainDateTime or *undefined*,
+          _largestUnit1_: a Temporal unit,
+          _largestUnit2_: a Temporal unit,
+        ): either a normal completion containing either *-1*<sub>𝔽</sub>, *+0*<sub>𝔽</sub>, or *1*<sub>𝔽</sub>, or a throw completion
```
ptomato: Nitpick: IIRC it's preferred to use real numbers (1, 0, -1) inside spec algorithms and convert them to 𝔽 at the point where they get sent to JS. But on the other hand, it doesn't matter that much here, since the 𝔽 numbers are not used for any calculations.

(Either way, something has to change because the description refers to the operation returning real numbers)

polyfill/lib/ecmascript.mjs
```diff
@@ -3190,13 +3218,29 @@ function NudgeToCalendarUnit(
   // Create a duration with smallestUnit trunc'd towards zero
   // Create a separate duration that incorporates roundingIncrement
   let r1, r2, startDuration, endDuration;
+  var didExpandCalendarUnit = false;
+  const compare = (d1, d2) => CompareDurations(d1, d2, undefined, ToTemporalDate(isoDateTime.isoDate), unit, unit);
```
ptomato: I missed this in the previous review but this will treat `isoDateTime.isoDate` as a property bag, and create an ISO8601 date even if the calendar is not ISO8601.
```suggestion
  const compare = (d1, d2) => CompareDurations(d1, d2, undefined, CreateTemporalDate(isoDateTime.isoDate, calendar), unit, unit);
```

(But see subsequent comments)

spec/duration.html
```diff
@@ -1586,18 +1604,39 @@ <h1>
         </dd>
       </dl>
       <emu-alg>
+        1. Let _didExpandCalendarUnit_ be *false*.
+        1. Let _date_ be ? ToTemporalDate(_isoDateTime_.[[ISODate]]).
```
ptomato: This line doesn't work - ToTemporalDate takes an ECMAScript language value, and ISO Date Record is a spec value. It should be:
```suggestion
        1. Let _date_ be ! CreateTemporalDate(_isoDateTime_.[[ISODate]], _calendar_).
```

But, in the past we've tried to get rid of creating Temporal objects internally that aren't exposed to user code. So this makes me think we probably need to refactor CompareDurations. Maybe we need to pass _originEpochNs_, _isoDateTime_.[[ISODate]], _timeZone_ , and _calendar_ to get the same info that we'd otherwise get by passing in _zonedRelativeTo_ and _plainRelativeTo_.

This unfortunately probably also means refactoring DateDurationDays to take an ISO Date Record and a calendar type, but that seems like an improvement anyway.

spec/duration.html
```diff
@@ -1586,18 +1604,39 @@ <h1>
...
         1. If _unit_ is ~year~, then
           1. Let _years_ be RoundNumberToIncrement(_duration_.[[Date]].[[Years]], _increment_, ~trunc~).
           1. Let _r1_ be _years_.
           1. Let _r2_ be _years_ + _increment_ × _sign_.
           1. Let _startDuration_ be ? CreateDateDurationRecord(_r1_, 0, 0, 0).
           1. Let _endDuration_ be ? CreateDateDurationRecord(_r2_, 0, 0, 0).
+          1. Let _cmpResult_ be ? CompareDurations(_endDuration_, _duration_, *undefined*, _date_, _unit_, _unit_).
+          1. If _sign_ > 0 and _cmpResult_ != 1 or _sign_ < 0 and _cmpResult_ != -1, then
```
ptomato: Ecmarkup formatting nitpicks
```suggestion
          1. If _sign_ > 0 and _cmpResult_ ≠ 1, or _sign_ &lt; 0 and _cmpResult_ ≠ -1, then
```

spec/duration.html
```diff
@@ -1586,18 +1604,39 @@ <h1>
...
+            1. Set _didExpandCalendarUnit_ to *true*.
+            1. Set _r1_ to _r2_.
+            1. Set _r2_ to _r1_ + _increment_ × _sign_.
+            1. Set _endDuration_ to ? CreateDateDurationRecord(_r2_, 0, 0, 0).
+            1. Set _cmpResult_ to ? CompareDurations(_endDuration_, _duration_, *undefined*, _date_, _unit_, _unit_).
+            1. Assert: _sign_ > 0 and _cmpResult_ is 1 or _sign_ < 0 and _cmpResult_ is -1.
```
ptomato: Ecmarkup formatting and notation nitpicks
```suggestion
            1. Assert: _sign_ > 0 and _cmpResult_ = 1, or _sign_ &lt; 0 and _cmpResult_ = -1.
```

spec/duration.html
```diff
@@ -1586,18 +1604,39 @@ <h1>
...
           1. Let _endDuration_ be ? AdjustDateDurationRecord(_duration_.[[Date]], 0, 0, _r2_).
+          1. Let _cmpResult_ be ? CompareDurations(_endDuration_, _duration_, *undefined*, _date_, _unit_, _unit_).
+          1. If _sign_ > 0 and _cmpResult_ != 1 or _sign_ < 0 and _cmpResult_ != -1, then
+            1. Set _didExpandCalendarUnit_ to *true*.
+            1. Set _r1_ to _r2_.
+            1. Set _r2_ to _r1_ + _increment_ × _sign_.
+            1. Set _endDuration_ to ? AdjustDateDurationRecord(_duration_.[[Date]], 0, 0, _r2_).
+            1. Set _cmpResult_ to ? CompareDurations(_endDuration_, _duration_, *undefined*, _date_, _unit_, _unit_).
+            1. Assert: _sign_ > 0 and _cmpResult_ is 1 or _sign_ < 0 and _cmpResult_ is -1.
+            1. Set _startDuration_ to ? CreateDateDurationRecord(_r1_, 0, 0, 0).
```
ptomato: This line needs to be deleted I think?

spec/duration.html
```diff
@@ -1586,18 +1604,39 @@ <h1>
...
         1. Else if _unit_ is ~month~, then
           1. Let _months_ be RoundNumberToIncrement(_duration_.[[Date]].[[Months]], _increment_, ~trunc~).
           1. Let _r1_ be _months_.
           1. Let _r2_ be _months_ + _increment_ × _sign_.
           1. Let _startDuration_ be ? AdjustDateDurationRecord(_duration_.[[Date]], 0, 0, _r1_).
           1. Let _endDuration_ be ? AdjustDateDurationRecord(_duration_.[[Date]], 0, 0, _r2_).
+          1. Let _cmpResult_ be ? CompareDurations(_endDuration_, _duration_, *undefined*, _date_, _unit_, _unit_).
+          1. If _sign_ > 0 and _cmpResult_ != 1 or _sign_ < 0 and _cmpResult_ != -1, then
```
ptomato: Ecmarkup formatting nitpicks
```suggestion
          1. If _sign_ > 0 and _cmpResult_ ≠ 1, or _sign_ &lt; 0 and _cmpResult_ ≠ -1, then
```

spec/duration.html
```diff
@@ -1586,18 +1604,39 @@ <h1>
...
+            1. Set _didExpandCalendarUnit_ to *true*.
+            1. Set _r1_ to _r2_.
+            1. Set _r2_ to _r1_ + _increment_ × _sign_.
+            1. Set _endDuration_ to ? AdjustDateDurationRecord(_duration_.[[Date]], 0, 0, _r2_).
+            1. Set _cmpResult_ to ? CompareDurations(_endDuration_, _duration_, *undefined*, _date_, _unit_, _unit_).
+            1. Assert: _sign_ > 0 and _cmpResult_ is 1 or _sign_ < 0 and _cmpResult_ is -1.
```
ptomato: Ecmarkup formatting and notation nitpicks
```suggestion
            1. Assert: _sign_ > 0 and _cmpResult_ = 1, or _sign_ &lt; 0 and _cmpResult_ = -1.
```

polyfill/lib/ecmascript.mjs
```diff
@@ -3278,8 +3335,8 @@ function NudgeToCalendarUnit(
   assert(MathAbs(r1) <= MathAbs(total) && MathAbs(total) <= MathAbs(r2), 'r1 ≤ total ≤ r2');
 
   // Determine whether expanded or contracted
-  const didExpandCalendarUnit = roundedUnit === MathAbs(r2);
-  duration = { date: didExpandCalendarUnit ? endDuration : startDuration, time: TimeDuration.ZERO };
+  didExpandCalendarUnit |= roundedUnit === MathAbs(r2);
```
ptomato: I think we want the short-circuiting here?
```suggestion
  didExpandCalendarUnit ||= roundedUnit === MathAbs(r2);
```

spec/duration.html
catamorphism: Fixed in [4aa5f06](https://github.com/tc39/proposal-temporal/pull/3172/commits/4aa5f062635a9b03fd2c5ea7cd1ecbb7f4b43101)

spec/duration.html
catamorphism: Did the refactorings you suggested in [4aa5f06](https://github.com/tc39/proposal-temporal/pull/3172/commits/4aa5f062635a9b03fd2c5ea7cd1ecbb7f4b43101)

spec/duration.html
catamorphism: Fixed in [4aa5f06](https://github.com/tc39/proposal-temporal/pull/3172/commits/4aa5f062635a9b03fd2c5ea7cd1ecbb7f4b43101)

polyfill/lib/ecmascript.mjs
```diff
@@ -3169,6 +3168,39 @@ export function DifferenceZonedDateTime(ns1, ns2, timeZone, calendar, largestUni
...
+  duration2,
+  originEpochNs,
+  timeZone,
+  calendar,
+  isoDate,
+  largestUnit1,
+  largestUnit2
+) {
+  if (
+    originEpochNs &&
```
ptomato: I'd suggest using `if (timeZone &&` for the "zoned relativeTo" condition, since originEpochNs will be defined in some places even when relativeTo was not zoned.

polyfill/lib/ecmascript.mjs
```diff
@@ -3190,13 +3222,29 @@ function NudgeToCalendarUnit(
   // Create a duration with smallestUnit trunc'd towards zero
   // Create a separate duration that incorporates roundingIncrement
   let r1, r2, startDuration, endDuration;
+  var didExpandCalendarUnit = false;
+  const compare = (d1, d2) => CompareDurations(d1, d2, undefined, undefined, calendar, isoDateTime.isoDate, unit, unit);
```
ptomato: Do we not need to pass `originEpochNs` and `timeZone` into this function? (Or is it only supposed to use the PlainRelativeTo comparison by definition?)

spec/duration.html
```diff
@@ -1481,14 +1479,53 @@ <h1>
...
         1. Return _dateDuration_.[[Days]] + _yearsMonthsWeeksInDays_.
       </emu-alg>
     </emu-clause>
 
+    <emu-clause id="sec-temporal-comparedurations" type="abstract operation">
+      <h1>
+        CompareDurations (
+          _duration1_: an Internal Duration Record,
+          _duration2_: an Internal Duration Record,
+          _originEpochNs_: a BigInt or *undefined*,
```
ptomato: Would be consistent with the other parameters to use `~unset~` here instead of `*undefined*`.

spec/duration.html
```diff
@@ -1481,14 +1479,53 @@ <h1>
...
+          _largestUnit2_: a Temporal unit,
+        ): either a normal completion containing either -1, 0, or 1, or a throw completion
+      </h1>
+      <dl class="header">
+        <dt>description</dt>
+        <dd>It compares _duration1_ with _duration2_ and returns -1 if _duration1_ represents a shorter time duration than _duration2_; 0 if the two durations are equal; or 1 if _duration1_ represents a longer time duration than _duration2_, relative to _zonedRelativeTo_ or _plainRelativeTo_.</dd>
+      </dl>
+      <emu-alg>
+        1. If _originEpochNs_ is not *undefined*, and either TemporalUnitCategory(_largestUnit1_) or TemporalUnitCategory(_largestUnit2_) is ~date~, then
+          1. Assert: _timeZone_ is not ~unset~.
```
ptomato:
```suggestion
        1. If _timeZone_ is not ~unset~, and either TemporalUnitCategory(_largestUnit1_) or TemporalUnitCategory(_largestUnit2_) is ~date~, then
```
(to align with my comment on the polyfill above)

polyfill/lib/ecmascript.mjs
catamorphism: I haven't thought of a test case where the time component matters. But there might be one.

polyfill/lib/ecmascript.mjs
catamorphism: Fixed in [ec89346](https://github.com/tc39/proposal-temporal/pull/3172/commits/ec8934655e9d3d0f350e04e75bbf104e6d7cfca2)

spec/duration.html
catamorphism: Fixed in [ec89346](https://github.com/tc39/proposal-temporal/pull/3172/commits/ec8934655e9d3d0f350e04e75bbf104e6d7cfca2)

spec/duration.html
```diff
@@ -1481,14 +1479,53 @@ <h1>
...
         1. Return _dateDuration_.[[Days]] + _yearsMonthsWeeksInDays_.
       </emu-alg>
     </emu-clause>
 
+    <emu-clause id="sec-temporal-comparedurations" type="abstract operation">
+      <h1>
+        CompareDurations (
+          _duration1_: an Internal Duration Record,
+          _duration2_: an Internal Duration Record,
+          _originEpochNs_: a BigInt or ~unset~.,
```
ptomato:
```suggestion
          _originEpochNs_: a BigInt or ~unset~,
```

spec/duration.html
```diff
@@ -1481,14 +1479,53 @@ <h1>
...
+          _timeZone_: an available time zone identifier or ~unset~,
+          _calendar_: a calendar type or ~unset~,
+          _isoDate_: an ISO Date record or ~unset~,
+          _largestUnit1_: a Temporal unit,
+          _largestUnit2_: a Temporal unit,
+        ): either a normal completion containing either -1, 0, or 1, or a throw completion
+      </h1>
+      <dl class="header">
+        <dt>description</dt>
+        <dd>It compares _duration1_ with _duration2_ and returns -1 if _duration1_ represents a shorter time duration than _duration2_; 0 if the two durations are equal; or 1 if _duration1_ represents a longer time duration than _duration2_, relative to _zonedRelativeTo_ or _plainRelativeTo_.</dd>
```
ptomato:
```suggestion
        <dd>It compares _duration1_ with _duration2_ and returns -1 if _duration1_ represents a shorter time duration than _duration2_; 0 if the two durations are equal; or 1 if _duration1_ represents a longer time duration than _duration2_, relative to the starting point given by _originEpochNs_, _isoDate_, _timeZone_, and _calendar_.</dd>
```

polyfill/lib/ecmascript.mjs
ptomato: Probably for consistency we should pass them in. (Otherwise it won't follow the same code path as Duration.compare I think?)

catamorphism:
> Probably for consistency we should pass them in. (Otherwise it won't follow the same code path as Duration.compare I think?)

@ptomato Okay, done in [62e1ed2](https://github.com/tc39/proposal-temporal/pull/3172/commits/62e1ed2ec39fe9dd4f5374d26004c113baa53e85)

spec/duration.html
```diff
@@ -1586,18 +1623,37 @@ <h1>
...
           1. Let _r2_ be _years_ + _increment_ × _sign_.
           1. Let _startDuration_ be ? CreateDateDurationRecord(_r1_, 0, 0, 0).
           1. Let _endDuration_ be ? CreateDateDurationRecord(_r2_, 0, 0, 0).
+          1. Let _cmpResult_ be ? CompareDurations(_endDuration_, _duration_, _originEpochNs_, _timeZone_, _calendar_, _isoDateTime_.[[ISODate]], _unit_, _unit_).
+          1. If _sign_ > 0 and _cmpResult_ ≠ 1, or _sign_ &lt; 0 and _cmpResult_ ≠ -1, then
+            1. Set _didExpandCalendarUnit_ to *true*.
+            1. Set _r1_ to _r2_.
+            1. Set _r2_ to _r1_ + _increment_ × _sign_.
+            1. Set _endDuration_ to ? CreateDateDurationRecord(_r2_, 0, 0, 0).
+            1. Set _cmpResult_ to ? CompareDurations(_endDuration_, _duration_, *undefined*, ~unset~, _calendar_, _isoDateTime_.[[ISODate]], _unit_, _unit_)..
```
ptomato:
```suggestion
            1. Set _cmpResult_ to ? CompareDurations(_endDuration_, _duration_, _originEpochNs_, _timeZone_, _calendar_, _isoDateTime_.[[ISODate]], _unit_, _unit_)..
```

spec/duration.html
```diff
@@ -1586,18 +1623,37 @@ <h1>
...
+            1. Assert: _sign_ > 0 and _cmpResult_ = 1, or _sign_ &lt; 0 and _cmpResult_ = -1.
+            1. Set _startDuration_ to ? CreateDateDurationRecord(_r1_, 0, 0, 0).
         1. Else if _unit_ is ~month~, then
           1. Let _months_ be RoundNumberToIncrement(_duration_.[[Date]].[[Months]], _increment_, ~trunc~).
           1. Let _r1_ be _months_.
           1. Let _r2_ be _months_ + _increment_ × _sign_.
           1. Let _startDuration_ be ? AdjustDateDurationRecord(_duration_.[[Date]], 0, 0, _r1_).
           1. Let _endDuration_ be ? AdjustDateDurationRecord(_duration_.[[Date]], 0, 0, _r2_).
+          1. Let _cmpResult_ be ? CompareDurations(_endDuration_, _duration_, *undefined*, ~unset~, _calendar_, _isoDateTime_.[[ISODate]], _unit_, _unit_).
```
ptomato:
```suggestion
          1. Let _cmpResult_ be ? CompareDurations(_endDuration_, _duration_, _originEpochNs_, _timeZone_, _calendar_, _isoDateTime_.[[ISODate]], _unit_, _unit_).
```

spec/duration.html
```diff
@@ -1586,18 +1623,37 @@ <h1>
...
+          1. If _sign_ > 0 and _cmpResult_ ≠ 1, or _sign_ &lt; 0 and _cmpResult_ ≠ -1, then
+            1. Set _didExpandCalendarUnit_ to *true*.
+            1. Set _r1_ to _r2_.
+            1. Set _r2_ to _r1_ + _increment_ × _sign_.
+            1. Set _endDuration_ to ? AdjustDateDurationRecord(_duration_.[[Date]], 0, 0, _r2_).
+            1. Set _cmpResult_ to ? CompareDurations(_endDuration_, _duration_, *undefined*, ~unset~, _calendar_, _isoDateTime_.[[ISODate]], _unit_, _unit_)..
```
ptomato:
```suggestion
            1. Set _cmpResult_ to ? CompareDurations(_endDuration_, _duration_, _originEpochNs_, _timeZone_, _calendar_, _isoDateTime_.[[ISODate]], _unit_, _unit_)..
```

ptomato (approved):

arshaw:
I finally got a chance to look at this, sorry for the delay. I think it's really good and the same general approach I would have taken, which is to shift the r1->r2 window by one increment if an overflow situation occurs. I feel there might be room for some minor code DRY-ing and optimization however.

I see that an out-of-bounds situation is preemptively detected via `compare > CompareDurations`. This will call `DateDurationDays > CalendarDateAdd`. Regardless of whether an overflow happened, when `end` is computed towards the end of `NudgeToCalendarUnit`, `CalendarDateAdd` will be called again. This second call could be avoided if the prior result where somehow saved.

Also, I see that IF an overflow is detected, the logic for shifting the window by an increment is rather repetitive across the unit:year and unit:month cases. It'd be nice if we could unify that.

So, instead of a new `CompareDurations` abstract operation to achieve unification, I'd propose a different new abstract operation called something like `ComputeNudgeWindow`. It would pull out most of the logic [from here to here](https://github.com/tc39/proposal-temporal/blob/c1f7bd564466ebbdcfc16efbdb9d3d8d9dc379d6/polyfill/lib/ecmascript.mjs#L3192-L3258) and yield a result like `{ r1, r2, startEpochNs, endEpochNs }`. If the `NudgeToCalendarUnit` caller detects that an overflow happens, a "retry" call to `ComputeNudgeWindow` could be called with an additional flag/int to imply an additional shift should occur.

I'm interested to hear of @catamorphism and @ptomato think this refactor would be worthwhile. Unfortunately I don't have time to try this myself, and who knows, maybe this new abstraction makes things look overly confusing after all is said and done, but could be worth a shot.

arshaw:
@catamorphism the new code looks great to me!

I'll leave the nitpicking & bike shedding to @ptomato :)

spec/duration.html
```diff
@@ -1564,38 +1564,42 @@ <h1>Duration Nudge Result Records</h1>
...
           _originEpochNs_: a BigInt,
-          _destEpochNs_: a BigInt,
           _isoDateTime_: an ISO Date-Time Record,
           _timeZone_: an available time zone identifier or ~unset~,
           _calendar_: a calendar type,
           _increment_: a positive integer,
           _unit_: a date unit,
-          _roundingMode_: a rounding mode,
-        ): either a normal completion containing a Record with fields [[NudgeResult]] (a Duration Nudge Result Record) and [[Total]] (a mathematical value), or a throw completion
+          _additionalShift_: either *true* or *false*,
```
ptomato:
```suggestion
          _additionalShift_: a Boolean,
```

spec/duration.html
```diff
@@ -1564,38 +1564,42 @@ <h1>Duration Nudge Result Records</h1>
...
-          It implements rounding a duration to an increment of a calendar unit, relative to a starting point, by calculating the upper and lower bounds of the starting point added to the duration in epoch nanoseconds, and rounding according to which one is closer to _destEpochNs_.
+          It implements calculating the upper and lower bounds of the starting point added to the duration in epoch nanoseconds.
         </dd>
       </dl>
       <emu-alg>
         1. If _unit_ is ~year~, then
           1. Let _years_ be RoundNumberToIncrement(_duration_.[[Date]].[[Years]], _increment_, ~trunc~).
-          1. Let _r1_ be _years_.
-          1. Let _r2_ be _years_ + _increment_ × _sign_.
+          1. If _additionalShift_ is not *true*, then
```
ptomato:
```suggestion
          1. If _additionalShift_ is *false*, then
```

(also below)

spec/duration.html
```diff
@@ -1564,38 +1564,42 @@ <h1>Duration Nudge Result Records</h1>
...
-          _destEpochNs_: a BigInt,
           _isoDateTime_: an ISO Date-Time Record,
           _timeZone_: an available time zone identifier or ~unset~,
           _calendar_: a calendar type,
           _increment_: a positive integer,
           _unit_: a date unit,
-          _roundingMode_: a rounding mode,
-        ): either a normal completion containing a Record with fields [[NudgeResult]] (a Duration Nudge Result Record) and [[Total]] (a mathematical value), or a throw completion
+          _additionalShift_: either *true* or *false*,
+        ): either a normal completion containing a Record with fields [[r1]] (a mathematical value), [[r2]] (a mathematical value), [[startEpochNs_ (a BigInt), _endEpochNs_ (a BigInt), _startDuration_ (an Internal Duration Record), and _endDuration_ (an InternalDurationRecord), or a throw completion
```
ptomato: Record fields are usually initial caps:

```suggestion
        ): either a normal completion containing a Record with fields [[R1]] (a mathematical value), [[R2]] (a mathematical value), [[StartEpochNs]] (a BigInt), [[EndEpochNs]] (a BigInt), [[StartDuration]] (an Internal Duration Record), and [[EndDuration]] (an InternalDurationRecord), or a throw completion
```

(Need to change the field names in the return statement and the caller as well)

spec/duration.html
```diff
@@ -1632,10 +1636,52 @@ <h1>
...
+      <emu-alg>
+        1. Let _didExpandCalendarUnit_ be *false*.
+        1. Let _nudgeWindow_ be ? ComputeNudgeWindow(_sign_, _duration_, _originEpochNs_, _isoDateTime_, _timeZone_, _calendar_, _increment_, _unit_, *false*).
+        1. Let _startEpochNs_ be _nudgeWindow_.[[startEpochNs]].
+        1. Let _endEpochNs_ be _nudgeWindow_.[[endEpochNs]].
         1. If _sign_ is 1, then
-          1. Assert: _startEpochNs_ ≤ _destEpochNs_ ≤ _endEpochNs_.
+          1. If _startEpochNs_ ≤ _destEpochNs_ ≤ _endEpochNs_ is *false*, then
+            1. Set _nudgeWindow_ to ? ComputeNudgeWindow(_sign_, _duration_, _originEpochNs_, _isoDateTime_, _timeZone_, _calendar_, _increment_, _unit_, *true*).
+            1. Assert: _startEpochNs_ ≤ _destEpochNs_ ≤ _endEpochNs_.
```
ptomato:
```suggestion
            1. Assert: _nudgeWindow_.[[StartEpochNs]] ≤ _destEpochNs_ ≤ _nudgeWindow_.[[EndEpochNs]].
```
(also below)

spec/duration.html
```diff
@@ -1632,10 +1636,52 @@ <h1>
...
+            1. Set _didExpandCalendarUnit_ to *true*.
         1. Else,
-          1. Assert: _endEpochNs_ ≤ _destEpochNs_ ≤ _startEpochNs_.
+          1. If _endEpochNs_ ≤ _destEpochNs_ ≤ _startEpochNs_. is *false*, then
+            1. Set _nudgeWindow_ to ? ComputeNudgeWindow(_sign_, _duration_, _originEpochNs_, _isoDateTime_, _timeZone_, _calendar_, _increment_, _unit_, *true*).
+            1. Assert: _endEpochNs_ ≤ _destEpochNs_ ≤ _startEpochNs_.
+            1. Set _didExpandCalendarUnit_ to *true*.
+        1. Let _r1_ be _nudgeWindow_[[r1]].
+        1. Let _r2_ be _nudgeWindow_[[r2]].
```
ptomato:
```suggestion
        1. Let _r1_ be _nudgeWindow_.[[R1]].
        1. Let _r2_ be _nudgeWindow_.[[R2]].
```

spec/duration.html
```diff
@@ -1632,10 +1636,52 @@ <h1>
...
-          1. Assert: _startEpochNs_ ≤ _destEpochNs_ ≤ _endEpochNs_.
+          1. If _startEpochNs_ ≤ _destEpochNs_ ≤ _endEpochNs_ is *false*, then
+            1. Set _nudgeWindow_ to ? ComputeNudgeWindow(_sign_, _duration_, _originEpochNs_, _isoDateTime_, _timeZone_, _calendar_, _increment_, _unit_, *true*).
+            1. Assert: _nudgeWindow_.[[StartEpochNs]] ≤ _destEpochNs_ ≤ _nudgeWindow_.[[EndEpochNs]].
+            1. Set _didExpandCalendarUnit_ to *true*.
         1. Else,
-          1. Assert: _endEpochNs_ ≤ _destEpochNs_ ≤ _startEpochNs_.
+          1. If _endEpochNs_ ≤ _destEpochNs_ ≤ _startEpochNs_. is *false*, then
+            1. Set _nudgeWindow_ to ? ComputeNudgeWindow(_sign_, _duration_, _originEpochNs_, _isoDateTime_, _timeZone_, _calendar_, _increment_, _unit_, *true*).
+            1. Assert: _endEpochNs_ ≤ _destEpochNs_ ≤ _startEpochNs_.
```
ptomato:
```suggestion
            1. Assert: _nudgeWindow_.[[EndEpochNs]] ≤ _destEpochNs_ ≤ _nudgeWindow_.[[StartEpochNs]].
```

spec/duration.html
```diff
@@ -1564,38 +1564,42 @@ <h1>Duration Nudge Result Records</h1>
...
+          1. Else,
+            1. Let _r1_ be _years_ + _increment_ × _sign_.
+          1. Let _r2_ be _r1_ + _increment_ × _sign_.
           1. Let _startDuration_ be ? CreateDateDurationRecord(_r1_, 0, 0, 0).
           1. Let _endDuration_ be ? CreateDateDurationRecord(_r2_, 0, 0, 0).
         1. Else if _unit_ is ~month~, then
           1. Let _months_ be RoundNumberToIncrement(_duration_.[[Date]].[[Months]], _increment_, ~trunc~).
-          1. Let _r1_ be _months_.
-          1. Let _r2_ be _months_ + _increment_ × _sign_.
+          1. If _additionalShift_ is not *true*, then
```
ptomato:
```suggestion
          1. If _additionalShift_ is *false*, then
```

spec/duration.html
```diff
@@ -1632,10 +1636,52 @@ <h1>
...
+        1. Let _endEpochNs_ be _nudgeWindow_.[[EndEpochNs]].
         1. If _sign_ is 1, then
-          1. Assert: _startEpochNs_ ≤ _destEpochNs_ ≤ _endEpochNs_.
+          1. If _startEpochNs_ ≤ _destEpochNs_ ≤ _endEpochNs_ is *false*, then
+            1. Set _nudgeWindow_ to ? ComputeNudgeWindow(_sign_, _duration_, _originEpochNs_, _isoDateTime_, _timeZone_, _calendar_, _increment_, _unit_, *true*).
+            1. Assert: _nudgeWindow_.[[StartEpochNs]] ≤ _destEpochNs_ ≤ _nudgeWindow_.[[EndEpochNs]].
+            1. Set _didExpandCalendarUnit_ to *true*.
         1. Else,
-          1. Assert: _endEpochNs_ ≤ _destEpochNs_ ≤ _startEpochNs_.
+          1. If _endEpochNs_ ≤ _destEpochNs_ ≤ _startEpochNs_. is *false*, then
```
ptomato:
```suggestion
          1. If _endEpochNs_ ≤ _destEpochNs_ ≤ _startEpochNs_ is *false*, then
```

ptomato (approved):
A couple of stragglers from the last round of comments.

PR #3175, opened 2025-11-04
Editorial: Allow formatting with default options when passing 'era' to formatter

catamorphism:
If 'era' is used as a standalone option, it should still trigger default formatting for PlainTime and PlainMonthDay.

See https://github.com/tc39/proposal-temporal/issues/3049 for more details.


catamorphism:
Draft test262 PR: https://github.com/tc39/test262/pull/4634

ptomato (approved):
Thanks!

PR #3179, opened 2025-11-07
Editorial: Remove early return in BubbleRelativeDuration

anba:
`BubbleRelativeDuration` can be called `largestUnit` being a smaller unit than `smallestUnit`, but the algorithm only returns early when both units are equal to each other. Removing the early return and changing the loop condition and loop counter simplifies this operation a bit.

### DIFF
--- a/spec/duration.html
+++ b/spec/duration.html
@@ -1767,12 +1767,11 @@
         </dd>
       </dl>
       <emu-alg>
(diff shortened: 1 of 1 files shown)


ptomato (commented):
This doesn't seem like much of an improvement? The Repeat clause is currently written in such a way that it's obvious and easy to implement with a `for` loop, but this refactor makes it harder since the increment no longer happens at the end of the loop.

Issue #3180, opened 2025-11-07
Is ParseMonthCode in CalendarResolveFields really fallible?


anba:
CalendarResolveFields, steps 1.d and 1.f-g:
> d. Let monthCode be fields.[[MonthCode]].
> ...
> f. Assert: monthCode is a String.
> g. Let parsedMonthCode be ? ParseMonthCode(monthCode).

`fields` is a [Calendar Fields Record](https://tc39.es/proposal-temporal/#sec-temporal-calendar-fields-records), which means `monthCode` is a [month code String](https://tc39.es/proposal-temporal/#sec-temporal-month-codes). But if it is a month code String, how can `ParseMonthCode` fail here?

Issue #3181, opened 2025-11-07
ParseMonthCode: Step 8 is a redundant


anba:
ParseMonthCode, step 8:
> 8. If monthNumber is 0 and isLeapMonth is false, throw a RangeError exception.

This condition can never occur when parsing using the [MonthCode](https://tc39.es/proposal-temporal/#prod-MonthCode) grammar.

PR #3182, opened 2025-11-07
Editorial: Month code spec text improvements

ptomato:
...and align reference code with spec

Closes: #3180, #3181


polyfill/lib/calendar.mjs
```diff
@@ -1761,8 +1761,8 @@ const helperCoptic = ObjectAssign(
 // Anchor is currently the older era to match ethioaa, but should it be the newer era?
 // See https://github.com/tc39/ecma402/issues/534 for discussion.
```
Ms2ger: Looks like this needs updating

PR #3184, opened 2025-11-13
Replace BalanceISODate(Time) and rearrange time zone offset checks

catamorphism:
This takes https://github.com/tc39/proposal-temporal/pull/3014 and adds a commit implementing each spec change in the polyfill, following each of the spec commits.

Some test262 failures were revealed in the process, and the spec is changed accordingly, in separate commits so that it's easy to track what's different from #3014.


ptomato:
Closing, see https://github.com/tc39/proposal-temporal/pull/3014#issuecomment-3643459375

PR #3185, opened 2025-11-14
Polyfill fixups for dealing with calendars

ptomato:
Found these as I wrote tests for Intl Era Monthcode.

### DIFF
--- a/polyfill/lib/calendar.mjs
+++ b/polyfill/lib/calendar.mjs
@@ -1243,7 +1243,9 @@ const helperHebrew = makeNonISOHelper([{ code: 'am', isoEpoch: { year: -3760, mo
     // Given that these can be calculated by counting the number of days in
     // those months, I assume that these DO NOT need to be exposed as
     // Hebrew-only prototype fields or methods.
-    return (7 * year + 1) % 19 < 7;
+    let cycleYear = (7 * year + 1) % 19;
+    if (cycleYear < 0) cycleYear += 19;
+    return cycleYear < 7;
   },
   monthsInYear(calendarDate) {
     return this.inLeapYear(calendarDate) ? 13 : 12;


polyfill/lib/calendar.mjs
```diff
@@ -625,7 +626,9 @@ const nonIsoHelperBase = {
         }
       }
       if (type === 'month') {
-        const matches = Call(RegExpPrototypeExec, /^([0-9]*)(.*?)$/, [value]);
+        // Newer ICU data has some formats with "Mo11" / "Mo9bis" for Chinese
```
justingrant, line 629: Will this change break compatibility with older Node versions?  If so, maybe we should also retain support for the older format too?  This I assume won't matter for the reference polyfill but would make it easier to port this commit to the prod polyfill.

polyfill/lib/calendar.mjs
```diff
@@ -672,6 +675,14 @@ const nonIsoHelperBase = {
         result.era = value;
       }
     }
+    if (hasEra && !result.era) {
+      // ICU bug that neglects to provide an era code for negative eraYear in
+      // the coptic calendar
+      assert(this.id === 'coptic', 'Unable to discern era from Intl.DateTimeFormat.formatToParts in Coptic calendar.');
```
justingrant: In other places where we bump into ICU bugs or limitations, we throw a RangeError. Should we do the same here?

justingrant (approved):
Cool, thanks for catching these. I added a few questions, otherwise looks good.

polyfill/lib/calendar.mjs
ptomato: Sure. I notice I got the error message backwards here anyway, so I'll update this.

polyfill/lib/calendar.mjs
ptomato: The `Mo` part is optional in this regexp, so it shouldn't affect older versions. It looks like the new ICU stuff has been backported to Node 22 anyway, but I did a quick check with Node 20 and it still works.

PR #3192, opened 2025-11-28
Editorial: Add explanatory note to CalendarMonthDayFromFields

catamorphism:
Explain why the ISODateWithinLimits check is necessary.

### DIFF
--- a/spec/calendar.html
+++ b/spec/calendar.html
@@ -762,7 +762,7 @@
       <emu-alg>
         1. Perform ? CalendarResolveFields(_calendar_, _fields_, ~month-day~).
         1. Let _result_ be ? CalendarMonthDayToISOReferenceDate(_calendar_, _fields_, _overflow_).
-        1. Assert: ISODateWithinLimits(_result_).
+        1. Assert: ISODateWithinLimits(_result_) is *true*.
         1. Return _result_.
       </emu-alg>
     </emu-clause>


ptomato (commented):
I didn't clock this earlier but now that I think about it some more, this check should probably turn into an assertion after a table of reference years is added in https://github.com/tc39/proposal-intl-era-monthcode/issues/60. Then the note will no longer be necessary.

Let's hold off on this until that issue is solved and see if this still applies.

catamorphism:
> @catamorphism I think [tc39/proposal-intl-era-monthcode#60](https://github.com/tc39/proposal-intl-era-monthcode/issues/60) has been hashed out sufficiently that it's clear that no far-future or far-past reference years can be returned. So we can go ahead with turning this check into an assertion.

@ptomato Done in [65be65](https://github.com/tc39/proposal-temporal/pull/3192/commits/f65be65c009d659746284667d5f7126d575f93b6)

spec/calendar.html
```diff
@@ -762,10 +762,13 @@ <h1>
       <emu-alg>
         1. Perform ? CalendarResolveFields(_calendar_, _fields_, ~month-day~).
         1. Let _result_ be ? CalendarMonthDayToISOReferenceDate(_calendar_, _fields_, _overflow_).
-        1. If ISODateWithinLimits(_result_) is *false*, throw a *RangeError* exception.
+        1. Assert: ISODateWithinLimits(_result_).
```
ptomato: ECMASpeak doesn't have auto-conversion to boolean :smile: 
```suggestion
        1. Assert: ISODateWithinLimits(_result_) is *true*.
```

PR #3196, opened 2025-12-04
Polyfill: Don't allow offsets of the form +00:0000

catamorphism:
This offset is not allowed because the separator for hours/minutes and the separate for minutes/second must be consistent.

To make this change, I refactored most of the regexps used for time/date parsing to use named groups. This was to make the code easier to follow The change in semantics is found in the `offsetWithParts` and `offset` regexes, which previously made the ':' separator optional in both of these cases, allowing +00:0000 or +0000:00. The fix is to use a disjunction (introducing new regexps optionalMinSecWithSep and optionalMinSecNoSep), to enforce that the separator or lack thereof is consistent.

### DIFF
--- a/polyfill/lib/regex.mjs
+++ b/polyfill/lib/regex.mjs
@@ -103,10 +103,14 @@ const numberWithOptionalFraction = /(\d+)(?:[.,](\d{1,9}))?/;
 
 const durationDate = /(?:(\d+)Y)?(?:(\d+)M)?(?:(\d+)W)?(?:(\d+)D)?/;
 const durationTime = new RegExpCtor(
-  Call(ArrayPrototypeJoin, [
-    `(?:${numberWithOptionalFraction.source}H)?`,
-    `(?:${numberWithOptionalFraction.source}M)?`,
-    `(?:${numberWithOptionalFraction.source}S)?`
-  ])
+  Call(
+    ArrayPrototypeJoin,
+    [
+      `(?:${numberWithOptionalFraction.source}H)?`,
+      `(?:${numberWithOptionalFraction.source}M)?`,
+      `(?:${numberWithOptionalFraction.source}S)?`
+    ],
+    ['']
+  )
 );
 export const duration = new RegExpCtor(`^([+-])?P${durationDate.source}(?:T(?!$)${durationTime.source})?$`, 'i');


polyfill/lib/regex.mjs
```diff
@@ -64,8 +103,8 @@ export const monthday = new RegExpCtor(
   `^(?:--)?(${monthpart.source})-?(${daypart.source})(?:\\[!?${timeZoneID.source}\\])?((?:${annotation.source})*)$`
 );
 
-const fraction = /(\d+)(?:[.,](\d{1,9}))?/;
+const fraction1 = /(\d+)(?:[.,](\d{1,9}))?/;
```
ptomato: I'd prefer `numberWithOptionalFraction` or something like that, to distinguish it better from the `fraction` introduced above.

polyfill/lib/regex.mjs
```diff
@@ -12,12 +12,51 @@ const yearpart = /(?:[+-]\d{6}|\d{4})/;
...
+    ],
+    ['']
+  )
+);
+const sep = /:/;
+const hourminute = new RegExpCtor(`(?<hour>\\d{2})(?:(?:${sep.source})?(?<minute>\\d{2}))?`);
+const second2 = /(?<second>\d{2})/;
+const fraction = /(?:[.,](?<fraction>\d{1,9}))/;
+const secondsfraction = new RegExpCtor(`(?:${sep.source}?${second2.source}${fraction.source}?)`);
+const millismicrosfraction = /(?<millis>\d{2})(?:(?<micros>\d{2}))(?:[.,](?<nanos>\d{1,9}))?/;
```
ptomato: The names don't seem right here, it's 3 digits each of ms and µs and they should come after the decimal separator. Is it possible these capture groups should actually be referring to hours and minutes?

polyfill/lib/regex.mjs
```diff
@@ -12,12 +12,51 @@ const yearpart = /(?:[+-]\d{6}|\d{4})/;
...
+);
+const optionalMinSecNoSepNoCapture = new RegExpCtor(`(?:${minute.source})(?:${second.source}${subseconds.source}?)?`);
+const optionalMinSecNoCapture = new RegExpCtor(
+  `${optionalMinSecWithSepNoCapture.source}|${optionalMinSecNoSepNoCapture.source}`
+);
+export const offsetWithParts = new RegExpCtor(
+  `^(?<sign>${sign.source})(?<hour>${hour.source})(?:${optionalMinSec.source})?$`
+);
+export const offset = new RegExpCtor(
+  `(?<offset>(?:${sign.source})(?:${hour.source})(?:${optionalMinSecNoCapture.source})?)`
```
ptomato: It seems possible that we could just drop the non-capturing alternatives since we are now using named capture groups instead of numbered ones, so extra captures won't mess up the numbering? Or are they necessary to distinguish the hour/minute/second captures in offsets from the hour/minute/second captures in time parts?

polyfill/lib/regex.mjs
catamorphism: Fixed in [32d1133](https://github.com/tc39/proposal-temporal/pull/3196/commits/32d1133171ecaf9b6527190b31bb5875ba85ea73)

polyfill/lib/regex.mjs
catamorphism: Fixed in [32d1133](https://github.com/tc39/proposal-temporal/pull/3196/commits/32d1133171ecaf9b6527190b31bb5875ba85ea73)

polyfill/lib/regex.mjs
catamorphism: Yes, we can drop them. The capture groups did have to be renamed to distinguish the offset captures from the time captures, but I think that's nicer than duplicating these regexps. Fixed in [32d1133](https://github.com/tc39/proposal-temporal/pull/3196/commits/32d1133171ecaf9b6527190b31bb5875ba85ea73)

PR #3198, opened 2025-12-05
Normative: Constrain internal days automatically in PlainYearMonth subtraction

### LINKED ISSUE #3197: Surprising behavior with PlainYearMonth subtraction and constraining
(Initially discovered by @ptomato)
 
In the reference implementation, `Temporal.PlainYearMonth.from({ year: 2024, monthCode: 'M02' }).subtract({ years: 1 }, { overflow: 'reject' })` throws:

```
Uncaught RangeError: value out of range: 1 <= 29 <= 28
    at RejectToRange (file:///home/tjc/proposal-temporal/polyfill/lib/ecmascript.mjs:2784:41)
    at RejectISODate (file:///home/tjc/proposal-temporal/polyfill/lib/ecmascript.mjs:2789:3)
    at Module.RegulateISODate (file:///home/tjc/proposal-temporal/polyfill/lib/ecmascript.mjs:697:7)
    at Object.dateAdd (file:///home/tjc/proposal-temporal/polyfill/lib/calendar.mjs:202:32)
    at CalendarDateAdd (file:///home/tjc/proposal-temporal/polyfill/lib/ecmascript.mjs:1713:46)
    at Module.AddDurationToYearMonth (file:///home/tjc/proposal-temporal/polyfill/lib/ecmascript.mjs:4144:21)
    at Temporal.PlainYearMonth.subtract (file:///home/tjc/proposal-temporal/polyfill/lib/plainyearmonth.mjs:96:15)
```

This behavior is consistent with the spec, because `AddDurationToYearMonth` constructs a full date from a YearMonth and passes it to CalendarDateAdd; the date's day is the last day of the same month (when subtracting). So in this case, the date 2024-02-29 is passed to CalendarDateAdd; because the overflow setting is `reject`, it's rejected because there is no 2023-02-29 in the ISO calendar.

This seems like it would be surprising to users. For YearMonth, `overflow` would be expected to apply to years and months, but not days, since there is no day visible to the user.

catamorphism:
Fixes #3197

### DIFF
--- a/spec/calendar.html
+++ b/spec/calendar.html
@@ -962,7 +962,7 @@
         NonISOMonthDayToISOReferenceDate (
           _calendar_: a calendar type that is not *"iso8601"*,
           _fields_: a Calendar Fields Record,
-          _overflow_: ~constrain~ or ~reject~
+          _overflow_: ~constrain~ or ~reject~,
         ): either a normal completion containing an ISO Date Record or a throw completion
       </h1>
       <dl class="header">


polyfill/lib/ecmascript.mjs
```diff
@@ -691,16 +691,9 @@ function ParseTemporalDurationString(isoString) {
...
-      break;
-    case 'constrain':
-      ({ year, month, day } = ConstrainISODate(year, month, day));
-      break;
-  }
-  return { year, month, day };
+export function RegulateISODate(year, month, day, overflowMonths, overflowDays) {
+  RejectISODate(year, month, day, overflowMonths, overflowDays);
+  return ConstrainISODate(year, month, day);
 }
```
ptomato, line 697: For this function, we don't have to make the distinction between the two types of overflow, because they don't matter in the ISO calendar. So RegulateISODate can just be left as is.

polyfill/lib/ecmascript.mjs
```diff
@@ -2784,9 +2783,13 @@ export function RejectToRange(value, min, max) {
...
-export function RejectISODate(year, month, day) {
-  RejectToRange(month, 1, 12);
-  RejectToRange(day, 1, ISODaysInMonth(year, month));
+export function RejectISODate(year, month, day, overflowMonths, overflowDays) {
+  if (overflowMonths === 'reject') {
+    RejectToRange(month, 1, 12);
+  }
+  if (overflowDays === 'reject') {
+    RejectToRange(day, 1, ISODaysInMonth(year, month));
+  }
```
ptomato, line 2792: Likewise, RejectISODate can just stay as is.

polyfill/lib/calendar.mjs
```diff
@@ -171,12 +171,18 @@ impl['iso8601'] = {
...
+  monthDayToISOReferenceDate(fields, overflowMonths, overflowDays) {
     const referenceISOYear = 1972;
-    const { month, day } = ES.RegulateISODate(fields.year ?? referenceISOYear, fields.month, fields.day, overflow);
+    const { month, day } = ES.RegulateISODate(
+      fields.year ?? referenceISOYear,
+      fields.month,
+      fields.day,
+      overflowMonths,
+      overflowDays
+    );
```
ptomato, line 185: Likewise, I think `monthDayToISOReferenceDate` can take just one overflow parameter since it is only ever called in `lib/ecmascript.js` with `overflow, overflow`.

polyfill/lib/calendar.mjs
```diff
@@ -171,12 +171,18 @@ impl['iso8601'] = {
     }
     ObjectAssign(fields, resolveNonLunisolarMonth(fields, 'iso8601'));
   },
-  dateToISO(fields, overflow) {
-    return ES.RegulateISODate(fields.year, fields.month, fields.day, overflow);
+  dateToISO(fields, overflowMonths, overflowDays) {
```
ptomato: Even `dateToISO` could have just one overflow parameter since it's called from `PlainDate.from` and `PlainYearMonth.from`, and `PlainYearMonth.from` won't pass in a fields object with a `day` property. (Although I don't think the same argument extends to `calendarToIsoDate`, because that's used in more places including in `regulateDate` which is then used in `addMonthsCalendar` which seems relevant to this change..)

ptomato (commented):
This seems like it's going to be a difficult-to-deal-with PR any way we slice it, but here are some suggestions that may make it at least a bit smaller and easier.

Also check the test262 and snapshot test results — are these legitimate failures in the sense that the previous behaviour did actually already have test coverage and those tests need to be updated,  or are they regressions?

It occurred to me that a simpler approach to implementing this might be to introduce a CalendarYearMonthAdd similar to but simpler than CalendarDateAdd. That could go either way, it might result in a bunch of duplication.

polyfill/lib/ecmascript.mjs
catamorphism: It doesn't work with the implementation of RejectISODate being as it is, because there's no way for it to distinguish between the case where it should (effectively) constrain the day despite `overflow` being `reject`, and the case where it should reject the day, without taking an overflow argument.

polyfill/lib/ecmascript.mjs
catamorphism: It needs to take these arguments because they have to be passed to `RejectISODate` (see below).

polyfill/lib/calendar.mjs
catamorphism: Done in [b54e152](https://github.com/tc39/proposal-temporal/pull/3198/commits/b54e1522e87693a0d218edde3516e36025586380)

polyfill/lib/calendar.mjs
catamorphism: Done in [b54e152](https://github.com/tc39/proposal-temporal/pull/3198/commits/b54e1522e87693a0d218edde3516e36025586380)

PR #3199, opened 2025-12-06
Bring polyfill in line with Intl Era Monthcode and update test262 commit

ptomato:
This updates the pinned test262 commit and makes the necessary fixes in the reference polyfill for the tests to pass. (Also fixes a straggler from https://github.com/tc39/proposal-temporal/pull/3175.)

PR #3201, opened 2025-12-09
Thorough tests for ISO-like calendar math

ptomato:
Splits test/thorough/gregorian.mjs into two new test files:

- calendarisolike.mjs tests that the year and month differences for the four ISO-like calendars are equal to the same differences calculated in the ISO calendar.

- calendardaymath.mjs tests that week and day differences for all non-ISO calendars (except, temporarily, 'chinese' and 'dangi') are equal to the same differences calculated in the ISO calendar. (Days are always equivalent, and weeks are always 7 days in all currently supported calendars.)

This is in order to test the simplest invariants in other non-ISO calendars and not just 'gregory'.


ptomato:
Note: we run into a polyfill bug here that is fixed by #3199, so the CI won't pass until that PR has been merged.

PR #3205, opened 2025-12-11
Polyfill: Validate result in UTC fast path in GetPossibleEpochNanoseconds

catamorphism:
The `IsValidEpochNanoseconds` check was previously omitted when the time zone was UTC.

Closes https://github.com/tc39/proposal-temporal/issues/3203


ptomato (approved):
:+1:
This brings it in line with the specification.

PR #3209, opened 2025-12-12
Polyfill: Reject time strings with unbalanced separators

catamorphism:
Previously, strings like "00:000" or "0000:00" were being accepted by PlainTime.from.


ptomato (approved):
Thanks! My goodness this is much easier to deal with now that you refactored it to use named capture groups.

I noticed that there is a pre-existing bug here but I'll open a new issue for that.

PR #3213, opened 2025-12-18
Polyfill: Replace check with assertion in calendarDateWeekOfYear

catamorphism:
(lastRelDow < 0) can never be true, so replace the check with an assertion.


ptomato (approved):
:+1:

PR #3216, opened 2025-12-19
Delete unreachable code

catamorphism:
This code is unreachable because it's inside an `else` clause attached to `if (year !== undefined)`, and `year` equals `calendarDate.year`.


ptomato (approved):
:+1:

PR #3218, opened 2025-12-19
Editorial: Add early return in CalendarDateUntil

catamorphism:
This change makes it possible to change a check to an assertion in the polyfill.

### DIFF
--- a/polyfill/lib/calendar.mjs
+++ b/polyfill/lib/calendar.mjs
@@ -1056,8 +1056,7 @@ const nonIsoHelperBase = {
       case 'year': {
         // Sign is -1 if calendarTwo < calendarOne, 1 if calendarTwo > calendarOne
         const sign = this.compareCalendarDates(calendarTwo, calendarOne);
-        // If dates were equal, this would have been checked in DifferenceTemporalPlainDate /
-        // DifferenceTemporalPlainDateTime / DifferenceTemporalYearMonth
+        // If dates were equal, this would have been checked in CalendarDateUntil
         assert(sign, 'equal dates should have been checked earlier');
         // Take the difference between the years of the two dates
         const diffYears = calendarTwo.year - calendarOne.year;


polyfill/lib/calendar.mjs
```diff
@@ -1056,10 +1056,9 @@ const nonIsoHelperBase = {
       case 'year': {
         // Sign is -1 if calendarTwo < calendarOne, 1 if calendarTwo > calendarOne
         const sign = this.compareCalendarDates(calendarTwo, calendarOne);
-        // If dates are equal, return 0 date duration
-        if (!sign) {
-          return { years: 0, months: 0, weeks: 0, days: 0 };
-        }
+        // If dates were equal, this would have been checked in DifferenceTemporalPlainDate /
+        // DifferenceTemporalPlainDateTime / DifferenceTemporalYearMonth
+        assert(sign);
```
ptomato: Prefer assertions with an assertion message, e.g. `assert(sign, 'equal dates should have been checked earlier')`

catamorphism: Fixed in [c3c93c9](https://github.com/tc39/proposal-temporal/pull/3218/commits/c3c93c96ca9e2de7ca3c9dfe999501a33632e000)

polyfill/lib/calendar.mjs
```diff
@@ -1056,10 +1056,9 @@ const nonIsoHelperBase = {
       case 'year': {
         // Sign is -1 if calendarTwo < calendarOne, 1 if calendarTwo > calendarOne
         const sign = this.compareCalendarDates(calendarTwo, calendarOne);
-        // If dates are equal, return 0 date duration
-        if (!sign) {
-          return { years: 0, months: 0, weeks: 0, days: 0 };
-        }
+        // If dates were equal, this would have been checked in DifferenceTemporalPlainDate /
+        // DifferenceTemporalPlainDateTime / DifferenceTemporalYearMonth
+        assert(sign, 'equal dates should ahve been checked earlier');
```
ptomato:
```suggestion
        assert(sign, 'equal dates should have been checked earlier');
```

polyfill/lib/calendar.mjs
```diff
@@ -1056,10 +1056,9 @@ const nonIsoHelperBase = {
       case 'year': {
         // Sign is -1 if calendarTwo < calendarOne, 1 if calendarTwo > calendarOne
         const sign = this.compareCalendarDates(calendarTwo, calendarOne);
-        // If dates are equal, return 0 date duration
-        if (!sign) {
-          return { years: 0, months: 0, weeks: 0, days: 0 };
-        }
+        // If dates were equal, this would have been checked in DifferenceTemporalPlainDate /
+        // DifferenceTemporalPlainDateTime / DifferenceTemporalYearMonth
```
ptomato:
```suggestion
        // If dates were equal, this would have been checked in CalendarDateUntil
```

polyfill/lib/calendar.mjs
ptomato: Still present, maybe you forgot to push this change?

polyfill/lib/calendar.mjs
ptomato: Ditto

spec/calendar.html
```diff
@@ -630,9 +630,11 @@ <h1>
         </dd>
       </dl>
       <emu-alg>
+        1. Let _sign_ be CompareISODate(_one_, _two_).
+        1. If _sign_ is 0, then
+          1. Return ZeroDateDuration().
         1. If _calendar_ is *"iso8601"*, then
-          1. Let _sign_ be -CompareISODate(_one_, _two_).
-          1. If _sign_ = 0, return ZeroDateDuration().
+          1. Set _sign_ to -sign.
```
ptomato, line 637: This is a bit odd-looking but I'm not sure how else to do it that would make sense.

PR #3220, opened 2025-12-19
Polyfill: Change check to assertion in minMaxMonthLength for Hebrew calendar

catamorphism:
This is called on dates that have already been validated, so it should be an internal error if there is no length for a valid month code.


ptomato (approved):
Makes sense.

PR #3221, opened 2025-12-19
Polyfill: Change check to assertion in adjustCalendarDate

catamorphism:
This should be guaranteed by the previous call to validateCalendarDate.

### DIFF
--- a/polyfill/lib/calendar.mjs
+++ b/polyfill/lib/calendar.mjs
@@ -1373,7 +1373,8 @@ const helperHebrew = makeNonISOHelper([{ code: 'am', isoEpoch: { year: -3760, mo
           // if leap month is before this one, the month index is one more than the month code
           if (this.inLeapYear({ year }) && month >= 6) month++;
           const largestMonth = this.monthsInYear({ year });
-          if (month < 1 || month > largestMonth) throw new RangeErrorCtor(`Invalid monthCode: ${monthCode}`);
+          // The earlier call to validateCalendarDate guarantees this.
+          assert(month >= 1 && month <= largestMonth, `invalid monthCode: ${monthCode}`);
         }
       } else {
         if (overflow === 'reject') {


polyfill/lib/calendar.mjs
```diff
@@ -1373,7 +1373,8 @@ const helperHebrew = makeNonISOHelper([{ code: 'am', isoEpoch: { year: -3760, mo
           // if leap month is before this one, the month index is one more than the month code
           if (this.inLeapYear({ year }) && month >= 6) month++;
           const largestMonth = this.monthsInYear({ year });
-          if (month < 1 || month > largestMonth) throw new RangeErrorCtor(`Invalid monthCode: ${monthCode}`);
+          // The earlier call to validateCalendarDate guarantees this.
+          assert(month >= 1 && month <= largestMonth, `invalid monthCode: ${monthCode}`);
```
ptomato, line 1377: This isn't quite the same as what's guaranteed in `validateCalendarDate`. I see there that the number part of the month code is validated to be between 1 and 13 inclusive, but in a 12-month year I can trigger this assertion:
```js
Temporal.PlainDate.from({ calendar: 'hebrew', year: 5781, monthCode: 'M13', day: 1 })
```

catamorphism: Ah, I see. There's already a test262 test: https://github.com/tc39/test262/blob/main/test/intl402/Temporal/PlainDate/from/invalid-month-codes-hebrew.js

and I see that codecov.io shows this line as green now (not sure why it was red before, as that test was added in Nov. 2025), so I think I can just close this PR.

ptomato: Probably it was red because 5781 is a common year and 5779 is a leap year, which happen to be different code paths. So it still might be worth adding a line to that test file. (I'm not sure why the line is green now.)

catamorphism: Done in https://github.com/tc39/test262/pull/4860

PR #3223, opened 2025-12-19
Polyfill: Change checks to assertions in adjustEras

catamorphism:
It would be an internal error to call adjustEras on invalid eras data, so change all these checks to assertions.

### DIFF
--- a/polyfill/lib/calendar.mjs
+++ b/polyfill/lib/calendar.mjs
@@ -1593,18 +1593,12 @@ const helperIndian = makeNonISOHelper([{ code: 'shaka', isoEpoch: { year: 79, mo
  * ```
  * */
 function adjustEras(eras) {
-  if (eras.length === 0) {
-    throw new RangeErrorCtor('Invalid era data: eras are required');
-  }
-  if (eras.length === 1 && eras[0].reverseOf) {
-    throw new RangeErrorCtor('Invalid era data: anchor era cannot count years backwards');
-  }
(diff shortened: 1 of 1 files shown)


polyfill/lib/calendar.mjs
```diff
@@ -1613,12 +1607,11 @@ function adjustEras(eras) {
...
     (e) => {
       if (e.isAnchor || (!e.anchorEpoch && !e.reverseOf)) {
-        if (anchorEra) throw new RangeErrorCtor('Invalid era data: cannot have multiple anchor eras');
+        assert(!anchorEra, 'Invalid era data: cannot have multiple anchor eras');
         anchorEra = e;
         e.anchorEpoch = { year: e.hasYearZero ? 0 : 1 };
-      } else if (!e.code) {
-        throw new RangeErrorCtor('If era name is blank, it must be the anchor era');
       }
+      assert(e.code, 'Invalid era data: if era name is blank, it must be the anchor era');
```
ptomato: This needs to be inside an `else` I think?

ptomato (changes requested):
One doubt, otherwise LGTM. Thanks!

polyfill/lib/calendar.mjs
catamorphism: Fixed in [1cbde9b](https://github.com/tc39/proposal-temporal/pull/3223/commits/1cbde9bd2330c6300e7928cf6731cc170e40fe02)

PR #3224, opened 2025-12-19
Polyfill: In Chinese calendar getMonthList, change check to assertion

catamorphism:
This is only called on valid dates, so year should always be defined.


ptomato (approved):
This one's a bit less straightforward to prove but I agree with you, I don't _think_ I see any case where `calendarYear` could be undefined.

PR #3226, opened 2025-12-19
Polyfill: Remove extra undefined checks in Duration constructor

catamorphism:
The arguments all default to 0 already.


ptomato (approved):
TIL that you get the default parameter even if you explicitly pass `undefined`. Thanks!

PR #3230, opened 2025-12-20
Polyfill: Check that time is not start-of-day in ToTemporalInstant

catamorphism:
The TemporalInstantString production requires a Time, so `time` will never be `start-of-day` in this code.

### DIFF
--- a/spec/instant.html
+++ b/spec/instant.html
@@ -414,6 +414,7 @@
         1. Assert: Either _parsed_.[[TimeZone]].[[OffsetString]] is not ~empty~ or _parsed_.[[TimeZone]].[[Z]] is *true*, but not both.
         1. If _parsed_.[[TimeZone]].[[Z]] is *true*, let _offsetNanoseconds_ be 0; otherwise, let _offsetNanoseconds_ be ! ParseDateTimeUTCOffset(_parsed_.[[TimeZone]].[[OffsetString]]).
         1. Let _time_ be _parsed_.[[Time]].
+        1. Assert: _time_ is not ~start-of-day~.
         1. Let _balanced_ be BalanceISODateTime(_parsed_.[[Year]], _parsed_.[[Month]], _parsed_.[[Day]], _time_.[[Hour]], _time_.[[Minute]], _time_.[[Second]], _time_.[[Millisecond]], _time_.[[Microsecond]], _time_.[[Nanosecond]] - _offsetNanoseconds_).
         1. Perform ? CheckISODaysRange(_balanced_.[[ISODate]]).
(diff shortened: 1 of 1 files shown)


polyfill/lib/ecmascript.mjs
```diff
@@ -1266,14 +1266,8 @@ export function ToTemporalInstant(item) {
...
-  const {
-    hour = 0,
-    minute = 0,
-    second = 0,
-    millisecond = 0,
-    microsecond = 0,
-    nanosecond = 0
-  } = time === 'start-of-day' ? {} : time;
+  assert(time !== 'start-of-day', 'Instant string must include a time');
+  const { hour = 0, minute = 0, second = 0, millisecond = 0, microsecond = 0, nanosecond = 0 } = time;
```
ptomato:
```suggestion
  const { hour, minute, second, millisecond, microsecond, nanosecond } = time;
```
I could be wrong but I'm reasonably sure `time` should have all the properties at this point.

ptomato (changes requested):
Looks like we can make the same change in the spec text, in [ToTemporalInstant](https://tc39.es/proposal-temporal/#sec-temporal-totemporalinstant) step 6.

catamorphism:
> Looks like we can make the same change in the spec text, in [ToTemporalInstant](https://tc39.es/proposal-temporal/#sec-temporal-totemporalinstant) step 6.

Done in [4392a0a](https://github.com/tc39/proposal-temporal/pull/3230/commits/4392a0ae5d19be3501313d8375fdef0afb537751)

spec/instant.html
```diff
@@ -413,7 +413,7 @@ <h1>
         1. Let _parsed_ be ? ParseISODateTime(_item_, « |TemporalInstantString| »).
         1. Assert: Either _parsed_.[[TimeZone]].[[OffsetString]] is not ~empty~ or _parsed_.[[TimeZone]].[[Z]] is *true*, but not both.
         1. If _parsed_.[[TimeZone]].[[Z]] is *true*, let _offsetNanoseconds_ be 0; otherwise, let _offsetNanoseconds_ be ! ParseDateTimeUTCOffset(_parsed_.[[TimeZone]].[[OffsetString]]).
-        1. If _parsed_.[[Time]] is ~start-of-day~, let _time_ be MidnightTimeRecord(); else let _time_ be _parsed_.[[Time]].
+        1. Let _time_ be _parsed_.[[Time]].
```
ptomato, line 416: Just for good measure:
```suggestion
        1. Let _time_ be _parsed_.[[Time]].
        1. Assert: _time_ is not ~start-of-day~.
```

ptomato (approved):

PR #3232, opened 2025-12-20
Polyfill: Change check to assertion in LeapYear

catamorphism:
As far as I can see, this function shouldn't be called on undefined.

### DIFF
--- a/polyfill/test/validStrings.mjs
+++ b/polyfill/test/validStrings.mjs
@@ -293,7 +293,7 @@ const timeSpec = (extended) =>
 const time = choice(timeSpec(true), timeSpec(false));
 
 function validateDayOfMonth(result, { year, month, day }) {
-  if (year !== undefined && day > ES.ISODaysInMonth(year, month)) throw SyntaxError('retry if bad day of month');
+  if (day > ES.ISODaysInMonth(year ?? 1972, month)) throw SyntaxError('retry if bad day of month');
 }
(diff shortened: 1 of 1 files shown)


polyfill/test/validStrings.mjs
```diff
@@ -293,7 +293,7 @@ const timeSpec = (extended) =>
 const time = choice(timeSpec(true), timeSpec(false));
 
 function validateDayOfMonth(result, { year, month, day }) {
-  if (day > ES.ISODaysInMonth(year, month)) throw SyntaxError('retry if bad day of month');
+  if (year !== undefined && day > ES.ISODaysInMonth(year, month)) throw SyntaxError('retry if bad day of month');
```
ptomato: Minor since this is only in the grammar testing code, but I think this will skip validation for MM-DD strings. Maybe something like this?
```suggestion
  if (day > ES.ISODaysInMonth(year ?? 1972, month)) throw SyntaxError('retry if bad day of month');
```

ptomato (approved):
Change looks good, thanks. Really I'm not sure why that undefined check is there!

polyfill/test/validStrings.mjs
catamorphism: Done in [0adb490](https://github.com/tc39/proposal-temporal/pull/3232/commits/0adb4903978f5f601445b543675ef3696933f56f) (accidentally made the change separately instead of clicking the button).

PR #3234, opened 2025-12-20
Editorial: Change check to assertion in IsValidDuration

catamorphism:
This shouldn't be called with infinity as the value of any field, as far as I can tell.


ptomato (changes requested):
Sorry, clicked wrong review conclusion

PR #3238, opened 2026-01-05
Polyfill: Change check to assertion in sameTemporalType

catamorphism:
This function is only called when isTemporalObject(x) is true and isTemporalObject(x) === isTemporalObject(y). Therefore, change the check for isTemporalObject(x) && isTemporalObject(y) to an assertion.


ptomato (approved):
Thanks!

PR #3241, opened 2026-01-05
Polyfill: Remove check for offset time zone identifiers in toLocaleString

catamorphism:
https://github.com/tc39/ecma402/issues/683 has been resolved, so this check can be removed.


catamorphism:
The accompanying test262 change is https://github.com/tc39/test262/pull/4808

PR #3245, opened 2026-01-08
Editorial: Refactor ISODateSurpasses parallel to NonISODateSurpasses

ptomato:
See https://github.com/tc39/proposal-intl-era-monthcode/pull/101. This changes the general outline of ISODateSurpasses so that it parallels the outline of NonISODateSurpasses in the Intl Era Monthcode proposal as proposed in the PR.

Also refactors the polyfill code, but keeps the optimizations we already had regarding not passing weeks and days to ISODateSurpasses.

~~(Draft; merge if https://github.com/tc39/proposal-intl-era-monthcode/pull/101 is adopted)~~


spec/plaindate.html
```diff
@@ -691,6 +691,38 @@ <h1>
...
     </emu-clause>
 
+    <emu-clause id="sec-temporal-comparesurpasses" type="abstract operation">
+      <h1>
+        CompareSurpasses (
+          _sign_: -1 or 1,
+          _year_: an integer,
+          _monthOrCode_: an integer or a month code,
+          _day_: an integer,
+          _target_: a Calendar Fields Record,
```
trflynn89, line 701: Should this be a `Calendar Date` record? It is invoked with the result of `CalendarISOToDate`, which returns a  `Calendar Date` record rather than `Calendar Fields`

ptomato: Thanks, added to #3264

PR #3250, opened 2026-01-14
Update status of implementations as of early January

megla-tlanghorst:
Chrome has officially shipped now, I've tested that it's present in stable Chrome myself this morning, now that 144 seems to be updating for everyone. And Graal have reportedly finished their implementation and will ship it with their next minor release as well.

> Significant in-the-field experience with shipping implementations, such as that provided by two independent VMs

IIRC this means that Temporal can move to Stage 4 now, right? I don't know about the Test262 results of course, but the we do have 2 big VMs shipping it now in stable without any flags.

> March TC39 Meeting: Stage 4 proposals are incorporated, final semantics are approved, and the new spec version is branched from main. Only editorial changes are accepted from this point forward.

Which would mean that it could be incorporated officially in ECMA Script 2026 in time if I'm not mistaken.


ptomato (approved):
Thanks for updating the table!

PR #3256, opened 2026-01-19
Polyfill: Fix two tests related to date/time formatting

catamorphism:
1. Remove 'timeZoneName' from hasAnyDateTimeOptions. This matches the spec ( https://tc39.es/proposal-temporal/#sec-getdatetimeformat step 14).
2. Add a case to timeAmend to throw an exception for the 'dateStyle' option. This matches the spec
( https://tc39.es/proposal-temporal/#sec-createdatetimeformat step 45c).

This allows the following tests to pass:
* intl402/DateTimeFormat/prototype/{format, formatToParts, formatRange, formatRangeToParts}/temporal-plaindatetime-unsupported-options.js
* intl402/Temporal/PlainTime/prototype/toLocaleString/datestyle-and-timestyle.js


polyfill/lib/intl.mjs
```diff
@@ -555,8 +557,7 @@ function hasAnyDateTimeOptions(originalOptions) {
     hasDateOptions(originalOptions) ||
     hasTimeOptions(originalOptions) ||
     'dateStyle' in originalOptions ||
-    'timeStyle' in originalOptions ||
-    'timeZoneName' in originalOptions
+    'timeStyle' in originalOptions
```
ptomato: I noticed that `hasDateOptions` and `hasTimeOptions` already check for `dateStyle` and `timeStyle`. So I think this function can just collapse to `hasDateOptions(...) || hasTimeOptions(...)`.

ptomato (changes requested):
I dug into this a bit today and found that it requires a bit more refactoring.

You can see in the CI log that intl402/DateTimeFormat/prototype/format/temporal-plaintime-formatting-datetime-style.js unexpectedly fails here. I initially thought this test was incorrect, but by my reading of the spec it is actually correct — `dateStyle` and `timeStyle` together should throw when passed to the `toLocaleString` method of PlainDate, PlainTime, PlainYearMonth, or PlainMonthDay, but should be accepted in the Intl.DateTimeFormat constructor.

I've kind of half-done the refactoring already while trying to figure out what was going on. Let me know if you want to go over it together so you can take it over to finish this PR, or alternatively I can clean up what I have and push it to this PR.

Either way, we'll also need to update the pinned test262 commit to include Anba's modifications to the tests.

PR #3277, opened 2026-02-18
Polyfill: fix Chinese/Dangi calendar at extreme date ranges

jessealama:
Attempt to fix #3081 by using the 19-year Metonic cycle to shift extreme ISO years into ICU4C's safe range before calling Intl.DateTimeFormat, then adjusting the calendar year back. This produces approximate but non-throwing results for dates outside ICU4C's supported range, which should be sufficient for the test262 extreme-dates tests that only check that no error is thrown. Happy to adjust the approach if there's a better way to handle this.

### DIFF
--- a/polyfill/lib/calendar.mjs
+++ b/polyfill/lib/calendar.mjs
@@ -1920,12 +1920,6 @@ const helperJapanese = ObjectAssign(
 const CHINESE_ICU_SAFE_LOW = -29000;
 const CHINESE_ICU_SAFE_HIGH = 70000;
 
-function chineseMetonicOffset(helper, year) {
-  if (!helper.isVulnerableTo70000Bug()) return 0;
-  if (year >= CHINESE_ICU_SAFE_LOW && year <= CHINESE_ICU_SAFE_HIGH) return 0;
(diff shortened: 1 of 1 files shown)


polyfill/lib/calendar.mjs
```diff
@@ -1920,16 +1920,30 @@ const helperJapanese = ObjectAssign(
 const CHINESE_ICU_SAFE_LOW = -29000;
 const CHINESE_ICU_SAFE_HIGH = 70000;
 
-function chineseMetonicOffset(year) {
+function chineseMetonicOffset(helper, year) {
+  if (!helper.isVulnerableTo70000Bug()) return 0;
   if (year >= CHINESE_ICU_SAFE_LOW && year <= CHINESE_ICU_SAFE_HIGH) return 0;
   return MathRound((year - 2000) / 19) * 19;
 }
```
ptomato: Now that we pass the helper in, it would make sense to make this a method of `helperChinese`, and use `this` instead. (It'll be inherited by `helperDangi`)

jessealama: Nice idea. Done.

PR #3278, opened 2026-02-19
Polyfill: Fix Chinese/Dangi calendar crash when Feb 17 doesn't land in month 1

### LINKED ISSUE #3158: Reference code has some weird case in the Chinese calendar in year 7253
Not a huge problem as the `chinese` calendar is only well-defined between ISO years 1900 and 2100 and implementation-approximated outside that range, as per https://tc39.es/proposal-intl-era-monthcode/#table-calendar-types. But since we are using the implementation's approximation taken from Intl.DateTimeFormat, we should at least not crash when handling dates that are covered by that approximation.

Also not a spec problem, as this is all covered by the implementation-approximated language.

Reproducer:
```js
const calendar = 'chinese';
const a = new Temporal.PlainDate(7252, 12, 31, calendar);
const b = new Temporal.PlainDate(7254, 1, 1, calendar);
a.until(b, { largestUnit: 'years' });
```
This crashes:
```
file:///.../proposal-temporal/polyfill/lib/calendar.mjs:1929
    result[oldMonthString].daysInMonth = oldCalendarDay + 30 - calendarDay;
                                       ^

TypeError: Cannot set properties of undefined (setting 'daysInMonth')
```

I did a bit of digging and it's because of the loop in `helperChinese.getMonthList()`. For some reason we have not backed all the way up to `M01` in year 7253 at the beginning of the loop. I tested that it's not a regression from #3156.

I'm opening this issue so I don't forget to investigate this at some later time.

jessealama:
Fixes #3158.

When Feb 17 of the ISO year falls in Chinese month `1bis` (a leap month after month 1) or month `2`, `getMonthList()` crashes because it assumes Feb 17 is either in month `1` or in the previous year's final month.

The fix checks for `2` first (backing up into month 1 or theoretically `1bis`), then for `1bis` (backing up to the previous year's last month), letting the existing `!== '1'` logic advance forward into month 1 as expected.

Known affected years: 7253 (`1bis`), 7625 (`2`). Thanks to @fabon-f for digging into this!

### DIFF
--- a/polyfill/lib/calendar.mjs
+++ b/polyfill/lib/calendar.mjs
@@ -1988,7 +1988,7 @@ const helperChinese = ObjectAssign({}, nonIsoHelperBase, {
 
     // Reuse the same local object for calendar-specific results, starting with
     // a date close to Chinese New Year. Feb 17 will either be in month 1 of the
-    // new year, in a leap month immediately after month 1 (rare), or near the
+    // new year, in month 2 or a leap month after month 1 (rare), or near the
     // end of the previous year's final month.
     let daysPastJan31 = 17;
     const calendarFields = { day: undefined, monthString: undefined, relatedYear: undefined };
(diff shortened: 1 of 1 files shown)


polyfill/lib/calendar.mjs
```diff
@@ -2022,6 +2023,13 @@ const helperChinese = ObjectAssign({}, nonIsoHelperBase, {
 
     // Ensure that we're in the first month.
     updateCalendarFields();
+    if (calendarFields.monthString === '1bis') {
```
fabon-f, line 2032: I think we should do the same when `monthString` is `"2"`:

```javascript
new Intl.DateTimeFormat("en-u-ca-chinese").formatToParts(new Date("7625-02-17"));
// 7625-M02-1 in latest ICU4C

Temporal.PlainDate.from('7624-12-31[u-ca=chinese]')
  .until('7627-01-01[u-ca=chinese]', { largestUnit: 'months' });
// still raise an assertion error in some JS engines
```

PR #3279, opened 2026-02-19
Polyfill: Fix extreme PlainYearMonth crash in non-ISO calendars

### LINKED ISSUE #3251: Polyfill: Exception incorrectly thrown with maximum allowed PlainYearMonth in some non-ISO calendars
```
> Temporal.PlainYearMonth.from ( { calendar: "coptic", year: 275471, era: "am", eraYear: 275471, month: 6, monthCode: "M06"})
Uncaught RangeError: Invalid ISO date: +275760-09-21T00:00Z
    at Object.getCalendarParts (file:///home/tjc/proposal-temporal/polyfill/lib/calendar.mjs:607:46)
    at Object.isoToCalendarDate (file:///home/tjc/proposal-temporal/polyfill/lib/calendar.mjs:618:24)
    at Object.calendarToIsoDate (file:///home/tjc/proposal-temporal/polyfill/lib/calendar.mjs:930:32)
    at Object.dateToISO (file:///home/tjc/proposal-temporal/polyfill/lib/calendar.mjs:2158:32)
    at CalendarYearMonthFromFields (file:///home/tjc/proposal-temporal/polyfill/lib/ecmascript.mjs:1776:31)
    at Module.ToTemporalYearMonth (file:///home/tjc/proposal-temporal/polyfill/lib/ecmascript.mjs:1375:21)
    at PlainYearMonth.from (file:///home/tjc/proposal-temporal/polyfill/lib/plainyearmonth.mjs:142:15)
```
The ISO date +275760-09-21 should not be rejected here, because the year/month +275760-09 is within range. This happens because `getCalendarParts` calls `IntlDateTimeFormatPrototypeFormatToParts`, which throws on an out-of-range ISO date.

~~I haven't checked if the bug affects other calendars.~~ The other affected calendars are ethioaa, ethiopic, hebrew, indian, islamic-civil, islamic-tbla, and islamic-umalqura.

jessealama:
`calendarToIsoDate()`'s bisection loop can produce intermediate ISO dates beyond the legacy `Date` boundary, causing `isoToCalendarDate()` to fail via `Intl.DateTimeFormat.formatToParts()`.

The fix is to override `isoToCalendarDate` in affected calendar helpers to (1) shift out-of-range ISO dates by a calendar-appropriate cycle, (2) convert, then (3) adjust the year back. Another wrinkle is that Coptic/Ethiopian and Islamic calendars use day-based shifting with exact cycle lengths to avoid Julian–Gregorian drift; Indian uses year-based shifting since its months start at fixed Gregorian dates; Hebrew doesn't need an override since its extreme dates stay within range during bisection, though I'm not fully sure this holds in all cases.

Fixes #3251

### DIFF
--- a/polyfill/lib/calendar.mjs
+++ b/polyfill/lib/calendar.mjs
@@ -561,26 +561,22 @@ function simpleDateDiff(one, two) {
 }
 
 function clampISODate(iso) {
-  if (iso.year < -271821 || (iso.year === -271821 && (iso.month < 4 || (iso.month === 4 && iso.day < 19)))) {
-    return { year: -271821, month: 4, day: 19 };
-  }
-  if (iso.year > 275760 || (iso.year === 275760 && (iso.month > 9 || (iso.month === 9 && iso.day > 13)))) {
-    return { year: 275760, month: 9, day: 13 };
-  }
+  const cmp = compareISODateToLegacyDateRange(iso);
+  if (cmp < 0) return { year: -271821, month: 4, day: 19 };
+  if (cmp > 0) return { year: 275760, month: 9, day: 13 };
   return iso;
 }
 
-function isOutOfLegacyDateRange(isoDate) {
+function compareISODateToLegacyDateRange(isoDate) {
   const { year, month, day } = isoDate;
-  if (year < -271821 || year > 275760) return true;
-  if (year === -271821) return month < 4 || (month === 4 && day < 19);
-  if (year === 275760) return month > 9 || (month === 9 && day > 13);
-  return false;
+  if (year < -271821 || (year === -271821 && (month < 4 || (month === 4 && day < 19)))) return -1;
+  if (year > 275760 || (year === 275760 && (month > 9 || (month === 9 && day > 13)))) return 1;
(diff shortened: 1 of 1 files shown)


polyfill/lib/calendar.mjs
```diff
@@ -570,6 +570,74 @@ function clampISODate(iso) {
   return iso;
 }
 
+function isOutOfLegacyDateRange(isoDate) {
+  const { year, month, day } = isoDate;
+  if (year < -271821 || year > 275760) return true;
+  if (year === -271821) return month < 4 || (month === 4 && day < 19);
+  if (year === 275760) return month > 9 || (month === 9 && day > 13);
+  return false;
+}
```
ptomato: Can we reuse this logic in `clampISODate()` above?
Maybe if this function returns -1/0/1 for too early / within range / too late?

jessealama: Neat idea! I'll refactor things.

jessealama: Done. Care to take another look?

PR #3280, opened 2026-02-19
Polyfill: Fix precision loss rounding large durations to days

### LINKED ISSUE #3259: polyfill: rounding large duration to days causes 1 day error
```javascript
// expected value is P100017D, but the reference polyfill returns P100016D
Temporal.Duration.from('P100016DT0.0000001S').round({ smallestUnit: 'day', roundingMode: 'ceil' });
```

I didn't read the reference polyfill code, but I suspect that the culprit is a floating-point error in `NudgeToDayOrTime` AO. Note that this also happens in temporal-polyfill too (cf. https://github.com/fullcalendar/temporal-polyfill/issues/84).

jessealama:
Fixes #3259.

Adds defensiveness for handling non-integer numbers of days by using bigInt-backed rounding for the `smallestUnit === 'day'` path, matching how other time units are already handled. Other time units don't appear to be affected.

### DIFF
--- a/polyfill/lib/duration.mjs
+++ b/polyfill/lib/duration.mjs
@@ -287,11 +287,12 @@ export class Duration {
     assert(!ES.IsCalendarUnit(smallestUnit), 'smallestUnit was larger than largestUnit');
     let internalDuration = ES.ToInternalDurationRecordWith24HourDays(this);
     if (smallestUnit === 'day') {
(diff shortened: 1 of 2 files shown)


ptomato (approved):
OK, I've convinced myself this code is equivalent to what the spec says. Thanks!

Issue #3269, opened 2026-02-05
Polyfill: String parsing cases not handled correctly


ptomato:
While working on snapshot tests for string parsing, I found a few cases that should be permitted by the spec text, but the polyfill does not handle correctly.
```js
Temporal.Instant.from('1976-11-18T15+000000,0')
Temporal.ZonedDateTime.from('1976-11-18T15-040000,0[-04:00]')
Temporal.PlainMonthDay.from('11-18[+01:00][u-ca=ISO8601]')
Temporal.PlainYearMonth.from('1976-11[u-ca=ISO8601]')
```
I would guess these are two separate bugs. The PlainMonthDay and PlainYearMonth bug is probably that we forgot to check case-independently that the calendar is iso8601 when accepting the YYYY-MM or MM-DD format. I'm not sure what the other bug is but it's obviously something with UTC offsets of the form ±HHMMSS,0.

These cases should be put into test262 as well.

PR #3282, opened 2026-02-26
Canonicalize calendar before checking YYYY-MM or MM-DD format

ioannad:
This fixes one of the two bugs reported in
https://github.com/tc39/proposal-temporal/issues/3269

In particular, without this fix the following two statements throw incorrectly:

```
Temporal.PlainMonthDay.from('11-18[+01:00][u-ca=ISO8601]')
Temporal.PlainYearMonth.from('1976-11[u-ca=ISO8601]')
```

### DIFF
--- a/polyfill/lib/ecmascript.mjs
+++ b/polyfill/lib/ecmascript.mjs
@@ -550,8 +550,11 @@ export function ParseTemporalYearMonthString(isoString) {
     year = +yearString;
     month = +match[2];
     referenceISODay = 1;
-    if (calendar !== undefined && calendar !== 'iso8601') {
-      throw new RangeErrorCtor('YYYY-MM format is only valid with iso8601 calendar');
+    if (calendar !== undefined) {
+      calendar = ASCIILowercase(calendar);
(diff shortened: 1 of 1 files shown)


polyfill/lib/ecmascript.mjs
```diff
@@ -550,8 +550,11 @@ export function ParseTemporalYearMonthString(isoString) {
     year = +yearString;
     month = +match[2];
     referenceISODay = 1;
-    if (calendar !== undefined && calendar !== 'iso8601') {
-      throw new RangeErrorCtor('YYYY-MM format is only valid with iso8601 calendar');
+    if (calendar !== undefined) {
+      calendar = CanonicalizeCalendar(calendar);
```
ptomato: It's sufficient to do `ASCIILowercase(calendar)` here (and below) instead of CanonicalizeCalendar. CanonicalizeCalendar does ASCIILowercase and then searches the list of allowed calendars, but we already check below if it is `iso8601`.

ioannad: Done

PR #3284, opened 2026-02-26
Polyfill: Avoid TimeClip at edge of DateTimeFormat supported range

ptomato:
In the reference polyfill we have to simulate formatting Plain types by using the existing DateTimeFormat methods to format epoch milliseconds.

Since Plain types have a slightly larger range than the exact times that DateTimeFormat could previously format, we need to anticipate when they correspond to an epoch milliseconds value that's out of range. In this case, we calculate a "day adjustment" and apply it to the 'day' part of formatToParts() or formatRangeToParts() to fudge the resulting formatted string back to the day that it should have represented.

This is probably not 100% format-accurate. For one thing, it'll lose leading zeroes in the day, if the format calls for them. This workaround is mainly to make sure we can handle the full range of the type, and assumes that the workaround path will rarely be hit in practice.

### DIFF
--- a/polyfill/test/expected-failures.txt
+++ b/polyfill/test/expected-failures.txt
@@ -1,9 +1,6 @@
 # Caused by https://issues.chromium.org/issues/40893567
 # Remove these lines after that bug is fixed.
 staging/Intl402/Temporal/old/date-time-format.js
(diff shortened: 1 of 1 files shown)


ptomato:
See https://github.com/tc39/test262/pull/4954 for test coverage

PR #3287, opened 2026-02-28
Thorough Duration rounding tests (without relativeTo)

ptomato:
Snapshot tests for the Temporal.Duration.p.round() method, covering hundreds of thousands of combinations of time-only duration, largest unit of days or smaller, smallest unit of days or smaller, rounding increment, and rounding mode.

Also verifies the invariant that for these durations, the presence or absence of `relativeTo` doesn't matter for the result, except if the duration itself or the nudge window would be unable to be added to the `relativeTo`, or in the case of `halfEven` rounding mode.

Also includes 2 commits fixing bugs in the reference code, found with these snapshot tests. One floating-point overflow that should've been calculated with bigints, and one overflow of TimeDuration - turns out we can skip that calculation in the cases where TimeDuration would overflow. (I checked, neither of these are spec bugs.)

I hope to follow this up next week with another PR with snapshot tests for `relativeTo`, but the snapshot file is too large and I'll need to figure out a way to shrink it.

### DIFF
--- a/polyfill/ci_codecov_test262.sh
+++ b/polyfill/ci_codecov_test262.sh
@@ -19,7 +19,6 @@ node runtest262.mjs "test262/test/built-ins/Date/*/toTemporalInstant/*.js" || [ 
 for subdir in $subdirs; do
   node runtest262.mjs "$subdir/**" || failed=1
 done
-node runtest262.mjs "test262/test/staging/Intl402/Temporal/**/*.js" || failed=1
 node runtest262.mjs "test262/test/intl402/DateTimeFormat/**/*.js" || failed=1
 node runtest262.mjs "test262/test/built-ins/Date/*/toLocale*String/*.js" || failed=1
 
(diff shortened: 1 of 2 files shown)


polyfill/lib/ecmascript.mjs
```diff
@@ -3342,10 +3342,15 @@ function NudgeToCalendarUnit(
...
 
-  // Trick to minimize rounding error, due to the lack of fma() in JS
-  const fakeNumerator = new TimeDuration(denominator.totalNs.times(r1).add(numerator.totalNs.times(increment * sign)));
-  const total = fakeNumerator.fdiv(denominator.totalNs);
-  assert(MathAbs(r1) <= MathAbs(total) && MathAbs(total) <= MathAbs(r2), 'r1 ≤ total ≤ r2');
+  let total = null;
+  if (increment === 1) {
+    // Trick to minimize rounding error, due to the lack of fma() in JS
+    const fakeNumerator = new TimeDuration(
+      denominator.totalNs.times(r1).add(numerator.totalNs.times(increment * sign))
```
Ms2ger: `increment * sign` can just be `sign` here, no?

PR #3291, opened 2026-03-05
Polyfill: Replace unreachable retry with assertion in NudgeToCalendarUnit

### LINKED ISSUE #3235: Test coverage gap in NudgeToCalendarUnit
codecov reports the following code (currently lines 3313-3330 of ecmascript.mjs) as a coverage gap:

```
else if (sign == -1) {
    if (!(nudgeWindow.endEpochNs.leq(destEpochNs) && destEpochNs.leq(nudgeWindow.startEpochNs))) {
      // Retry nudge window if it's out of bounds
      nudgeWindow = ComputeNudgeWindow(
        sign,
        duration,
        originEpochNs,
        isoDateTime,
        timeZone,
        calendar,
        increment,
        unit,
        true
      );
      assert(
        nudgeWindow.endEpochNs.leq(destEpochNs) && destEpochNs.leq(nudgeWindow.startEpochNs),
        `${unit} was 0 days long`
      );
      didExpandCalendarUnit = true;
```

and indeed, we don't have tests that exercise the case where the duration is negative and we need to retry the nudge window.

I haven't been able to construct an example to exercise this code. In the `(sign == 1)` case, the nudge window may need to be retried if (for example) the duration being added could be expressed as either 1 month, 10 hours or 29 days, 10 hours (such as when the relativeTo date is 2020-01-31). The same case doesn't seem to happen going backwards as the duration is unambiguously a month.

However, I haven't convinced myself that this is always true.

jessealama:
The retry path for negative durations appears unreachable: end-of-month constraining reduces the day, which for negative durations pushes the endpoint further from the origin, so the nudge window always contains the destination.

This isn't a rigorous proof of unreachability, but we haven't been able to construct a test case that exercises this path. Replace the retry block with an assertion guarding the invariant, so that if we're wrong, we'll get a clear error rather than silent misbehavior.

Closes #3235

### DIFF
--- a/polyfill/lib/ecmascript.mjs
+++ b/polyfill/lib/ecmascript.mjs
@@ -3308,25 +3308,11 @@ function NudgeToCalendarUnit(
       didExpandCalendarUnit = true;
     }
   } else if (sign == -1) {
-    if (!(nudgeWindow.endEpochNs.leq(destEpochNs) && destEpochNs.leq(nudgeWindow.startEpochNs))) {
-      // Retry nudge window if it's out of bounds
-      nudgeWindow = ComputeNudgeWindow(
-        sign,
-        duration,
-        originEpochNs,
-        isoDateTime,
-        timeZone,
-        calendar,
-        increment,
-        unit,
-        true
-      );
-      assert(
-        nudgeWindow.endEpochNs.leq(destEpochNs) && destEpochNs.leq(nudgeWindow.startEpochNs),
-        `${unit} was 0 days long`
-      );
-      didExpandCalendarUnit = true;
-    }
+    /* c8 ignore next */
+    assert(
+      nudgeWindow.endEpochNs.leq(destEpochNs) && destEpochNs.leq(nudgeWindow.startEpochNs),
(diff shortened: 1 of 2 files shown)


jessealama:
> I think if we assume this code can't be reached by any JS code, we should change the spec text accordingly to have an assertion there as well, and submit it as an editorial change. (Could possibly also be done after stage 4.)
> 
> I have a hunch that the reason the nudge window only needs to be retried when moving forwards, is because the reason for the retry is when the edge of the window lands in a DST gap. The `disambiguation: "compatible"` behaviour always picks the later time for DST gaps, so that might be why. It's just a hunch, but maybe that helps ruling out why this doesn't happen?

I've tweaked the spec text to add an assertion, matching the polyfill code. I've added a note there. I'm happy to open an issue (which this PR would close) where we can try to think out loud about why this behavior is impossible. How does that sound?

ptomato:
That sounds great.

ptomato:
Closing as per https://github.com/tc39/proposal-temporal/issues/3235#issuecomment-4410646115 - this code is not dead after all.

Issue #3311, opened 2026-04-29
Assertion failure and shady results in DifferenceZonedDateTime across backward UTC shift


ptomato:
Discovered thanks to Firefox fuzzing: https://bugzilla.mozilla.org/show_bug.cgi?id=2028872

Reproducer:
```js
const zdt1 = Temporal.ZonedDateTime.from('1867-10-18T12:44:35.000000001-11:47[America/Adak]');
const zdt2 = Temporal.ZonedDateTime.from('1867-10-19T00:00:00+12:13[America/Adak]');
const realDuration = zdt1.since(zdt2);

zdt1.since(zdt2, { largestUnit: 'days' });
zdt2.until(zdt1, { largestUnit: 'days' });
realDuration.round({ largestUnit: 'days', relativeTo: zdt2 });
realDuration.total({ unit: 'days', relativeTo: zdt2 });
```

The four method calls (`since`, `until`, `round`, `total`) all hit the assertion in step 3 of CombineDateAndTimeDuration ("Assert: If _dateSign_ ≠ 0 and _timeSign_ ≠ 0, _dateSign_ = _timeSign_.") via step 15 of DifferenceZonedDateTime ("Return CombineDateAndTimeDuration(_dateDifference_, _timeDuration_)"). The cause is pretty simple. For the input data above, `zdt1` is chronologically later than `zdt2` but earlier on the wall-clock, because of the backwards time zone shift. The day correction loop in DifferenceZonedDateTime handles that fine if the two times occur within the same calendar day, but here they don't and that's exactly the case that fails the assertion.

I'm not 100% sure what the results of the above calls should be, but my first guess would be that `since`, `until`, and `round` should all return a duration of equal length to `realDuration`. We should figure this out first, and then add test262 tests and adjust the assertion accordingly, since it is invalid.

In addition while investigating this, I also found shady results when rounding relative to the later ZonedDateTime (`zdt1` instead of `zdt2`):
```js
// Note, realDuration is PT12H44M35.000000001S
realDuration.round({ largestUnit: 'days', relativeTo: zdt1 });
  // PT36H44M35.000000001S
realDuration.total({ unit: 'days', relativeTo: zdt1 });
  // 0.5309606481481597
```
[...]

Issue #3312, opened 2026-04-29
Assertion failure in ZonedDateTime.prototype.round for backward UTC shift bridging midnight


ptomato:
Discovered thanks to Firefox fuzzing: https://bugzilla.mozilla.org/show_bug.cgi?id=2029756

Reproducer:
```js
const z = Temporal.ZonedDateTime.from("2010-03-04T23:10:00+08:00[Antarctica/Casey]");
z.round({ smallestUnit: 'day' });
```

What's going on here is that there was a backward UTC shift that crossed midnight. 23:10+08:00 is the repeated 23:10 wall-clock time, so it's in a piece of 2010-03-04 that occurs after a piece of 2010-03-05. In ZonedDateTime.prototype.round, we call GetStartOfDay to find the beginning and the end of the day to snap to. For the latter, GetStartOfDay disambiguates to the earlier midnight (at the end of the first piece of 2010-03-04). The assertion fails because we were not expecting the ZonedDateTime to be _after_ the end of the day that it's supposed to round to.

Note that the relatively much more common phenomenon of a UTC shift that starts or ends at midnight, such as Brazil pre-2018, is not affected.

Also note that we use GetStartOfDay similarly to find the start of the _next_ day in ZonedDateTime.prototype.hoursInDay. That does not fail any assertion; we already discussed it in https://github.com/tc39/proposal-temporal/issues/2938, but forgot to consider how it would affect ZonedDateTime.prototype.round. Applied to this case, `z.hoursInDay` could plausibly return either 24 (length of the first piece of 2010-03-04) or 27 (length from the first moment of 2010-03-04 until the last moment, including a discontiguous piece of 2010-03-05). We chose the former, and documented it as such. Whatever resolution we pick here should be consistent with hoursInDay (or else we should consider changing the hoursInDay behaviour as well.)

ptomato:
Trying to figure out what the correct result should be, I drew this crappy diagram:

<img width="1000" height="600" alt="Image" src="https://github.com/user-attachments/assets/854cceda-42a8-4577-a1f1-0f89492ac432" />

I'd say to be consistent with `hoursInDay`, we'd want the result of `round({ smallestUnit: 'day' })` to be the start of the first 03-05 (short blue line).

If we decide that 27 is a better result for `hoursInDay`, we'd want the result of `round({ smallestUnit: 'day' })` to be the end of the second 03-04 (short red line).

ptomato:
[...]

PR #3315, opened 2026-05-08
Handle double midnight in ZonedDateTime.round

ptomato:
Tentative solution to #3312. If we like this solution, I'll make the change in the spec text.


polyfill/lib/zoneddatetime.mjs
```diff
@@ -301,7 +301,8 @@ export class ZonedDateTime {
       assert(thisNs.geq(startNs), 'cannot produce an instant during a day that occurs before start-of-day instant');
 
       const endNs = ES.GetStartOfDay(timeZone, dateEnd);
-      assert(thisNs.lt(endNs), 'cannot produce an instant during a day that occurs on or after end-of-day instant');
+      // Handle interleaved calendar days due to repeated midnight
```
justingrant: Either an example of when this happens, or a link to a GH issue, might be warranted here, because the use case is so weird and obscure.

justingrant (approved):
I think this is fine?  Not 100% confident but it seems reasonable to handle the one known case of this.

justingrant:
Although the snapshot tests fail. Is that expected?

ptomato:
> Although the snapshot tests fail. Is that expected?

Well, I did forget to update the snapshot files, which I've done now. But one of them still fails because the Antarctica/Casey case _also_ hits the assertion in #3311, so we won't get a clean run until we have a solution for that.

ptomato:
Proposing the spec change in https://ptomato.name/talks/tc39-2026-05/#6.

ptomato:
The needs-consensus PR that corresponds to this fix reached TC39 consensus in the meeting of 2026-05-19.

ptomato:
I removed the new case from the snapshot files, since it doesn't fully pass them until we also fix #3311. I'll re-add it then.

PR #3317, opened 2026-05-15
Fix ZDT diff issue with ymd travel with opposite sign of ns travel

arshaw:
Fix for #3311

Just some strategically placed short-circuits guarding against weird scenarios, falling back to time-only math.

Output:

```js
const zdt1 = Temporal.ZonedDateTime.from('1867-10-18T12:44:35.000000001-11:47[America/Adak]');
const zdt2 = Temporal.ZonedDateTime.from('1867-10-19T00:00:00+12:13[America/Adak]');
const realDuration = zdt1.since(zdt2); // PT12H44M35.000000001S

zdt1.since(zdt2, { largestUnit: 'days' }); // PT12H44M35.000000001S
zdt2.until(zdt1, { largestUnit: 'days' }); // PT12H44M35.000000001S
realDuration.round({ largestUnit: 'days', relativeTo: zdt2 }); // PT12H44M35.000000001S
realDuration.total({ unit: 'days', relativeTo: zdt2 }); // 0.26548032407407984

// from ptomato's follow-up investigation
realDuration.round({ largestUnit: 'days', relativeTo: zdt1 }); // PT12H44M35.000000001S
realDuration.total({ unit: 'days', relativeTo: zdt1 }); // 0.5309606481481597
```

All results look reasonable, except for maybe the first .total one. Gotta think through if that's correct. Might be related to another PR I'm about to make...

### DIFF
--- a/polyfill/lib/ecmascript.mjs
+++ b/polyfill/lib/ecmascript.mjs
@@ -3104,16 +3104,19 @@ export function DifferenceZonedDateTime(ns1, ns2, timeZone, calendar, largestUni
   // Convert start/end instants to datetimes
   const isoDtStart = GetISODateTimeFor(timeZone, ns1);
   const isoDtEnd = GetISODateTimeFor(timeZone, ns2);
+  const isoDtLexSign = CompareISODate(isoDtStart.isoDate, isoDtEnd.isoDate);
 
-  // If year-month-day values are same-day,
-  // there's no point involving the calendar/timezone for zdt->pdt conversions.
-  // It also avoids a situation where dayCorrection backs up too far on same-day diffs
-  // with reverse-direction wallclock delta due to DST:
-  // https://github.com/tc39/proposal-temporal/issues/3141
   if (
-    isoDtStart.isoDate.year === isoDtEnd.isoDate.year &&
-    isoDtStart.isoDate.month === isoDtEnd.isoDate.month &&
-    isoDtStart.isoDate.day === isoDtEnd.isoDate.day
+    // If year-month-day values are same-day,
+    // there's no point involving the calendar/timezone for zdt->pdt conversions.
+    // It also avoids a situation where dayCorrection backs up too far on same-day diffs
+    // with reverse-direction wallclock delta due to DST:
+    // https://github.com/tc39/proposal-temporal/issues/3141
+    isoDtLexSign === 0 ||
+    // If year-month-day diff is opposite sign from overall nanosecond diff, then the time zone
+    // likely repeated a whole day (ex: America/Adak). In this case, do all math in time realm.
+    // https://github.com/tc39/proposal-temporal/issues/3311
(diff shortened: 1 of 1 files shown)


polyfill/lib/ecmascript.mjs
```diff
@@ -3104,16 +3104,19 @@ export function DifferenceZonedDateTime(ns1, ns2, timeZone, calendar, largestUni
...
+    // there's no point involving the calendar/timezone for zdt->pdt conversions.
+    // It also avoids a situation where dayCorrection backs up too far on same-day diffs
+    // with reverse-direction wallclock delta due to DST:
+    // https://github.com/tc39/proposal-temporal/issues/3141
+    isoDtLexSign === 0 ||
+    // If year-month-day diff is opposite sign from overall nanosecond diff, then the time zone
+    // likely repeated a whole day (ex: America/Adak). In this case, do all math in time realm.
+    // https://github.com/tc39/proposal-temporal/issues/3311
+    isoDtLexSign !== sign
   ) {
```
ptomato, line 3120: These two conditions can probably collapse into `if (CompareISODate(isoDtStart.isoDate, isoDtEnd.isoDate) === -sign)`?

ptomato (commented):
And another big thank you for looking into this one as well!

I haven't yet had a chance to dig as deep into this issue as I have into the other ones, so I'll have less to say.

Firefox's [proposed patch](https://bugzilla.mozilla.org/page.cgi?id=splinter.html&ignore=&bug=2028872&attachment=9563616) also had something similar to the first part of yours, falling back to time math if isoDtStart/isoDtEnd passed each other. (They had a second fix in ZonedDateTime.round which I didn't believe was correct though.)

This PR passes the tentative test262 tests I've written, but note that I still need to validate those against the ideas that Justin wrote down in https://github.com/tc39/proposal-temporal/issues/3311#issuecomment-4427936643.

I threw a bunch of snapshot tests at this PR and noticed that it does change one result that I wasn't expecting:
```js
const earlier = Temporal.ZonedDateTime.from('2011-12-29T23:59:59.999999999-10:00[Pacific/Apia]');
const later = Temporal.ZonedDateTime.from('2011-12-31T00:00+14:00[Pacific/Apia]');
earlier.since(later, { largestUnit: 'days' })
  // was    : P1DT0.000000001S
  // becomes: PT0.000000001S
```
This is indeed falling back to time math, but here I'm not sure that it should, since it's a +24h transition where the direction of travel of wall clock and exact time is the same.

arshaw:
@ptomato are you sure?

> I threw a bunch of snapshot tests at this PR and noticed that it does change one result that I wasn't expecting

On current live https://tc39.es/proposal-temporal/docs/ (shim impl) and current live Chrome (native impl), i'm getting `PT0.000000001S` for all.

Screenshot 2026-06-29 at 9 07 45 PM

PR #3319, opened 2026-05-19
Polyfill: Fix detection of bounding window start

### LINKED ISSUE #3149: Duration.p.round, when relativeTo with second wallclock in DST transition, wrong results
Similar root cause as #3148

Recreation:

```js
// the second 01:00 time in the DST transition. 01:00-07:00 happens just before this
const origin = Temporal.ZonedDateTime.from('2025-11-02T01:00:00-08:00[America/Vancouver]')
const dur = Temporal.Duration.from({ hours: 11, minutes: 30 })
const roundedDur = dur.round({
  largestUnit: 'days',
  smallestUnit: 'days',
  relativeTo: origin,
  roundingMode: 'halfExpand',
})
```

ACTUAL: `'P1D'`
EXPECTED: `'PT0S'`

Why? Using the spirit of the algorithm, we can do step-by-step:

```js
const origin = Temporal.ZonedDateTime.from('2025-11-02T01:00:00-08:00[America/Vancouver]')
const dur = Temporal.Duration.from({ hours: 11, minutes: 30 })
const dest = origin.add(dur) // '2025-11-02T12:30:00-08:00[America/Vancouver]'
const spanStart = origin.add({ days: 0 }) // '2025-11-02T01:00:00-08:00[America/Vancouver]'
const spanEnd = origin.add({ days: 1 }) // '2025-11-03T01:00:00-08:00[America/Vancouver]'
const numerator = dest.epochNanoseconds - spanStart.epochNanoseconds
const denominator = spanEnd.epochNanoseconds - spanStart.epochNanoseconds
const frac = Number(numerator) / Number(denominator)
// 0.4791666666666667
// 0 with halfExpand
```

### LINKED ISSUE #3316: Shady results when rounding duration with smallestUnit/largestUnit
While investigating #3310 I stumbled upon another result that looks dubious to me.
```js
const relativeTo = new Temporal.ZonedDateTime(0n, "UTC");
const d = Temporal.Duration.from('P7D');
d.round({ smallestUnit: 'days', largestUnit: 'weeks', relativeTo })
  // => P1W1D, should be P1W?
```
This seems plain wrong, and unfortunately does not even involve weirdness with a time zone transition.

SpiderMonkey and V8 also produce this result.

LLM disclosure: I found this while prototyping other solutions with a bot, and prompted the bot to write an isolated test case. The risk of hallucination is minimal, because the test case is simple enough and the expected result is obvious.

ptomato:
Seems to be a regression from https://github.com/tc39/proposal-temporal/pull/3147. There is a condition triggering the special case for avoiding a disambiguation when calculating the startEpochNs of the bounding window (added to fix https://github.com/tc39/proposal-temporal/issues/3149). This condition is not correct when largestUnit ≠ smallestUnit, and the smallest unit is 0 while a larger unit is nonzero. The condition only checks that the smallest unit (r1) is 0, while it should check that all larger units are 0 as well.

Updates the newly added snapshot tests to show where the differences in output are.

This fix corresponds to a needs-consensus PR that reached TC39 consensus in the meeting of 2026-05-19.

Closes: https://github.com/tc39/proposal-temporal/issues/3316


Ms2ger (approved):
Thanks!

